PR #11, opened 2015-06-03
build compatible w/windows; some color for webpack

mikekidder:


package.json
```diff
@@ -4,10 +4,11 @@
   "description": "An experiment in fully hot-reloadable Flux",
   "main": "lib/index.js",
   "scripts": {
-    "start": "cd examples && node server.js",
+    "build": "babel src --out-dir lib",
+    "clean": "rimraf lib",
```
gaearon: Hmm, this seems to assume `rimraf`. Needs to be added to `devDependencies`?

gaearon: I added https://github.com/gaearon/redux/commit/059f4dd379baa3380ebedba59e7c84054e849193

mikekidder: oops, yes... sorry about that!

PR #12, opened 2015-06-03
Reduce context footprint

ooflorent:
Wrapping context would prevent collision with 3rd party libraries.

### DIFF
--- a/src/Root.js
+++ b/src/Root.js
@@ -4,14 +4,16 @@ import createDispatcher from './createDispatcher';
 export default function root(DecoratedComponent) {
   return class ReduxRoot extends Component {
     static childContextTypes = {
-      observeStores: PropTypes.func.isRequired,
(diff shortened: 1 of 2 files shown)


src/Root.js
```diff
@@ -4,14 +4,16 @@ import createDispatcher from './createDispatcher';
...
   return class ReduxRoot extends Component {
     static childContextTypes = {
-      observeStores: PropTypes.func.isRequired,
-      wrapActionCreator: PropTypes.func.isRequired
+      redux: PropTypes.object.isRequired
     };
 
     getChildContext() {
+      const { observeStores, wrapActionCreator } = this.dispatcher
       return {
```
ooflorent: Could be changed to:

```js
return {
  redux: this.dispatcher
}
```

gaearon:
Good point. :+1:

PR #15, opened 2015-06-03
Resurrect decorators

ooflorent:
Bring back decorator APIs:

```js
import { container, root } from 'redux';

@root
class App {
  render() {
    return (
      <Counter />
    );
  }
}

@container({
  actions: { increment, decrement },
  stores: counterStore
})
class Counter {
  static propTypes = {
    increment: PropTypes.func.isRequired,
    decrement: PropTypes.func.isRequired,
    counter: PropTypes.number.isRequired
  };

  render() {
    const { increment, decrement, counter } = this.props;
    return (
      <p>
        Clicked: {counter} times
        {' '}
        <button onClick={() => increment()}>+</button>
        {' '}
        <button onClick={() => decrement()}>-</button>
      </p>
    );
  }
}
```

### DIFF
--- a/src/Root.js
+++ b/src/Root.js
@@ -1,23 +1,32 @@
-import React, { Component, PropTypes } from 'react';
+import React, { PropTypes } from 'react';
 import createDispatcher from './createDispatcher';
 
-export default function root(DecoratedComponent) {
-  return class ReduxRoot extends Component {
-    static childContextTypes = {
-      redux: PropTypes.object.isRequired
-    };
-
-    getChildContext() {
-      return { redux: this.dispatcher };
-    }
-
-    constructor(props, context) {
-      super(props, context);
-      this.dispatcher = createDispatcher();
-    }
-
-    render() {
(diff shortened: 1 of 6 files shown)


src/Root.js
```diff
@@ -1,7 +1,7 @@
 import React, { Component, PropTypes } from 'react';
 import createDispatcher from './createDispatcher';
 
-export default function root(DecoratedComponent) {
+export default function Root(DecoratedComponent) {
```
ooflorent: There was a typo here.

src/decorators/root.js
```diff
@@ -0,0 +1,3 @@
+import Root from '../Root';
+
+export default Root;
```
ooflorent: Current `Root` is a decorator. It may change in the future.

ooflorent:
@gaearon What do you think of it? I haven't done the `addons` stuff.

gaearon:
Looks good to me. Let me play with it & get back to you!

gaearon:
Out in 0.4.0.

PR #18, opened 2015-06-03
Fix decorator name

ooflorent:


gaearon:
Whoops, I just did it in another commit. Thanks for noticing ;-)

PR #19, opened 2015-06-03
Allow children to be React element

ooflorent:
The following PR allows `ReduxContainer` to contain React elements.
It does not break compatibility.

```js
<Container {...opts} Wrapper={SomeFancyComponent}>
  <CounterRed />
  <CounterBlue />
</Container>
```

Feedback is welcomed :wink:

### DIFF
--- a/src/Container.js
+++ b/src/Container.js
@@ -1,7 +1,9 @@
-import { Component, PropTypes } from 'react';
+import React, { Children, Component, PropTypes } from 'react';
+import invariant from 'invariant';
 import mapValues from 'lodash/object/mapValues';
 import identity from 'lodash/utility/identity';
 import invariant from 'invariant';
+import isFunction from 'lodash/lang/isFunction';
 import isPlainObject from 'lodash/lang/isPlainObject';
 
 export default class ReduxContainer extends Component {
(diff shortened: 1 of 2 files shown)


src/addons/container.js
```diff
@@ -2,13 +2,13 @@ import React from 'react';
 import Container from '../Container';
 import getDisplayName from './getDisplayName';
 
-export default function container({ actions, stores }) {
+export default function container(options) {
```
ooflorent: Maybe we should find a better name

src/Container.js
```diff
@@ -71,9 +80,29 @@ export default class ReduxContainer extends Component {
...
-    });
+    };
+
+    if (isFunction(children)) {
+      return children(props);
+    }
+
+    if (Array.isArray(children)) {
+      return (
+        <Wrapper>
```
ooflorent: I was not able to get rid of this...

gaearon:
What does this buy us over the function form?

ooflorent:
With `react@0.14` and parent-based context we no longer need to be careful with the `owner`. Using `children` as a function into `Root` will be useless. Allowing the same syntax for `Container` would reinforce code consistency between the 2 components.

gaearon:
Good point about `Root` in React 0.14.

I think the need for `Wrapper` is what makes this proposal not very appealing to me. With the function form, we need to teach people once: put a function inside and you'll be fine. With this form, I'd need to consult the docs: why can't I just put two things in there? Where does the extra span come from? Oh wait, so there's `Wrapper`. But I used a custom component, and now I want to props to my wrapper. How do I do that? I'd need `wrapperProps`. Or else they'd be magically inferred from `props` I passed to the `<Component>`...

Another problem is cloning. I don't see cloning as the right solution. A designer might wrap the insides of `<Container>` in a `<div>` to change the layout, and the data no longer comes through because only the immediate children are cloned. Oops! Also, I specify my props with `isRequired`, but propTypes validation will fail initially because when I create the element, the props are not there yet.

The alternative seems to be more consistent. No edge cases, and a single trick to learn instead of a dozen. These are the reasons I'm inclined to not accept this.

ooflorent:
Perfectly legit. It was just a suggestion.

ooflorent:
However be sure to drop the function form for `Root` with 0.14 :wink:

gaearon:
Yep!

PR #26, opened 2015-06-03
Adds in-publish script which prevents running prepublish on npm insall

boopathi:
Your call - If you'd like to have `npm install` not run `prepublish` script.

More discussion here:
https://github.com/npm/npm/issues/3059

Note: The eslint step fails now.


boopathi:
Actually no!
You need to run babel to run on preinstall and prepublish.

Closing this.

Issue #7, opened 2015-06-02
Implement devtool hooks


gaearon:
When #6 is done, I want to use it to implement hooks for debugging and time travelling from console.

![](https://pbs.twimg.com/media/CGhoexjVAAE5V4k.png:large)

The API could be something like

```js
ReduxDebug.printLastActions();
ReduxDebug.printLastStates();
ReduxDebug.jumpTo(stateId);
```

etc.

dariocravero:
It would be nice to model this in a flux way so that apps (like a devtools app) can be plugged in on top of it. [Marty has done that already](http://martyjs.org/devtools/). We can use [panels](https://github.com/UXtemple/panels-ui) to present it. I'll be glad to help out with that :)

gaearon:
Yeah, I definitely want to potentially support a DevTools app. I'm not going to write it myself tho, as it's crazy effort and people trying it generally say it wasn't worth it. A separate “debugger” app communicating by sockets that can be displayed in a DevTools container is more like what I envision.

dariocravero:
Good stuff :) I think that in redux's case it might be a different story regarding its use as _hot reloading all the things_ changes the scenario quite a lot and a clear UI to it might be very beneficial to do all that time travelling :)

gaearon:
Agree here. :-)
I'm aiming for something runnable in a month for my talk at ReactEurope so I'm keeping things simple for now.

fibric:
it looks like redux is now turning into an 'isomorphic thing' (or it ever was but api changes make it clearer now). then i suggest to check out https://github.com/caiogondim/logdown

gaearon:
Superseded by #195.

tyleranton:
>A separate “debugger” app communicating by sockets that can be displayed in a DevTools container is more like what I envision.

I'm currently working on that [here](https://github.com/tyleranton/redux-devtools-electron). So far it has been doable (and difficult). At the moment I'm trying to reason about getting Commit, Revert, etc. to work back to the client from the desktop app. 

I plan on building in more devtools features as well (eg. The ability to add action buttons that will trigger your actions in your application).

Issue #4, opened 2015-06-02
Action types as symbols


ooflorent:
Allowing this would be awesome.
Why did you enforce `type` as `string`? ([createDispatcher.js#L67](https://github.com/gaearon/redux/blob/789b548f065b16b1ca3c1057bf97e0a23610e087/src/createDispatcher.js#L67))

```js
export const INCREMENT_COUNTER = Symbol()
export const DECREMENT_COUNTER = Symbol()
```

gaearon:
I thought of that, but this kind of screws the replay thing. (AFAIK you can't serialize and then deserialize symbols.) I want to pass them around for devtools, localStorage persistence, etc.

ooflorent:
You're right you cannot deserialize a `Symbol`. I've missed that point.

gaearon:
I say let's keep them strings for now and revisit if there are ways to get around (de)serialization problem.

tbranyen:
I might be confused here, but what do you mean by "cannot deserialize a Symbol"?

``` javascript
const INCREMENT_COUNTER = Symbol('INCREMENT_COUNTER');

console.log(String(INCREMENT_COUNTER)); // Symbol(INCREMENT_COUNTER)
```

Wouldn't this be perfectly ideal for Redux? The ability to have identically named action types that are namespaced. I'm absolutely hating the fear that action types are overlapping since everything is global.

tbranyen:
I just tried using a Symbol and thankfully it works excellent with Redux! The only error that came up was with redux-logger, so I'll narrow down the error and submit a patch (if one isn't already there).

gaearon:
You can't take an array of actions that use symbols, serialise them to JSON, save to localStorage, and later deserialise and replay them in a different browser session.

tbranyen:
@gaearon I see what you mean, that is very true. For my use case it won't be a problem, and I've submitted a patch request for redux-logger here: https://github.com/fcomb/redux-logger/pull/106

(Sorry for spamming this particular issue, but I think it's useful for others looking to use Symbols).

PR #28, opened 2015-06-03
Allow symbols to be actionTypes

boopathi:
So you can do,

```js
// actionTypes.js
export const MY_ACTION = Symbol();
```


gaearon:
I'm not sure I'm going to support this because symbols are not serializable AFAIK.
See also #4.

gaearon:
I'm closing for now (see reasons in #4) but may revisit later.
Need to understand our tooling story (#7) first.
Thanks for the PR though!

PR #30, opened 2015-06-03
make consistent with other decorator

mikekidder:


gaearon:
I'd rather change the other one. When `()` are unnecessary, let's drop them.

gaearon:
Closed via 77351708dcb01368e790ffbde24ceb14cf8eebe4

PR #31, opened 2015-06-03
Accept stores as an object of prop key to store

### LINKED ISSUE #22: Shape of props depends on number of stores
Changing the names of props based on number of stores subscribed to seems tricky.

Let's say I have a dumb component that takes a prop `board` ( `{ name: 'My Board' }`), which should come from `boardStore`:
```js
<Container stores={[boardStore]}
           actions={{ addList }}>
  {(props) => <Board {...props} /> }
</Container>
```
But because I only subscribed to one store, it gets flattened and the component only receives `name` prop.

If I need to add a subscription to another store `userStore`, the props change to:
```js
{
  boardStore: { name: 'My Board' },
  userStore: ...,
  fooStore: ... // (This actually includes data from all stores, not just subscribed.)
}
```

I think it would be simplest, if the data would never be flattened. Then I can always just do:
```js
<Container stores={[boardStore]}
           actions={{ addList }}>
  {(props) => <Board board={props.boardStore} /> }
</Container>
```

Another way could be to pass another function or a map of functions to map the state from stores to
props, but I think I actually prefer the former because of its simplicity.

fson:
This makes `stores` and `actions` symmetric and lets us specify custom
prop names easily.

`container` decorator now takes a second argument `transformProps`,
which is a function that is used to transform the state and actions
props before passing them to the component. It defaults to
`({ state, actions }) => ({ ...state, ...actions })`

I removed some nesting from the states of `counterStore` and `todoStore` examples, because it didn't seem necessary after this change and made prop passing simpler.

Fixes #22.

### DIFF
--- a/src/Container.js
+++ b/src/Container.js
@@ -1,6 +1,6 @@
 import { Component, PropTypes } from 'react';
+import values from 'lodash/object/values';
 import mapValues from 'lodash/object/mapValues';
-import identity from 'lodash/utility/identity';
 import invariant from 'invariant';
 import isPlainObject from 'lodash/lang/isPlainObject';
 
@@ -12,11 +12,11 @@ export default class ReduxContainer extends Component {
   static propTypes = {
     children: PropTypes.func.isRequired,
     actions: PropTypes.object.isRequired,
-    stores: PropTypes.arrayOf(PropTypes.func.isRequired).isRequired
+    stores: PropTypes.object.isRequired
   }
 
   static defaultProps = {
-    stores: [],
+    stores: {},
     actions: {}
   };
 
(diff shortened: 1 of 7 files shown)


gaearon:
Neat! Gonna test this tomorrow morning!

ooflorent:
LGTM. Maybe replacing `typeof` checks by `isFunction` would improve readability.

fson:
@ooflorent: Maybe we could replace those invariants altogether, with stricter propTypes?
```js
static propTypes = {
  children: PropTypes.func.isRequired,
  actions: PropTypes.objectOf(PropTypes.func).isRequired,
  stores: PropTypes.objectOf(PropTypes.func).isRequired
}
```
We would not get a custom error message, but I think it would still be pretty obvious what's wrong, `propTypes` are the standard way to validate props after all. That approach doesn't let us validate that the object is a plain object though, unless we make a custom propType validator for it.

ooflorent:
@json I haven't thought about that. Seems to be a viable option. Let's see Dan's opinion.

src/Container.js
```diff
@@ -42,23 +42,26 @@ export default class ReduxContainer extends Component {
       '"actions" must be a plain object with functions as values. Did you misspell an import?'
     );
     invariant(
-      Array.isArray(stores) &&
-      stores.every(s => typeof s === 'function'),
-      '"stores" must be an array of functions. Did you misspell an import?'
+      isPlainObject(stores) &&
+      Object.keys(stores).every(key => typeof stores[key] === 'function'),
```
ooflorent: Addind:

```js
import every from 'lodash/collection/every';
import isFunction from 'lodash/lang/isFunction';
```

and changing the test to:

```js
isPlainObject(stores) && every(stores, isFunction)
```

Would make code more readable.

gaearon:
I'd rather keep the invariants. Prop types are cool but out code will crash if you supply something else. The actual error will be more noticeable than propTypes warning and you might struggle longer than you have to.

gaearon:
Although I sure don't mind *stricter* propTypes. Just want to also have invariants instead of weird crashes if user doesn't notice the warning.

gaearon:
This is awesome. Thanks to everybody for a thoughtful discussion.

PR #32, opened 2015-06-04
Allow to get action result

vslinko:
I want to wait for an async action before React.renterToString on server. Example
```js
function fetchDataForHomePageActionCreator() {
  return async dispatch => {
    await timeout(100)
    dispatch({type: HOMEPAGE_DATA_LOADED})
  }
}

async function homepageHandler(request) {
  const dispatcher = createDispatcher()
  const prepare = dispatcher.wrapActionCreator(fetchDataForHomePageActionCreator)
  await prepare()
  return React.renderToString(<Dispatcher dispatcher={dispatcher}><HomePage /></Dispatcher>)
}
```

Issue #6, opened 2015-06-02
Implement middleware


gaearon:
Similar to https://github.com/rpominov/fluce/issues/4

gaearon:
This is how Redux works now:

```
----[action]---->  [Redux] ----[state]---->
               (calls stores)
```

This is how I want it to work:

```
 ----[action]----> | ? | ----[action]---> [Redux] ----[state]----> | ? | ----[state]---->
                   | ? |                                           | ? |
                   | ? |                                           | ? |
                   | ? |                                           | ? |
```

Things in the middle are called **interceptors**. They are async black boxes. They are the extension points of Redux. There may be an interceptor that logs whatever goes through it. There may be an interceptor that transforms the values passing through it, or delays them. An interceptor should be able to fake the data, i.e. fire an arbitrary action at any point of time even if there's nothing being processed at the moment. This will be useful for time travel.

I think I found a nice API for interceptors. I'm going to rewrite Redux dispatcher to use them now. Even observing changes to stores can then be implemented as a (built-in) interceptor.

The interceptor signature is `sink => payload => ()`. See the proof of concept below.

```js
// Utilities
function noop() { }
function compose(...interceptors) {
  return sink => interceptors.reduceRight(
    (acc, next) => next(acc),
    sink
  );
}
function seal(interceptor) {
  return interceptor(noop);
}

// Dispatcher implementation
function getFirstAtom(stores) {
  return new Map([
    for (store of stores)
      [store, undefined]
  ]);
}
function getNextAtom(stores, atom, action) {
  if (!action) {
    return getFirstAtom(stores);
  }
  return new Map([
    for ([store, state] of atom)
      [store, store(state, action)]
  ]);
}
// Dispatcher is an interceptor too
function dispatcher(...stores) {
  stores = new Set(stores);
  let atom = getFirstAtom(stores);

  return sink => action => {
    atom = getNextAtom(stores, atom, action);
    sink(atom);
  };
}

// What can interceptors do?
// These are simple examples, but they show the power of interceptors.
function logger(label) {
  return sink => payload => {
    console.log(label, payload);
    sink(payload);
  };
}
function repeater(times) {
  return sink => payload => {
    for (let i = 0; i < times; i++) {
      sink(payload);
    }
  };
}
function delay(timeout) {
  return sink => payload => {
    setTimeout(() => sink(payload), timeout);
  };
}
function filter(predicate) {
  return sink => payload => {
    if (predicate(payload)) {
      sink(payload);
    }
[...]

PR #33, opened 2015-06-04
Allow to pass custom dispatcher to Root component

vslinko:
Will be helpful for #32

### DIFF
--- a/src/Root.js
+++ b/src/Root.js
@@ -3,6 +3,7 @@ import createDispatcher from './createDispatcher';
 
 export default class ReduxRoot {
   static propTypes = {
+    dispatcher: PropTypes.object,
     stores: PropTypes.object.isRequired,
     children: PropTypes.func.isRequired
   };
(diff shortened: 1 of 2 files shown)


src/Root.js
```diff
@@ -12,12 +13,19 @@ export default class ReduxRoot {
   };
 
   constructor(props) {
-    this.dispatcher = createDispatcher();
-    this.dispatcher.receiveStores(props.stores);
+    this.dispatcher = props.dispatcher || createDispatcher();
```
ooflorent: Couldn't `createDispatcher()` be called in `static defaultProps`?

vslinko: It isn't good for two reasons:
* Dispatcher from defaultProps will be shared between all Root instances.
* It will be created even if we don't use it.

ooflorent: Well, yes :+1:

vslinko:
@gaearon

gaearon:
I'm going to change quite a few things inside with #6 (work in progress) so hard to say now. I'll keep your use case from #32 in mind.

gaearon:
Closed via #46

PR #36, opened 2015-06-04
Allow to transform container props based on component props

vslinko:


gaearon:
This is good, thanks!

PR #39, opened 2015-06-04
Hydration / Dehydration

vslinko:
I need to transfer state from server to client.

### DIFF
--- a/src/createDispatcher.js
+++ b/src/createDispatcher.js
@@ -128,11 +128,21 @@ export default function createDispatcher() {
       }
     };
   }
+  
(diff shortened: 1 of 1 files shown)


gaearon:
Fair. I don't want to expand the API until I'm done with #6, as if people depend on it now, they'll have more stuff to change.

I intend to keep this open until #6 is solved, and then revisit. (But rest assured something like this will be merged afterwards.)

vslinko:
Thats ok, I just merged all my PRs into my master

gaearon:
Closed via #46

PR #42, opened 2015-06-04
Allow to return from actionCreator without action

vslinko:


gaearon:
Curious, what is the use case? Usually you'd want to exit early when some state-dependent condition is true, but *to read the state, you'd need to use the callback form*, from which you can already exit early.

vslinko:
Yes, I'm already understood that.
I receive problem when started new action with console.log and no return statement.

vslinko:
Could be closed

PR #44, opened 2015-06-05
Allow to read state by store on action creator callback

emmenko:
@gaearon Is that what you meant? It looked too much easier to implement... O.O

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,10 +62,10 @@ export function incrementAsync() {
   };
 }
 
-// Could also look into state in the callback form
+// Could also read state of a store in the callback form
 export function incrementIfOdd() {
-  return (dispatch, state) => {
-    if (state.counterStore % 2 === 0) {
+  return (dispatch, read) => {
+    if (read(counterStore) % 2 === 0) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


gaearon:
Yeah, this seems right! Let's keep this out of the examples though. It's fine in README, I might have a proper example later. It's OK that the method is unused for now.

README.md
```diff
@@ -1,9 +1,9 @@
 redux
 =========================
 
-An experiment in fully hot-reloadable Flux.
```
gaearon: Whitespace is significant in markdown files. Otherwise it'll remove the line breaks where I want them.

emmenko: Oh sorry, I have `"trim_trailing_white_space_on_save": true` on my sublime :)
Will try to revert it...

emmenko:
Regarding the example, sure, I was just making sure it was working :)

I'll revert that as well

emmenko:
@gaearon done :)

gaearon:
Good, can you please squash into a single commit?

emmenko:
Here we go :)

gaearon:
:+1:

gaearon:
Out in 0.7.0.

Issue #45, opened 2015-06-06
remove actions from props


threepointone:
Was wondering how you handle clashing action and store names in containers. But it also strikes me that since actions are just functions *without* a current state, they don't have to be passed down from the top. Instead, as long as the dispatch function is available as a prop, you'll still be able to do full dispatches. to explain, instead of - 
```js 
@inject({
  actions: CounterActions,
  stores: { counter: counterStore }
})
export default class Counter {
  render() {
    const { increment, decrement, counter } = this.props;
    return (
      <p>
        <button onClick={increment}>+</button>
      </p>
    );
  }
```
you could do 

```js 
@inject({
  stores: { counter: counterStore }
})
export default class Counter {
  static propTypes = {
    dispatch: PropTypes.func.isRequired
  };

  render() {
    const { counter, dispatch } = this.props;
    return (
      <p>
        Clicked: {counter} times
        {' '}
        <button onClick={() => dispatch(increment())}>+</button>
        {' '}
        <button onClick={() => dispatch(decrement())}>-</button>
      </p>
    );
  }
```
alternately

```js 
        <button onClick={() => dispatch(increment)}>+</button>
```

but I'm not clear on the benefits (scheduling?) 

This would clear up the name clash problem, and make it more 'primitive', if that make sense.

threepointone:
also, you could then read state from stores when calling `dispatch`, preventing the need for `read`/`storeKeys`

threepointone:
thinking more, anyone could implement any action creator system on top of this. promises, observables, etc. 

[it's starting to look like cycle, but more accessible]

gaearon:
cc @acdlite who was also [rethinking Redux](https://gist.github.com/acdlite/9f1b5883d132ad242323). Indeed, I want to move this action creator <-> dispatch thing out of the core.

gaearon:
This is done now in #46, let's discuss there.

gaearon:
https://github.com/gaearon/redux/pull/46#issuecomment-109595200

threepointone:
Wow lots going on there. I want to hack on a project tonight with this, will come back with thoughts. Cheers.

gaearon:
Out in 0.8.0

PR #41, opened 2015-06-04
Expose createDispatcher

vslinko:
I want to create dispatcher on server

PR #46, opened 2015-06-06
Feeback wanted: Rewrite with ideas from @acdlite

gaearon:
So @acdlite got me a birthday present (yay it's my birthday today :-): [this gist](https://gist.github.com/acdlite/9f1b5883d132ad242323).

It shows how much easier Redux can be implemented if we embrace single-atomness and instead of hiding it, *make it explicit*.

I changed a few things from his implementation:

* Instead of pushing `atom` down the `context`, I push `subscribe()` to avoid [shouldComponentUpdate problems](https://github.com/facebook/react/issues/2517);
* Instead of relying on `setState`, maintain `this.atom` in Dispatcher so `setState` is called both on the root and the children at the same time and the atom is never stale in the action creators;
* Add a convenience `combineStores` higher-order store (lol) that converts an object map of Stores into a Store function that combines them.

I combined our two examples to better show how this works together.
I like it.

Would love to hear your thoughts.

@acdlite @ooflorent @threepointone @vslinko 

## TODO

- [x] Initial implementation of Store composition
- [x] Performance work
- [x] Extract dispatcher for non-React use, make it part of public API
- [x] Figure out the API for dispatching actions (#45, #48)
- [x] Support isomorphic apps @acdlite 
- [x] Update README

### DIFF
--- a/examples/containers/App.js
+++ b/examples/containers/App.js
@@ -1,10 +1,10 @@
 import React from 'react';
-import { Dispatcher, Provider, composeStores } from 'redux';
+import { createDispatcher, Provider, composeStores } from 'redux';
 import CounterApp from './CounterApp';
 import TodoApp from './TodoApp';
 import * as stores from '../stores/index';
 
-const dispatcher = new Dispatcher(composeStores(stores));
+const dispatcher = createDispatcher(composeStores(stores));
 
 export default class App {
   render() {
--- a/examples/containers/CounterApp.js
+++ b/examples/containers/CounterApp.js
@@ -1,18 +1,17 @@
 import React from 'react';
-import { connect } from 'redux';
+import { connect, bindActions } from 'redux';
 import Counter from '../components/Counter';
 import * as CounterActions from '../actions/CounterActions';
 
-@connect({
-  actions: CounterActions,
-  select: state => ({
-    counter: state.counter
-  })
-})
+@connect(state => ({
+  counter: state.counter
+}))
 export default class CounterApp {
   render() {
+    const { counter, dispatcher } = this.props;
     return (
-      <Counter {...this.props} />
+      <Counter counter={counter}
+               {...bindActions(CounterActions, dispatcher)} />
     );
   }
 }
--- a/src/Dispatcher.js
+++ b/src/Dispatcher.js
@@ -26,9 +26,7 @@ export default class Dispatcher {
     this.setAtom(nextAtom);
   }
 
-  perform(actionCreator, ...args) {
-    const action = actionCreator(...args);
-
+  perform(action) {
     return typeof action === 'function'
       ? action(this.dispatch, this.atom)
       : this.dispatch(action);
--- a/src/Dispatcher.js
+++ b/src/Dispatcher.js
@@ -3,17 +3,22 @@ function dispatch(store, atom, action) {
 }
 
 export default class Dispatcher {
-  constructor(store, atom) {
+  constructor(store) {
     this.store = store;
+    this.hydrate();
+  }
+
+  hydrate({ atom, subscriptions = [] } = {}) {
     this.atom = atom;
-    this.subscriptions = [];
+    this.subscriptions = subscriptions;
     this.dispatch({});
   }
 
-  receive(dispatcher) {
-    this.atom = dispatcher.atom;
-    this.subscriptions = dispatcher.subscriptions;
-    this.dispatch({});
+  dehydrate() {
+    const { atom, subscriptions } = this;
+    delete this.atom;
+    this.subscriptions = [];
+    return { atom, subscriptions };
   }
 
   dispatch(action) {
@@ -39,10 +44,5 @@ export default class Dispatcher {
   emitChange() {
     const { atom, subscriptions } = this;
     subscriptions.forEach(listener => listener(atom));
-  }
-
-  dispose() {
-    this.atom = undefined;
-    this.subscriptions = [];
   }
 }
--- a/src/Dispatcher.js
+++ b/src/Dispatcher.js
@@ -26,6 +26,14 @@ export default class Dispatcher {
     this.setAtom(nextAtom);
   }
 
+  perform(actionCreator, ...args) {
+    const action = actionCreator(...args);
+
+    return typeof action === 'function'
+      ? action(this.dispatch, this.atom)
+      : this.dispatch(action);
+  }
+
   setAtom(atom) {
     this.atom = atom;
     this.emitChange();
--- a/src/createDispatcher.js
+++ b/src/createDispatcher.js
@@ -1,5 +1,12 @@
 import Dispatcher from './Dispatcher';
 
 export default function createDispatcher(...args) {
-  return new Dispatcher(...args);
+  const dispatcher = new Dispatcher(...args);
+
+  return {
+    subscribe: ::dispatcher.subscribe,
+    perform: ::dispatcher.perform,
+    hydrate: ::dispatcher.hydrate,
+    dehydrate: ::dispatcher.dehydrate,
+  };
 }
--- a/src/createDispatcher.js
+++ b/src/createDispatcher.js
@@ -1,0 +1,5 @@
+import Dispatcher from './Dispatcher';
+
+export default function createDispatcher(...args) {
+  return new Dispatcher(...args);
+}
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Read **[The Evolution of Flux Frameworks](https://medium.com/@dan_abramov/the-ev
 
 * Hot reloading of everything.
 * A hook for the future devtools to "commit" a state, and replay actions on top of it during hot reload.
-* No `createAction`, `createStores`, `wrapThisStuff`. Your stuff is your stuff.
+* No wrapper calls in your stores and actions. Your stuff is your stuff.
 * Super easy to test things in isolation without mocks.
 * I don't mind action constants. Seriously.
 * Keep Flux lingo. No cursors or observables in core.
@@ -53,23 +53,24 @@ export function decrement() {
 }
 
 // Can also be async if you return a function
-// (wow, much functions, so injectable :doge:)
 export function incrementAsync() {
-  return dispatch => {
+  return perform => {
     setTimeout(() => {
-      dispatch(increment());
+      // Yay! Can invoke sync or async actions with `perform`
+      perform(increment());
     }, 1000);
   };
 }
 
+
 // Could also read state of a store in the callback form
 export function incrementIfOdd() {
-  return (dispatch, read) => {
-    if (read(counterStore) % 2 === 0) {
+  return (perform, { counter }) => {
+    if (counter % 2 === 0) {
       return;
     }
 
-    dispatch(increment());
+    perform(increment());
   };
 }
 ```
@@ -86,15 +87,15 @@ import { INCREMENT_COUNTER, DECREMENT_COUNTER } from '../constants/ActionTypes';
 // and the state shape is up to you: you can use primitives,
 // objects, arrays, or even ImmutableJS objects.
 
-export default function counterStore(counter = 0, action) {
+export default function counter(state = 0, action) {
   // this function returns the new state when an action comes
   switch (action.type) {
   case INCREMENT_COUNTER:
-    return counter + 1;
+    return state + 1;
   case DECREMENT_COUNTER:
-    return counter - 1;
+    return state - 1;
   default:
-    return counter;
+    return state;
   }
 
   // BUT THAT'S A SWITCH STATEMENT!
@@ -135,25 +136,28 @@ export default class Counter {
 #### Smart Components
 
 ```js
-// The smart component may inject actions
-// and observe stores using <Injector />:
+// The smart component may observe stores using `<Connector />`,
+// and bind actions to the dispatcher with `bindActions`.
 
-import React, { Component } from 'react';
-import { Injector } from 'redux';
-import * as CounterActions from './actions/CounterActions';
-import counterStore from './stores/counterStore';
-import Counter from './Counter';
+import React from 'react';
+import { Connector, bindActions } from 'redux';
+import Counter from '../components/Counter';
+import * as CounterActions from '../actions/CounterActions';
 
-export default class CounterContainer {
+function select(state) {
+  return { counter: state.counter };
+}
+
+export default class CounterApp {
   render() {
-    // stores and actions must both be string -> function maps.
-    // props passed to children will combine these actions and state.
     return (
-      <Injector stores={{ counter: stores.counterStore }}
-                 actions={CounterActions}>
-        {/* Yes this is a function as a child. Bear with me. */}
-        {({ state, actions }) => <Counter {...state} {...actions} />}
-      </Injector>
+      <Connector select={select}>
+        {({ counter, dispatcher }) =>
+          /* Yes this is child as a function. */
+          <Counter counter={counter}
+                   {...bindActions(CounterActions, dispatcher)} />
+        }
+      </Connector>
     );
   }
 }
@@ -161,36 +165,23 @@ export default class CounterContainer {
 
 #### Decorators
 
-Don't want to separate dumb and smart components just yet? Use the decorators!  
-They work exactly the same as the container components, but are lowercase:
+The `@connect` decorator lets you create smart components less verbosely:
 
 ```js
-import React, { PropTypes } from 'react';
-import * as CounterActions from './actions/CounterActions';
-import { inject } from 'redux';
-import counterStore from './stores/counterStore';
+import React from 'react';
+import { connect, bindActions } from 'redux';
+import Counter from '../components/Counter';
+import * as CounterActions from '../actions/CounterActions';
 
-@inject({
-  actions: CounterActions,
-  stores: { counter: counterStore }
-})
-export default class Counter {
-  static propTypes = {
-    increment: PropTypes.func.isRequired,
-    decrement: PropTypes.func.isRequired,
-    counter: PropTypes.number.isRequired
-  };
-
+@connect(state => ({
+  counter: state.counter
+}))
+export default class CounterApp {
   render() {
-    const { increment, decrement, counter } = this.props;
+    const { counter, dispatcher } = this.props;
     return (
-      <p>
-        Clicked: {counter} times
-        {' '}
-        <button onClick={increment}>+</button>
-        {' '}
-        <button onClick={decrement}>-</button>
-      </p>
+      <Counter counter={counter}
+               {...bindActions(CounterActions, dispatcher)} />
     );
   }
 }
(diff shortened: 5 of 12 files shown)


examples/containers/App.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { dispatch, composeStores } from 'redux';
+import CounterApp from './CounterApp';
+import TodoApp from './TodoApp';
+import * as stores from '../stores/index';
+
+@dispatch(composeStores(stores))
```
gaearon: `composeStores` will turn a `stores` exported object into a normal Store with `state, action => state` signature that would combine them. It's a good shortcut, and you may replace it with a custom mechanism like a single Immutable Map if you want to. cc @fisherwebdev

examples/containers/CounterApp.js
```diff
@@ -0,0 +1,16 @@
+import React, { Component } from 'react';
+import { Injector } from 'redux';
+import Counter from '../components/Counter';
+import * as CounterActions from '../actions/CounterActions';
+
+export default class CounterApp {
+  render() {
+    return (
+      <Injector actions={CounterActions}>
+        {({ atom: { counter }, actions }) =>
```
gaearon: Every `Injector` is now given the root `atom`. This encourages few `Injector`s at the top. For better performance, just implement `shouldComponentUpdate` on the contained components, or even implement your own finer-grained `Injector`.

examples/stores/counter.js
```diff
@@ -1,12 +1,12 @@
 import { INCREMENT_COUNTER, DECREMENT_COUNTER } from '../constants/ActionTypes';
 
-export default function counterStore(counter = 0, action) {
+export default function counter(state = 0, action) {
```
gaearon: Inspired by @kevinrobinson's tweets, I dropped the damned `Store` suffix. Let's just call it `counter`!

src/Injector.js
```diff
@@ -14,94 +9,50 @@ export default class ReduxInjector extends Component {
...
-          'Expected "%s" in "stores" to be a function. Instead received: %s.',
-          key,
-          stores[key]
-        )
-      );
-    }
-
-    const { wrapActionCreator, observeStores } = this.context.redux;
-    this.actions = mapValues(props.actions, wrapActionCreator);
+  performAction(actionCreator, ...args) {
```
gaearon: You could write a custom `Injector` (should we provide `makeInjector`?) that does something fancy to action creators, like handling promises. Relevant for #45.

src/addons/composeStores.js
```diff
@@ -0,0 +1,8 @@
+export default function composeStores(stores) {
```
gaearon: A higher-order Store! I love it.

src/addons/dispatch.js
```diff
@@ -2,13 +2,13 @@ import React from 'react';
...
-  return DecoratedComponent => class ReduxDispatcherDecorator {
-    static displayName = `ReduxDispatcher(${getDisplayName(DecoratedComponent)})`;
+export default function dispatch(store) {
+  return DecoratedComponent => class DispatcherDecorator {
+    static displayName = `Dispatcher(${getDisplayName(DecoratedComponent)})`;
 
     render() {
       return (
-        <Dispatcher stores={stores}>
+        <Dispatcher store={store}>
```
gaearon: The whole act of hot reloading any Store now is just replacing this single prop.

src/createDispatcher.js
```diff
@@ -1,143 +0,0 @@
...
-    return currentState[getStoreKey(store)];
-  }
-
-  return {
-    wrapActionCreator,
-    observeStores,
-    receiveStores,
-    getStoreKey
-  };
-}
```
gaearon: Good riddance.

src/Dispatcher.js
```diff
@@ -1,9 +1,12 @@
-import { PropTypes } from 'react';
-import createDispatcher from './createDispatcher';
+import { Component, PropTypes } from 'react';
 
-export default class ReduxDispatcher {
+export default function dispatch(store, atom, action) {
+  return store(atom, action);
+}
+
+export default class Dispatcher extends Component {
```
ooflorent: You cannot `export default` twice

gaearon: Oh yeah. Copy-paste typo, worked by miracle.

vslinko:
Happy Birthday!
Right now I don't understand how to make this isomorphic — how to prepare atom before render?

gaearon:
@vslinko 

Thanks!

@acdlite, counting on your help with this :-). I don't have experience with isomorphic apps but Andrew built Flummox so he should know.

It's a good point. I want to merge this after it supports in some form all of the use cases currently PR'd by @vslinko: #41 #39 #33 #32.

src/Injector.js
```diff
@@ -14,94 +9,50 @@ export default class ReduxInjector extends Component {
...
   render() {
-    const { children } = this.props;
-    return children({
-      state: this.state,
-      actions: this.actions
-    });
+    const { children, actions: _actions } = this.props;
+    const { atom } = this.state;
+
+    const actions = Object.keys(_actions).reduce((result, key) => {
```
ooflorent: Lodash's `mapValue` would be more readable.

gaearon: Yeah, might use that. I kinda like the rawness though (I didn't realize I could use `reduce` like this).
I don't have strong opinion here. (Note it's also used in `combineStores`)

src/Dispatcher.js
```diff
@@ -11,17 +14,49 @@ export default class ReduxDispatcher {
...
-    return { redux: this.dispatcher };
+  dispatch(action) {
+    const nextAtom = dispatch(this.props.store, this.atom, action);
+    this.setAtom(nextAtom);
+  }
+
+  setAtom(atom) {
+    this.atom = atom;
+    if (this.state) {
+      this.setState({ atom });
```
ooflorent: _Here_ is my only concern with this rewrite. Each `dispatch` call will trigger a `render` of this component. The developer will need to implement a lot of `shouldComponentUpdate` to not lose any perf.

gaearon: Now that I think of it.. Do we even need to keep it in `state`? I don't think so! `Injector`s subscribe to the updates via `context.subscribe()`.

Now, `Injector`s, too receive the full `atom`, but we can easily implement more efficient / fine-grained `Injector`s in the future that only check their part of the tree.

ravicious: Maybe we could incorporate immutable data here to ease the pain, but then one would also need to make use of some kind of cursors ([like Om](https://github.com/omcljs/om/wiki/Cursors)) or FRP-like reactions ([like re-frame](https://github.com/Day8/re-frame#how-flow-happens-in-reagent)).

gaearon:
@ooflorent I addressed both your comments

gaearon:
So, the things still bugging me:

* Lack of support for isomorphic (progressive, universal, whatever) apps — hoping for help from @acdlite
* Should we call it `atom`? I don't want to scare off people used to Flux terminology who feel things like Om, NuclearJS or Omniscient are above their head.

src/Dispatcher.js
gaearon: Redux already assumes you're never mutating the state, but I don't want it to assume any library. Also, I will avoid any fancy functional concepts that may alienate the users. We *can* keep it simple, so we should.

gaearon:
@ooflorent 

I pushed some perf optimizations: b70739a0f18f3b4d197cf20a75e8a1f5f42bbd17.
It's possible to optimize further (in fact even more!) if we add a `select` prop to `Injector`:

```js
// Before
  render() {
    return (
      <Injector actions={CounterActions}>
        {({ state: { counter }, actions }) =>
          <Counter counter={counter} {...actions} />
        }
      </Injector>
    );
  }
}


// After
function select(state) {
  return state.counter;
}

  render() {
    return (
      <Injector actions={CounterActions}
                select={select}>
        {({ state, actions }) =>
          <Counter counter={state} {...actions} />
        }
      </Injector>
    );
  }
}
```

Note that this also frees you from destructuring inside `Injector` child function.

The `select` prop is just a function that picks the relevant part of the tree. The results of previous and next `select` runs would be compared shallowly, and if they are equal, rendering would be skipped. This is *kinda* like cursors, but without the word “cursor” and string keypath arrays.

What do you say? If you're fine, I'd be happy to merge a PR in this branch implementing `select`.

gaearon:
Note: I renamed `atom` to `state`. No functional lingo for now. :-)

fibric:
Happy birthday!

You must be truly coming from future or different universe...
After renaming atom to state, its easier to understand. At least I think I've got it ;-)

gaearon:
@ooflorent 

I added a `select` prop, this should solve your perf issues: 18b0695e7b6d067d29c31b65a777e29374f6aaf7

This lets you reach into the global state as deep as you want and return an object that will be shallowly compared on each change. If the object is shallowly equal, `shouldComponentUpdate` bails out. This is actually a potentially *more* performant solution than what we had in Redux before.

By default, `select` is just an identity function.

src/Dispatcher.js
ravicious: I think these concepts are neither fancy nor complicated, but I understand your reasoning. I was once scared by those ideas, too; if avoiding the functional lingo means the library is more approachable to more people, then it's okay I guess. :+1: 

Too bad we have this FP-stigma, though. JS community could learn so much from FP (as it already did).

gaearon: It’s not exactly FP-stigma.. I also love this ideas. But to bring them to JS community means creating APIs that explain these concepts instead of using strict terminology in README documents.

src/Injector.js
```diff
@@ -1,107 +1,79 @@
...
 
-export default class ReduxInjector extends Component {
+export default class Injector extends Component {
   static contextTypes = {
     redux: PropTypes.object.isRequired
   };
 
   static propTypes = {
     children: PropTypes.func.isRequired,
+    select: PropTypes.func.isRequired,
```
ooflorent: What about:

```js
select: oneOfType([ string, func ]).isRequired
```

And then:

```js
import get from 'lodash/object/get'
import isFunction from 'lodash/lang/isFunction'

const slice = isFunction(this.props.select)
  ? this.props.select(atom)
  : get(atom, this.props.select)
```

gaearon: I thought about that.. But then, people are going to want arrays of string keys for keyPaths and we're back to cursors :-). I think functions (or lambdas) work great here.

gaearon:
So here's what I got API-wise:

### Decorator version

```js
import React from 'react';
import { inject } from 'redux';
import Counter from '../components/Counter';
import * as CounterActions from '../actions/CounterActions';

@inject({
  actions: CounterActions,
  select: state => ({
    counter: state.counter
  })
})
  render() {
    return (
      <Counter {...this.props} />
    );
  }
}
```

### Component version

```js
import React from 'react';
import { Injector } from 'redux';
import AddTodo from '../components/AddTodo';
import TodoList from '../components/TodoList';
import * as TodoActions from '../actions/TodoActions';

export default class TodoApp {
  render() {
    return (
      <Injector actions={TodoActions}
                select={state => state.todos}>
        {this.renderChild}
      </Injector>
    );
  }

  renderChild({ state, actions }) {
    return (
      <div>
        <AddTodo {...actions} />
        <TodoList todos={state} {...actions} />
      </div>
    );
  }
}
```

Is the state/action merging too implicit? Can we make it better now that we have `select`?
@threepointone

gaearon:
Since the `actions` references don't change, I wonder if it's too much to put them in `select` too? Then `select` can become the only parameter. (See also #45)

```js
import React from 'react';
import { inject } from 'redux';
import Counter from '../components/Counter';
import * as CounterActions from '../actions/CounterActions';

@inject((state, wrap) => ({
  counter: state.counter,
  ...wrap(CounterActions)
}))
```

vslinko:
:+1: for select

gaearon:
I don't *quite* like this `wrap` thing. It would need to keep some kind of memoized cache of bound action creators, or otherwise it would be really hard to implement a sensible `shouldComponentUpdate` for the components.

Maybe do it like this?

```js
import React from 'react';
import { inject, perform } from 'redux';
import Counter from '../components/Counter';
import * as CounterActions from '../actions/CounterActions';

@inject(state => ({
  counter: state.counter
}))
  const { dispatch, counter } = this.props;
  render() {
    return (
      <Counter counter={this.props.counter}
               {...perform(dispatch, CounterActions)} />
    );
  }
}
```

```js
import React from 'react';
import { Injector, perform } from 'redux';
import AddTodo from '../components/AddTodo';
import TodoList from '../components/TodoList';
import * as TodoActions from '../actions/TodoActions';

export default class TodoApp {
  render() {
    return (
      <Injector select={state => state.todos}>
        {this.renderChild}
      </Injector>
    );
  }

  renderChild({ state, dispatch }) {
    const actions = perform(dispatch, TodoActions);
    return (
      <div>
        <AddTodo {...actions} />
        <TodoList todos={state} {...actions} />
      </div>
    );
  }
}
```

This would also fix #45.

The difference with the [previous suggestion](https://github.com/gaearon/redux/pull/46#issuecomment-109573133) is that, because `perform` is called *inside* the container component, this let us have the efficient `shouldComponentUpdate` without any memoizing at the `Injector` level. You are also free to pass `dispatch` down as is and call action creators from the leaf components.

`perform(dispatch, actionCreator)` would work both with a single function and with an object map of functions, and would wrap the action creator.

vslinko:
Since you are extracted `Dispatcher` then `redux` is isomorphic now.
Are you sure Dispatcher should be a class? I prefer `createDispatcher`

vslinko:
> Is this sufficient for (de)hydration?

Yes, thank you.

gaearon:
I can make it a factory again, don't have a strong opinion here.

vslinko:
I'll be glad.

gaearon:
Okay, so here is new `select` API. I like it.

#### Decorator version

```js
import React from 'react';
import { connect, bindActions } from 'redux';
import Counter from '../components/Counter';
import * as CounterActions from '../actions/CounterActions';

@connect(state => ({
  counter: state.counter
}))
  render() {
    const { counter, dispatcher } = this.props;
    return (
      <Counter counter={counter}
               {...bindActions(CounterActions, dispatcher)} />
    );
  }
}
```

#### Component version

```js
import React from 'react';
import { bindActions, Connector } from 'redux';
import AddTodo from '../components/AddTodo';
import TodoList from '../components/TodoList';
import * as TodoActions from '../actions/TodoActions';

export default class TodoApp {
  render() {
    return (
      <Connector select={state => ({ todos: state.todos })}>
        {this.renderChild}
      </Connector>
    );
  }

  renderChild({ todos, dispatcher }) {
    const actions = bindActions(TodoActions, dispatcher);
    return (
      <div>
        <AddTodo {...actions} />
        <TodoList todos={todos} {...actions} />
      </div>
    );
  }
}
```

Note that `bindActions` is a simpler helper that could be written in userland:

```js
import mapValues from 'lodash/object/mapValues';

export default function bindActions(actionCreators, dispatcher) {
  return mapValues(actionCreators, actionCreator =>
    (...args) => dispatcher.perform(actionCreator, ...args)
  );
}
```

In fact `perform` is not that special either:

```js
  perform(actionCreator, ...args) {
    const action = actionCreator(...args);

    return typeof action === 'function'
      ? action(this.dispatch, this.atom)
      : this.dispatch(action);
  }
```

If you want, you can write a custom `bindActions`, and/or a custom `perform`, and use that.
This addresses #45.

@vslinko Can you help me here? What do I need to do to address your concerns from https://github.com/gaearon/redux/issues/48#issuecomment-109578338?

src/Dispatcher.js
```diff
@@ -1,31 +1,56 @@
...
+    return () => {
+      const index = this.subscriptions.indexOf(listener);
+      this.subscriptions.splice(index, 1);
+    };
   }
 
-  render() {
-    const { children } = this.props;
-    return children();
+  emitChange() {
```
acdlite: What's the advantage of emitting changes manually like this instead of using `setState()` on the root + context in the provider? Is it to avoid a context bug?

acdlite:
This looks really great. I dig the `select` API — that's the perfect verb, I think, to describe what that does.

The only thing that's stopping this from being 100% isomorphic is when you do `createDispatcher()` outside the lifecycle of the root component.

vslinko:
> Can you help me here? What do I need to do to address your concerns from #48 (comment)?

`perform` should pass itself as first argument into action instead of `dispatch`:

```js
 perform(actionCreator, ...args) {
    const action = actionCreator(...args);

    return typeof action === 'function'
      ? action(this.perform, this.atom)
      : this.dispatch(action);
  }
```

src/Dispatcher.js
gaearon: Yes. It works around the `shouldComponentUpdate` context bug. Also context is not meant for passing often changing data due to other performance implications (I believe @sebmarkbage warned me about this). By using a manual system, we're only using the context for creating a wormhole, but allow fine-grained subscriptions at any level of the tree.

gaearon: There are also other problems with using just `setState`. For example, the state passed to action creator might be stale because there are pending operation that have not yet been rendered. We are able to sidestep this problem completely by maintaining the atom manually.

vslinko:
Missed that perform shouldn't call actionCreator itself, it already takes action:

```js
 perform(action, ...args) {
    return typeof action === 'function'
      ? action(this.perform, this.atom)
      : this.dispatch(action);
  }
```

gaearon:
@vslinko I need to go now, can you PR your suggested changes to action creator calls relative to this branch?

vslinko:
@gaearon I'll try

gaearon:
>The only thing that's stopping this from being 100% isomorphic is when you do createDispatcher() outside the lifecycle of the root component.

Can you elaborate on this? What do I need to change?

gaearon:
Anything else left re: isomoprhic apps? @acdlite @vslinko

vslinko:
> Anything else left re: isomoprhic apps?

No, It's ready.

acdlite:
As is, the isomorphism story is already way better than any other Flux library, but if you want out-of-the-box isomorphism with no extra set-up — the same way hot reloading "just works" — then you'd have to move `createDispatcher()` inside the component lifecycle of the root component and interact with it via props.

gaearon:
Unfortunately I don't have expertise here so can't comment. Can you give me an idea of how that would be used on the server side? I know how (de)hydrate API can be used, but I'm not sure what you're suggesting. (on server, there is no mounting so no prop lifecycle, is there?)

acdlite:
It's not a huge deal, but currently on the server the user would need to make certain they do `createDispatcher()` on every request.

```js
app.get("/", function(req, res) {
  const dispatcher = createDispatcher(store);

  res.send(
    React.renderToString(<Provider dispatcher={dispatcher} />)
  );
});
```

Not bad at all, but someone could accidentally making the mistake of calling `createDispatcher()` outside of the request handler.

I think it would nice to have this option

```js
app.get("/", function(req, res) {
  res.send(
    React.renderToString(<Provider store={store} />)
  );
});
```

and call `createDispatcher()` internally. You'll still probably want to provide the first option so you can easily access dispatcher methods for dehydration etc., but I like how this option "just works."

gaearon:
How would you solve this API-wise while still providing the (de)hydrate API?
I'd prefer `Provider` to accept a single thing (either dispatcher or store) for API consistency.

vslinko:
:-1: for `<Provider store={store} />`

acdlite:
For extra APIs like dehydrate, then yes, passing the dispatcher is the way to go. I personally don't think supporting both APIs is that bad (it's only two props total!), and I like the simplicity of just passing a store, but it's not a huge deal to me.

acdlite:
Also, my perspective is admittedly a bit skewed because in my gist, literally the only stateful part is the atom, whereas Redux also keeps track of subscriptions (for good reasons) and needs to provide a direct interface for dehydration, etc.

gaearon:
This is out as 0.8.0.
*Huge* thanks to everyone, especially @acdlite.

PR #53, opened 2015-06-06
Connect decorator should pass props into select function

vslinko:


vslinko:
Connector component shouldn't pass props because user have direct access to props:

```js
// The smart component may observe stores using `<Connector />`,
// and bind actions to the dispatcher with `bindActions`.

import React from 'react';
import { Connector, bindActions } from 'redux';
import Counter from '../components/Counter';
import * as CounterActions from '../actions/CounterActions';

// You can optionally specify `select` for finer-grained subscriptions
// and retrieval. Only when the return value is shallowly different,
// will the child component be updated.
function select(state, props) {
  return { counter: state.counter };
}

  render() {
    return (
      <Connector select={state => select(state, this.props)}>
        {({ counter, dispatcher }) =>
          /* Yes this is child as a function. */
          <Counter counter={counter}
                   {...bindActions(CounterActions, dispatcher)} />
        }
      </Connector>
    );
  }
}
```

gaearon:
This sounds right! Should you also remove {...props} inside renderChild?

vslinko:
Done

gaearon:
Thank you!

PR #55, opened 2015-06-07
Middleware all the things

acdlite:
# Composable middleware

Attempts to address #6.

Currently, the dispatcher does not publicly expose its `dispatch()` method. Instead, it provides a `perform()` method that serves the same purpose, except if the action is a function it calls the function and passes `dispatch` as a callback. This is a clever way of supporting async action creators. However, I believe this feature does not belong in the dispatcher, especially not as the sole interface for dispatching actions. Conceptually, this makes more sense implemented as middleware.

I propose that we remove the `perform()` method and export `dispatch()` in its place. `dispatch()` works the same way as before: it passes an action through the store to calculate the next atom, then emits a change.

To continue supporting async actions, and to provide an extensibility point for external plugins and tools, we can provide some common action middleware, a helper for composing middleware, and documentation for  how extension authors can easily create their own.

An action middleware is simply a higher-order function. It returns a function with the same interface as `dispatch()`. For example, here's how you'd implement the `perform()` method described above:

```js
// `next()` is either the next middleware in the chain or the `dispatch()` method
function callbackMiddleware(next) {
  return action =>
    typeof action === 'function'
      ? action(next)
      : next(action);
}

const perform = callbackMiddleware(dispatch);
```

Now we can use `perform()` just like before:

```js
perform(action);
```

Of course, things really get interesting when you compose multiple middleware together. Here's another middleware that adds basic support for Flummox-like async action creators:

```js
function promiseMiddleware(next) {
  return action =>
    action && typeof action.then === 'function'
      ? action.then(next)
      : next(action)
}
```

Combine this with the callback middleware we defined above:

```js
const perform = callbackMiddleware(promiseMiddleware(dispatch));
```

This is very flexible. You can mix and match middleware, and even change middleware on a per-action basis. We should provide an interface to specify global middleware that gets applied to every action.

Like we do with `composeStores()`, we can also provide a `composeMiddleware()` helper as a convenience:

```js
function composeMiddleware(...middlewares) {
  return next =>
    middlewares.reduceRight((_next, middleware) => middleware(_next), next);
}
```

Here's the above example, rewritten to use our "higher-order middleware" (hehe):

```js
const middleware = composeMiddleware(promiseMiddleware, callbackMiddleware);
const perform = middleware(dispatch);
```

@gaearon has a good example of how this compose function would work with interceptors (which are just functions that return middleware): https://github.com/gaearon/redux/issues/6

Yay!

In summary:

- [X] Remove `perform()` method and export `dispatch()` instead.
- [X] Rewrite `bindActions()` so the final parameter is `dispatch` rather than the dispatcher. (Minor nitpick: Also rename it to `bindActionCreators` for consistency. I made this mistake with Flummox and came to regret it quickly).
- [X] Add `callbackMiddlware()` for async functionality previously provided by `perform()`. There's no `atom` parameter, because I don't think it makes sense: you can just pass the atom directly to the action creator. I rewrote the Counter action creator `incrementIfOdd` as an example. Also, I wasn't sure if this should go on the main export or not, so for now I've put it at `redux/middleware/callback`.
- [X] Add `composeMiddleware()` util for easily composing multiple action middlewares.
- [X] Rewrite examples, including `incrementIfOdd` and `incrementAsync` action creators. `incrementAsync` illustrates how to implement basic promise middleware.
- [ ] Provide an API for global middleware.
- [ ] Update README (if/when this is ready to be merged)?

Thoughts?

### DIFF
--- a/examples/actions/CounterActions.js
+++ b/examples/actions/CounterActions.js
@@ -17,11 +17,12 @@ export function incrementIfOdd(counter) {
 }
 
 export function incrementAsync() {
-  return perform => {
-    setTimeout(() => {
-      perform(increment());
-    }, 1000);
-  };
+  return new Promise(resolve =>
+    setTimeout(
+      () => resolve(increment()),
+      1000
+    )
+  );
 }
 
 export function decrement() {
--- a/examples/actions/CounterActions.js
+++ b/examples/actions/CounterActions.js
@@ -6,8 +6,8 @@ export function increment() {
   };
 }
 
-export function incrementIfOdd() {
-  return (perform, { counter }) => {
+export function incrementIfOdd(counter) {
+  return perform => {
     if (counter % 2 === 0) {
       return;
     }
--- a/examples/components/Counter.js
+++ b/examples/components/Counter.js
@@ -5,11 +5,17 @@ export default class Counter {
     increment: PropTypes.func.isRequired,
     decrement: PropTypes.func.isRequired,
     incrementIfOdd: PropTypes.func.isRequired,
+    incrementAsync: PropTypes.func.isRequired,
     counter: PropTypes.number.isRequired
   };
 
   render() {
-    const { increment, decrement, incrementIfOdd, counter } = this.props;
+    const {
+      increment,
+      decrement,
+      incrementIfOdd,
+      incrementAsync,
+      counter } = this.props;
     return (
       <p>
         Clicked: {counter} times
@@ -19,6 +25,7 @@ export default class Counter {
         <button onClick={decrement}>-</button>
         {' '}
         <button onClick={incrementIfOdd}>Increment if odd</button>
+        <button onClick={incrementAsync}>Increment async</button>
       </p>
     );
   }
--- a/examples/components/Counter.js
+++ b/examples/components/Counter.js
@@ -4,11 +4,12 @@ export default class Counter {
   static propTypes = {
     increment: PropTypes.func.isRequired,
     decrement: PropTypes.func.isRequired,
+    incrementIfOdd: PropTypes.func.isRequired,
     counter: PropTypes.number.isRequired
   };
 
   render() {
-    const { increment, decrement, counter } = this.props;
+    const { increment, decrement, incrementIfOdd, counter } = this.props;
     return (
       <p>
         Clicked: {counter} times
@@ -16,6 +17,8 @@ export default class Counter {
         <button onClick={increment}>+</button>
         {' '}
         <button onClick={decrement}>-</button>
+        {' '}
+        <button onClick={incrementIfOdd}>Increment if odd</button>
       </p>
     );
   }
--- a/examples/components/Transactions.js
+++ b/examples/components/Transactions.js
@@ -11,15 +11,13 @@ export default class Transactions {
 
   render() {
     const { status, commit, begin, rollback } = this.props;
+
     return (
       <p>
-        <DOMify value={status} />
-        {' '}
         <button onClick={begin}>Begin</button>
-        {' '}
         <button onClick={commit}>Commit</button>
-        {' '}
         <button onClick={rollback}>Rollback</button>
+        <DOMify value={status} />
       </p>
     );
   }
--- a/examples/components/Transactions.js
+++ b/examples/components/Transactions.js
@@ -1,0 +1,26 @@
+import React, { PropTypes } from 'react';
+import DOMify from 'react-domify';
+
+export default class Transactions {
+  static propTypes = {
+    status: PropTypes.object.isRequired,
+    commit: PropTypes.func.isRequired,
+    begin: PropTypes.func.isRequired,
+    rollback: PropTypes.func.isRequired
+  };
+
+  render() {
+    const { status, commit, begin, rollback } = this.props;
+    return (
+      <p>
+        <DOMify value={status} />
+        {' '}
+        <button onClick={begin}>Begin</button>
+        {' '}
+        <button onClick={commit}>Commit</button>
+        {' '}
+        <button onClick={rollback}>Rollback</button>
+      </p>
+    );
+  }
+}
--- a/examples/containers/App.js
+++ b/examples/containers/App.js
@@ -1,13 +1,27 @@
 import React from 'react';
-import { createDispatcher, Provider, composeStores, reduceStore } from 'redux';
+import { createDispatcher, Provider, composeStores, reduceStore, compose } from 'redux';
 import CounterApp from './CounterApp';
 import TodoApp from './TodoApp';
 import * as stores from '../stores/index';
 import createTransactor from '../middleware/createTransactor';
+import callbackMiddleware from 'redux/middleware/callback';
+
+// Naive implementation of promise middleware
+// Leave full implementation to userland
+function promiseMiddleware(next) {
+  return action =>
+    action && typeof action.then === 'function'
+      ? action.then(next)
+      : next(action);
+}
 
 const store = composeStores(stores);
 const transactor = createTransactor();
-const dispatcher = createDispatcher(getAtom => transactor(getAtom, store));
+const dispatcher = createDispatcher(getAtom => compose(
+  promiseMiddleware,
+  callbackMiddleware,
+  transactor(getAtom, store)
+));
 
 global.transactor = transactor;
 
--- a/examples/containers/App.js
+++ b/examples/containers/App.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { createDispatcher, Provider, composeStores, reduceStore, compose } from 'redux';
+import { createDispatcher, Provider, composeStores, compose } from 'redux';
 import CounterApp from './CounterApp';
 import TodoApp from './TodoApp';
 import * as stores from '../stores/index';
--- a/examples/containers/App.js
+++ b/examples/containers/App.js
@@ -18,18 +18,13 @@ function promiseMiddleware(next) {
 
 const store = composeStores(stores);
 const transactor = createTransactor();
-// const dispatcher = createDispatcher(compose(
-//   promiseMiddleware,
-//   callbackMiddleware,
-//   transactor(store)
-// ));
+
 const dispatcher = createDispatcher({
-  middleware: compose(promiseMiddleware, callbackMiddleware),
-  reducer: transactor(store)
+  store,
+  reducer: transactor,
+  middleware: [ promiseMiddleware, callbackMiddleware ]
 });
 
-
-global.transactor = transactor;
 
 export default class App {
   render() {
--- a/examples/containers/App.js
+++ b/examples/containers/App.js
@@ -3,9 +3,13 @@ import { createDispatcher, Provider, composeStores, reduceStore } from 'redux';
 import CounterApp from './CounterApp';
 import TodoApp from './TodoApp';
 import * as stores from '../stores/index';
+import createTransactor from '../middleware/createTransactor';
 
 const store = composeStores(stores);
-const dispatcher = createDispatcher(getAtom => reduceStore(getAtom, store));
+const transactor = createTransactor();
+const dispatcher = createDispatcher(getAtom => transactor(getAtom, store));
+
+global.transactor = transactor;
 
 export default class App {
   render() {
--- a/examples/containers/App.js
+++ b/examples/containers/App.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { createDispatcher, Provider, composeStores, compose } from 'redux';
 import CounterApp from './CounterApp';
 import TodoApp from './TodoApp';
+import TransactionsApp from './TransactionsApp';
 import * as stores from '../stores/index';
 import createTransactor from '../middleware/createTransactor';
 import callbackMiddleware from 'redux/middleware/callback';
@@ -17,11 +18,16 @@ function promiseMiddleware(next) {
 
 const store = composeStores(stores);
 const transactor = createTransactor();
-const dispatcher = createDispatcher(getAtom => compose(
-  promiseMiddleware,
-  callbackMiddleware,
-  transactor(getAtom, store)
-));
+// const dispatcher = createDispatcher(compose(
+//   promiseMiddleware,
+//   callbackMiddleware,
+//   transactor(store)
+// ));
+const dispatcher = createDispatcher({
+  middleware: compose(promiseMiddleware, callbackMiddleware),
+  reducer: transactor(store)
+});
+
 
 global.transactor = transactor;
 
@@ -33,6 +39,7 @@ export default class App {
           <div>
             <CounterApp />
             <TodoApp />
+            <TransactionsApp transactor={transactor} />
           </div>
         }
       </Provider>
--- a/examples/containers/App.js
+++ b/examples/containers/App.js
@@ -1,10 +1,11 @@
 import React from 'react';
-import { createDispatcher, Provider, composeStores } from 'redux';
+import { createDispatcher, Provider, composeStores, reduceStore } from 'redux';
 import CounterApp from './CounterApp';
 import TodoApp from './TodoApp';
 import * as stores from '../stores/index';
 
-const dispatcher = createDispatcher(composeStores(stores));
+const store = composeStores(stores);
+const dispatcher = createDispatcher(getAtom => reduceStore(getAtom, store));
 
 export default class App {
   render() {
--- a/examples/containers/CounterApp.js
+++ b/examples/containers/CounterApp.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { connect, bindActionCreators } from 'redux';
 import Counter from '../components/Counter';
 import * as CounterActions from '../actions/CounterActions';
+import callbackMiddleware from 'redux/middleware/callback';
 
 @connect(state => ({
   counter: state.counter
@@ -9,9 +10,12 @@ import * as CounterActions from '../actions/CounterActions';
 export default class CounterApp {
   render() {
     const { counter, dispatcher } = this.props;
+    const dispatch = callbackMiddleware(dispatcher.dispatch);
+    const actionCreators = bindActionCreators(CounterActions, dispatch);
     return (
       <Counter counter={counter}
-               {...bindActionCreators(CounterActions, dispatcher.dispatch)} />
+               {...actionCreators}
+               incrementIfOdd={() => actionCreators.incrementIfOdd(counter)} />
     );
   }
 }
--- a/examples/containers/CounterApp.js
+++ b/examples/containers/CounterApp.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { connect, bindActions } from 'redux';
+import { connect, bindActionCreators } from 'redux';
 import Counter from '../components/Counter';
 import * as CounterActions from '../actions/CounterActions';
 
@@ -11,7 +11,7 @@ export default class CounterApp {
     const { counter, dispatcher } = this.props;
     return (
       <Counter counter={counter}
-               {...bindActions(CounterActions, dispatcher)} />
+               {...bindActionCreators(CounterActions, dispatcher.dispatch)} />
     );
   }
 }
--- a/examples/containers/CounterApp.js
+++ b/examples/containers/CounterApp.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { connect, bindActionCreators, composeMiddleware } from 'redux';
+import { connect, bindActionCreators, compose } from 'redux';
 import Counter from '../components/Counter';
 import * as CounterActions from '../actions/CounterActions';
 import callbackMiddleware from 'redux/middleware/callback';
@@ -13,7 +13,7 @@ function promiseMiddleware(next) {
       : next(action);
 }
 
-const middleware = composeMiddleware(callbackMiddleware, promiseMiddleware);
+const middleware = compose(callbackMiddleware, promiseMiddleware);
 
 @connect(state => ({
   counter: state.counter
--- a/examples/containers/CounterApp.js
+++ b/examples/containers/CounterApp.js
@@ -1,8 +1,19 @@
 import React from 'react';
-import { connect, bindActionCreators } from 'redux';
+import { connect, bindActionCreators, composeMiddleware } from 'redux';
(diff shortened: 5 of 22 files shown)


acdlite:
I'll wait for further discussion before starting on a global middleware API. (Perhaps dispatcher-level is the more accurate descriptor.)

@gaearon

acdlite:
I *think* this would address all the concerns in #6.

dariocravero:
Wow, this is neat :)! :+1:

acdlite:
Just in case it isn't clear, this is fundamentally the same concept as the interceptors described in @gaearon. I'm using slightly different terminology so that we can distinguish between stuff like the callback middleware and side-effect-y middleware like the logger interceptor. Comparing to Dan's example:

```
function logger(label) {
  return sink => payload => {
    console.log(label, payload);
    sink(payload);
  };
}
```

- `sink` === `next`
- `payload` === `action`

acdlite:
Perhaps the api for global middleware/interceptors could be as simple as this:

```
createDispatcher({
  store,
  dispatch: dispatch => compose(
    callbackMiddleware,
    promiseMiddleware,
    logger('dispatchingAction'),
    recordAction,
    dispatch, // the actual dispatch to the stores
    emitChange // perhaps?
  )
})
```

acdlite:
Now that I think about it we probably should settle on a single term, because they're the same thing. "Interceptor" sounds awesome but "middleware" is clearer, I think — anyone who has used Express, Rails, etc. already knows what it means.

acdlite:
Oh boy, we could even model the store as middleware! Then we don't even need a `store` param!

```js
createDispatcher(
    callbackMiddleware,
    promiseMiddleware,
    logger('Dispatching action...'),
    createStore(store),
    logger('Finished dispatching action.')
    logger('Emitting new atom...'),
    emitChange,
    logger('Done emitting.')
)
```

Um, yes. I like this. Too crazy?

acdlite:
Well damn, now `dispatcher` feels like the totally wrong word for this... It's the thing that holds the state and emits changes... so shouldn't we call it a *store*?? Then our store function just becomes... a mapping function?!?!

```js
createStore(emitChange => [
  callbackMiddleware,
  promiseMiddleware,
  logger('Dispatching action...'),
  map(store), // wtf should we call this?
  logger('Finished dispatching action.')
  logger('Emitting new atom...'),
  emitChange,
  logger('Done emitting.')
])
```

Does anyone else like this? Have I taken up residence in Crazytown?

acdlite:
Actually, a simple `map()` won't do because we also need reference to the atom. Updated:

```js
createStore(({ emitChange, reduceState }) => compose(
  // ...middleware...
  reduceState(reducer), // "store" is now "reducer"
  // ...more middleware...
));
```

where `reduceState()` looks like this:

```js
function reduceState(reducer) {
  return next => action => next(reducer(getAtom(), action))
}
```

`getAtom()` is a bound reference to `store.getAtom()`

vslinko:
In general it's looks great! It could be a 1.0.0 release.

vslinko:
>  Rewrite examples, including incrementIfOdd and incrementAsync action creators. incrementAsync illustrates how to implement basic promise middleware.

I'm not sure about that. Action should to be able to dispatch several actions. There is my example:
```js

export function submit() {
  return async (perform, state) => {
    perform(setDisabled(true))
    perform(setError(null))

    try {
      await perform(authorize(
        state.signIn.data.username,
        state.signIn.data.password
      ))
    } catch (error) {
      perform(setError(error))
    } finally {
      perform(setDisabled(false))
    }
  }
}
```

vslinko:
I see you create callbackMiddleware to pass `next` into action.
It works when your action is pretty simple.
But it ins't solves my previous example because `perform` from master allows to call async actions by chain.
Here is my full example:

```js
export function submit() {
  return async (perform, state) => {
    perform(setDisabled(true))
    perform(setError(null))

    try {
      await perform(authorize(
        state.signIn.data.username,
        state.signIn.data.password
      ))
    } catch (error) {
      perform(setError(error))
    } finally {
      perform(setDisabled(false))
    }
  }
}


export function authorize(username, password) {
  return async perform => {
    await new Promise(resolve => setTimeout(resolve, 1000))

    if (Math.random() > 0.5) {
      throw new Error('Invalid credentials')
    }

    const user = {
      type: 'users',
      id: '1',
      attributes: {
        username,
      },
    }

    const userLink = {
      type: user.type,
      id: user.id,
    }

    perform(mergeResources([user]))
    perform(setCurrentUser(userLink))

    return user
  }
}
```

rpominov:
How do we implement "transactions" middleware with it? Such middleware need to be able to:

 - [ ] save current state on start of transaction
 - [x] log actions
 - [ ] replay several actions at arbitrary time
 - [ ] replace current state directly (without dispatching actions)

vslinko:
I could create PR for master branch with example, if you want to fulfil this usage pattern.

gaearon:
Interesting! I wonder if this and the transaction thing are two different extension points, or the same one. Not sure yet.

Another idea I've been toying with is "shadow Flux" under the hood. Imagine

gaearon:
Oh. GitHub mobile is ridiculous with its comment button. :-(

gaearon:
Anyway. I'll tell you about this "shadow flux" later tomorrow :-)

gaearon:
https://github.com/gaearon/redux/issues/6#issuecomment-109768420

acdlite:
Okay, I just pushed a basic transaction middleware using the API I described above (not sure if this should be the final API, but I'm using for now because it cleanly expresses what's going on at a conceptual level).

Here's what it looks like:

```js
const transactor = createTransactor();
const dispatcher = createDispatcher(getAtom => compose(
  promiseMiddleware,
  callbackMiddleware,
  transactor(getAtom, store)
));
```

As you can see, the transactor middleware composes with the async middleware discussed above. `transactor` has methods `begin()`, `commit()`, `rollback()`, and `getStatus()`. This is just a proof of concept — additional features like time travel would be relatively straightforward to implement from there (it's more a UX problem than a technical problem at this point).

@gaearon Any red flags? I don't like the `getAtom()` part, but it occurs to me that if we enforce that the store/transactor middleware comes last (or that any middleware that follows should not transform the state, only perform side effects), then we shouldn't need `getAtom()`, because the most recent atom is just the last atom that was passed through the final middleware.

acdlite:
More explanation of the previous comment:

With this api, `createDispatcher()` accepts a single middleware. At the minimum level, you need a reducer middleware where you apply the store. (This is point 4 from @rpominov's [comment](https://github.com/gaearon/redux/issues/6#issuecomment-109787788):

```js
const dispatcher = createDispatcher(getAtom => reduceActions(getAtom, store));
```

The reducer is special because that's where the true Flux cycle occurs. The basic reducer middleware provided by `reduceActions()` simply passes the action through the store like normal. The `transactor()` middleware works by taking the place of `reduceActions()`:

```js
const dispatcher = createDispatcher(getAtom => transactor(getAtom, store));
``

Again, we can probably dispense with `getAtom()` by enforcing that the reducer must come last in the chain before `setAtom()` is called by the dispatcher. From an API perspective, it may be best to have the user specify the reducer separately from the other types of middleware, even if internally they're all just composed together.

**Comment by rpominov** [comment]
I like this. Only thing that troubles me is that we can have only one "powerful" middleware, that can do stuff like transactions. [My proposal](https://github.com/rpominov/fluce/issues/4#issuecomment-107402900) should allow to use multiple "powerful" middlewares. But I'm not sure we __need__ multiple  "powerful" middlewares, and that they __will work nice__ with each other (or at least could work nice). In this case my proposal looks rather overcomplicated, and this or what @gaearon proposes is better.

**Comment by acdlite** [comment]
@rpominov Still experimenting, but I believe we can compose "reducer" middleware as well.

**Comment by gaearon** [comment]
I agree that it doesn't make a lot of sense to compose powerful middleware. Time travel and transactions can't play nicely together if they don't know about each other.

I'll look tomorrow again at all proposals. I'd rather focus on a single extension point that allows free experimentation outside the core.

**Comment by gaearon** [comment]
@acdlite @rpominov @vslinko 

First of all, I'm humbled by all the code, suggestions & discussion here. Keep it coming!

At this point I think we see that there are many different ways to implement middleware, and it's hard to define what  middleware actually is until we make some. For this reason I want to make the extension point as small as possible, but as flexible as possible too. It's best if that would be a single method (no helpers), so that we can experiment with different ways of implementing the actual middleware without touching the core or changing the extension point.

Let's see if we agree on the following points:

**(When I say “middleware”, I mean “single function”, not “composable middleware”. This single function may be *implemented* as something composable inside, but just like Stores, at the very end, middleware is a function.)**

- The middleware receive actions from `dispatch` and it decides how they translate into `atom` changes.
- “Run reducers over the current atom and set the result to it” logic *is* the default middleware.
- We don't need to include extra composition helpers in the first release of middleware support because we're not sure of the best ways to compose it yet, and it's easy to do in userland.
- It's fine that we provide powerful mutative API like `setAtom` to the middleware “entry point” because it's always possible to implement composable APIs with granular control on top of it, as demonstrated by this PR.
- The internal state of the middleware needs to be “transferable” to the “next instance” of it so that hot reloading works. For example, transactions “remember” the pending actions, so if the middleware is reinitialized during hot reload (likely, because `store` will change), this internal state needs to be rehydrated into the new middleware.
- When somebody calls `setAtom`, the middleware is re-initialized. Otherwise, if it keeps any state inside, it may become stale.

If these are all true, it seems to me that the middleware does not need `getAtom`: it receives the initial atom and it is the only thing that emits next atoms. It always knows the last atom. If somebody calls `setAtom` imperatively (rare?), the middleware is just reinitialized with the new initial atom.

**What middleware signature satisfies this conditions?**

I don't mind using Observables (or at least Observable signatures, not Rx itself) *inside* the lib if it makes things clearer.

Can we describe the middleware as `(initialAtom: Atom, actions: Observable<Action>) => Observable<Atom>` or something similar? How would we (re)hydrate the middleware's state between hot reloads? Should we change `initialAtom` to `initialState` and let middleware also provide `getAtom(middlewareState)`?

**Comment by gaearon** [comment]
How about

```
type MiddlewareState = any
type Atom = any

type Middleware = {
  map(state: MiddlewareState): Atom
  run(initialState: MiddlewareState, actions: Observable<Action>): Observable<MiddlewareState>
}
```

**Comment by vslinko** [comment]
As I understood middleware is store that have private state and it could map its private state to atom.
Am I right?

**Comment by gaearon** [comment]
Not really a store.

(I think) it could take the initial state (`undefined` on the initial run) and an Observable stream of Actions. It returns an Observable of states. It doesn't matter what “state” truly is. This depends on the middleware's implementation. In the simplest case, the state is just the new Atom, so the middleware is like `(initialAtom, Observable<Action>) => Observable<Atom>`. However, for stateful middleware, it is necessary to have some way to rehydrate this state. That's why I added `map` function that “extracts” the state atom from the middleware's “opaque” state object.

I *think* this satisfies all proposals both here and in #6 because it looks to me like the broadest solution possible, with the least set of assumptions about how middleware functions inside. I'd love to hear your thoughts on this!

**Comment by vslinko** [comment]
Sorry, I haven't any comments because I'm satisfied by current API.
You can do anything but save `perform` method :smile:

**Comment by acdlite** [comment]
Alright, I got hot reload *almost* working with the transactor... the UI updates properly, but the buttons stop working until you do a full page refresh. I'm not familiar with the mechanics of how hot reloading works. There are few other remaining issues, but it mostly works. I'll give a look again tomorrow.

Anyway, here's a previous of what I have so far:

**Comment by acdlite** [comment]
I will say that the more I mess around with this, the more I like @gaearon's concept of "lifted" or "meta" actions. My preference would be for a slightly more generalized approach — for instance, how would meta actions work with action middleware (the first part of this PR)? Ideally we could find a way to combine both concepts.

**Comment by rpominov** [comment]
> how would meta actions work with action middleware (the first part of this PR)?

As shown [here](https://github.com/gaearon/redux/issues/6#issuecomment-109775294) we can stack multiple middlewares on top of each other. And we can create a middleware that transform actions from a different format. But we can't support async formats like function or promise.

>
```js
type Middleware = {
  map(state: MiddlewareState): Atom
  run(initialState: MiddlewareState, actions: Observable<Action>): Observable<MiddlewareState>
}
```

This should work. It support asynchrony i.e., a middleware can introduce actions or replace state at an arbitrary time, without "meta" actions, but we can still use "meta" actions. And I think we could imagine a way to stack multiple middlewares on top of each other. And we could implement support of async actions (functions and promises) as one of those stacked middlewares.

But if we to implement internal stuff as middlewares based on observables, we'll probably need to bring Rx as dependency.

Also observables can be stateful itself, for instance `foo.take(5)` has a counter of how many items it already took as a state. All states like that will be reset on a hot reload.

@gaearon Frankly I don't have a strong opinion on what to choose at this point. So many different ideas. But if I had to choose, I'd go with what you propose in #6, it has limitation of not supporting asynchrony naturally, but it will serve as API for transactions, and for many other purposes. So I'd go with it, and if we'll need more from the middleware API in the future we can change the API.

acdlite:
Okay, so after reviewing both this PR and #6 again, here's when my thoughts currently stand.

I don't think @gaearon and are proposing conflicting options. Upon close inspection, they're actually quite complementary, but it's a bit hard to tell due to the lack of clarity around terminology. Let me try to fix that.

Let's think of a Redux dispatch cycle as an asynchronous stream of actions, which is transformed into an asynchronous stream of states. At a high level, when we talk about "middleware," we mean anything that modifies that stream.

```
action    --[ middleware ]-->   state
```

So at the beginning of the stream, you have "action middleware," like the promise and callback middlewares I describe above. These middleware operate asynchronously. They take a stream of "raw actions" (where an "raw action" is the raw value that is sent via `Dispatcher.dispatch()`, which could, per the above examples, possibly include functions and promises) and turn them into "action payloads," which are proper action objects

At the end of the stream, you have "state middleware," which must be synchronous to fit with the Flux/Redux architecture. We can call this middleware the "reducer". Expanding on the above diagram:

```
raw action    --[ action middleware ]-->    action payload    --[ reducer ]-->    state
```

As @rpominov [noted](https://github.com/gaearon/redux/pull/55#issuecomment-109793459), it's really only possible to have a single state middleware that does anything powerful. Within that single reducer, however, you have a lot of freedom.

That's where Dan's concept of "lifted actions" and "Flux inside Flux" fit into this equation. Those strategies can be implemented as reducers. The observables API described by Dan? Also can be implemented as a reducer.

As for Observables/streams, I'm a fan, and they are an excellent at modeling the Flux flow. But I don't think they should be the primary API for constructing middleware in Redux (I believe @gaearon agrees with me here). Using observables inside middleware, though, is perfectly fine! You can even use it inside the reducer, so long as it's synchronous.

As Dan said:

> I'd rather focus on a single extension point that allows free experimentation outside the core.

What we want is a high-level, minimal API for creating generalized middleware, preferably one that supports "intercepting" both actions and store atoms, since they are fundamentally all part of the same stream. I believe the approach laid out in this PR may qualify, as shown by my transactions example.

### Middleware "specification" (lol)

Middleware are higher-order functions. The reducer is one level higher, a function that returns a middleware. I don't know the best way to document these using formal notation.

Could someone help me out? Here's my attempt to describe it in English, which you can cross reference with the code above.

- **Middleware**: A higher-order function with the same signature as `dispatch()`. It accepts a single parameter, `action`. It returns a function whose input is `next`, where `next` is either the next middleware in the chain, or `dispatch()` if it is at the end of the chain. Middleware is both composable and asynchronous.
- **Reducer**: A higher-order function which returns a synchronous middleware. Its parameters are `store`, `state` (the current state atom at the time the reducer is invoked), and `dispatch` (so that reducers can trigger dispatches). The dispatcher is responsible for sending these values at the time the reducer is invoked.  I believe this is the minimal interface most reducers will need. Because a reducer returns a middleware function, it can be composed in exactly the same way as other kinds of middleware, where `next` is the dispatch method.

Blah, I hope this all makes sense.

acdlite:
Just cleared up with @gaearon that in his [above comment](https://github.com/gaearon/redux/pull/55#issuecomment-109801533) he was using the word "middleware" to refer to a "dispatcher strategy." Clearing that up helped us get on the same page.

Going forward, I propose that we use "middleware" to refer to composable middleware, like the ones in this PR, and "dispatcher strategy" to describe the higher-level object which accepts new actions and keeps track of state (and which, as an implementation detail, may make use of composable middleware).

emmenko:
Thanks for clarifying! :+1: 
I'm still trying to get the all picture though...

Also, quick question: how does this affect the HOC (`Provider`, `Connector`) ? 
From your changes it looks like they behave like before. Is that correct?

gaearon:
I'm drafting a new PR for custom dispatchers. That's what my “dispatch strategy” really is. Then things like this PR will be implementable outside the core.

gaearon:
I propose to close this PR and instead merge https://github.com/gaearon/redux/pull/60 which allows this PR to be implemented in userland.

gaearon:
(#60 wouldn't happen without the discussion here. Huge props to @acdlite!)

gaearon:
Redux 0.9.0 is out with the following default dispatcher function:

```js
export default function createDispatcher(store) {
  return function dispatcher(initialState, setState) {
    let state = store(initialState, {});
    setState(state);

    function dispatchSync(action) {
      state = store(state, action);
      setState(state);
      return action;
    }

    function dispatch(action) {
      return typeof action === 'function' ?
        action(dispatch, state) :
        dispatchSync(action);
    }

    return dispatch;
  };
}
```

It should be possible now to implement this PR as a custom dispatching function.

gaearon:
I think the next step could be to have `createDispatcher` accept optional `scheduler` and `reducer` arguments:

```js
function createDispatcher(store, schedule = createDefaultScheduler(), reduce = createDefaultReducer(store)) {
  return function dispatcher(initialState, setState) {
    let state = reduce(initialState, {});
    setState(state);

    function dispatchRaw(action) {
      state = reduce(state, action);
      setState(state);
      return action;
    }

    function dispatch(action) {
      return schedule(dispatchRaw, () => state, action);
    }

    return dispatch;
  };
}
```

A default scheduler might look like

```js
function createDefaultScheduler() {
  return function schedule(dispatch, getState, action) {
    return typeof action === 'function' ?
      action(schedule, getState()) :
      dispatch(action);
  }
}
```

A default reducer might look like:

```js
function createDefaultReducer(store) {
  return function reduce(state, action) {
    return store(state, action);
  };
}
```

Then we can expose an utility for chaining schedulers, like in this PR.
Thoughts?

@acdlite

gaearon:
An alternative dispatcher that replays all actions on top of the initialState:

```js
export default function createReplayingDispatcher(store) {
  return function dispatcher(initialState = {}, setState) {
    let actions = initialState.actions || [];
    let state = { initialState, ...initialState, actions };

    function dispatchSync(action) {
      actions.push(action);
      state = { ...state, ...actions.reduce(store, state.initialState) };
      setState(state);
      return action;
    }

    function dispatch(action) {
      return typeof action === 'function' ?
        action(dispatch, state) :
        dispatchSync(action);
    }

    dispatch({});
    return dispatch;
  };
}
```

You can swap the default dispatcher with this one on hot reload, and it will “just work”.
Would be great if this was expressible with `scheduler` + `reducer` thing I wrote above..

gaearon:
So I'm playing around with an “uber dispatcher” (no composition for now, just boolean options). What I realized is ES6 Symbols are *awesome*! The let us hide the middleware state *inside* the state atom!

```js
function createDefaultReducer(store) {
  return store;
}

const REPLAY_STATE = Symbol('Replay State');

function createReplayingReducer(store) {
  return (state = {}, action) => {
    let {
      [REPLAY_STATE]: { actions = [], initialState = state } = {},
      ...appState
    } = state;

    actions = [...actions, action];
    appState = actions.reduce(store, initialState);

    return {
      [REPLAY_STATE]: { actions, initialState },
      ...appState
    }
  };
}

function createDefaultScheduler(dispatch, getState) {
  function schedule(action) {
      return action(schedule, getState());
    } else {
      return dispatch(action);
    }
  }

  return schedule;
}

export default function createDispatcher(store, {
  log = false,
  replay = false
}: options = {}) {
  return function dispatcher(initialState, setState) {
    const reduce = replay ?
      createReplayingReducer(store) :
      createDefaultReducer(store);

    let state = reduce(initialState, {});
    setState(state);

    function dispatch(action) {
      if (log) {
        console.groupCollapsed(action);
        console.log('State before:', state);
      }

      state = reduce(state, action);
      setState(state);

      if (log) {
        console.log('State after:', state);
        console.groupEnd(action);
      }

      return action;
    }

    return createDefaultScheduler(dispatch, () => state);
  };
}
```

acdlite:
Sorry I'm just now looking at this, but yay! Looks great.

The point about using Symbols to hide "meta" state on the state atom smart. Small caveat, though. This

```js
let {
  [REPLAY_STATE]: { actions = [], initialState = state } = {},
  ...appState
} = state;
```

assumes that the state atom is a plain object, rather than something like ImmutableJS. Which isn't a huge deal, since in the 99% case that will be true (because most people will use `createStores()`) but we'll need to document assumptions like that.

gaearon:
Hmm, good point. We should probably just put the app's state in a separate field instead, and expose just it to the app in `redux.getState()`. This way it's never exposed to the user, we can still use Symbols to avoid clashes between middleware, and the user can even use a primitive as the atom.

gaearon:
@acdlite I'd be happy to have you playing with https://github.com/gaearon/redux/pull/62

acdlite:
Closing this per #60. Eventually once the API is more settled I'll probably open a similar PR to add an easy integration point for global action middleware and get rid of `perform()`. Because action middleware like `callbackMiddleware()` and `promiseMiddleware()` merely wrap around a `dispatch()` method, they have the special property of being compatible with any implementation of the dispatcher, because every dispatcher will have this fundamental API. In other words, wherever you can do this

```js
dispatch(action);
```

you can also do this

```js
callbackMiddleware(dispatch)(action)
```

For that same reason, it's also very easy to implement this in userland for the time being.

gaearon:
:+1: Thanks a lot for your work here. I feel that us approaching this from two different sides (close to the action vs close to the setState) really helped figure out the common base API.

acdlite:
:+1: Agreed!

emmenko:
Awesome work guys!!! Keep going :+1:

Issue #17, opened 2015-06-03
Tests


gaearon:
Tests are sorely missing. Gonna try to fill that gap today.

fibric:
what tools do you wanna use here?

gaearon:
I'd probably use Mocha + `createShallowRenderer()` from React's TestUtils.

emmenko:
I think at this point a basic testing environment is quite important. Makes refactoring safer...

emmenko:
Should we close this?

gaearon:
Yep! Great job.

PR #56, opened 2015-06-07
Test env setup

emmenko:
Refs #17 

@gaearon I've provided an initial test environment with a couple of tests already. I just wanted to hear from you first if it's somehow ok for you like this. 

If it's acceptable I would suggest to merge it, setup CI on travis and simply continue adding (and improving) the remaining tests. 
This makes refactoring safer, especially if there will be releases. 

Thanks :)


gaearon:
Wow, thank you! I'll merge on Monday (and add some tests) :+1:

gaearon:
Sorry for dragging my feet here. Can I ask you to update this to 0.9.0 API?
I also would prefer [expect](https://github.com/mjackson/expect) to be used for assertions.

emmenko:
Sure, will do that

gaearon:
Thank you!

emmenko:
@gaearon @acdlite basic test setup is ready-to-merge :)

gaearon:
Awesome! Thank you so much!

@acdlite Do you think it's better to merge as is, or replace stubs with some real tests firsts?

test/Redux.spec.js
```diff
@@ -0,0 +1,47 @@
...
+    expect(redux.state).toEqual({ fakeStore: fakeState });
+    expect(redux.listeners).toEqual([]);
+    expect(redux.dispatcher).toBeA('function');
+    expect(redux.dispatchFn).toBeA('function');
+  });
+
+  it('should subscribe to changes', done => {
+    redux.subscribe(() => {
+      const state = redux.getState();
+      expect(state).toEqual(fakeState);
```
acdlite: This test isn't sufficient because the updated state is the same as the initial state, since `fakeStore()` always returns the same thing.

acdlite:
@gaearon I say merge as is. It's better than what we currently have :)

Thanks @emmenko!

acdlite:
Oh actually, could you fix that issue from my comment? Then I'll merge so I can update the tests for  https://github.com/gaearon/redux/pull/63

emmenko:
@acdlite fixed the test. How is it now?

gaearon:
Thank you so much!

emmenko:
Actually I'm fixing the tests because of the changes here #63 :)

Will do another PR

gaearon:
OK great! I wanted to ask you to test only the public API.

e.g. `createRedux` doesn't expose `setState`, so it doesn't need to be tested. Instead, to test subscriptions, you can give it a fake dispatcher.

gaearon:
Also, it would help the tests if they used something like the Counter example store instead of `FOO` etc.

emmenko:
> Also, it would help the tests if they used something like the Counter example store instead of `FOO` etc.

Ok, I will do another PR for that.

> I wanted to ask you to test only the public API

Shouldn't we test also the internal stuff?

gaearon:
>Shouldn't we test also the internal stuff?

It's a tricky issue. We should test _the “public” API_ of the internal stuff. Even if some class or object is internal only, it still has a “public” API, and has implementation details. We should check that it behaves as a unit correctly, but we should not call its methods that are not meant to be called from outside.

emmenko:
Ok, I will try to change that. Also use useful test data :)

PR #59, opened 2015-06-08
[WIP] Example with react-router and async data fetching

emmenko:
@gaearon this is not a real PR (at least for now). I just wanted to try an actual real-world example by doing some data fetching and using `react-router` and get feedback from you.

It would be good if you can have a look if the setup of the components is actually correct. There are a couple of things that are not working yet (e.g.: navigating to post detail).

I also have a couple of questions:
- if I use the `connect` decorator I don't get any `props` that are passed to the component, the decorator just wraps the `dispatcher` and the related `store`
- to solve this I'm using the `Connector` component, so I can manually pass down the `props` like this https://github.com/emmenko/redux/blob/80593cd5bbe391332f67ac2643ceead0217a976a/examples/containers/BlogApp.js#L21-L28
- since `actions` have to be bound to the `dispatcher`, in case I use the `Connector` component I can only do that in the `render` method, causing to pass the `actions` to the dumb component (not the container) and call the action from there (e.g.: https://github.com/emmenko/redux/blob/80593cd5bbe391332f67ac2643ceead0217a976a/examples/components/PostList.js#L6-L8). Is that how it's supposed to be?

Maybe you point me in the right direction? Thanks :)


emmenko:
@gaearon also if you have 5min, can you have a look at that? I have to migrate to the new API but still conceptually I would like to hear your opinion.
Thanks :)

gaearon:
@emmenko 

Sorry, I'm crazy busy the following couple of weeks with preparing for my ReactEurope talk. Redux is ready enough for me to polish my slides and examples now. So I'd love to take a better look but I just can't afford to at the moment. Maybe somebody else can?

emmenko:
@gaearon don't worry then. Actually I'd love to have a quick chat with you in person at ReactEurope ;)
So keep preparing your talk, it's more important!

gaearon:
Oh, you're coming? See you there then!

emmenko:
Of course I'm coming ;)

Looking forward to it then!!

emmenko:
@gaearon just FYI (don't want to waste your time): I realised I had a typo in the import, so now the example works :)

@acdlite if you have time to review it, I would be happy. I'm not asking this to be merged, just want to get your feedback if the setup is actually correct. 

And there are still a couple of things to be clarified I guess:
- if I use the `connect` decorator I don't get any `props` that are passed to the component, the decorator just wraps the `dispatcher` and the related `store`
- to "solve" this I'm using the `Connector` component, so I can manually pass down the `props` like this 
https://github.com/emmenko/redux/blob/async-with-router/examples/containers/BlogApp.js#L21-L28
- since `actions` have to be bound to the dispatcher, in case I use the `Connector` component I can only do that in the `render` method, causing to pass the `actions` to the "dumb component" (not the container) and call the action from there (e.g.: https://github.com/emmenko/redux/blob/async-with-router/examples/components/PostList.js#L6-L9). Is that how it's supposed to be??

Thanks!

acdlite:
This looks great.

We'll definitely need to update the HoCs so that they pass through props.

I think the solution to your action creator problem is to create another container component in between BlogApp and PostList — PostListContainer, perhaps — and call fetch from there.

On a related note, I think maybe we should add back the `actions` prop from previous versions, except call it `actionCreators` instead. It will reduce the boilerplate of having to call `bindActionCreators()` inside the render method, and has the (slight) performance benefit of not having to execute on ever render. We should keep passing `dispatch()` to the child function, though, so this wouldn't be a breaking change.

emmenko:
@acdlite so I've simplified the example setup.

Also I'm using decorators as intermediate components to fetch the data, which I think it's more elegant. 
https://github.com/gaearon/redux/commit/38610b1bb254024a43b56c9edfda3eb1334e167b

> I think maybe we should add back the actions prop from previous versions, except call it actionCreators instead. It will reduce the boilerplate of having to call bindActionCreators() inside the render method

Right, currently the implementation is like this

```js
@connect(state => ({
  blog: state.blog
}))
export default class BlogDetailApp {
  render() {
    const { dispatch } = this.props;
    const actions = bindActionCreators(BlogActions, dispatch);
    return <PostDetail actions={actions} {...this.props} />;
  }
}
```

but it can be simplified to this

```js
@connect(state => ({
  blog: state.blog
})) // TODO: somehow pass the action creators to be bound here?
export default class BlogDetailApp {
  render() {
    return <PostDetail {...this.props} />;
  }
}
```

So the binding will be done in the decorator, right?

Any feedback would be very welcomed :)

gaearon:
If you have some more time to work on this, I'd appreciate you porting this to use React Router 1.0 beta.

emmenko:
@gaearon it was my next step yes :)

Just wanted to be sure I was on the right path (e.g.: see comment above regarding `fetch` decorators). 
If you guys can give me some feedback here it would be awesome.

PS: also I would like to have tests for the examples. It think it would be very useful for people setting up a project to see how to write tests for react components and redux. But we can discuss this in a separate issue.

gaearon:
I'm not super convinced about things `fetch` but I like the direction.
I think it's fine to show how custom stuff like this can be easily created around Redux.

@acdlite?

emmenko:
Yeah, I'm not super happy about it either, but haven't found a more elegant solution for now...maybe you have a better idea?

bradgignac:
@gaearon @emmenko I'm using something similar to the `fetch` decorator in a couple of my applications. I'd also love to know if anyone has found a better pattern.

bradgignac:
@emmenko I'm planning on porting an existing application to React Router v1.0 and Redux, so I'm extremely interested in this example. I'm really curious to see how you would handle adding a post and calling `router.transitionTo(...)` once it has completed. If there's anything I can do to help out with this example, let me know! Thanks for all your hard work!

emmenko:
@bradgignac sure, I'll try to do the migration by tomorrow. I'll let you know if I encounter any problems.

As a next step we could then enhance the example by allowing to **add/remove** posts. But I would merge this first and provide another PR then.

gaearon:
If you port to RR 1.0 I'd like to see proper Flux (it should support this). That is, subscribe to `history` object for changes and turn them into actions with `path` payload, then have the root component subscribe to the “store” that contains the current path, and render `<Router>` with it.

cc @mjackson @ryanflorence in case I said something stupid

gaearon:
Also I'm not sure if we want to do it here or spin up `redux-react-router-exaple`.

emmenko:
> Also I'm not sure if we want to do it here or spin up `redux-react-router-exaple`

We could yes. In that case should it be the same example with the github repos? I think so...

> That is, subscribe to `history` object for changes and turn them into actions with `path` payload, then have the root component subscribe to the “store” that contains the current path, and render `<Router>` with it.

I'm a bit confused here. I thought this is what RR does :) 
Maybe can you elaborate a bit more what's the purpose of this?

gaearon:
>We could yes. In that case should it be the same example with the github repos? I think so...

That might be good for comparison. On the other hand, there's no POST/PUT action there. Maybe we could add actual authorization (also a good use case) and a “Star” button. *Then* it would be awesome. But it's more work.

>I'm a bit confused here. I thought this is what RR does :) 
>Maybe can you elaborate a bit more what's the purpose of this?

I want every change in the app to flow through Redux. Normally RR encourages you to just call `React.render` again. I want to turn this into actions instead so that it works with record/replay.

emmenko:
Hmm ok, I feel this would be better done in a different PR though (so migrating to RR 1.x + refactor redux flow). Would be that ok for you?

> Maybe we could add actual authorization (also a good use case) and a “Star” button. Then it would be awesome.

:+1: :+1: :+1:

gaearon:
>Hmm ok, I feel this would be better done in a different PR though (so migrating to RR 1.x + refactor redux flow). Would be that ok for you?

To be honest I really think it's better to split into several repo. We can link to it!

emmenko:
Oh ok, I wasn't sure you actually wanted to do that...should we keep this example then or go with `*-react-router-example` version 2?

So what's the plan? Do you want to setup the repo?

gaearon:
I'd say close this PR and create a repo of your own :-). That's your work after all!

emmenko:
If you're ok with that, then sure :)

I'll let you know then.

emmenko:
@gaearon the repo for this example will be https://github.com/emmenko/redux-react-router-async-example

I'll add you as a contributor.

gaearon:
:+1:

tsingson:
+1

ajaybeniwal:
+1

ryanflorence:
Do it!

SudoPlz:
+1

netfishx:
+1

lancetw:
:+1:

PR #62, opened 2015-06-09
Experiment with uber dispatcher

gaearon:
This is not really a PR, just something for everyone to play with.
I'm creating a custom dispatcher that accepts options:

```js
// You can modify these options while the code is running:
const dispatcher = createDispatcher(store, { log: true, replay: false });
const redux = createRedux(dispatcher);
```

To have fun, you may start in normal mode, make a few actions, then set `replay: true`, fire an action, change a Store and see that particular action replayed. Then to “commit” the state, you can set `replay` back to `false`.

### DIFF
--- a/examples/containers/App.js
+++ b/examples/containers/App.js
@@ -1,10 +1,14 @@
 import React from 'react';
 import CounterApp from './CounterApp';
 import TodoApp from './TodoApp';
-import { createRedux, Provider } from 'redux';
+import { createRedux, composeStores, Provider } from 'redux';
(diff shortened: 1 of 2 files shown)


acdlite:
Great, I'll give this a look . The magical re-playing behavior on hot reload is super cool. Discovered that when I was working on #60. The more familiar I become FP the more it astounds me that it isn't a more dominant paradigm.

Also, just imagine working on this library without ES6. The mind reels :)

gaearon:
Closing, as it's not a real PR :-).
I might come up with something more real later..

PR #64, opened 2015-06-09
Use function bind syntax

jamiehodge:
Since you're using the newer syntax elsewhere...

### DIFF
--- a/src/components/Connector.js
+++ b/src/components/Connector.js
@@ -35,8 +35,7 @@ export default class Connector extends Component {
   constructor(props, context) {
     super(props, context);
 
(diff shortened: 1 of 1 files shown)


gaearon:
While it's not exactly the same, this should still work I think, right? We're not relying on `handleChange` being bound elsewhere.

acdlite:
Right, I'm just worried it may lead to a bug later. Also, like you said in https://github.com/gaearon/redux/pull/66, we should resist merging changes like this until we have tests.

gaearon:
This doesn't merge cleanly now, but for consistency I made this change in 24975c846d97b040c05de3dac1226c760749b367.

PR #66, opened 2015-06-09
Set initial state in constructor

jamiehodge:
Replace if statement and extract getState method.

### DIFF
--- a/src/components/Connector.js
+++ b/src/components/Connector.js
@@ -37,7 +37,8 @@ export default class Connector extends Component {
 
     this.handleChange = this.handleChange.bind(this);
     this.unsubscribe = context.redux.subscribe(this.handleChange);
-    this.handleChange();
+
+    this.state = this.getState();
   }
 
   componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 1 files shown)


acdlite:
I don't like this because the method names conflate React component state with the state from Redux. In other words, `this.getState()` and `this.setState()` are not symmetrical.

jamiehodge:
Corrected

jamiehodge:
I assume this is no-go:

````javascript
this.state = this.getState(props, context);

...

getState(props = this.props, context = this.context) {
  const state = context.redux.getState();
  return props.select(state);
}
````

acdlite:
That's not the worst idea, but I still don't like how `this.setState()` and `this.getState()` refer to two different things.

jamiehodge:
Well, given the method, I guess one could name it `selectState`.

acdlite:
Thanks for the PR, but I'm going to close this since it's mostly style issue, and there's a reason for the way it currently is. If you can think of a way to change it that would improve clarity in a significant way, I'll re-open.

acdlite:
Actually, yeah, I would consider `selectState()` if you want to try that.

gaearon:
I'd avoid any purely stylistic changes until we have the tests.

emmenko:
Tests are coming :) #56

PR #67, opened 2015-06-09
Extract selectState and set initial state in constructor

jamiehodge:
Another attempt at https://github.com/gaearon/redux/pull/66.

This has the benefit of associating `this.props.select` with its use (`selectState`).


acdlite:
I like this but let's wait until we have test coverage.

gaearon:
>I like this but let's wait until we have test coverage.

Yep

PR #73, opened 2015-06-10
Add support for React Native

leoasis:
Basically using the approach from [Flummox](https://github.com/acdlite/flummox), changed all component modules to be factories that get the React instance injected, and the decorator modules to be factories that get the React instance and the component class injected.

Then we have `'redux/react'` and `'redux/react-native'` as separate imports in order to use the corresponding components and decorators using those factories with the correct React instance.

Also removed `react` from the dependencies, and into the devDependencies, since either `react` or `react-native` are needed. Perhaps they should be put in `optionalDependencies`? Not really sure how to handle that correctly. In practice, webpack handles this correctly.

Should we also add a `react-native` tag to the package.json?

This fixes https://github.com/gaearon/redux/issues/71


gaearon:
This is gold, thank you. Let this be reviewed by @acdlite. He should know how to deal with deps..

acdlite:
Looks great!

I agree that keeping it in devDependencies is the way to go. Technically, neither React nor React Native are *required*. You could subscribe to the Redux instance directly and use it with Backbone, Angular, whatever.

peerDependencies has only ever given me problems.

I wasn't familiar with optionalDependencies, but I looked it up:

> If a dependency can be used, but you would like npm to proceed if it cannot be found or fails to install, then you may put it in the optionalDependencies object.

That's not what we want, because this will still attempt to install React.

So we're good! :+1:

leoasis:
Thanks guys! :smile:

emmenko:
Cool :+1:

gaearon:
Thank you!

PR #74, opened 2015-06-10
Test only public API and use better test data

emmenko:
@gaearon @acdlite it should be better now.

Refs https://github.com/gaearon/redux/pull/56#issuecomment-110797801


emmenko:
@acdlite aaargh, too many breaking changes here :)

I will adjust the tests...

acdlite:
:) You should just have to change the import paths.

emmenko:
@acdlite done, ready-to-merge ;)

acdlite:
Thanks!

emmenko:
:+1:

emmenko:
I don't know if you guys want to take over now, otherwise I can see if I can provide some tests for the React components...just let me know :)

gaearon:
I'd appreciate component tests. You can use [shallow rendering](https://facebook.github.io/react/docs/test-utils.html#shallow-rendering) for that to avoid jsdom.

emmenko:
Ok, I'll see what I can do. Besides that, I think you guys can already setup CI on travis...

gaearon:
I'm going to be very busy in the next three weeks and won't participate as much. (Don't worry, I'll be back after ReactEurope :-). So things like setting up Travis, etc, are very welcome if @acdlite finds time, but I won't be able to focus on something like this just yet.

acdlite:
I can handle that

PR #75, opened 2015-06-11
WIP: Component tests

acdlite:
I started writing some tests for the components, but ran into some issues with React's shallow renderer related to its non-support for context: https://github.com/facebook/react/issues/3696

Anyone have any ideas? @gaearon would prefer not to use jsdom if we can avoid it, and I agree.

Either way, I'll track progress here.


emmenko:
Does it work with the workaround described in the issue?

gaearon:
Might help: https://github.com/facebook/react/issues/3721#issuecomment-106318499

acdlite:
I'll try that out, thanks.

acdlite:
I couldn't get any the workarounds to work :(

I'm going to use jsdom for now (via [mocha-jsdom](https://github.com/rstacruz/mocha-jsdom), which is nice) since it's better than nothing. If anyone can figure out how to use shallow rendering for our tests that deal with context, that'd be awesome.

gaearon:
:+1:

acdlite:
Component tests are done. Tests for shallowEqualScalar and shallowEqual are remaining, but I think we're good to release 0.10.

@gaearon Could you explain to me how shallowEqualScalar is different from shallowEqual, and why you'd use one over the other? It's not clear to me from reading the source.

leoasis:
Awesome stuff! :+1:

gaearon:
>Could you explain to me how shallowEqualScalar is different from shallowEqual, and why you'd use one over the other? It's not clear to me from reading the source.

`shallowEqualScalar` bails out and returns `false` if any of the values is an object.

`shallowEqual` is not a safe default to use in our `shouldComponentUpdate` because it assumes the user never does deep mutations (which they might). On the other hand, `shallowEqualScalar` is safe. If your props are all scalar (no objects, just primitives, such as numbers, boolean, strings or functions), then you'll get some perf benefit for free.

acdlite:
Aha, makes sense. Thanks!

gaearon:
```js
shallowEqual({ a: 3 }, { a: 3 }) // true
shallowEqualScalar({ a: 3 }, { a: 3 }) // true

const b = {};
shallowEqual({ a: 3, b: b  }, { a: 3, b: b }) // true
shallowEqualScalar({ a: 3, b: b  }, { a: 3, b: b }) // false
```

gaearon:
>Question: do we want to pass dispatch to the decorated component?

I'd do that.

gaearon:
Do we want to fix #76 before 1.0?

gaearon:
Oh wait, I misread `0.10` as `1.0`. Anyway, do you want to get a fix for it in before cutting a release?

emmenko:
Yes please :)

acdlite:
Actually I did that already since I was writing a test for it, anyway. I'll close it.

emmenko:
:+1:

acdlite:
@gaearon I started writing some release notes for 0.10. I'll finish up when I get home and push to npm. Then I think we should wait until we get a universal app example finished before releasing 1.0. I'll work on that this weekend.

src/components/createConnectDecorator.js
```diff
@@ -13,14 +15,10 @@ export default function createConnectDecorator(React, Connector) {
       render() {
         return (
           <Connector select={state => select(state, this.props)}>
-            {this.renderChild}
+            {stuff => <DecoratedComponent {...stuff} {...this.props} />}
```
gaearon: Can we avoid allocating a function here?

gaearon: Probably not.. Since it depends on `this.props`, we have to make it a lambda, or else the props will be stale.

acdlite: Yep, beat me to it :)

gaearon:
@acdlite 

Overall, awesome job as always. :+1:

emmenko:
@acdlite @gaearon I was looking into the shallow rendering "problem" (also because jsdom works only with iojs) and found something weird. 
If I run the [test example](https://github.com/facebook/react/blob/master/src/test/__tests__/ReactTestUtils-test.js#L162-L178) from the react repository, it fails.

You can see it here https://gist.github.com/emmenko/d842e62513d6fb08aa93

Not sure if I'm doing something wrong or there is actually a problem somewhere...any idea?

gaearon:
Maybe they fixed it in master? Were you running master?

emmenko:
Hmm, I haven't tried master, just pulling latest version from npm

emmenko:
@gaearon you're right, using `react` from master works. I guess we just have to wait for the next release, then we can use shadow rendering.

PR #81, opened 2015-06-13
compiling react.js and react-native.js before npm publish

skevy:
The root react.js and react-native.js files need to be compiled with Babel before pushing to npm.

It's kind of nasty to have these long pre- and post-publish commands in package.json...so if you guys have interest in moving to a Makefile approach, let me know and I'll send another PR.


acdlite:
Yeah, we should think about this. I like Make but when I used it in Flummox I got complaints from people that it made it harder for Windows users to contribute. Maybe a `scripts` folder with bash scripts would be better? I don't think we need the incremental build features of Make, just a way to run some predefined commands.

skevy:
Scripts are fine too. I like `make`, but I understand the concerns from Windows users. I'll do something with scripts now and update this PR.

dariocravero:
You may find these useful then :)
https://github.com/UXtemple/panels-router/tree/master/.bin we took it a
step further and use the same paths for compiled and source sending only
the necessary parts to npm and git but it's easy not to do that too if you
want to keep the current src structure.

gaearon:
Let's keep ES6 versions of `react.js` and `react-native.js` in `src`.

>Maybe a scripts folder with bash scripts would be better?

Let's do this. The `build` script should put the compiled versions of `react.js` and `react-native.js` into the root folder. They should be in `.gitignore`.

gaearon:
A temporary fix is out in https://github.com/gaearon/redux/commit/1ed784cc025d835d8fdf3812db2b8c2ec9d0fe2f.

gaearon:
Meaning, it's out in 0.10.1..

gaearon:
Do you mind rebasing this? Feel free to get rid of 1ed784cc025d835d8fdf3812db2b8c2ec9d0fe2f again, I just wanted to get a fix out and didn't notice you already updated the PR..

skevy:
No problem. Will do after I push updates to #86.

gaearon:
:+1:

skevy:
@gaearon rebased.

gaearon:
Cool, can you squash into a single commit please?

```
git checkout fix-packaging
git merge origin/master
git reset --soft origin/master
git commit -am "message"
git push -f
```

skevy:
Sorry about that. Done.

gaearon:
:+1: Thanks!

Issue #78, opened 2015-06-13
We need to throw if `select` returns something other than a plain object


gaearon:
`<Connector select={fn}>` contract is that `fn` takes the root state and returns a plain object. It is *essential* for the result to be a plain object because:

* When using `@connect`, its result *is* what you'll get as injected props, and React only supports plain object props at the moment.
* Redux shallowly compares `select` return values to determine if an update is needed (a perf optimization).

I think we need to enforce that `select` returns a plain object. We can use `lodash/lang/isPlainObject` to perform this check.

PR is welcome (with a test please!).

gaearon:
Closed via #85, thanks!

PR #85, opened 2015-06-13
Throws an error when `select` not returns an object

rwillrich:
On Connector, verify if `select` prop, which is a function, returns an
object, and if not, throws an error.

Issue #78

### DIFF
--- a/src/components/createConnector.js
+++ b/src/components/createConnector.js
@@ -1,5 +1,6 @@
 import identity from 'lodash/utility/identity';
 import shallowEqual from '../utils/shallowEqual';
+import isPlainObject from 'lodash/lang/isPlainObject';
 
 export default function createConnector(React) {
   const { Component, PropTypes } = React;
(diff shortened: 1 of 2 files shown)


src/components/createConnector.js
```diff
@@ -59,6 +60,11 @@ export default function createConnector(React) {
     selectState({ context, props } = this) {
       const state = context.redux.getState();
       const slice = props.select(state);
+
+      if (!isPlainObject(slice)) {
```
gaearon: Great! Let's use `invariant`?

```js
import invariant from 'invariant';

// ...

invariant(
  isPlainObject(slice),
  'The return value of `select` prop must be an object. Instead received %s.',
  slice
);
```

rwillrich: Done!

emmenko:
:+1:

gaearon:
Awesome, thank you!

istarkov:
Mb i'm wrong, but you need to add babel transform to strip invariant message in production like this https://github.com/facebook/react/blob/master/vendor/constants.js

gaearon:
I actually don't mind having message in production (it's not that long and we only have very few).

Issue #84, opened 2015-06-13
Explore passing actions in Connector and @connect decorator?


skevy:
It would be more convenient to not have to call `bindActionCreators` directly inside the component, so I was thinking that we might able to add something to `Connector` and the `@connect` decorator to make this nicer? In addition, that would prevent calling `bindActionCreators` in render every time (which would provide an *extremely slight* performance benefit :)).

It seems like the idea of "connecting" could mean connecting to stores AND/OR connecting to actions, if the user wanted. Seems like `Connector` is perfect to just mean..."hey, connect this component (or child component) to Redux".

Thoughts?

skevy:
We could also continue to pass `dispatch`...no real reason NOT to do that I don't think.

emmenko:
See https://github.com/gaearon/redux/pull/59#issuecomment-111214797

skevy:
Oh cool! Didn't see that. Yah that's exactly what I'm talking about.

emmenko:
:+1: :)

skevy:
So, even though I think this is still a good idea, I think that even if we abstract binding of action creators into Connector...we're still going to have to bind on every render due to the fact that `dispatch` is passed through props. Am I right about @acdlite ?

gaearon:
>we're still going to have to bind on every render due to the fact that dispatch is passed through props. 

Yeah..

Maybe we can introduce some binding mechanism closer to the root, so all actions end up in `this.props.actionCreators`?


```js
import * as actionCreators from './actions/index';
<Provider redux={redux} actionCreators={actionCreators}>
```

Decorator version:

```js
@connect(state => ({
  blog: state.blog
}))
export default class BlogDetailApp {
  render() {
    return <PostDetail {...this.props.actionCreators.BlogActions} {...this.props} />;
  }
}
```

Container version:

```js
<Connector select={select}>
  {({ someState, actionCreators }) => <PostDetail {...actionCreators.BlogActions} {...this.props} />}
</Connector>
```

The `dispatch` function would also be available (and also passed from the top).

skevy:
Is there any bad thing about binding them higher up? In the current version, did you purposely choose *not* to do that? Or was it just something you hadn't gotten to yet?

I have a working version of a modified `Connector` and `@connect` that you can pass actions to...seems like a decent API and keeps actions modularized and localized to the components that need them, but has a downside of still binding in `render` all the time (though that's not a huge performance hit, obviously).

[...]

PR #86, opened 2015-06-13
Allow for passing of action creators through `Connector` and `@connect` decorator

skevy:
A possible solution to #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@ redux
 [![build status](https://img.shields.io/travis/gaearon/redux.svg?style=flat-square)](https://travis-ci.org/gaearon/redux)
 [![npm version](https://img.shields.io/npm/v/redux.svg?style=flat-square)](https://www.npmjs.com/package/redux)
 
-An experiment in fully hot-reloadable Flux.  
+An experiment in fully hot-reloadable Flux.
 
-**The API might change any day.**  
+**The API might change any day.**
 _**Don't use in production just yet.**_
 
 ## Why another Flux framework?
@@ -140,10 +140,8 @@ export default class Counter {
 
 ```js
 // The smart component may observe stores using `<Connector />`,
-// and bind actions to the dispatcher with `bindActionCreators`.
 
 import React from 'react';
-import { bindActionCreators } from 'redux';
 import { Connector } from 'redux/react';
 import Counter from '../components/Counter';
 import * as CounterActions from '../actions/CounterActions';
@@ -158,11 +156,13 @@ function select(state) {
 export default class CounterApp {
   render() {
     return (
-      <Connector select={select}>
-        {({ counter, dispatch }) =>
+      // Passing an object of action creators here automatically binds them 
+      // with Redux's dispatcher and passes them to the child function
+      // (so they can be passed as props to child components)
+      <Connector select={select} actionCreators={CounterActions}>
+        {({ counter, actions }) =>
           /* Yes this is child as a function. */
-          <Counter counter={counter}
-                   {...bindActionCreators(CounterActions, dispatch)} />
+          <Counter counter={counter} {...actions} />
         }
       </Connector>
     );
@@ -176,22 +176,23 @@ The `@connect` decorator lets you create smart components less verbosely:
 
 ```js
 import React from 'react';
-import { bindActionCreators } from 'redux';
 import { connect } from 'redux/react';
 import Counter from '../components/Counter';
 import * as CounterActions from '../actions/CounterActions';
 
+// Pass an object of action creators you want to pass 
+// as the `actions` prop to your component as the (optional) 
+// second parameter of @connect
 @connect(state => ({
   counter: state.counter
-}))
+}), CounterActions)
 export default class CounterApp {
   render() {
-    const { counter, dispatch } = this.props;
-    // Instead of `bindActionCreators`, you may also pass `dispatch` as a prop
-    // to your component and call `dispatch(CounterActions.increment())`
+    const { counter, actions, dispatch } = this.props;
+    // Instead of passing action creators as a parameter to @connect to autobind them, you may also
+    // pass `dispatch` as a prop to your component and call `dispatch(CounterActions.increment())`
     return (
-      <Counter counter={counter}
-               {...bindActionCreators(CounterActions, dispatch)} />
+      <Counter counter={counter} {...actions} />
     );
   }
 }
--- a/src/components/createConnector.js
+++ b/src/components/createConnector.js
@@ -1,5 +1,7 @@
 import identity from 'lodash/utility/identity';
 import shallowEqual from '../utils/shallowEqual';
+import bindActionCreators from '../utils/bindActionCreators';
+
 
 export default function createConnector(React) {
   const { Component, PropTypes } = React;
@@ -11,7 +13,11 @@ export default function createConnector(React) {
 
     static propTypes = {
       children: PropTypes.func.isRequired,
-      select: PropTypes.func.isRequired
+      select: PropTypes.func.isRequired,
+      actionCreators: PropTypes.oneOfType([
+        PropTypes.func,
+        PropTypes.object
+      ])
     };
 
     static defaultProps = {
@@ -62,12 +68,24 @@ export default function createConnector(React) {
       return { slice };
     }
 
+    bindActionCreators(dispatch) {
+      if (!this.props.actionCreators) {
+        return {};
+      }
+
+      if (typeof this.props.actionCreators === 'function') {
+        return { actions: this.props.actionCreators(dispatch) };
+      } else {
+        return { actions: bindActionCreators(this.props.actionCreators, dispatch) };
+      }
+    }
+
     render() {
       const { children } = this.props;
       const { slice } = this.state;
       const { redux: { dispatch } } = this.context;
 
-      return children({ dispatch, ...slice });
+      return children({ dispatch, ...slice, ...this.bindActionCreators(dispatch) });
     }
   };
 }
--- a/src/components/createConnector.js
+++ b/src/components/createConnector.js
@@ -14,14 +14,12 @@ export default function createConnector(React) {
     static propTypes = {
       children: PropTypes.func.isRequired,
       select: PropTypes.func.isRequired,
-      actionCreators: PropTypes.oneOfType([
-        PropTypes.func,
-        PropTypes.object
-      ])
+      actionCreators: PropTypes.object
     };
 
     static defaultProps = {
-      select: identity
+      select: identity,
+      actionCreators: {}
     };
 
     shouldComponentUpdate(nextProps, nextState) {
@@ -45,12 +43,17 @@ export default function createConnector(React) {
 
       this.unsubscribe = context.redux.subscribe(::this.handleChange);
       this.state = this.selectState({ context, props });
+      this.boundActions = this.bindActionCreators(props.actionCreators, context.redux.dispatch);
     }
 
     componentWillReceiveProps(nextProps) {
       if (nextProps.select !== this.props.select) {
         // Force the state slice recalculation
         this.handleChange();
+      }
+
+      if (nextProps.actionCreators !== this.props.actionCreators) {
+        this.boundActions = this.bindActionCreators(nextProps.actionCreators, this.context.redux.dispatch);
       }
     }
 
@@ -68,19 +71,10 @@ export default function createConnector(React) {
       return { slice };
     }
 
-    bindActionCreators(dispatch) {
-      if (!this.props.actionCreators) {
-        return {};
-      }
-
-      // Handle `function` version of prop that allows user
-      // to do fancy things, like change how actions are bound or
-      // change the shape of the `actions` object.
-      if (typeof this.props.actionCreators === 'function') {
-        return { actions: this.props.actionCreators(dispatch) };
-      } else {
-        return { actions: bindActionCreators(this.props.actionCreators, dispatch) };
-      }
+    bindActionCreators = (actionCreators, dispatch) => {
+      return {
+        actions: bindActionCreators(actionCreators, dispatch)
+      };
     }
 
     render() {
@@ -88,7 +82,7 @@ export default function createConnector(React) {
       const { slice } = this.state;
       const { redux: { dispatch } } = this.context;
 
-      return children({ dispatch, ...slice, ...this.bindActionCreators(dispatch) });
+      return children({ dispatch, ...slice, ...this.boundActions });
     }
   };
 }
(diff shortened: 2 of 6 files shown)


gaearon:
I like to have less boilerplate, but here's what I don't like about this API:

* `@connect` has just gotten harder to understand in my opinion because it no longer does single thing
* there is no way to implement efficient `shouldComponentUpdate` for the contained components because they'll get new bound functions every time

I can live with the first, but the second one is a red flag to me. If we're breaking `shouldComponentUpdate` of user's component, we're doing something wrong.

**This is also a problem with the current API and I agree it needs fixing.** I just want to add first-class action creator binding support if it *also* fixes the `shouldComponentUpdate` issue.

gaearon:
One way to work around the `shouldComponentUpdate` problem is to keep the bound functions in the `Connector`'s `state` and only rebind if:

- `dispatch` changed (which means Store got hot-reloaded)
- passed action creators are shallowly unequal (which means Actions got hot-reloaded)

acdlite:
@gaearon Yes, I think that's the way to go.

gaearon:
You can try exploring this API:

```js
@connect((state, bindActionCreators) => ({
  counter: state.counter,
  ...bindActionCreators(CounterActions)
}))
export default class Counter {
  render() {
  }
}
```

```js
function select(state, bindActionCreators) {
  return {
    todos: state.todos,
    actions: bindActionCreators(TodoActions)
  };
}

export default class TodoApp {
  render() {
    return (
      <Connector select={select}>
        {this.renderChild}
      </Connector>
    );
  }

  renderChild({ todos, actions }) {
    return (
      <div>
        <AddTodo {...actions} />
        <TodoList todos={todos} {...actions} />
      </div>
    );
  }
}
```

You'll need to figure out some way of memoizing the bound functions though, so that if the `dispatch` function is the same, and passed action creators are the same, they're not re-bound on each `select` call.

gaearon:
What I like about `select(state, bindActionCreators)`: one-stop shop for the Flux connect thing both for state and actions in one function. Easy to shape the resulting props the way you want to.

What I don't like: “smart” memoization logic it would need to be efficient. I'd rather have `select(state, dispatch)`, but if `bindActionCreators` was a separate utility, there would be no efficient way for memoizing.

```js
// I wish I could make this fast
function select(state, dispatch) {
  return {
    todos: state.todos,
    actions: bindActionCreators(TodoActions, dispatch)
  };
}
```

gaearon:
Also, as for my last suggestion—it's actually not much better than what we have now already.
So we're back to what you proposed initially..

src/components/createConnector.js
```diff
@@ -62,12 +68,27 @@ export default function createConnector(React) {
       return { slice };
     }
 
+    bindActionCreators(dispatch) {
+      if (!this.props.actionCreators) {
+        return {};
```
gaearon: I think a sane default is the current behavior: injecting just the `dispatch` prop. Let's remove this early exit, and instead specify `dispatch => ({ dispatch })` as the default prop value for `actionCreators`.

gaearon:
What is the use case for `actionCreators`-as-a-function? Do we need it? It seems to me that middleware already allows `dispatch`-level customization, so at this point I'd rather have `actionCreators` strictly be an object.

It's easy to pick just what you want for it:

```
import { increment } from './CounterActions';
import * as TodoActions from './TodoActions';

<Connector actionCreators={{ increment, ...TodoActions }}>
```

skevy:
I was thinking of namespaced actions when I thought of `actionCreators` as a function...something like:

```js
import * as CounterActions from './CounterActions';
import * as TodoActions from './TodoActions';

<Connector actionCreators={(dispatch) => ({
    counter: bindActionCreators(CounterActions, dispatch),
    todo: bindActionCreators(TodoActions, dispatch)
}) />
```

Anyway, of course this suffers from that `shouldComponentUpdate` problem as well, and complicates things. It's not really necessary.

skevy:
> Also, as for my last suggestion—it's actually not much better than what we have now already.
So we're back to what you proposed initially..

Were you referring to having the `select` function combine the actions and other props...or were you talking about https://github.com/gaearon/redux/pull/86#issuecomment-111742272?

gaearon:
Both. I think burdening `select` with this is not much better than doing it in `render` in the first place..
It makes `select` too magic because of the memoization trickery it would need.

I'm starting to like your original suggestion, given that you add some memoization, and remove the function overload. I think that, instead of namespacing, it's better to merge the actions, and if there are conflicts, you can always assemble a custom object and rename the functions as you need.

skevy:
Talk to me a little bit about how hot loader works with decorators right now...it reinitializes the whole component on every change, am I right?

gaearon:
>Talk to me a little bit about how hot loader works with decorators right now...it reinitializes the whole component on every change, am I right?

Yeah, that's the problem with decorators. I'll solve it eventually with [this plugin](https://github.com/gaearon/babel-plugin-react-hotify) but for now I recommend using container components for that very reason.

skevy:
As I go through this memoization process...I'm noticing this potential problem:

```javascript
<Connector actionCreators={{ ...CounterActions, ...TodoActions }}>
{() => [things]}
</Connector>
```

That's going to create a new object every `render()`...thus causing a rebind.

It would require the user to implement storing these elsewhere...which is completely fine...I think we just need to be sure we mention this somewhere.

gaearon:
>That's going to create a new object every render()...thus negating any prop checking in Connector and causing a rebind.

I'd say it's likely that the developer would measure where `shouldComponentUpdate` is failing if they have a performance problem, and notice that it's due to this object, and figure out to move it outside `render`.

```js
const actions = { ...CounterActions, ...TodoActions };

<Connector actionCreators={actions}>
  {() => [things]}
</Connector>
```

I think we can write an “Optimizing” doc later, but it's too early at this point as we're still figuring out the API.

gaearon:
(The same argument can be made for `select`. It's possible to write it as a lambda, and have potential perf problems because of that, but it would catch the eye of seasoned React dev looking for performance problems, so we're *just* like the other libraries in this regard.)

skevy:
Fair enough :)

skevy:
Just pushed a change the saves the bound action creators and only updates them when passed action creators changed. I know you also mentioned making sure they get updated when the dispatch method changes, which if I'm not mistaken, only happens when the entire redux instance is re-created, am I right? When would that happen? Only in hot-reloading?

gaearon:
>When would that happen? Only in hot-reloading?

Yeah, I think so.

gaearon:
Great. Can I ask you to update the examples and README to show how this API makes Redux better?

skevy:
Absolutely. Was already working on it.

gaearon:
Thanks!

gaearon:
Hi, do you need help there? Looking forward to seeing this!

skevy:
Sorry sorry. Crazy day at work today. I'll see if I can get to this later on.

gaearon:
Yeah no problem. Let us know if you don't have time, I'm sure somebody can finish that one. (Maybe even I can when I get some free time ;-)

skevy:
There's a first pass at the README update (and I updated the examples). We can probably say more if we want to break out of the "commenting code" model of the README and into more explanation...but I'm not sure it's necessary. Happy to continue adding more info if you feel we need to.

README.md
```diff
@@ -4,9 +4,9 @@ redux
 [![build status](https://img.shields.io/travis/gaearon/redux.svg?style=flat-square)](https://travis-ci.org/gaearon/redux)
 [![npm version](https://img.shields.io/npm/v/redux.svg?style=flat-square)](https://www.npmjs.com/package/redux)
 
-An experiment in fully hot-reloadable Flux.
```
gaearon: Whitespace is significant in Markdown :-). I have these double spaces for a reason here. Can you please change them back?

README.md
```diff
@@ -158,11 +156,13 @@ function select(state) {
 export default class CounterApp {
   render() {
     return (
-      <Connector select={select}>
-        {({ counter, dispatch }) =>
+      // Passing an object of action creators here automatically binds them 
+      // with Redux's dispatcher and passes them to the child function
+      // (so they can be passed as props to child components)
+      <Connector select={select} actionCreators={CounterActions}>
+        {({ counter, actions }) =>
```
gaearon: Shouldn't this be `actionCreators` for consistency then? cc @acdlite

dariocravero: Quick question, isn't `select` a bit too generic? Flummox is calling the `stores` and `actions`, I think the distinction is worth it. The way I see it, it makes the overall code simpler to read and follow. Am I missing something and is there any specific reasons for the current names?

gaearon: I think if we call it “stores” people will be tempted to pass actual Store functions there.

dariocravero: I see your point, I just think that `select` is probably too broad and can work for both: actions and stores and that might be confusing too

gaearon:
I would like to have some more eyes on this API-wise.
Is there anything we can do to make/keep the API simpler?

@ooflorent @acdlite @johanneslumpe?

gaearon:
What about the case where a component *only* needs to fire actions, but doesn't care about stores? The default `select` function would then have it to subscribe to every single root atom change. This doesn't seem good.

dariocravero:
Perhaps `select` should be explicitly set at all times?

gaearon:
@dariocravero 

But what would you specify in case you only want the actions? `select={() => ({})}`? Sounds like working around the problem.

Here's another proposal: use a special component for binding actions.

#### Decorator version

```js
@bind(CounterActions)
@connect(state => ({
  counter: state.counter
}))
  render() {
    return (
      <Counter {...this.props} />
    );
  }
}
```

#### Container version

```js
export default class TodoApp {
  render() {
    return (
      <Bind actionCreators={TodoActions}>{(actionCreators) =>
        <Connect select={state => ({ todos: state.todos })}>{({ todos }) =>
          <div>
            <AddTodo {...actionCreators} />
            <TodoList todos={todos} {...actionCreators} />
          </div>
        }</Connect>
      }</Bind>
    );
  }
}
```

It would do a single thing, it would not complicate the existing `@connect` API by introducing a second parameter, and it would be usable directly without `<Connect>` if the user doesn't need to observe state, but needs to fire actions in a particular component.

Thoughts?

gaearon:
Don't forget the user can always create their own `<Container>` component that unifies `<Bind>` and `<Connect>` if they wish so.

johanneslumpe:
I think that we definitely need a way to allow a component only to dispatch actions but to not get notified by store changes. Of course it's very useful for us lazy people that we can just use `@connect` without a select function and get all the state automatically injected, but it as soon as this brings performance-implications with it, that's not great.

What about if `connect` were to work like this:
```javascript
@connect({
 select: (state) => {...},
 actions: actionObject
})
```
The `actionObject` could automatically be spread as props onto the wrapped component and this would allow users to specify whether they want to select data or not. I would rather just pass actions verbatim, without binding them, too.

dariocravero:
@gaearon the fact that state isn't set should be enough, shouldn't it?
@johanneslumpe like flummox :)
@gaeron's approach of having two decorators/components is probably a good thing anyway to separate responsibilities and minimise the impact

johanneslumpe:
That would work, yeah. I'm just not very keep on binding the actions, as it introduces some overhead. But  since that stuff is optional anyway...

gaearon:
>What about if connect were to work like this:

That's how it used to be but I really dig the simplicity of a single function parameter. I feel like having `select` and `actions` in one place conflates two different needs: a real need for subscribing to updates, and a cosmetic need to bind actions. I'd rather have separate unopinionated components, and maybe some kind of `<Container>` that combines them if people want ready solutions.

johanneslumpe:
@gaearon fair enough. maybe instead of `bind` it could be named `injectBoundActions` or something like that? It would be nice if the name would reflect both the things this decorator does.

README.md
emmenko: Maybe something with a more contextual name (e.g.: `selectStoresByName`) ?

emmenko:
I like the `decorator` solution

leoasis:
:+1: to the separate decorators and components

tsingson:
I like the `decorator` solution too, and ,
```
@connect({
 select: (state) => {...},
 actions: actionObject
})

```
is good and clear.
look like   `connect` is automatic bridge between actions and stores for component.

the render function in the `decorator' version:
```
  render() {
    return (
      <Counter {...this.props} />
    );
  }
}
```
is more familiar.

skevy:
To those of you who are saying not to bind the actions...why not? The overhead is muted by the fact that we're memoizing the bind operation unless 1) the actions change or 2) the redux instance changes.

skevy:
As far as the multiple decorator/component approach, I'm down with that. I think having the `<Container>` would be useful as well...but perhaps unnecessary.

One problem that I've run into when using multiple decorators on a component is that if I need access to the ref of the component I'm decorating, many decorators makes that difficult to do (without touching internal APIs). So if we just keep adding decorators upon decorators (as awesome as they are)...it can break component-to-component communication (a good use case for this, before everyone tells me I'm an idiot and shouldn't access components this way...is animation, where an inner component knows how to tween itself, but the parent container component actually handles synchronizing the tweens). It's very possible this doesn't matter...and there may be a BETTER way to handle this type of use case...but just thought I'd bring it up.

I can work later on the multiple decorator approach, if that's the direction we want to explore.

gaearon:
I'd like other collaborators to express their thoughts here and give it some time to rest.
I'll get back to re-evaluating after my current time travel work.

skevy:
thought I'd ping this to see if any of the other collabs have had time to review? I'm not particularly attached to my idea over anyone else's, but it would be cool to get something like the ideas presented merged before 1.0

gaearon:
I'll definitely consider this before 1.0. :-)

guillaume86:
I'm all for https://github.com/gaearon/redux/pull/86#issuecomment-112386904
Before finding this thread I was thinking actions could be bound at the Provider level and retrieved with some decorator in child components but this is even better.

cef62:
I give my +1 to the decorator solution, could be really useful!

gaearon:
I'd love to hear more proposals for the alternative to the current `<Provider>` and `<Connector>` API: https://github.com/gaearon/react-redux/issues/1

gaearon:
Here's a more radical variation on this, with a very similar API but slightly different prescribed usage.
It also completely drops `<Connector>` in favor of `connect()`: https://github.com/gaearon/react-redux/issues/1#issuecomment-120677352

I'd appreciate your comments and ideas there.

gaearon:
I'm closing this in favor of https://github.com/gaearon/react-redux/issues/1#issuecomment-120729921.
Let me know if you'd like to implement it!

gaearon:
This is done in https://github.com/gaearon/react-redux/releases/tag/v0.5.0.

skevy:
Wooooo :)

Issue #83, opened 2015-06-13
Investigate Flow support


acdlite:
The ES6 support for Flow seems to have improved quite a bit. Not sure if it covers every feature we're currently using (e.g. the bind operator), but it may be worth giving up a few niceties in exchange for type checking.

https://github.com/facebook/flow/blob/master/Changelog.md#v0120

gaearon:
Agreed. Type-checking the insides would be awesome.

emmenko:
Good idea, I was thinking the same after seeing latest release...

gaearon:
Superseded by #290.

PR #87, opened 2015-06-13
Flow type annotations

acdlite:
Opening this PR to track adding Flow annotations. So far I've completed Redux.js. Looks like the biggest thing we'll have to give up is `const`, as Flow supports neither that nor `let`. I think that trade off is worth it.

We'll also want to integrate Flow into our automated tests.

Prior to this my experience with Flow was the "Hello, World!" example on their website, so if I do something wrong please let me know :)

Addresses #83.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -14,7 +14,8 @@
     "strict": [2, "never"],
     "react/jsx-uses-react": 2,
     "react/jsx-uses-vars": 2,
-    "react/react-in-jsx-scope": 2
+    "react/react-in-jsx-scope": 2,
+    "no-unused-vars": 0
   },
   "plugins": [
     "react"
--- a/.flowconfig
+++ b/.flowconfig
@@ -5,6 +5,5 @@
 [include]
 
 [libs]
-./type-definitions
 
 [options]
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,10 @@
+[ignore]
+.*/lib
+.*/test
+
+[include]
+
+[libs]
+./type-definitions
+
+[options]
--- a/src/Redux.js
+++ b/src/Redux.js
@@ -1,9 +1,19 @@
+/* @flow */
+
 import createDispatcher from './createDispatcher';
 import composeStores from './utils/composeStores';
 import thunkMiddleware from './middleware/thunk';
 
+type Dispatch = (action: mixed) => mixed;
+type Dispatcher = (state: mixed, setState: (nextState: mixed) => mixed) => Dispatch;
+
 export default class Redux {
-  constructor(dispatcher, initialState) {
+  state: mixed;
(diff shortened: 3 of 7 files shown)


gaearon:
Everything is `mixed` haha.

I like this. Want to see more :-)

acdlite:
@gaearon Haha, that's a good thing in this case! Not making any assumptions about action or state types :)

acdlite:
@gaearon How about this: I'll create a `types.js` file with all the basic types we deal with in Redux. We will have an Action and State type that simply aliases to `mixed`, for clarity.

gaearon:
I like this.

acdlite:
Still can't figure out how to use union types to implement function overloading.

skevy:
I haven't tested this...but have you guys considered running the code base through babel first (and blacklisting flow so that it keeps the types) and then running it through the flow type checker? If that worked; it would allow all ES2015+ features to stay and still get the benefits of type checking.

skevy:
Looks like this guy is doing it http://www.keendevelopment.ch/flow-babel-gulp-es6/

ooflorent:
Why did you convert `const` to `var`?

gaearon:
>Why did you convert const to var?

See the first post:

>Looks like the biggest thing we'll have to give up is const, as Flow supports neither that nor let.

Although it may be possible to do it some another way:

http://www.keendevelopment.ch/flow-babel-gulp-es6/

acdlite:
I'm kind of stuck on this PR until someone can explain to me how to properly use union types to implement function overloading. Any Flow experts out there?

taylorhakes:
I am not very familiar with Flow, but the issue is being caused by this line in createDispatcher.
```
return compose(...middlewares, dispatch);
```
If you take out dispatch, it stops complaining. I am not sure this is an issue with function overloading, but I am having a hard time figuring out what it wants.

I noticed https://github.com/gaearon/redux/pull/166 will conflict with this pull request. The issue may be solved because the createDispatcher function is being removed. 

Do you want me to take this over or would you like to continue? I can fix the code conflicted, add the code not covered in 166 and resubmit.

acdlite:
We should probably wait until 1.0 when the API is stable. After that, feel free!

winkler1:
Here's a way to run Flow when files change... very useful when tweaking the config, fast feedback loop.

`npm i --save-dev onchange && ./node_modules/onchange/cli.js '**/*' -- flow check`

hzoo:
@acdlite About function overloading - http://flowtype.org/docs/union-intersection-types.html you can use intersections (`&`)

```js
// Intersections are well-suited to mimic function overloading.
declare var f: ((x: number) => void) & ((x: string) => void);
```

gaearon:
Superseded by #87.

gaearon:
Whoops, inception! I meant by #254.

PR #94, opened 2015-06-15
use shelljs for shell like commands

mikekidder:
This is preliminary;  unsure of the **test:cov** and **test:watch**..  The latter is causing a loop

### DIFF
--- a/scripts/prepublish.js
+++ b/scripts/prepublish.js
@@ -1,0 +1,13 @@
+var sh = require('shelljs');
+
+sh.exec('eslint src test', function(code, output) {
+
+  if(code === 0){
+    sh.exec('npm run clean');
+    sh.exec('npm run build');
+  } else {
+    sh.echo('Exit code:', code);
+    sh.echo('Program output:', output);
+  }
+
+});
--- a/scripts/prepublish.js
+++ b/scripts/prepublish.js
@@ -1,13 +1,12 @@
 var sh = require('shelljs');
 
 sh.exec('eslint src test', function(code, output) {
-
-  if(code === 0){
+  if(code === 0) {
     sh.exec('npm run clean');
     sh.exec('npm run build');
+    sh.exec('npm run test');
   } else {
     sh.echo('Exit code:', code);
     sh.echo('Program output:', output);
   }
-
 });
--- a/package.json
+++ b/package.json
@@ -4,14 +4,14 @@
   "description": "An experiment in fully hot-reloadable Flux",
   "main": "lib/index.js",
   "scripts": {
-    "build": "scripts/build",
-    "clean": "scripts/clean",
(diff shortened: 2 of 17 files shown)


mikekidder:
Also note that ansi codes currently will not come through on output.

gaearon:
@mikekidder Do you need help fixing those problems?

mikekidder:
1) **test:cov** - may be a mocha issue w/windows, doesn't appear you are having error. Correct?
2) **test:watch** - may be related to **test:cov**, but the --watch may probablematic for shelljs
3) colors output - shelljs using **child-process** internally; known issue https://github.com/arturadib/shelljs/issues/86

If you want to get away from npm script, as you mention in #94, could also look into gulp.

gaearon:
Woah, color support is a bummer! The issue has been up for two years..

Can you help me with understanding what exactly fails on Windows? How are shell scripts different from the same commands right inside `package.json`? Why do these commands work, but shell scripts don't work?

mikekidder:
No difference in running npm script vs. shelljs.  The test scripts were put in after my initial PR's, and hadn't ran them yet until yesterday.

This fixes **test:cov** as mentioned in https://github.com/gotwarlost/istanbul/issues/90. Path to mocha should work on Mac as well.   And looks like **test:watch** is working.

This really leaves no color output as the issue.

gaearon:
Thanks! I'll take another look later today.

scripts/prepublish.js
```diff
@@ -0,0 +1,13 @@
+var sh = require('shelljs');
+
+sh.exec('eslint src test', function(code, output) {
+
+  if(code === 0){
```
gaearon: Nit pick: missing space before `{` and unneeded newlines before and after `if` block.

scripts/prepublish.js
```diff
@@ -0,0 +1,13 @@
+var sh = require('shelljs');
+
+sh.exec('eslint src test', function(code, output) {
+
```
gaearon: Let's also run the tests here.

gaearon:
This PR should also remove the old scripts, right?

gaearon:
Sorry for dragging my feet here. I'm very busy atm, but will be more available in a week or so.
I'll definitely either merge this or go back to npm scripts.

Just give me some more time before looking at it!
Cheers

mikekidder:
Oh, no worries from me. I know how to work around the shell scripts. :)  
FWIW, the npm scripts worked and has color output.

gaearon:
Closing in favor of #267.   
Nevertheless thanks for working on this!

gaearon:
Fixed by #267, the fix is merged into `breaking-changes-1.0` branch. Feel free to give it a go!

PR #95, opened 2015-06-15
Update README.md

DenisIzmaylov:


gaearon:
I don't mind omitting parens when there is a single argument.

Do you folks think it reads better?

emmenko:
I would also prefer no brackets for only one argument

DenisIzmaylov:
Yes, I think it's more readable for almost people.
But we can vote in the community.
I fixed it according https://github.com/airbnb/javascript#arrow-functions which is basically using is default code style. Or do you prefer other?

gaearon:
Let's go with Airbnb's style guide. But if we do that, I'd like to see eslint configuration to enforce as much of it as possible, and also update the source code.

gaearon:
Closing in favor of #97.

PR #98, opened 2015-06-15
Test shallowEquality

vramana:


src/utils/shallowEqual.js
```diff
@@ -19,12 +19,6 @@ export default function shallowEqual(objA, objB) {
       return false;
     }
 
-    var valA = objA[keysA[i]];
-    var valB = objB[keysA[i]];
-
-    if (valA !== valB) {
-      return false;
-    }
   }
```
vramana: This is dead code. This is already present in the if statement.

emmenko:
The file should be named `shallowEquality.spec.js`

vramana:
@emmenko Okay.Sorry I didn't notice it.

emmenko:
Np :)

src/utils/shallowEqual.js
gaearon: Yep.

test/utils/shallowEquality.js
```diff
@@ -0,0 +1,68 @@
+import expect from 'expect';
+import shallowEqualScalar from '../../src/utils/shallowEqualScalar';
+import shallowEqual from '../../src/utils/shallowEqual';
+
+describe('Utils', () => {
+  describe('shallowEqualScalar', () => {
+    it('should compare the given arguments shallowly', () => {
+
+      expect(shallowEqualScalar(null, undefined)).toBe(false);
```
gaearon: Let's only test things whose `typeof` is `object`. This includes `null`, but not `undefined`.

test/utils/shallowEquality.js
```diff
@@ -0,0 +1,68 @@
...
+      expect(shallowEqualScalar(a, a)).toBe(true);
+      expect(shallowEqualScalar(a, b)).toBe(true);
+
+    });
+  });
+
+  describe('shallowEqual', () => {
+    it('should compare the given objects shallowly', () => {
+
+      // expect(shallowEqual(null, undefined)).toBe(true);
```
gaearon: Here, too, we should assume `typeof` of whatever passed is `object`. This includes `null`, but not `undefined`.

test/utils/shallowEquality.spec.js
```diff
@@ -0,0 +1,71 @@
+import expect from 'expect';
+import shallowEqualScalar from '../../src/utils/shallowEqualScalar';
+import shallowEqual from '../../src/utils/shallowEqual';
+
+describe('Utils', () => {
+  // More info: https://github.com/gaearon/redux/pull/75#issuecomment-111635748
+  describe('shallowEqualScalar', () => {
+    it('should compare the given objects shallowly', () => {
```
vramana: @gaearon I don't know how to concisely describe shallowEqualScalar. Help please? or just leave it be since I have left the comment?

test/utils/shallowEquality.spec.js
```diff
@@ -0,0 +1,71 @@
...
+
+      expect(shallowEqualScalar(m, n)).toBe(false);
+
+      const a = {
+        one: 1,
+        two: 2
+      };
+
+      const b = a;
+      b.three = 3;
```
gaearon: Hmm, this would mutate `a` object. I don't think it's intended in your test?

test/utils/shallowEquality.spec.js
gaearon: “should compare shallowly but bail out on object fields”

test/utils/shallowEquality.spec.js
```diff
@@ -0,0 +1,71 @@
...
+      expect(shallowEqualScalar([], [])).toBe(true);
+      expect(shallowEqualScalar([], [1, 2])).toBe(false);
+      expect(shallowEqualScalar([3, 4], [3, 4])).toBe(true);
+      expect(shallowEqualScalar([3, 4], [3, 5])).toBe(false);
+      expect(shallowEqualScalar([[]], [[]])).toBe(false);
+      expect(shallowEqualScalar([m], [m])).toBe(false);
+
+      const n = {two: 2};
+
+      expect(shallowEqualScalar(m, n)).toBe(false);
```
gaearon: I'd prefer to put these objects inline into method calls so it's easier to see what's being compared.

test/utils/shallowEquality.spec.js
```diff
@@ -0,0 +1,71 @@
...
+      expect(shallowEqualScalar({}, null)).toBe(false);
+      const m = {one: 1};
+
+      expect(shallowEqualScalar([], [])).toBe(true);
+      expect(shallowEqualScalar([], [1, 2])).toBe(false);
+      expect(shallowEqualScalar([3, 4], [3, 4])).toBe(true);
+      expect(shallowEqualScalar([3, 4], [3, 5])).toBe(false);
+      expect(shallowEqualScalar([[]], [[]])).toBe(false);
+      expect(shallowEqualScalar([m], [m])).toBe(false);
+
```
gaearon: There's a lot of tests for the arrays, but I'd expect to see more tests with objects.
For some inspiration: https://github.com/facebook/react/blob/master/src/shared/utils/__tests__/shallowEqual-test.js

test/utils/shallowEquality.spec.js
```diff
@@ -0,0 +1,138 @@
+import expect from 'expect';
+import shallowEqualScalar from '../../src/utils/shallowEqualScalar';
+import shallowEqual from '../../src/utils/shallowEqual';
+
+describe('Utils', () => {
+  // More info: https://github.com/gaearon/redux/pull/75#issuecomment-111635748
+  describe('shallowEqualScalar', () => {
+    it('returns true if both arguments are the same object referentially', () => {
+      const o = {a: 1, b: 2};
```
gaearon: Please tweak the style to use spaces inside object definitions. (`{a : 1}` => `{ a: 1 }`). This doesn't concern arrays, just objects.

test/utils/shallowEquality.spec.js
```diff
@@ -0,0 +1,138 @@
...
+      expect(shallowEqualScalar(o, o)).toBe(true);
+    });
+
+    it('returns false if either argument is null', () => {
+      expect(shallowEqualScalar(null, {})).toBe(false);
+      expect(shallowEqualScalar({}, null)).toBe(false);
+    });
+
+    it('returns false if only one argument is not an object', () => {
```
gaearon: Do we need this test? The way we use them in Redux, we make a hard assumption that both arguments have `'object'` `typeof`.

vramana:
@gaearon Can you take a look now ?

ooflorent:
Currently the following test fails:

```js
expect(shallowEqual({a: NaN}, {a: NaN})).toBe(true)
```

Maybe we should fix that.

src/utils/shallowEqual.js
```diff
@@ -19,12 +19,6 @@ export default function shallowEqual(objA, objB) {
       return false;
     }
```
ooflorent: Can you remove this line?

emmenko:
@ooflorent :+1:

vramana:
@ooflorent Done. Anything else?

PR #101, opened 2015-06-15
remove rimraf

chicoxyzzy:
because we can just `rm -rf`


gaearon:
We're actually probably just going to switch to shelljs: #94
Rimraf was there for Windows compatibility but I broke it later (hence #94)

PR #107, opened 2015-06-16
Connector#componentWillReceiveProps should use the new value of `select`

gaearon:
This fixes a bug that causes the state passed by `Connector` to stay stale when a new `select` function is passed. This PR includes a failing test and a fix that makes it pass.

### DIFF
--- a/test/components/jsdomReact.js
+++ b/test/components/jsdomReact.js
@@ -1,0 +1,7 @@
+import ExecutionEnvironment from 'react/lib/ExecutionEnvironment';
+import jsdom from 'mocha-jsdom';
+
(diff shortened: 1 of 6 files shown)


test/components/jsdomReact.js
```diff
@@ -0,0 +1,7 @@
+import ExecutionEnvironment from 'react/lib/ExecutionEnvironment';
+import jsdom from 'mocha-jsdom';
+
+export default function jsdomReact() {
+  jsdom();
+  ExecutionEnvironment.canUseDOM = true;
```
gaearon: Can't use `setState` if I don't add this.

PR #110, opened 2015-06-16
Gracefully handle nested dispatches

leoasis:
Currently, when we happen to have nested dispatches, the last action's returned state prevails, even if they belong to different stores. The nested dispatch still sees the state as if the outer dispatch wasn't executed yet, so it ends up overwriting whatever that outer dispatch did.

Perhaps it's more clear with the failing test that I'm providing here. I'd expect there to either have both dispatches applied, or the dispatcher to throw an exception or a warning, as the original Facebook's Flux dispatcher.

The use case for running into nested dispatches is the classic "need data in componentWillMount", which triggers an action creator that dispatches an action immediately, and is executed as a consequence of a previous dispatch.

I'm leaving this unsolved until we decide the way to go, after that I can fix the test with the expected behavior

### DIFF
--- a/test/createRedux.spec.js
+++ b/test/createRedux.spec.js
@@ -69,4 +69,29 @@ describe('createRedux', () => {
 
     expect(state).toEqual(redux.getState().todoStore);
   });
+
+  it('should handle nested dispatches gracefully', () => {
+    function foo(state = 0, action) {
+      return action.type === 'foo' ? 1 : state;
+    }
+
+    function bar(state = 0, action) {
+      return action.type === 'bar' ? 2 : state;
+    }
+
+    const redux = createRedux({ foo, bar });
+
+    redux.subscribe(() => {
+      // What the Connector ends up doing.
+      const state = redux.getState();
+      if (state.foo !== 0) {
+        redux.dispatch({type: 'bar'});
+      }
+    });
+
+    redux.dispatch({type: 'foo'});
+
+    // Either this or throw an error when nesting dispatchers
+    expect(redux.getState()).toEqual({foo: 1, bar: 2});
+  });
 });


gaearon:
Thanks for the failing test case. Best way to start a conversation :-)

>The use case for running into nested dispatches is the classic "need data in componentWillMount"

Yes. Let's solve this somehow.

I'm currently busy with time travel but I'm sure @acdlite, @johanneslumpe, @goatslacker could provide valuable insight from their experience building Flux libraries (even on top of FB's dispatcher)..

johanneslumpe:
I do not have a solution for this now but I remember that nested dispatches are something that's difficult to handle on the server-side. Let's say we have the classic react-router example where we want to fetch some data prior to rendering. We added some static handlers to our container components, and inject redux etc etc. to fetch some data. Let's use a promise middleware too. It returns the promise from `dispatch` so we can wait for all async actions to be completed prior to rendering. If we now have dispatches in `componentWillMount` which might be async as well, then those would not be caught and we would render our application with possibly half the state we'd need/want.

gaearon:
We could do this:

* Throw on nested dispatch just like FB dispatcher.
* Let user specify a custom “update scheduler” function to `createRedux`. They can use it to wrap the dispatch in `React.addons.batchedUpdates` (#125) -or- to schedule all updates inside `requestAnimationFrame`. This way dispatching from lifecycle hooks won't be an issue.

Am I making sense?

johanneslumpe:
I think we should probably just throw for now, so that a user at least knows what's going on and isn't stumped by the fact that the state isn't what they expected.

I do like the the `requestAnimationFrame` idea too. The question is whether it is a good thing to allow that, because those applications would most likely just explode once somebody tries to change to a "normal" scheduler again. But then again, if you do that you most likely know what you're in for.

leoasis:
I _think_ the problem with the last action overriding the entire atom of the first dispatch gets fixed with https://github.com/gaearon/redux/pull/119, since the problem is that the dispatcher is using its internal state to update, which is set _after_ setting the state in the redux instance's state and calling the listeners.

I'm saying this because after that PR gets merged, the nested dispatches would set the whole state as expected, so it could be an option for desired behavior.

Having said that, I'm not sure if this would cause any edge cases, or if it will increase the mental model both on developing redux (taking into account supporting nested dispatches), and in userland.

leoasis:
@gaearon @johanneslumpe just to let you know that it seems `rAF` may have further issues: https://github.com/facebook/react/issues/3570. In any case, that would be an optimization or a strategy to be decided by the user

gaearon:
>I think we should probably just throw for now, so that a user at least knows what's going on and isn't stumped by the fact that the state isn't what they expected.

@leoasis Would you like to implement this as part of this PR?

leoasis:
@gaearon yes, sure! Have you considered what I said in [this](https://github.com/gaearon/redux/pull/110#issuecomment-112799258) previous comment? Just to make sure we _don't_ want to go with that route

gaearon:
@leoasis 

Oh, I haven't. Hmm.

@acdlite Any opinion here?

acdlite:
This is actually caused by a subtle bug in the current version where `setState()` triggers the subscribers before assigning to the dispatcher's internal copy of the current state. https://github.com/gaearon/redux/blob/master/src/createDispatcher.js#L9

This is kind of bug that https://github.com/gaearon/redux/pull/119 is intended to avoid by moving all state into the Redux instance. I've added the test to that PR to show that it passed. (Modified slightly — the original version causes an infinite loop because this line is always true:  https://github.com/leoasis/redux/commit/e23b56bf295e0cb649927425692ef85ae408991f#diff-2dce452f5f745e87ca49080c71d81a21R87.)

See the passing test here: https://github.com/gaearon/redux/commit/b71a7fc0ca64a8ce86b67b178ab602d065c6492e

acdlite:
So long story short, nested dispatches in Redux aren't a problem like they are in other libraries, because our "dispatcher" is just a reduce operation.

leoasis:
@acdlite awesome, yeah that's what I thought in https://github.com/gaearon/redux/pull/110#issuecomment-112799258, so let's keep nested dispatches supported and fix that bug in your PR!

Only thing curious about why you had an infinite loop in my test, since that was not happening without the fix. Just asking since it may be a hint of another bug perhaps

acdlite:
@leoasis The infinite loop never kicked off in the original case for the same reason the test failed: the call to `getState()` inside the subscriber returned the old (original) state, so the nested dispatch was never fired.

leoasis:
@acdlite yeah, realised that right after sending the comment :P

acdlite:
@leoasis :)

leoasis:
Perfect then, since this will be fixed in https://github.com/gaearon/redux/pull/119, I'm closing this one

Issue #88, opened 2015-06-13
Investigate the library size and dependencies


gaearon:
I haven't given much attention to library size or dependencies yet, as we're still figuring out the right API. However, as people start to depend on it, it's a good idea to check whether we're dragging unnecessary polyfills (babel-runtime?) when it's easily avoidable, and whether there's something we could do to reduce the library size (we can start by adding a UMD build and measuring it! ;-)

I'm currently busy but I'd be happy to see somebody investigate this. If you decide to do it, please let us know in this issue!

dariocravero:
@gaearon I think we can close this now :)

PR #111, opened 2015-06-16
Remove unnecessary browserify config in package.json and envify dependency

dariocravero:
This PR removes the `browserify` config in `package.json` and the dependency to `envify` (#88) which doesn't seem to be used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,13 +54,7 @@
   },
   "dependencies": {
     "babel-runtime": "^5.5.8",
-    "envify": "^3.4.0",
     "invariant": "^2.0.0",
     "lodash": "^3.9.3"
-  },
-  "browserify": {
-    "transform": [
-      "envify"
-    ]
   }
 }


dariocravero:
(let me clean the commit history before merging it in)

cdevienne:
Unless I am mistaken, the browserify+envify config is useful.

In my case, when I assemble my app with browserify, even if I setup envify in my own project, the substitution is not done in the redux source. The result is that the dev version of redux, and not the production one, is enabled in my application.

To solve this, the only solution I found so far is to revert this patch on my locally installed redux.

Could you consider re-add the option ? The extra dependency looks not a big issue since react already pulls it for the same reason.

Also, you may want to use loose-envify instead of envify, because it is much faster.

gaearon:
I totally forgot about this PR! We have re-added the config since then. Now I understand how Browserify works a bit better :-). Apparently many people don't run envify with a global flag in their project and this is why they used to get development build instead of production build of Redux.

PR #120, opened 2015-06-17
Pass dispatch to higher-order middleware and re-implement thunk

acdlite:
See https://github.com/gaearon/redux/issues/113#issuecomment-112603386 for context.

### DIFF
--- a/src/createDispatcher.js
+++ b/src/createDispatcher.js
@@ -14,7 +14,7 @@ export default function createDispatcher(store, middlewares = []) {
     }
 
     const finalMiddlewares = typeof middlewares === 'function' ?
-      middlewares(getState) :
+      middlewares({ getState, dispatch }) :
       middlewares;
 
     return composeMiddleware(...finalMiddlewares, dispatch);
--- a/src/middleware/thunk.js
+++ b/src/middleware/thunk.js
@@ -1,10 +1,6 @@
-export default function thunkMiddleware(getState) {
-  return (next) => {
-    const recurse = (action) =>
-      typeof action === 'function' ?
-        action(recurse, getState) :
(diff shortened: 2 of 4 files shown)


src/middleware/thunk.js
```diff
@@ -1,10 +1,6 @@
-export default function thunkMiddleware(getState) {
-  return (next) => {
-    const recurse = (action) =>
-      typeof action === 'function' ?
-        action(recurse, getState) :
-        next(action);
-
-    return recurse;
-  };
+export default function thunkMiddleware(getState, dispatch) {
```
gaearon: Oh, so if the thunk ain't the first one, `dispatch` will re-start the chain right?

src/createDispatcher.js
```diff
@@ -14,7 +14,7 @@ export default function createDispatcher(store, middlewares = []) {
     }
 
     const finalMiddlewares = typeof middlewares === 'function' ?
-      middlewares(getState) :
+      middlewares({ getState, dispatch }) :
```
leoasis: Isn't the idea to pass the full dispatch middleware chain instead of the default fn?

acdlite: D'oh, you're right. I'll update.

src/middleware/thunk.js
acdlite: Yep!

src/createDispatcher.js
acdlite: Actually I'll update it if/when we merge #119 since there will be conflicts.

gaearon:
Implemented in #195. (Sorry, it was a pain to merge so I just copied the idea.)

PR #126, opened 2015-06-17
Ignorning non-func values in composeStores

### LINKED ISSUE #124: Let `composeStores` ignore everything that isn't a function
If using loose mode with babel, `composeStores` trips up on `__esModule:true`.

johanneslumpe:
Fixes #124


gaearon:
:+1:

emmenko:
:+1:

PR #127, opened 2015-06-17
Give the initial action a special type

gaearon:
Supersedes #114.

I want to establish a convention for “built-in” and “devtools” action types (more will come in #113) so that the tools are able to distinguish them, and so the user doesn't feel compelled to handle them and can clearly see that they are not from their app.

Thanks to @ooflorent for the `@@` naming tip.

### DIFF
--- a/src/createDispatcher.js
+++ b/src/createDispatcher.js
@@ -1,8 +1,12 @@
 import composeMiddleware from './utils/composeMiddleware';
 
+const INIT_ACTION = {
+  type: '@@INIT'
+};
+
 export default function createDispatcher(store, middlewares = []) {
   return function dispatcher(initialState, setState) {
-    let state = setState(store(initialState, {}));
+    let state = setState(store(initialState, INIT_ACTION));
 
     function dispatch(action) {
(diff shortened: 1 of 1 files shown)


src/createDispatcher.js
```diff
@@ -1,8 +1,12 @@
 import composeMiddleware from './utils/composeMiddleware';
 
+const INIT_ACTION = {
+  type: '@@INIT'
```
ooflorent: Camelcase would be sexier ([example on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/iterator#Description))

gaearon: I was actually kinda torn on this one. :-)
Action types are all caps by convention.

I'd rather keep `@@` as the marker of “special stuff” and keep using all caps for action types and camel for state keys:

```js
export const StateKeys = {
  LOG: '@@log'
};

export const ActionTypes = {
  JUMP_TO_STATE: '@@JUMP_TO_STATE'
};
```

PR #112, opened 2015-06-16
Revisit dependencies and add browser ready build

dariocravero:
This PR:

- cleans up `package.json` by removing the unused `browserify` section;
- removes `envify` and `babel-runtime` as they aren't being used;
- adds a dist build script (borrowed from flummox) to create distribution ready builds of `redux` that can be used in a browser directly, put into a cdn, into a live playground :), etc. @acdlite is there any specific reason why `flummox` doesn't save (nor ignore on .gitignore) `dist/`?

Regarding sizes, we're looking at `56k` for a regular build and `~12k` for a minified one, when gzipped it's a bit below `4k`! Pretty impressive! :)

PR #129, opened 2015-06-17
Replace lodash's mapValues

frankychung:
From #112's discussion, this replaces lodash's `mapValues` to reduce the number of lodash dependencies.


gaearon:
Great job, thanks!

PR #131, opened 2015-06-18
Replace lodash utils

frankychung:
From #112 -- Removing lodash utils decreases the build size significantly as described by @dariocravero so this PR replaces the remaining three lodash modules with basic utils.

### DIFF
--- a/src/utils/identity.js
+++ b/src/utils/identity.js
@@ -1,0 +1,3 @@
+export default function identity(value) {
+  return value;
+}
--- a/src/utils/isPlainObject.js
+++ b/src/utils/isPlainObject.js
@@ -1,0 +1,3 @@
+export default function isPlainObject(obj) {
+  return typeof obj == 'object' && Object.getPrototypeOf(obj) === Object.prototype;
+}
--- a/src/utils/isPlainObject.js
+++ b/src/utils/isPlainObject.js
@@ -1,3 +1,3 @@
 export default function isPlainObject(obj) {
-  return typeof obj == 'object' && Object.getPrototypeOf(obj) === Object.prototype;
+  return typeof obj === 'object' && Object.getPrototypeOf(obj) === Object.prototype;
 }
--- a/src/utils/isPlainObject.js
+++ b/src/utils/isPlainObject.js
@@ -1,3 +1,3 @@
 export default function isPlainObject(obj) {
-  return typeof obj === 'object' && Object.getPrototypeOf(obj) === Object.prototype;
+  return obj ? typeof obj === 'object' && Object.getPrototypeOf(obj) === Object.prototype : false;
 }
--- a/package.json
+++ b/package.json
@@ -55,8 +55,7 @@
   "dependencies": {
     "babel-runtime": "^5.5.8",
     "envify": "^3.4.0",
-    "invariant": "^2.0.0",
-    "lodash": "^3.9.3"
+    "invariant": "^2.0.0"
   },
   "browserify": {
     "transform": [
--- a/src/components/createConnector.js
+++ b/src/components/createConnector.js
@@ -1,6 +1,6 @@
-import identity from 'lodash/utility/identity';
+import identity from '../utils/identity';
 import shallowEqual from '../utils/shallowEqual';
-import isPlainObject from 'lodash/lang/isPlainObject';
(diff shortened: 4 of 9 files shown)


src/utils/identity.js
```diff
@@ -0,0 +1,3 @@
+export default function identity(value) {
+  return value;
+}
```
chicoxyzzy: actually lodash's identity function is the same as yours

goatslacker: It's probably best to put these all in a single file and export them individually that way you avoid the overhead of including each file, I think it adds ~3 lines and a bunch of bytes for every file you include.

This also has the benefit of keeping all the utils together.

goatslacker: As an aside: identity would then be something like:

`export const identity = x => x;` which is just :sunglasses:

taylorhakes:
These functions are only used in one spot each. I think it would be better to keep the code inline and not create an unnecessary abstraction in another file. If they start to get used a lot or become a source of bugs, then put them in another file.

src/utils/identity.js
frankychung: @goatslacker Grouping these in one file sounds like a good idea to me however I worry it doesn't follow the style of the rest of utils of exporting one function per module (or as @taylorhakes mentions inlining some of them, especially simple ones like identity). Also, exporting arrow functions is pleasantly succinct but also another question of style :). @gaearon thoughts?

src/utils/isPlainObject.js
```diff
@@ -0,0 +1,3 @@
+export default function isPlainObject(obj) {
+  return typeof obj === 'object' && Object.getPrototypeOf(obj) === Object.prototype;
```
ooflorent: This one does not work.

```js
isPlainObject(null) // TypeError: Object.getPrototypeOf called on non-object
```

src/utils/identity.js
ooflorent: As @goatslacker suggested, using an arrow function would be better because:

- `identity` is not `new`-able
- `identity` has no `prototype`

src/utils/isPlainObject.js
frankychung: Thank you, added another test case for calling with `null`

src/utils/isPlainObject.js
```diff
@@ -0,0 +1,3 @@
+export default function isPlainObject(obj) {
+  return obj ? typeof obj === 'object' && Object.getPrototypeOf(obj) === Object.prototype : false;
```
goatslacker: A simple way to detect a pojo is to do Object.prototype.toString.call(object) === '[object Object]' or something like that

ooflorent: I won't work:

```js
var F = (function(){})
var f = new F
Object.prototype.toString.call(f) // [object Object]
```

src/utils/identity.js
emmenko: > It's probably best to put these all in a single file and export them individually

:+1: 

> using an arrow function would be better

I think it's just a matter whether we want to have it as a named function or not.

ooflorent: > I think it's just a matter whether we want to have it as a named function or not.

It is not the only implication of arrow functions ;)

src/utils/isPlainObject.js
taylorhakes: @ooflorent Are we trying to avoid the object above or `Object.create(null)`? It is perfectly fine for someone to have a React state that is not a plain object. ImmutableJS is used for state and it is not a plain object.

gaearon: >It is perfectly fine for someone to have a React state that is not a plain object.

I thought that broke in some recent version, didn't it? Either way they don't officially support it yet.
https://github.com/facebook/react/issues/3303

gaearon: What I want from `isPlainObject`:

* no to arrays
* no to null or undefined
* no to ImmutableJS records

I don't really care for edge cases like `Object.create(null)`.

taylorhakes: @gaearon I retract my earlier comment. You are correct

gaearon:
This looks good to me. I want to go ahead with #112 so merging.
Thanks!

PR #132, opened 2015-06-18
Fixed code style to match airbnb eslint

taylorhakes:
https://github.com/gaearon/redux/issues/97 This changes style to match airbnb style guide. There were a couple of code changes to make this work. All the tests are passing.

It seems the eslint does not fully support decorators. I had to disable the padding rule in 2 tests.


src/index.js
```diff
@@ -1,8 +1,16 @@
...
-export createRedux from './createRedux';
-export createDispatcher from './createDispatcher';
+import createRedux from './createRedux';
+import createDispatcher from './createDispatcher';
 
 // Utilities
-export composeMiddleware from './utils/composeMiddleware';
-export composeStores from './utils/composeStores';
-export bindActionCreators from './utils/bindActionCreators';
+import composeMiddleware from './utils/composeMiddleware';
```
taylorhakes: eslint was complaining about the import and export in the same statement. I had to split it up.

src/index.js
```diff
@@ -1,8 +1,16 @@
...
+import composeStores from './utils/composeStores';
+import bindActionCreators from './utils/bindActionCreators';
+
+export {
+	createRedux,
+	createDispatcher,
+	composeMiddleware,
+	composeStores,
+	bindActionCreators
+};
```
chicoxyzzy: this is exactly what it was before

chicoxyzzy: https://github.com/leebyron/ecmascript-more-export-from

chicoxyzzy: so maybe we just need custom eslint plugin/rule for that

taylorhakes: I know it may be possible in ES7, but eslint complains. Is it worth waiting to create a plugin? Or should I ignore this file? Or just accept this change until eslint is updated?

src/index.js
```diff
@@ -1,8 +1,16 @@
...
 // Utilities
-export composeMiddleware from './utils/composeMiddleware';
-export composeStores from './utils/composeStores';
-export bindActionCreators from './utils/bindActionCreators';
+import composeMiddleware from './utils/composeMiddleware';
+import composeStores from './utils/composeStores';
+import bindActionCreators from './utils/bindActionCreators';
+
+export {
+	createRedux,
```
ooflorent: Tabs?

.eslintrc
```diff
@@ -19,4 +13,4 @@
   "plugins": [
     "react"
   ]
-}
+}
```
ooflorent: Can you add a new line at the end of the file?

src/index.js
ooflorent: Keep `babel-eslint` and everything should be fine.

DenisIzmaylov:
Hmm. I fixed it here https://github.com/gaearon/redux/pull/121 yesterday

src/index.js
gaearon: ESLint will not complain if you use `"parser": "babel-eslint"` like in the original config.
It is essential!

_But_ I heard that `export` syntax is not going to make it in ES7 after all. So I'm fine with replacing this with `export { default as createDispatcher } from './createDispatcher`, etc.

.eslintrc
```diff
@@ -1,17 +1,11 @@
...
-    "jsx": true,
-    "modules": true
-  },
+  "extends": "eslint-config-airbnb",
   "env": {
     "browser": true,
     "mocha": true,
     "node": true
   },
-  "parser": "babel-eslint",
```
gaearon: We need `parser` here.

emmenko: Isn't it already defined [here](https://github.com/airbnb/javascript/blob/master/linters/.eslintrc#L2) ?

taylorhakes:
@DenisIzmaylov I thought you were just submitting an ignore of all rules. You just changed your pull request a couple of hours ago. Since you started working on this, I will close this pull request.

PR #133, opened 2015-06-18
Document the `@provide` decorator in the README

ide:
Added a little section showing how to use `@provide(redux)`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ Atomic Flux with hot reloading.
     - [Decorators](#decorators)
   - [React Native](#react-native)
   - [Initializing Redux](#initializing-redux)
+    - [Decorators](#decorators-1)
   - [Running the same code on client and server](#running-the-same-code-on-client-and-server)
   - [Additional customization](#additional-customization)
 - [FAQ](#faq)
@@ -207,7 +208,7 @@ export default class CounterApp {
 
 #### Decorators
 
-The `@connect` decorator lets you create smart components less verbosely:
+The `@connect` decorator lets you create smart components more concisely:
 
 ```js
 import React from 'react';
@@ -265,6 +266,21 @@ export default class App {
         }
       </Provider>
     );
+  }
+}
+```
+
+#### Decorators
+
+The `@provide` decorator lets you inject a Redux instance more concisely:
+
+```js
+import { provide } from 'redux/react';
+
+@provide(redux)
+export default class App {
+  render() {
+    return <CounterApp />;
   }
 }
 ```


README.md
```diff
@@ -269,6 +269,21 @@ export default class App {
...
+
+The `@provide` decorator lets you inject a Redux instance more succinctly:
+
+```js
+import { provide } from 'redux/react';
+
+@provide(redux)
+export default class App {
+  render() {
+    return <CounterApp />;
```
taylorhakes: CounterApp is not imported at the top

ide: I kept the example brief for focus and there's a very similar example with `<Provider>` right above this one anyway... which also doesn't import CounterApp for that matter.

README.md
```diff
@@ -269,6 +269,21 @@ export default class App {
 }
 ```
 
+#### Decorators
+
+The `@provide` decorator lets you inject a Redux instance more succinctly:
```
taylorhakes: uses `more succinctly` here and `less verbosely` above. Probably best to keep consistent language.  I like more `more succinctly`, personally, but I think the connect decorator language should be changed to match.

README.md
taylorhakes: Correct, I would assume the above is a bug. For me, it is a little confusing when missing imports. I would not understand where CounterApp came from. I feel it's better to error on the side of clarity instead of brevity.

ide: i'm going to defer to @gaearon who wrote most of the readme on this one. When reading through the docs the CounterApp is already referenced several times. So holistically I find it easier to learn about the `@provide` feature if the example focuses on the salient details...

README.md
ide: Ya - will change the example above.

README.md
```diff
@@ -4,9 +4,9 @@ redux
 [![build status](https://img.shields.io/travis/gaearon/redux.svg?style=flat-square)](https://travis-ci.org/gaearon/redux)
 [![npm version](https://img.shields.io/npm/v/redux.svg?style=flat-square)](https://www.npmjs.com/package/redux)
 
-Atomic Flux with hot reloading.  
+Atomic Flux with hot reloading.
 
-**The API is likely to change a few times before we reach 1.0.**
```
emmenko: Keep the whitespaces please.

README.md
```diff
@@ -207,7 +207,7 @@ export default class CounterApp {
 
 #### Decorators
 
-The `@connect` decorator lets you create smart components less verbosely:
+The `@connect` decorator lets you create smart components more succinctly:
```
emmenko: Strange word `succinctly`.

I think something like `... smart components in a more compact way` would be better.

README.md
```diff
@@ -269,6 +269,21 @@ export default class App {
 }
 ```
 
+#### Decorators
```
emmenko: It's already defined above

ide: This is the Decorators sub-section for the Providers topic. If you look at the doc as a whole it makes sense.

README.md
emmenko: Agree with @ide

README.md
ide: I can revert but just FYI it was my editor that did this automatically so other contributors will probably end up doing the same.

README.md
ide: i actually think "succinctly" is the one of the most succinct ways to write this. "concisely" could also work.

README.md
emmenko: I disagree. This section is about `Initializing Redux`, not about `decorators` which belongs to `Components`. And since there is already the description about `@connect`, I would expect `@provide` to be also there.

README.md
emmenko: I'm not native english speaker, but it's just weird to pronounce it for me. `concisely` is much better :)

README.md
emmenko: I know, I did the same "mistake" ;)

It's just that @gaearon wants to have it like this, so we should follow his conventions.

README.md
ide: The problem with putting it in the decorators section above is that the reader hasn't learned about Providers yet. So the logical structure I was going for is: (1) learn about Connector and (1a) learn about alternate syntax (`@connect`), then (2) learn about Provider and (2a) learn about alternative syntax (`@provide`).

I can see it would make sense to group the decorators together in an API docs page, but for this initial guide I think we want to group these sections by the concept that the reader is learning instead of grouping by syntax.

emmenko: Then make sure it's correctly linked in the "Table of Contents" section

ide: Good point

gaearon:
@acdlite's going to work on the new docs so let him decide the fate of this PR :-)
Could merge now, or could reuse for new docs..

gaearon:
We're writing new docs and moving React-specific wrappers to a new project at the same time, and there's even some [API churn](https://github.com/gaearon/react-redux/issues/1#issuecomment-120729921) there too so I'm closing this. Please let me know if you'd like to write the initial README (you can reuse your changes from this PR) for https://github.com/gaearon/react-redux/ describing the current React API, as right now it's non-existent.

PR #140, opened 2015-06-18
Docs improvements

acdlite:
PR to keep track of docs improvements, as discussed in https://github.com/gaearon/redux/issues/137

- [x] Rename Stores to Reducers
- [x] Rename the “Redux instance” to Store as per https://github.com/gaearon/redux/issues/137#issuecomment-113252359
- [ ] Explain that if you squint at traditional Flux Stores, you'll see Reducers trapped in them
- [x] Glossary of terms
- [ ] JSDoc (and/or Flow annotations)
- [ ] Design decisions
- [ ] Updated README (with focus on familiar Flux concepts; keep advanced stuff in docs)
- [x] "Getting Started" guide
- [ ] Explanation of how to subscribe to changes manually, without Provider and Connector https://github.com/gaearon/redux/issues/134#issuecomment-113272964
- [x] Higher-order stores (possibly call these "store transformers"?)
- [ ] Full API docs
- [ ] Touch on sugar like `createReducer`, [`createActionCreators` haha](https://github.com/gaearon/redux/commit/4b207670dc8988b2976973c30f345535d382fa45)

Focus right now is to get these items completed in Markdown form. Then (or in parallel) we will turn these into a proper docs site.


emmenko:
:+1:

acdlite:
This my weekend project. I started working on a new version of the README. I think that target audience for the README is developers who aren't super familiar with functional programming, and are skeptical of *yet another* Flux library. So, the focus should be on explaining how Redux significantly improves on the traditional Flux architecture, while at the same time not being too dissimilar.

I'm going to move most of the current README into separate documents and link to them from the table of contents. The rough outline will be

I. Key features (like an elevator pitch) (includes links, so also like a summarized table of contents)
2. Why Redux? (slightly longer elevator pitch)
3. Code sample
4. Link to tutorial
5. Link to example apps
6. Index for rest of docs

The reason the index goes at the end is because the current table of contents, while good, could be overwhelming for new visitors. The initial impression should be "hey, Redux is very simple and easy to learn."

gaearon:
>I think that target audience for the README is developers who aren't super familiar with functional programming, and are skeptical of yet another Flux library. So, the focus should be on explaining how Redux significantly improves on the traditional Flux architecture, but is still very familiar.

Yes!

gaearon:
Idea for fans of cursors: showing how to implement a cursor-like functionality with Redux using a single reducer and a single `SET` action. Use case: you like cursors but want benefits of Redux (time travel etc). Also because it's a fun experiment.

faassen:
What is the documentation technology intended to be used? I have a lot of experience with Python restructured text and Sphinx, but I take it the React Drag & Drop website is generated using a bunch of custom scripts from markdown. That helps with flexibility (the examples in different versions of JavaScript for instance) , though I like not having to maintain code to generate my websites.

dariocravero:
Hey guys, what do you think about adding a section on how to migrate from other (traditional) flux frameworks? I know that the migration path might not be that hard but I think it would help people get on board too. Plus, now that flummox is staying at version 4, I believe we owe it to our users... :) Thoughts?

faassen:
I think that's a good idea, @dariocravero, especially as it may help people understand Redux better as well -- it's another way into the documentation.

gaearon:
Great idea @dariocravero!

vladap:
I think people are used to mutability, hence I think it needs to be documented what pattern to use to handle large mutable data structures which can't be copied with every mutation. And if it is either supported by Redux anyhow. I assume that immutability is the way how Redux detects changes to update views, hence I can't return a reference to a mutable whatever and this need to be clear from a documentation.

Should I create some kind of an immutable companion object with getters for underlying mutable struct which I would recreate and return as state with every update? Would this even work or it breaks other features like replay?

An example could be [Crossfiler](http://square.github.io/crossfilter/). I'm not aware of an efficient immutable equivalent. And I don't have much idea how to integrate it with Redux.

vladap:
Here are my thoughts on documenting important differences between ActionCreators vs. Actions which everybody trying to use flux/redux should be clear about. The differences are subtle but important. Feel free to rephrase it or drop it all together, I just want to log what I think can help to new users.

ActionCreator tells what should happen while Action tells what happened. ActionCreator can be understood as a Command while an Action is an event (like mouseClicked in DOM). If you come from an analytical database modeling background you would call Action a Fact (and they would be stored in a fact table in a star schema).

ActionCreators can be processed concurrently (async) and emit Actions when appropriate. Actions are processed synchronously, they can be logged to create synchronous log which can be potentially replayed (if there is such a facility).

As a sidenote, because ActionCreator term is long, people tend to shortcut it to just Action term in a general discussion (or even in a API) and freely swap them, which might be confusing at times. It could be better to have Commands and Events, or Actions (as a synonym to commands) and Events (some other eventsourcing/commandsourcing systems use these terms). But I understand the desire to stick to the original Flux terminology. For me, used to another background, it was quite confusing to decode Flux at first even there was no new concept for me.

vladap:
My anecdotal theory why Facebook decided to use ActionCreators and Actions is that "event" term is too general and in web environment too close to DOM events which could cause another confusion. In the end my preferred terminology would be Actions/Commands (instead of ActionCreators) and Facts (instead of Actions). It is how I map them internally in my mind, but it might just because of my other background (backend world).

pascience:
Some thoughts in complement to @vladap's :
I have no "backend world" background but I use the same Action/Fact terminology and have found it to be useful both in everyday programming and when introducing the Flux architecture to coworkers.

One thing that reinforces that distinction is conjugating facts in the past tense. For instance, `dispatch({ type: Facts.ASKED_TO_FETCH_DOCUMENT, documentId })` rather than `dispatch({ type: ActionConstants.FETCH_DOCUMENT, ... })`. In the logs, the fact in this example would be followed shortly by either `FETCHED_DOCUMENT` or `FAILED_TO_FETCH_DOCUMENT`.
In my opinion, this better represents what just happened in the app, and also makes it clear that a) the demand may not succeed and b) at the call-site, we don't care whether or when the demand gets handled.

I'm not suggesting that `reduce` be renamed into "fold from the past" as in Elm, but I have the impression that `examples/counter` would benefit from the past-tense conjugation (ie `ASKED_TO_INCREMENT_COUNTER`).

vladap:
Past tense works well for facts which happened. But I was thinking about it yesterday and I think I switched my understanding of Flux.

As I understand:
Eventsourcing system is based on a log of facts which happened. Facts are logged after complex processing. Eventsourcing system process them synchronously and when it starts with state A and applies the log to it always ends with state B. Hence this system can be replayed.

Commandsourcing system is when I log a command before any processing happens. To replay such a system then processing of each command has to be deterministic. In today concurrent world with Service Oriented Architectures, microservices and such, it is impossible to build replayable commandsourcing system as a whole.

Flux:
Flux doesn't log commands (actions) before any processing happens but it logs commands after all non-deterministic processing is done. Subsequent commands (actions) processing is strictly synchronous and potentially replayable. Hence I don't have to log facts and I can log commands (actions).

With this understanding actions doesn't log fact what happened but actually what should happen.

The only issue is that different developers try to grasp Flux either as eventsourced app or general commandsourced app while it is a special case of a command sourced app. Flux can be a small part of my whole system which can afford to work as synchronous command sourced app. There still can be many other non-deterministic async services around, like WebApiTools.

Hope, it makes sense to anybody. Let me know if I should tweak my understanding more and I'm still off.

gaearon:
@acdlite Have you had any chance to work on this yet? Not pushing, just checking up. I'm going to be available to work on this right after React Europe.

emmenko:
@gaearon @acdlite FYI: I've started writing some JSDoc...

acdlite:
I just pushed up some docs on what we're currently calling "higher-order stores," which have this signature:

```
createStore => createStore'
```

Does anyone have any ideas for a better name? Perhaps "store transformer" (to borrow some transducer terminology)? I feel like the term “higher-order” gets redundant after a while in a functional library. Basically everything is a higher-order something :smile:

taylorhakes:
Store transformer makes the most sense to me. If there isn't an exact name, do we try to avoid creating new terms?

acdlite:
@taylorhakes My preference is to borrow from existing names and concepts. I believe @gaearon agrees.

`createStore => createStore'`  is very similar to a transducer. Transducers operate on transforms like `{ init, step, result }`, whereas we're operating on `{ dispatch, getState, setState, rest... }`.

Although now that I've typed it out, I remember that `{ init, step, result }` itself is technically the transformer. So I guess "store transformer" doesn't fit this metaphor, after all :/

Maybe "transtorer"? (kidding, of course :smile:)

I'll keep going with higher-order store until we come up with something better.

tappleby:
For some reason every time I see higher-order I think of [Star wars](http://starwars.wikia.com/wiki/Jedi_High_Council_(New_Jedi_Order)) :smile:

docs/higher-order-stores.md
```diff
@@ -0,0 +1,57 @@
...
+
+A higher-order store is a function that turns a store creating function into a new store creating function:
+
+```
+createStore => createStore'
+```
+
+Look familiar? It's just like the signature for [middleware](middleware.md), only we're wrapping `createStore()` instead of `dispatch()`.
+
+Higher-order stores are a pattern for creating composable Redux extensions. They're much the same as higher-order components in React. They can be as simple as `applyMiddleware()`, or as powerful as the Redux Devtools](https://github.com/gaearon/redux-devtools). There's no limit to the kinds of things you can create.
```
tcoopman: should be:  \[Redux Devtools](https://github.com/gaearon/redux-devtools)

missing _[_

docs/higher-order-stores.md
```diff
@@ -0,0 +1,57 @@
...
+      ...store,
+      dispatch: middleware(store.dispatch)
+    };
+  };
+}
+```
+
+`next` is the next store creating function in the chain — either the return value of another higher-order store, or `createStore()` itself.
+
+This design allows multiple higher-stores can be combined simply using function composition:
```
tcoopman: higher-order stores instead of higher-stores for consistency.

The sentence looks a bit strange to me, I would rephrase it like this: _This design allows the combination of multiple higher-order stores simply using function composition._

src/utils/bindActionCreators.js
```diff
@@ -1,5 +1,13 @@
 import mapValues from '../utils/mapValues';
 
+/**
+ * Given a list action creators, wrap them to the `dispatch` function
```
tcoopman: Given a list of action creators

src/utils/shallowEqual.js
```diff
@@ -1,3 +1,13 @@
+/**
+ * Given two objects, performs equality by iterating through keys
+ * on an object and returning `false` when any key has values which
+ * are not strictly equal between `objA` and `objB`.
+ * Returns `true` when the values of all keys are strictly equal.
+ *
+ * @param  {Object} objA
+ * @param  {Object} objB
+ * @return {Boolean}
+ */
```
tcoopman: Documentation is not the same as the implementation.
Implementation just checks `objA === objB`

gaearon:
Ref https://github.com/gaearon/redux/issues/139

docs/higher-order-stores.md
```diff
@@ -0,0 +1,57 @@
...
+  applyMiddleware(m1, m2, m3),
+  devTools,
+  createStore
+);
+```
+
+Now just pass your reducer (and an initial state, if desired) to your new store creating function:
+
+```js
+const store = createStore(reducer, intialState);
```
vladap: Shouldn't it be:
const store = newCreateStore(reducer, intialState);
?

acdlite:
I split the current README into multiple docs. Much of it is out of date as of 1.0, and some of it will be moved into a redux-react.

docs/types-and-terminology.md
```diff
@@ -0,0 +1,107 @@
...
+
+We must distinguish between dispatching functions in general and the base dispatch function provided by the Store interface. The base dispatch function *always* sends an action to the store's reducer, along with the previous state returned by the store, to calculate a new state. However, the application will rarely call the base dispatch function directly — usually, the base dispatch function is wrapped by middleware. See below for more information.
+
+### Reducer
+
+```js
+type Reducer<S,A> = (state: S, action: A) => S
+```
+
+A **reducer** or **reducing function** is a function that accepts an accumulation and a value and returns a new accumulation. They are used to reduce a collection of values down to a single value. Reducers are not unique to Redux — they are a fundamental concept in functional programming.  Even most non-functional languages, like JavaScript, have a built-in API for reducing. (In JavaScript, it's `Array.prototype.reduce()`.)
```
nbostrom: Since functions in JavaScript are closures and first-class citizens it is by all means a functional language by definition. Though it's not considered "pure functional", since functions can have side-effects.

vladap:
In Redux (or app built with Redux), what it the equivalent of [Cycle.js driver](http://cycle.js.org/drivers.html)? It is role to interact with side-efects is well defined. I think it is ActionCreator which is supposed to trigger side-effects and return action based on it. But, In the current docs there is explicit comment `// But action creators are pure functions returning actions` in Counter example (Actions sections). The new docs for ActionCreator term don't say much.

In my previous question regarding Crossfilter I now believe that it should be treated as an `util module` or I could call it a `service` or whatever and interact with it using a driver, ehm ActionCreator. I could think about it that I could implement Crossfiler funcionality as a server-side service. Then I would expose its webapi and use ActionCreator to talk to it. Even I will implement it client side, nothing stops me to think about it and handle it like it would completely independent and separated component of my system and handle it via ActionCreators. The same can apply for react-router (I understood that many are already exploring this), and even browser itself (to use its SessionStore, LocalStore and other services).

emmenko:
@acdlite let me know if/how I can help with the new docs please.

acdlite:
@emmenko I split the existing README into separate docs — you could help by updating the language & fleshing them out to match the new API.

I'm working on middleware & higher-order store docs right now.

ellbee:
@acdlite @emmenko I'd like to help with the docs too. I'll take a shot at updating store.md to the new api if that is not currently being worked on.

emmenko:
@acdlite sure, I'll give it a try.

PS: how is the experiment with Flow going? Is it usable? It would also help with the documentation I guess...

acdlite:
@emmenko Yeah the flow-types-take-2 branch works great

gaearon:
Something visual like image from [this tweet] (https://twitter.com/erikthedev_/status/620951888326856704) might also be helpful:

dariocravero:
cc @tomatuxtemple

docs/getting-started.md
```diff
@@ -0,0 +1,250 @@
...
+  type: REMOVE_TODO,
+  payload: 123
+}
+```
+
+`payload` in this case indicates the id of the todo we want to remove.
+
+## Action creators
+
+**Action creators** exactly that — functions that create actions. It's easy to conflate the terms "action" and "action creator," so do your best to use the proper term.
```
julioolvr: `**Action creators** ARE exactly that`?

acdlite: D'oh, yes

docs/getting-started.md
```diff
@@ -0,0 +1,250 @@
...
+const initialState = Immutable.Map({ todos: [], idCounter: 0 });
+
+function todoReducer(state = initialState, action) {
+  switch (action.type) {
+    case ADD_TODO:
+      return state
+        .update('todos',
+          todos => todos.push(Immutable.Map({
+            text: action.payload,
+            id: state.get('idCounter')
```
dferber90: Should this be `id: state.get('idCounter') + 1`?

docs/getting-started.md
```diff
@@ -0,0 +1,250 @@
...
+const store = applyMiddleware(promise, thunk, observable)(createStore)(reducer);
+```
+
+Yes, you read that correctly. [Read more about how middleware works here.](middleware.md)
+
+## Connecting to your views
+
+You'll rarely interact with the store object directly. Most often, you'll use some sort of binding to your preferred view library.
+
+Flux is most popular within the React community, but Redux works with any kind of view layer. The React bindings for Redux are available as react-redux — see that project for details on how to integrate with React.
```
dferber90: "Flux is most popular within the React community"
Did you mean "React is most popular within the Flux community," instead?

gaearon: Hmmm.. I still think React community is bigger than Flux community :-)

dferber90: That's true :-) I was suggesting the switch because of the second part of the sentence.
In my mind the current wording suggests Flux is the "view layer", which seems wrong.

*Flux is most popular [as a view layer] within the React community, but Redux works with any kind of view layer.*

Anyways, great work on Redux and the new docs :+1:

docs/getting-started.md
```diff
@@ -0,0 +1,250 @@
...
+```js
+dispatch(addTodo(text));
+dispatch(removeTodo(id));
+```
+
+Or create **bound action creator** that automatically dispatches:
+
+```js
+const boundAddTodo = text => dispatch(addTodo(text));
+const boundRemoveTodo = id => dispatch(addTodo(id));
```
hasdavidc:
```
const boundRemoveTodo = id => dispatch(removeTodo(id));
```

docs/getting-started.md
hasdavidc: agree with @dferber90 - the phrasing suggests that the React's most popular view layer is Flux

AprilArcus: "Although Flux architectures were originally popularized within the React community, Redux works with any kind of view layer."

docs/middleware.md
```diff
@@ -1,4 +1,56 @@
...
+```
+
+Yes, you read that correctly. If this looks strange to you, it may help to break the process down into multiple steps:
+
+```js
+const newCreateStore = applyMiddleware(thunk, promise, observable)(createStore);
+const store = newCreateStore(reducer);
+```
+
+If you
```
hasdavidc: incomplete phrase

docs/middleware.md
```diff
@@ -1,4 +1,56 @@
...
+Middleware
+==========
+
+A middleware in Redux is a function that turns a dispatching function into a new dispatching function:
+
+```
+dispatch => dispatch'
+```
+
+The key feature of middleware is that it is composable. Multiple middleware can be combined together, where each middleware requires no knowledge of the what comes before or after it in the chain.
```
hasdavidc: orphan ```the``` - "requires no knowledge of what comes"

docs/middleware.md
```diff
@@ -1,4 +1,56 @@
...
+
+Usage
+=====
+
+To enable middleware in your Redux app, use `applyMiddleware()`.
+
+### `applyMiddleware(...middlewares)`
+
+This function returns a [higher-order store](higher-order store). You don't need to worry about that if you're not interested — here's how you use it:
```
hasdavidc: link should be ```higher-order-stores.md```

docs/middleware.md
```diff
@@ -1,4 +1,56 @@
...
+
+```js
+const store = applyMiddleware(thunk, promise, observable)(createStore)(reducer);
+```
+
+Yes, you read that correctly. If this looks strange to you, it may help to break the process down into multiple steps:
+
+```js
```
hasdavidc: could use the same comment that was in getting-started.md:

```
// where promise, thunk, and observable are examples of middleware
```

docs/store.md
```diff
@@ -0,0 +1,157 @@
...
+You may optionally specify the initial state as the second argument to `createStore`. This is useful for hydrating the state of the client to match the state of a Redux application running on the server.
+
+```js
+// server
+const store = createStore(reducer);
+store.dispatch(MyActionCreators.doSomething()); // fire action creators to fill the state
+const state = store.getState(); // somehow pass this state to the client
+
+// client
+const initialState = window.STATE_FROM_SERVER;
```
hasdavidc: duplicated as in universal-rendering.md - intentional?

AprilArcus:
I remade Erik Aybar's diagram with the Counter example from 1.0rc. Atom + Babel-Sublime + Base16 Eighties + Menlo > OS X CoreText @ 2x HiDPI mode > a little fussing in photoshop > a bunch of vector overlays in Omnigraffle. Please use if useful or request corrections/updates if desired.

emmenko:
@AprilArcus wow, nice!! :+1:

AprilArcus:
I think this might look better with the long form of the <Connect> HOC, so we can see the dependency injection of `counter` into `CounterApp`'s props.

I think it might also be interesting to (maybe on another layer?) annotate the `import` tree in another color.

Maybe `App.js` should be included?

Maybe the spiral of perpetual-application-to-return-value would look good broken out into the flow of data through the Redux source code.

Maybe state, actioncreator, and action should be annotated in different colors or line styles, with a legend? Maybe color or line style should be used to distinguish "imports" from "dependency injects" from "calls"?

erikthedeveloper:
Very cool that you took that idea and [ran with it here](https://github.com/gaearon/redux/pull/140#issuecomment-121982266) @AprilArcus! :+1:

My original intention with that [skitch from Twitter](https://twitter.com/erikthedev_/status/620951888326856704) was to follow a single piece of data flowing end-to-end through the "Redux cycle" (Reducer -> Container -> Component -> Action -> Reducer [...])

> [...] request corrections/updates if desired

My only suggestion would be to simplify it as an illustration of the "flow of data" rather than trying to cover all the bases and highlighting too many details in one screen grab. I feel like the number/complexity of the arrows could be simplified (I prefer straight arrows w/ one arrow entering and one arrow exiting each area/concept) to illustrate the "flow" rather than "connecting all the dots".

I would also hide the line numbers and change the arrow color to contrast w/ the code/text. Just my preference :smile:

Merely a suggestion to simplify to better reflect the simplicity of Redux. Great to see it pop up in this conversation!

#### Original 

image

#### Possible suggested simplification of April's

_ignoring green arrows/boxes_

image

gaearon:
I agree it's better to focus on the flow than on connecting all the dots.

AprilArcus:
I see the appeal of simplifying, but I find the use of identical arrows to track values of constantly changing type (`state`, `(state, action) => state`, and `() => action`) to be really disorienting.

gaearon:
@AprilArcus Maybe it would help to have different pictures focusing on different arrows. Right now it's kind of hard to understand where to start and which arrow to follow.

docs/examples.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+### Async and Universal Examples with Routing
+
+These async and [universal (aka “isomorphic”)](https://medium.com/@mjackson/universal-javascript-4761051b7ae9) examples using React Router should help you get started:
+
+* [redux-react-router-async-example](https://github.com/emmenko/redux-react-router-async-example): Work in progress. Semi-official. Only the client side. Uses React Router.
+* [react-redux-universal-hot-example](https://github.com/erikras/react-redux-universal-hot-example): Universal. Uses React Router.
+* [redux-example](https://github.com/quangbuule/redux-example): Universal. Uses Immutable, React Router.
+* [isomorphic-counter-example](https://github.com/khtdr/redux-react-koa-isomorphic-counter-example): Universal. A bare-bone implentation of the [counter example app](https://github.com/gaearon/redux/tree/master/examples/counter). Uses promises-middleware to interact with API via Koa on the server.
```
mfunkie: implentation -> implementation

AprilArcus:
Okay, while I think about that idea, here is a non-defaced hi-res screencap of the five source files that people can mark up. Sorry I forgot to set my editor font to Consolas ^~

arstin:
FWIW, I had annotated a screenshot similar to how erikthedeveloper's orange lines to explain what was going on to my coworkers and it seemed helpful. I do think something similar focusing on the overall flow would be useful as a visual reference for an actual explanation. In my case, I also added numbers attached to comments describing what is happening, so the expectation wasn't that you'd be able to understand redux just from the image. On a wiki, it might make more sense to break this up into separate images with paragraphs below them?

gaearon:
Next week @acdlite is going on a vacation so I'll take over the docs.
Happily I don't have anything else to do at the moment!

tychotatitscheff:
Hi the docs from https://github.com/acdlite/flummox are hosted in github page http://acdlite.github.io/flummox . 

It used an isomorphic fluxmox app to dislpay it nicely https://github.com/acdlite/flummox/tree/master/docs.

What about making the same here, if one is interessted, i can carry on !

gaearon:
Superseded by #289. Docs written here will be gradually moved over there.

gaearon:
Want to help with the new docs?
I created the initial directory structure: https://github.com/gaearon/redux/tree/rewrite-docs/docs/

If you'd like to help and maybe write a few [recipes](https://github.com/gaearon/redux/tree/rewrite-docs/docs/Recipes), let me know in #289 so we can collaborate.

tsingson:
@erikthedeveloper @AprilArcus 

cool!!

PR #141, opened 2015-06-18
counter example in README should match code in examples/

acrispino:


gaearon:
Thanks, forgot to update this one!

PR #142, opened 2015-06-18
Revisit dependencies and add browser ready builds

### LINKED ISSUE #88: Investigate the library size and dependencies

### LINKED ISSUE #115: babel loose all
@acdlite flummox [is using `"loose": "all"` in `.babelrc`](https://github.com/acdlite/flummox/blob/master/.babelrc#L3). Babel [says that we would have some speed gains](https://babeljs.io/docs/advanced/loose/) from using it but that it might be risky. What are your thoughts/experience with it in flummox so far? Should we include that into redux as well?

dariocravero:
Added:
- Dist build script (borrowed from flummox) to create distribution ready builds of redux
  that can be used in a browser directly. `dist/` is ignored by git.
- Configuration for `cdnjs` autoimport from npm.

Changed:
- Clean up package.json by removing the unused browserify section;
- Remove envify and babel-runtime as they aren't being used;
- A change log :)

Fixes #112, Fixes #115. Fixes #88.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,0 +1,16 @@
+# Change Log
+All notable changes to this project will be documented in this file.
+This project adheres to [Semantic Versioning](http://semver.org/).
+
+## [Unreleased][unreleased]
+### Added
+- Dist build script (borrowed from flummox) to create distribution ready builds of redux
+  that can be used in a browser directly. `dist/` is ignored by git.
+- Configuration for `cdnjs` autoimport from npm.
+
+### Changed
+- Clean up package.json by removing the unused browserify section;
+- Remove envify and babel-runtime as they aren't being used;
+- A change log :)
+
+[unreleased]: https://github.com/olivierlacan/keep-a-changelog/compare/v0.11.1...HEAD
--- a/scripts/browser
+++ b/scripts/browser
@@ -1,0 +1,14 @@
+#!/bin/sh -e
+
+WEBPACK_CMD=node_modules/.bin/webpack
+
+mkdir -p dist
+
+$WEBPACK_CMD src/index.js dist/redux.js
+NODE_ENV=production $WEBPACK_CMD src/index.js dist/redux.min.js
+
+$WEBPACK_CMD src/react.js dist/redux-react.js
+NODE_ENV=production $WEBPACK_CMD src/react.js dist/redux-react.min.js
+
+$WEBPACK_CMD src/react-native.js dist/redux-react-native.js
+NODE_ENV=production $WEBPACK_CMD src/react-native.js dist/redux-react-native.min.js
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,0 +1,41 @@
+'use strict';
+
+var webpack = require('webpack');
+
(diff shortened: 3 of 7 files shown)


dariocravero:
@ooflorent could you double check that the cdnjs stuff is right? :)
@gaearon as I was saying on slack, I don't know if there's any difference with the compiled react and react-native version as both expect a global called `React`. Perhaps we can get rid of one of them afterwards? Maybe it's a good idea to keep them should the need to be different? Thoughts?

The uglified and gzipped version of the whole thing is at around 3kb (core (~1kb) + react (~2kb)). AMAZING :). Well done everyone for keeping `redux` so lean! :)

dariocravero:
@gaearon I think you have to [import it to cdnjs](https://github.com/cdnjs/cdnjs#adding-a-new-or-updating-an-existing-library)?

CHANGELOG
```diff
@@ -0,0 +1,16 @@
+# Change Log
```
gaearon: Do we need to fill this from Releases page?

scripts/browser
```diff
@@ -0,0 +1,14 @@
...
+
+mkdir -p dist
+
+$WEBPACK_CMD src/index.js dist/redux.js
+NODE_ENV=production $WEBPACK_CMD src/index.js dist/redux.min.js
+
+$WEBPACK_CMD src/react.js dist/redux-react.js
+NODE_ENV=production $WEBPACK_CMD src/react.js dist/redux-react.min.js
+
+$WEBPACK_CMD src/react-native.js dist/redux-react-native.js
```
gaearon: Nah, let's ditch it, it's the same thing.

webpack.config.js
```diff
@@ -0,0 +1,41 @@
...
+    'react-native': 'React'
+  },
+  module: {
+    loaders: [
+      { test: /\.js$/, loaders: ['babel-loader'], exclude: /node_modules/ }
+    ]
+  },
+  output: {
+    library: 'Redux',
+    libraryTarget: 'var'
```
gaearon: Maybe do a UMD build instead? Like React Router: https://github.com/rackt/react-router/blob/5d674c9515aa628b644c223b2e3f43bb28ac24bc/webpack.config.js#L7-L18

gaearon:
Merged via dd831886fa068a1a306844492d982fbcc9fb3ed2.

gaearon:
I slightly changed it: now `react-redux` exports everything there is. I don't think it makes sense to separate them. (And with the config in this PR, they exported to the same global which wouldn't work anyway.) It's like “with addons” vs “without addons” now.

dariocravero:
Brilliant! :) Thanks!!

PR #149, opened 2015-06-19
Create roadmap document

emmenko:
We have many parallel topics currently under discussion and it's easy to get lost in them.

I think it makes it also easier for people to have a quick overview on where we are heading.

/cc @gaearon @acdlite


dariocravero:
:+1:

PR #150, opened 2015-06-19
Add Footer to the TodoMVC and Bring back Counter example

vramana:


vramana:
@ooflorent Can you take a look at this PR ??

The current examples  fail to lint correctly due to a bug in babel-eslint I guess.  What is better way to go disable the rule for now or change the example source slightly ?

I prefer disabling the rule `block-scoped-vars` which is giving this error . Since we don't use var in  the source code, I think it is okay to turn it off.

examples/todomvc/components/MainSection.js
```diff
@@ -1,30 +1,73 @@
...
     actions: PropTypes.object.isRequired
   };
 
+  constructor(props, context) {
+    super(props, context);
+    this.state = { showing: 'all' };
+  }
+
+  handleClearMarked() {
+    const atleastOneMarked = this.props.todos.some(todo => todo.marked);
```
ooflorent: Can you rename the variable to `atLeastOneMarked `?

examples/todomvc/stores/todos.js
```diff
@@ -1,4 +1,5 @@
-import { ADD_TODO, DELETE_TODO, EDIT_TODO, MARK_TODO, MARK_ALL } from '../constants/ActionTypes';
+import { ADD_TODO, DELETE_TODO, EDIT_TODO } from '../constants/ActionTypes';
+import { MARK_TODO, MARK_ALL, CLEAR_MARKED } from '../constants/ActionTypes';
```
ooflorent: Can you combine those 2 lines?

examples/todomvc/components/MainSection.js
```diff
@@ -1,30 +1,73 @@
...
+    }
+
+    let footer = null;
+    if (markedCount || unmarkedCount) {
+      footer = (
+        <Footer markedCount={markedCount}
+                unmarkedCount={unmarkedCount}
+                showing={this.state.showing}
+                onClearMarked={::this.handleClearMarked}
+                onShow={(e, show) => this.handleShow(e, show)} />
```
ooflorent: `::this.handleShow` should be enough

examples/todomvc/stores/todos.js
emmenko: Or simply do 

```js
import * as constants from '../constants/ActionTypes'

constants.ADD_TODO
// ...
```

vramana: I prefer the one line but It kind of too long so I split it up into two lines.

examples/todomvc/components/MainSection.js
johanneslumpe: Actually rebinding in render is wasteful :P

gaearon: I'd say let's not care about micro-optimizations at this point.
Sadly ES7 property syntax won't work with hot reloading (at least yet).

examples/todomvc/components/MainSection.js
```diff
@@ -1,30 +1,73 @@
...
 
-export default class MainSection {
   static propTypes = {
     todos: PropTypes.array.isRequired,
     actions: PropTypes.object.isRequired
   };
 
+  constructor(props, context) {
+    super(props, context);
+    this.state = { showing: 'all' };
```
gaearon: Please make these strings constants.

PR #165, opened 2015-06-22
Added an auto-generated change log.

dariocravero:
Before we go into `1.0`, here's a change log autogenerated from releases.
It doesn't make use of sections like _Added_, _Changed_ and _Fixed_ as we should but it's probably ok because it's imported. I think we should stick to those from now on. Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,341 @@
+# Change log
+
+All notable changes to this project will be documented in this file.
+This project adheres to [Semantic Versioning](http://semver.org/).
+
+## [Unreleased][unreleased]
+
+## [0.12.0] - 2015/06/19
+No breaking changes this time.
+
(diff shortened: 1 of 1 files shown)


gaearon:
What does `[Unreleased][unreleased]` header mean?
How do we regenerate this changelog?

dariocravero:
I put together [a little script](https://github.com/dariocravero/releases-to-changelog) to get the bulk of it in. I guess from there on we should be writing it ourselves as we commit stuff. That's to me the best way to go about it.
`[Unreleased][unreleased]` is a convention from [keepachangelog.com](http://keepachangelog.com/#how-can-i-minimize-the-effort-required-) to keep the release-to-be in there as we go. I guess we can remove one of them. That reminds me I didn't update the bottom of the file (the link to the latest stuff).

dariocravero:
Updated!

gaearon:
Oh, okay. Thanks a lot!

Issue #99, opened 2015-06-15
Best async serverside loading technique?


erikras:
First of all, I love this library and the patterns you guys are using. 👍👍

I'm trying to use redux to build an isomorphic app. It's working beautifully so far, except that I need to figure out how to wait until my stores have loaded (on the server) before returning the initial page load. Ideally, the loading should take place in the stores themselves, but when I call `dispatch(userActions.load())`, the store has to return the new state (i.e. `return { ...state, loading: true };`), so it can't return a promise to wait for. `dispatch()` returns the action passed to it for some reason. I'd really like something like...

`dispatch(someAsyncAction, successAction, failureAction) => Promise`

...where the promise doesn't resolve until one of the other two actions is dispatched.

Is that the sort of thing that could be enabled with the middleware pattern?

Am I totally off base and there's a simple way to do this already?

Thanks.

gaearon:
Hey, thanks!

>Ideally, the loading should take place in the stores themselves

Redux enforces that Stores are completely synchronous. What you describe should happen in the action creator instead.

I *think* it may even be possible with the default thunk middleware. Your action creator would look like:

```js
export function doSomethingAsync() {
  return (dispatch) => {
    dispatch({ type: SOMETHING_STARTED });

    return requestSomething().then(
      (result) =>  dispatch({ type: SOMETHING_COMPLETED, result }),
      (error) =>  dispatch({ type: SOMETHING_FAILED, error })
    );
  };
}
```

and handling the actual (granular) actions in the Store.

It's also possible to write a custom middleware to remove the boilerplate.

erikras:
Genius! I figured I was overlooking something obvious. I like that separation of *doing* and *storing*.

I look forward to watching this library grow, although it's pretty doggone complete already. Cheers, @gaearon!

gaearon:
You can also write a custom middleware like this

```js
export default function promiseMiddleware() {
  return (next) => (action) => {
    const { promise, ...rest } = action;
    if (!promise) {
      return next(action);
    }

    next({ ...rest, readyState: 'request' );
    return promise.then(
      (result) => next({ ...rest, result, readyState: 'success' }),
      (result) => next({ ...rest, error, readyState: 'failure' })
    );
  };
}
```

and use it instead of the default one.

This will let you write async action creators like

```
function doSomethingAsync(userId) {
  return {
    type: SOMETHING,
    promise: requestSomething(userId),
    userId
  };
}
```

and have them turn into

```
[...]

PR #172, opened 2015-06-23
added link to erikras' starter kit example in README

erikras:
Added at @gaearon's request in #99.

Wasn't sure where exactly to put it, so I put it next to @emmenko's example.


gaearon:
Awesome job

PR #174, opened 2015-06-23
Checking that the return value of every store function inside compose…

honzatrtik:
…Stores is not undefined

### DIFF
--- a/src/utils/composeStores.js
+++ b/src/utils/composeStores.js
@@ -5,13 +5,20 @@ import invariant from 'invariant';
 export default function composeStores(stores) {
   const finalStores = pick(stores, (val) => typeof val === 'function');
 
+  const dummyAction = {};
+  const dummyStore = {};
+
   Object.keys(finalStores).forEach(key => {
-    invariant(
-        typeof finalStores[key]({}, {}) !== 'undefined',
-        'The return value of `%s` reducer function must not be undefined.',
-        key
-    );
+    if (finalStores[key](dummyStore, dummyAction) !== dummyStore) {
+      const message = `Your ${key} Store must return the state given to it for any unknown actions.`;
+      if (process.env.NODE_ENV === 'production') {
+        console.warn(message);
+      } else {
+        throw new Error(message);
+      }
+    }
   });
+
   return function Composition(atom = {}, action) {
     return mapValues(finalStores, (store, key) =>
       store(atom[key], action)
--- a/src/utils/composeStores.js
+++ b/src/utils/composeStores.js
@@ -1,8 +1,17 @@
 import mapValues from '../utils/mapValues';
 import pick from '../utils/pick';
+import invariant from 'invariant';
 
 export default function composeStores(stores) {
   const finalStores = pick(stores, (val) => typeof val === 'function');
+
+  Object.keys(finalStores).forEach(key => {
+    invariant(
+        typeof finalStores[key]({}, {}) !== 'undefined',
+        'The return value of `%s` reducer function must not be undefined.',
+        key
+    );
+  });
   return function Composition(atom = {}, action) {
     return mapValues(finalStores, (store, key) =>
       store(atom[key], action)
--- a/src/utils/composeStores.js
+++ b/src/utils/composeStores.js
@@ -1,6 +1,5 @@
 import mapValues from '../utils/mapValues';
 import pick from '../utils/pick';
-import invariant from 'invariant';
 
 export default function composeStores(stores) {
   const finalStores = pick(stores, (val) => typeof val === 'function');
--- a/src/utils/composeStores.js
+++ b/src/utils/composeStores.js
@@ -7,16 +7,13 @@ export default function composeStores(stores) {
   const dummyAction = {};
   const dummyStore = {};
 
-  Object.keys(finalStores).forEach(key => {
-    if (finalStores[key](dummyStore, dummyAction) !== dummyStore) {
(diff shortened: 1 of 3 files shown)


src/utils/composeStores.js
```diff
@@ -1,8 +1,17 @@
 import mapValues from '../utils/mapValues';
 import pick from '../utils/pick';
+import invariant from 'invariant';
 
 export default function composeStores(stores) {
   const finalStores = pick(stores, (val) => typeof val === 'function');
+
+  Object.keys(finalStores).forEach(key => {
+    invariant(
```
gaearon: Should this be a hard error, or a warning in dev?

src/utils/composeStores.js
```diff
@@ -1,8 +1,17 @@
...
+        typeof finalStores[key]({}, {}) !== 'undefined',
```
gaearon: Should be two less spaces I think.

src/utils/composeStores.js
```diff
@@ -1,8 +1,17 @@
...
+        'The return value of `%s` reducer function must not be undefined.',
```
gaearon: Until we adopt the new naming everywhere let's change the message to

"Your `%s` Store returns undefined when given an unknown action. Change it to return the `state` given to it for any unknown actions."

src/utils/composeStores.js
gaearon: I'd say let's compare reference equality here instead.
Something like

```js
const dummyState = {};
const dummyAction = {};
invariant(finalStores[key](dummyState, dummyAction) === dummyState, ...)
```

src/utils/composeStores.js
dariocravero: @gaearon I guess the answer to that is: would `undefined` be a valid state? If so, then I think that we should warn, otherwise fail :)

src/utils/composeStores.js
gaearon: Okay, I think I'm confident now I only want to do this in development. Let's add `process.env.NODE_ENV !== 'production'` check and use `console.warn` instead.

honzatrtik: @gaearon you mean that every reducer (store) called with dummyAction should act as identity function?

honzatrtik: If so, then i would omit 'undefined' and change message to something like "Your %s Store must return the state given to it for any unknown actions."

src/utils/composeStores.js
gaearon: I think it's a symptom of problem but I'm not sure it's worth failing in production. It's like wrong propTypes.

src/utils/composeStores.js
gaearon: Yes, let's use this phrasing.

taylorhakes:
@honzatrtik @gaearon I feel this is only needed during development. I think it would be best if we don't have this check in production. Wrap the whole thing in 

```
if (process.env.NODE_ENV === 'production') {
 // code here
}
```
Instead of just changing the error type. It will make production performant and not need to go through unnecessary code checks.

src/utils/composeStores.js
```diff
@@ -3,6 +3,21 @@ import pick from '../utils/pick';
...
 export default function composeStores(stores) {
   const finalStores = pick(stores, (val) => typeof val === 'function');
+
+  const dummyAction = {};
+  const dummyStore = {};
+
+  Object.keys(finalStores).forEach(key => {
+    if (finalStores[key](dummyStore, dummyAction) !== dummyStore) {
+      const message = `Your ${key} Store must return the state given to it for any unknown actions.`;
+      if (process.env.NODE_ENV === 'production') {
```
gaearon: Let's do it the other way.
In development, we warn. In production, we skip this whole check (lines 10 to 20).

taylorhakes:
I gave this some thought. This change might cause some issues. If someone has runtime type checking in their reducers, this will cause an error. 

Also, If they have some type of reducer that counts the number of actions that go through the reducer no matter the action. This change will make their code always show a warning.

If we don't think either of those are likely, then the benefits outweigh those downsides.

gaearon:
>If someone has runtime type checking in their reducers, this will cause an error.

What kind of type checking are you referring to?

>If they have some type of reducer that counts the number of actions that go through the reducer no matter the action.

Reducers must be pure in Redux. That's a hard constraint but it's not currently enforced. It's better if we log a warning than just behave weirdly.

taylorhakes:
> What kind of type checking are you referring to?
There are libraries that convert Flow/Typescript type definitions to runtime checks. 

I am not sure if default Flow would allow this to pass if your reducer function doesn't have type Object for state.

For instance in the counter example, the reducer always expects a number type state. This check is passing an object to all reducers.

gaearon:
Oh, I see now.. I haven't really thought about Flow in the context of this change.
Let's hear others' thoughts about this.

gaearon:
My suggestion: pass `undefined` as the state and make sure the returned result is *not* `undefined`.

gaearon:
Thanks for the food for thought.
I don't plan to merge this PR as is because it is superseded by #191.

Please let me know if you would like to implement #191 (throw if undefined is returned) instead.
Sorry for the churn!

honzatrtik:
Yup, i will try to implement #191 & apropriate tests tonight...

gaearon:
I think @taylorhakes is working on this already: #193

honzatrtik:
oh, i see! ok then.

PR #175, opened 2015-06-24
Removed .jsx from webpack config file, inside examples

djhojd:


gaearon:
We don't really use JSX extension so there's no point. But. There's indeed a typo there. :-) There's no need for extra `?` at the end. I'd merge it if you fixed that in both examples' configs.

djhojd:
Yes, here we do not use JSX but you have `extensions: ['', '.js', '.jsx']` and if someone would copy your webpack config file, take me for instance, and use .jsx it will end up patching your file :-).

johanneslumpe:
Wouldn't it then be more appropriate to remove `.jsx` from the webpack config instead?

emmenko:
@johanneslumpe yes

@djhojd please simply remove `.jsx` and the question mark in the regex

gaearon:
Thanks!

PR #179, opened 2015-06-25
subscribe moved from Connector constructor to componentDidMount

mattybow:
You shouldn't need to attach the Connector components to redux's setState call unless the component mounts which should only be on the browser side.  This will provide more flexibility in the way people decide to do server rendering because componentDidMount is not called on the server.

### DIFF
--- a/src/components/createConnector.js
+++ b/src/components/createConnector.js
@@ -38,8 +38,11 @@ export default function createConnector(React) {
     constructor(props, context) {
       super(props, context);
 
-      this.unsubscribe = context.redux.subscribe(::this.handleChange);
       this.state = this.selectState(props, context);
+    }
+
+    componentDidMount() {
+      this.unsubscribe = this.context.redux.subscribe(::this.handleChange);
     }
 
     componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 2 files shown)


src/components/createConnector.js
```diff
@@ -38,10 +38,14 @@ export default function createConnector(React) {
     constructor(props, context) {
       super(props, context);
 
-      this.unsubscribe = context.redux.subscribe(::this.handleChange);
+      this.redux = context.redux;
```
emmenko: You don't have to to this (see below)

src/components/createConnector.js
```diff
@@ -38,10 +38,14 @@ export default function createConnector(React) {
...
       this.state = this.selectState(props, context);
     }
 
+    componentDidMount() {
+      this.unsubscribe = this.redux.subscribe(::this.handleChange);
```
emmenko: You can access `redux` from the context

```js
this.context.redux.subscribe
```

mattybow:
ok, thanks makes sense, I've modified as suggested

gaearon:
Can you please add a unit test verifying this change, and then squash the commits into a single commit? Thanks.

mattybow:
@gaearon added unit test and squashed, very new to this, forgive me.  Let me know if I need to change anything

PR #182, opened 2015-06-26
TodoMVC - Do not save onBlur if newTodo

dbrans:
Currently if you blur the newTodo TextInput containing text you'll see a new todo item get added.
Blurring probably shouldn't have any effect on the newTodo TextInput.


vramana:
@dbrans Nice catch!

gaearon:
:+1:

PR #185, opened 2015-06-27
Add Slack-in Badge

listochkin:
Unfortunately, there's no good way to point to a specific channel on Slack so I use a link to Slack-in website instead. That's yet another reason why you should probably use Gitter instead.

I use React Blue color. Feel free to improve it.


gaearon:
:+1:

PR #188, opened 2015-06-28
Replace trailing whitespaces with a newline

clearjs:
This splits the lines just as well, and wouldn't get destroyed when someone has automatic trimming of trailing whitespace enabled or just doesn't know the intention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ redux
 
 Atomic Flux with hot reloading.
 
-**The API is likely to change a few times before we reach 1.0.**  
+**The API is likely to change a few times before we reach 1.0.**<br>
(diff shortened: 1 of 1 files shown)


gaearon:
Haha. This creates a paragraph :-). I want a line break like it is now.

clearjs:
You're right... Would <br> be fine instead?

Apparently Github doesn't use GitHub Flavored Markdown, despite the name: http://forums.apricitysoftware.com/t/github-flavored-markdown-normale-linebreak-does-not-exist/111

gaearon:
Let's br.

clearjs:
Amended.

Issue #191, opened 2015-06-29
Change composeStores to throw if undefined is returned


gaearon:
Right now it's easy to forget a `default` case and have your reducer return nothing in some circumstances. There have been some WIP attempts to address it: #174. Here's the issue that spawned that PR: #173.

But let's take a moment to think about what `composeStores` really is. It's a shortcut right? It's a bit opinionated already: it assumes plain objects as the initial state. See #153.

Can we make it a bit more opinionated for convenience? <s>We could make it return the previous `state` if the reducer returns `undefined`.</s>

**Edit: see [this](https://github.com/gaearon/redux/issues/191#issuecomment-116739085) instead.**

Pros:

* Solves issues like #173 nicely
* Slightly less boilerplate in the “vanilla Redux” code

Cons:

* Dilutes the contract of reducers by introducing some magic

Thoughts?

gaearon:
To be honest I can't imagine `undefined` being a valid state. If you want “nothing” just return `null`.
Therefore I think we should do something with `undefined`:

* Either interpret it as “keep the previous state” as suggested above
* Or throw.

gaearon:
Here's something wrong with this change. People will start with `composeStores` but might wrap their reducers into community-provided helpers (e.g. DevTools, pagination utils, etc) later. If all of these community-provided have to special-case `undefined`, it's going to be a pain. It's also easy to forget about this when writing a utility, and to interpret `undefined` as a value. Everyone's going to have to write special branches to handle this case.

Therefore I propose, instead of interpreting `undefined` as a special value, **narrow down** the `composeStores`-compatible reducer contract. That is, throw when `undefined` is return.

It's a breaking change, but one that is easy to fix with `switch` statements, and even easier with custom reducer factories (might require a single line change). I think it's worth it to keep the contract interoperable, but prevent mistakes like #173.

vslinko:
I'm have a stores with `undefined` as valid value. For example `tokenStore` stores `undefined` or `string`.
I don't want to store null or any other object.

johanneslumpe:
Do you really check for undefined or just !token ? Also, instead or just returning a single value you could wrap it in an object and keep your undefined value.

I would welcome this change. It is a bit more opinionated, but it will solve some common mistakes.

vslinko:
I don't want to wrap every single value with an object. I want to write `const {token} = getState()` instead of `const {tokenStore: {token}} = getState()` or `const {token: {hash}} = getState()`.

[...]

PR #193, opened 2015-06-30
Throw error on undefined value from store function

taylorhakes:
Updated composeStores to throw if undefined is returned. #191

### DIFF
--- a/src/utils/composeStores.js
+++ b/src/utils/composeStores.js
@@ -1,11 +1,18 @@
-import mapValues from '../utils/mapValues';
-import pick from '../utils/pick';
+import mapValues from './mapValues';
+import pick from './pick';
+
+// UNDEF = undefined;
+const UNDEF = void 0;
 
 export default function composeStores(stores) {
   const finalStores = pick(stores, (val) => typeof val === 'function');
   return function Composition(atom = {}, action) {
-    return mapValues(finalStores, (store, key) =>
-      store(atom[key], action)
-    );
+    return mapValues(finalStores, (store, key) => {
+      const state = store(atom[key], action);
+      if (state === UNDEF) {
+        throw new Error(`Store ${key} returns undefined. By default store should return original state.`);
+      }
+      return state;
+    });
   };
 }
--- a/test/composeStores.spec.js
+++ b/test/composeStores.spec.js
@@ -1,6 +1,9 @@
 import expect from 'expect';
 import { composeStores } from '../src';
 
+// UNDEF = undefined;
+const UNDEF = void 0;
+
 describe('Utils', () => {
   describe('composeStores', () => {
     it('should return a store that maps state keys to reducer functions', () => {
@@ -26,6 +29,30 @@ describe('Utils', () => {
       });
 
       expect(Object.keys(store({}, {type: 'push'}))).toEqual(['stack']);
+    });
+    it('should throw an error if undefined return from store', () => {
+      const store = composeStores({
+        stack: (state = []) => state,
+        bad: (state = [], action) => {
+          if (action.type === 'something') {
+            return state;
+          }
+        }
+      });
+      expect(() => store({}, {type: '@@testType'})).toThrow();
+    });
+    it('should throw an error if undefined return not by default', () => {
+      const store = composeStores({
+        stack: (state = []) => state,
+        bad: (state = 1, action) => {
+          if (action.type === 'something') {
+            return UNDEF;
+          }
+          return state;
+        }
+      });
+      expect(store({}, {type: '@@testType'})).toEqual({stack: [], bad: 1});
+      expect(() => store({}, {type: 'something'})).toThrow();
     });
   });
 });


src/utils/composeStores.js
```diff
@@ -1,11 +1,18 @@
-import mapValues from '../utils/mapValues';
-import pick from '../utils/pick';
+import mapValues from './mapValues';
+import pick from './pick';
 
 export default function composeStores(stores) {
   const finalStores = pick(stores, (val) => typeof val === 'function');
+
+  if (process.env.NODE_ENV !== 'production') {
+    Object.keys(finalStores).forEach(function(key) {
```
ooflorent: Why didn't you use an arrow function here?

gaearon:
This is "probing" a reducer but I actually want the check to run every time a reducer is called. So it needs to be inside the function actually calling it.

gaearon:
Also, because it is a hard constraint, it should be enforced in production too. No need for the if statement. Otherwise pros and dev bugs would differ.

taylorhakes:
I changed it because I was considering unexpected issues in production. For instance, if the default doesn't return `undefined`, but one of their case statements. This could cause their whole state not to update. It is debatable whether an `undefined` state is worse than causing a full update not to work.

I think it generally shouldn't be an issue in dev. I think that someone might run into the error at the wrong time in production, unfortunately.

taylorhakes:
I made that change to your suggestion. I could be wrong and people will prefer a hard error, even in production.

test/composeStores.spec.js
```diff
@@ -27,5 +27,29 @@ describe('Utils', () => {
...
+        }
+      });
+      expect(() => store({}, {type: '@@testType'})).toThrow();
+    });
+    it('should throw an error if undefined return not by default', () => {
+      const store = composeStores({
+        stack: (state = []) => state,
+        bad: (state = 1, action) => {
+          if (action.type === 'something') {
+            return undefined;
```
ooflorent: I'm not very fond of `undefined` since it may be overridden.

```js
var undefined = null
console.log(undefined) // null
```

taylorhakes: I thought about doing that, but I wasn't sure the stance of this repo. I would also need to change it in composeStores.js also. What are people's thoughts?

gaearon:
If something warns in development it must no-op in production.
If something throws in development it must throw in production.

You're not going to "save" the app by returning undefined in production. It's already unexpected what will happen (because it never happened in dev apparently). 

It's important to keep crashes consistent. Otherwise you won't be able to reproduce a production crash report in development. The different stack trace will mislead you and you'll have a hard time figuring out why you never have crashes with that stack trace in development.

src/utils/composeStores.js
```diff
@@ -1,11 +1,18 @@
-import mapValues from '../utils/mapValues';
-import pick from '../utils/pick';
+import mapValues from './mapValues';
+import pick from './pick';
+
+// UNDEF = undefined;
+const UNDEF = void 0;
```
gaearon: It looks like a hack. Can you please check typeof instead?

gaearon:
Here's a potential crazy issue you might encounter if undefined was tolerated in production: loss of user data. Instead of crashing, data is overridden, and the app saves the bad state.

Crashing isn't always bad :-)

taylorhakes:
Definitely, I hinted above that there is a trade off. If the app crashes in prod, losing state might be better than blocking the action. Depends on the app. 

I just trying to note that by fixing one issue, we are introducing another. Probably will be less an issue than the original. Time will tell.

I am fine making it a hard error. I understand your reasoning.

src/utils/composeStores.js
taylorhakes: This is a pretty common pattern. I will make the change. How should I fix the other UNDEF use in the test?

gaearon: You can replace `return UNDEF` with just `return`, and use `typeof x === 'undefined'` as a check.
Or are you referring to something else?

test/composeStores.spec.js
gaearon: Just `return` is enough :-).

gaearon:
@taylorhakes Once you get to that, would you please make another PR on top of https://github.com/gaearon/redux/pull/195? It contains the renaming of this function..

src/utils/composeStores.js
taylorhakes: Unfortunately, there is an ESLint rule that forbids `return;` when there is another type returned elsewhere. I think we can just use `undefined` though. We don't have to worry about users overwriting `undefined` in tests.

gaearon: I think you can also invert the condition.
Instead of returning undefined in some cases, return *something* in other cases.

```js
function bad() {
  if (action.type === 'something') {
    return something;
  }
}
```

taylorhakes: I thought about that. I was trying to make the test as obvious as possible. I will change it to above

taylorhakes:
I will need to close this pull request to make a pull request against the alpha branch.

Issue #164, opened 2015-06-22
Umbrella 1.0


gaearon:
- [x] Decide on API changes needed for time travel and similar tools (#113)
- [x] Remove the dispatcher (<s>#166</s> #195)
- [x] Pass `dispatch` to middleware (<s>#120</s> #195)
- [x] Make sure nested dispatches are fixed (needs a test from #119, should be solved by #195)
- [x] Finalize terminology (#140, #195)
- [x] Throw in `composeReducers` if either returns `undefined` (#191, #193)
- [x] Decide on API for binding action creators (#86)
- [x] Release DevTools (#113)
- [x] Fix build on Windows (#94)
- [x] Split packages (https://github.com/gaearon/redux/issues/230)
- [x] Merge https://github.com/gaearon/redux/pull/295
- [x] Merge https://github.com/gaearon/redux/pull/302
- [x] Create new docs (#140)

gaearon:
While most of these already have PRs ready to be merged, I would like to finish my work in #113 (I have a proof of concept and now need to take some time to figure out the right API). This is to avoid two API changes in a row. Thanks everyone for your hard work and please be patient for a few more days. ;-)

gaearon:
1.0 is out.

PR #195, opened 2015-06-30
Breaking API changes for 1.0

gaearon:
### Naming

* “Stateless Stores” are now called reducers. (https://github.com/gaearon/redux/issues/137#issuecomment-114178411)
* The “Redux instance” is now called “The Store”. (https://github.com/gaearon/redux/issues/137#issuecomment-113252359)
* The dispatcher is removed completely. (https://github.com/gaearon/redux/pull/166#issue-90113962)

### API changes

* <s>`composeStores`</s> is now `composeReducers`.
* <s>`createDispatcher`</s> is gone.
* <s>`createRedux`</s> is now `createStore`.
* `<Provider>` now accepts `store` prop instead of <s>`redux`</s>.
* The new `createStore` signature is `createStore(reducer: Function | Object, initialState: any, middlewares: Array | ({ getState, dispatch }) => Array)`.
* If the first argument to `createStore` is an object, `composeReducers` is automatically applied to it.
* The “smart” middleware signature changed. It now accepts an object instead of a single `getState` function. The `dispatch` function lets you “recurse” the middleware chain and is useful for async: #113 (comment).

### Correctness changes

* The `dispatch` provided by the default thunk middleware now walks the whole middleware chain.
* It is enforced now that raw Actions at the end of the middleware chain have to be plain objects.
* Nested dispatches are now handled gracefully. (#110, #119)

### Internal changes

* The object in React context is renamed from <s>`redux`</s> to `store`.
* Some tests are rewritten for clarity, focus and edge cases.
* Redux in examples is now aliased to the source code for easier work on master.


gaearon:
This PR supersedes #166, #119, and #120, and brings us much closer to 1.0 (#164).

The missing pieces for 1.0 are more strict `composeReducers` (#193), new docs (#140), and a more convenient action creator binding API (#86), if we are certain that we need it.

gaearon:
Released on NPM as `1.0.0-alpha`.

dariocravero:
:tada: :)

examples/counter/containers/CounterApp.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { bindActionCreators } from 'redux';
+import { bindActionCreators } from 'redux/index';
```
gaearon: Whoops, this slipped in by mistake when copy-pasting.

acdlite:
Yay! Looks good!

I've been wondering if `composeReducers()` is really the best name for that function. The name is very generic, whereas its implementation is very specific.

Consider another common way of composing reducers, which is to apply each one of them in sequence:

```js
(...reducers) => (prevState, action) =>
  reducers.reduce(
    (state, r) => r(state, action),
    prevState
  );
```

This could also be named `composeReducers()`. (I'd actually argue this is better fit for the name.)

Maybe we should think of a more specific name? Don't have any suggestions right now, but I wonder what everyone else's thoughts are.

gaearon:
@acdlite Any precedents in other libraries' APIs?

src/middleware/thunk.js
```diff
@@ -1,10 +1,6 @@
-export default function thunkMiddleware(getState) {
-  return (next) => {
-    const recurse = (action) =>
-      typeof action === 'function' ?
-        action(recurse, getState) :
-        next(action);
-
-    return recurse;
-  };
+export default function thunkMiddleware({ dispatch, getState }) {
```
dbrans: Where does the name "thunk" come from?

src/utils/composeReducers.js
```diff
@@ -0,0 +1,12 @@
+import mapValues from '../utils/mapValues';
+import pick from '../utils/pick';
+
+export default function composeReducers(reducers) {
+  const finalReducers = pick(reducers, (val) => typeof val === 'function');
+
+  return function Composition(atom = {}, action) {
```
dbrans: A few questions:
1. Why are you calling this "atom" and not "state" or "slice"?
2. Why does atom have a default value?
3. why is Composition capitalized?

src/middleware/thunk.js
```diff
@@ -1,10 +1,6 @@
...
+  return (next) => (action) =>
+    typeof action === 'function' ?
+      action(dispatch, getState) :
```
dbrans: Looks like thunkMiddleware gives the call-site the ability to look at the current state of the app and dispatch actions based on that state. Did I get that right? When / how would we use this?

src/Store.js
```diff
@@ -0,0 +1,50 @@
...
+  getReducer() {
+    return this.reducer;
+  }
+
+  replaceReducer(nextReducer) {
+    this.reducer = nextReducer;
+    this.dispatch({ type: '@@INIT' });
+  }
+
+  dispatch(action) {
```
dbrans: At this point this method is more like a "performAction" than a "dispatch".
The code reads like "perform an action which modifies state. Then notify listeners that state has changed".

src/Store.js
```diff
@@ -0,0 +1,50 @@
...
+    invariant(
+      isPlainObject(action),
+      'Actions must be plain objects. Use custom middleware for async actions.'
+    );
+
+    const { reducer } = this;
+    this.state = reducer(this.state, action);
+    this.listeners.forEach(listener => listener());
```
dbrans: The listener will probably turn around and call getState right away. Why not pass the state to the listener?

emmenko:
@gaearon one step forward, great work! :+1:

src/utils/composeReducers.js
gaearon: @dbrans 

1. “Atom” is leftover from earlier versions of API. Need to be changed to `state`.
2. We want the default values for reducers to be `undefined`. For `state[key]` to be undefined initially, `state` needs to be `{}`. If it's `undefined` itself, `state[key]` will just crash.
3. You're right, we should lower case that.

src/createStore.js
```diff
@@ -0,0 +1,45 @@
...
+
+  cookedDispatch = composeMiddleware(
+    ...finalMiddlewares,
+    rawDispatch
+  );
+
+  return {
+    dispatch: cookedDispatch,
+    subscribe: ::store.subscribe,
+    getState: ::store.getState,
```
iamdustan: I’m curious why a second `::store.getState` function is created instead of reusing the one from [#L20](https://github.com/gaearon/redux/pull/195/files#diff-01b6e016dc33dc6cf4466e7abd376200R20)?

src/middleware/thunk.js
iamdustan: as in so? https://en.wikipedia.org/wiki/Thunk

src/middleware/thunk.js
gaearon: For example, the action creator might want to exit early if some data is already available in the state. Also useful for dispatching an action and then reading the result (e.g. authentication workflow).

src/createStore.js
gaearon: Overlooked this :-). No real reason

src/Store.js
gaearon: @dbrans In fact this is how it used to be, but I felt it was redundant because the components would reach into `getState()` anyway in some cases. So there's no real reason other than making the API do less things. I don't have a strong opinion here though.

src/Store.js
gaearon: I agree in a way. I'm not sure it's worth changing though. It doesn't seem to cause much confusion (unlike “Stores” naming which did, so we're changing it to “Reducers”).

src/middleware/thunk.js
clearjs: https://github.com/tj/co#thunks

rpominov:
> I've been wondering if `composeReducers()` is really the best name for that function. 
> Maybe we should think of a more specific name? Don't have any suggestions right now, but I wonder what everyone else's thoughts are.

`combineReducers()` maybe?

gaearon:
I like combine.

acdlite:
:+1: for `combineReducers()`.

I think we should remove middleware as a parameter `createRedux()`. Because they have the potential to be async, middleware needs to be applied *before* being passed to higher-order stores like `createDebuggerRedux()`: https://github.com/gaearon/redux/pull/166#issuecomment-114222798

Instead, I think the API for middleware should be something like `applyMiddleware()` from that comment.

gaearon:
Oh, you're right!

gaearon:
@acdlite Please do!

acdlite:
@gaearon Hmm, just realized... we can't keep `thunkMiddleware()` as a default if we move middleware out of `createStore()`.

Also, should `applyMiddleware()` be composable using `compose()`? E.g.

```js
const store = compose(
  applyMiddleware(t => [ thunkMiddleware(t), promiseMiddleware ]),
  debugger,
  createStore
)(reducer, intialState)
```

or should it be applied to the store after it's created? E.g.

```js
const store = applyMiddleware(
  t => [ thunkMiddleware(t), promiseMiddleware ],
  debugger(createStore)(reducer, initialState)
)
```

The second version looks nicer and is less likely to confuse people, but the first version is more flexible and "correct" since it has the same interface as debugger.

acdlite:
(Hahaha I guess calling it `debugger` isn't an option :D)

acdlite:
Submitted a PR to this branch to update middleware API https://github.com/gaearon/redux/pull/213

skaapgif:
Like you said in your talk @gaearon, it often takes months to think about something and only a short time to code it once you've established a good mental model.

I think redux can benefit greatly from established backend mental models - specifically "event sourcing". @vladap Already hinted at the similarities in https://github.com/gaearon/redux/pull/140#issuecomment-115157831

Event sourcing has similar goals to redux:
 - By keeping a log of every event it's easy to "time travel" and evaluate the complete application state at different times.
 - In the event of a bug you can simply change the code, replay all your events and arrive at a consistent state
 - It separates writes from reads which reduces complexity. Instead of a component mutating state directly, it triggers an event (a write) which trickles down the system and eventually updates the state for future reads.

There are important architectural details which should be kept in mind with a event sourced system. One of the most important being the separation of commands / events. Commands are requests for state to be mutated. Depending on business logic these command requests can be denied. Events on the other hand can be thought of as approved commands and are written in the past tense to make this clear.

In my understanding of Redux these two concepts map nicely to `actionCreators` and `actions`. `actionCreators` receive commands, it might reject these commands silently, or it could create several events based on this command such as: `ADD_TODO_REQUESTED`, `ADD_TODO_FAILED`, `ADD_TODO_SUCCEEDED`.

`Reducers` take these events and synchronously modify the state of the application accordingly.

Do you think @vladap and I's analogy to event sourcing is valid for reflux? Would it help to use the command/event analogy? Since "event" is close to browser events, "facts" might be a better fit.

gaearon:
Separation of commands and events is something @glenjamin has been doing in [fluctuations](https://github.com/glenjamin/fluctuations):

```js
var interceptor = fluctuations.createInterceptor({
    FETCH_MESSAGE: function(emit, payload) {
        emit("FETCH_MESSAGE_BEGIN");
        setTimeout(function() {
            emit("CHANGE_MESSAGE", { value: "whatever" });
        }, 2000);
    }
});
```

I have no opinion on this yet. I'd probably stay with the current middleware approach for now, and see if something interceptor-like appears in the community.

skaapgif:
Thanks I had a look at fluctuations...

Middleware is a powerful concept since it allows you to create a chain of transformers on an event instead of sticking to a single function.

I'm not sure I completely understand the current middleware proposal. I read about
> "dispatch-level "middleware" and store-level "middleware"

from https://github.com/gaearon/redux/pull/166#issuecomment-114217907 

Can I think of these two middleware stacks as `command-middleware` and `fact-middleware`?

Using these terms has some advantages over redux inventing it's own... it becomes more apparent that the command-middleware has to come before the fact-middleware. It also becomes more obvious that you should store a stream of facts for your time machine and throw away all commands. If you only traverse facts you solve the problems with async dependencies as mentioned in https://github.com/gaearon/redux/issues/113#issuecomment-114210769 since the async commands are already resolved to synchronous facts.

Or am I misunderstanding the current architecture?

clearjs:
+1. IMO this naming would add clarity, provided that it's accurate.

acdlite:
@skaapgif I feel like those terms are too limited. What we're calling "higher-order stores" right now is not just about intercepting "commands" — the uses cases are far more general. It's much the same concept as higher-order components in React. Yes, you can use them to intercept actions, and that will probably be the most common use case. However, you can also use them to do things like add the transducer protocol to a store, like so: https://github.com/acdlite/redux-transducers.

The best API is often no API. The current proposals for middleware and higher-order stores have the tremendous benefit is that they receive no special treatment by the core API — they're just wrappers around `dispatch()` and `createStore()`, respectively. You can even use them today, before 1.0 is released. That's a huge win for extensibility and rapid innovation. We should favor *patterns and conventions* over *rigid, privileged APIs*.

[See a fuller explanation of higher-order stores here.](https://github.com/gaearon/redux/blob/improve-docs/docs/higher-order-stores.md)

skaapgif:
Thanks for the links @acdlite, I didn't notice the improve-docs branch, I'll have a read through it to make sure my understanding is correct.

skaapgif:
So am I correct in saying there's no such thing as dispatch-level "middleware" and store-level "middleware"... just middleware? Middleware always wraps `dispatch()` and higher-order stores are stores which have middleware applied to them?

The difference between commands and facts then are the input and output of a single dispatch function or a middleware chain wrapping the dispatch function. Commands go into this chain and facts come out. These fact(s) are then "reduced" into new state change(s).

acdlite:
@skaapgif I think the soft consensus is that when we say "middleware" in Redux, we're talking specifically about a function that wraps or proxies `dispatch()`. Higher-order stores often employ middleware, but they are a superset, because they also have the ability to proxy or amend other parts of the Redux interface. `applyMiddleware()` is a good example because it shows the relationship between the two concepts — it is a higher-order store that implements middleware.

grrowl:
For clarification, here's an example how to do "new" middleware:
```js
export default function actionDebugMiddleware(next) {
  return (action) => {
    console.info(action.type, action);
    next(action);
  };
}
```

acdlite:
@grrowl It will actually be this in 1.0, because the outer function receives `({ dispatch, getState })`:

```js
export default function actionDebugMiddleware() {
  return next => action => {
    console.info(action.type, action);
    next(action);
  };
}
```

grrowl:
Bizarre, that's what I had originally, but `next` was being filled with the `action`, with `action` undefined. On npm `1.0.0-alpha`. I'll dig deeper

acdlite:
Oh that hasn't been merged yet :) Should have clarified. https://github.com/gaearon/redux/issues/213

gaearon:
@grrowl Merged now (but not released yet).

vladap:
The main difference in command sourcing and event sourcing is how parts of the system are coupled. Event sourced system is highly de-coupled while command sourced system is coupled. I will try to show it on a contrived example.

ActionCreator buys pizzas. Each pizza comes with coffee and cola. There are 3 reducers, one counts bought pizzas, one counts coffees and one counts colas.

Now what action you should dispatch when AC finishes buying pizzas?

You can dispatch BUY_PIZZA but then you get a semantic problem that reducers were told to perform BUY_PIZZA action but they do something else. It is like you would tell me to buy pizza but actually wanted me to count them.

Or you want to dispatch actions describing what they should do. But which one? Each is doing something else. If you choose to dispatch all of them INC_PIZZA, INC_COFFEE, INC_COLA, it requires that AC knows what each reducer is doing. Even more worse is that if you want to add another reducer which would do something else when pizza is bought you are breaking Single Responsibility Principle, because by adding reducer you have to update ActionCreator as well if it requires new action!

Or you can invent some action covering all cases, you will probably end up with something generic non-descriptive like DO_WHATEVER_YOU_WANT.

With event sourced mental model AC writes facts and doesn't care at all what happens next. AC simply log PIZZA_BOUGHT. First reducer sees PIZZA_BOUGHT and thinks, when pizza is bought I'm responsible to increase pizza count. Next one inc coffee and the next inc cola. Everything is correct and it makes sense even under semantic lenses.

And you can add how many new reducers you want doing whatever you wish when pizza is bought. You don't need to touch AC. You do so only to extend its logging capability when your reducers need more facts.

Our teams writing transactional systems know nothing what our analytics team does. Command sourced mentality wouldn't work at all here. They just log what they did and sometimes we tell them to log some more.

Webapp components are very close to each other and they are developed by a single team/person so it is possible to get away with command sourced mentality. Game engines are (were) doing it for years (game replays). I still believe that thinking event sourced, understanding that past never changes but anybody can come and make a new decision/processing based on facts in the past is a good model to think in.

glenjamin:
My current thinking is that action creators act like commands, and actions act like events.

The redux approach to storage and replay seems closest to event sourcing to me.

The experiments with APIs I'm doing in fluctuations are largely aimed at solving two problems:
* reducing flux boilerplate
* connecting action creators to a dispatcher without using singletons

vladap:
Those damned ActionCreators. There is bindActionCreators() but dispatch(action) on a component level. We know that we actually dispatch (or something like that) actionCreator.

But isn't it confusing for somebody new this ActionCreator vs Action "fight"?

I'm fine with dispatch(action) and would like to have bindActions but it is not technically correct, at least not for async code. It works when there is 1:1 relation between AC and A, then AC can be seen as a hidden mediator between views and stores. It stops to work when AC starts to be more complex having quite a bit of logic and generates additional actions.

Saying that I'm binding AC and dispatching AC feels bad.

johanneslumpe:
@vladap No we do not dispatch an action creator, since what we do is this: `dispatch(actionCreator())`. We are dispatching the result of an action creator, which is an action.

gaearon:
>We are dispatching the result of an action creator, which is an action.

This is right.
See also https://github.com/gaearon/redux/blob/improve-docs/docs/types-and-terminology.md (work in progress).

src/Store.js
iclanzan: I really think state should be passed to the listener.

vladap:
Thanks for clarifying. The problem is then somewhere between my chair and keyboard :)

gaearon:
The version currently in this branch is likely to become the 1.0 release candidate.
If you wanna play with it ahead of the release, please do!

ryancole:
dat switch to `1.0.0-rc` :+1:

gaearon:
1.0 RC release notes: https://github.com/gaearon/redux/releases/tag/v1.0.0-rc

vslinko:
How to `bindActionCreators` right now? I don't get it :sos:

vslinko:
Looks like it's impossible in decorator.
Is that means I should use `Connect` component every time I need `bindActionCreators`?

vslinko:
@gaearon What if change `select` signature to `select(state: mixed, dispatch: Function): mixed` ?

vslinko:
Found #86 
Sorry for noise :smile:

gaearon:
@vslinko Here's the `@connect` API I have in mind: https://github.com/gaearon/react-redux/issues/1#issuecomment-120729921. Please let me know if you see ways to make it better (in that thread).

kagia:
I don't know if anyone  would find this useful, but here is the gist I use for setting up my store with the new API. [gist](https://gist.github.com/kagia/e953b2fd9897a2ba4146)

RnbWd:
Wow - I remember when this library was just a [pseudo-code](https://gist.github.com/gaearon/c02f3eb38724b64ab812)! I took a 2 month hiatus and now it's grown into something incredible :smirk:. TWO MONTHS!!! (I apologize if this is off topic).

[Smart/Dumb components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) was excellent article - but Rich Hickey's talks have had profound influences on myself:

[Simplicity Matters](https://www.youtube.com/watch?v=rI8tNMsozo0) --- [reducers](https://vimeo.com/45561411) --- [transducers](https://www.youtube.com/watch?v=6mTbuzafcII), to name a few. 

But the most powerful feature of redux - what I never expected - was the ecosystem. Projects that encourage plugins and have a minimal core-api seem to dominate the npm world: `express/connect` `gulp`, `grunt`, etc.  - *never imagined it would be used for flux-esc architecture* 

IMO simplify the core API as much as possible (like react-DND) - and encourage the plugin the ecosystem. I want to help contribute :)  -- (after I grok the concepts fully)

danmartinez101:
2.5 (insanely hectic for you)months from "Initial Commit" to "1.0" is kind of crazy... Nice work :)

PR #197, opened 2015-07-01
Throw error on undefined value from reducer function

taylorhakes:
Accidentally created against alpha before. Sorry about that.

Updated composeReducers to throw error when reducer returns undefined. Update to #193. #191

### DIFF
--- a/test/composeReducers.spec.js
+++ b/test/composeReducers.spec.js
@@ -29,5 +29,30 @@ describe('Utils', () => {
         Object.keys(reducer({}, { type: 'push' }))
       ).toEqual(['stack']);
     });
+
+    it('should throw an error if undefined return from reducer', () => {
+      const reducer = composeReducers({
+        stack: (state = []) => state,
+        bad: (state = [], action) => {
+          if (action.type === 'something') {
+            return state;
+          }
+        }
+      });
+      expect(() => reducer({}, {type: '@@testType'})).toThrow();
+    });
+    
+    it('should throw an error if undefined return not by default', () => {
+      const reducer = composeReducers({
+        stack: (state = []) => state,
+        bad: (state = 1, action) => {
+          if (action.type !== 'something') {
+            return state;
+          }
+        }
+      });
+      expect(reducer({}, {type: '@@testType'})).toEqual({stack: [], bad: 1});
+      expect(() => reducer({}, {type: 'something'})).toThrow();
+    });
   });
 });
--- a/src/utils/composeReducers.js
+++ b/src/utils/composeReducers.js
@@ -1,12 +1,18 @@
 import mapValues from '../utils/mapValues';
 import pick from '../utils/pick';
+import invariant from 'invariant';
 
 export default function composeReducers(reducers) {
   const finalReducers = pick(reducers, (val) => typeof val === 'function');
 
   return function composition(state = {}, action) {
-    return mapValues(finalReducers, (store, key) =>
-      store(state[key], action)
-    );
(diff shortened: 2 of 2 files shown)


taylorhakes:
@gaearon Is this good to merge?

gaearon:
Yeah, can you please rebase so it merges cleanly?

taylorhakes:
Just did the rebase. The tests pass locally, but are failing in Travis. Have you seen this before?

gaearon:
Something to do with linting. I'll merge soon, thanks.

acdlite:
Sorry for only now bringing this up, but I believe the check should go inside the Store class rather than `composeReducers()`. That way it works for all reducers, not just those created by that function. Also, we use the invariant module for consistency, like @gaearon does here: https://github.com/gaearon/redux/blob/breaking-changes-1.0/src/Store.js#L26-L29

acdlite:
^ In fact the new invariant should go just a few lines below the highlighted portion

gaearon:
>Sorry for only now bringing this up, but I believe the check should go inside the Store class rather than composeReducers().

But then we'll check the root value (which is unlikely to be undefined if you use `composeReducers`).

gaearon:
I think we're actually doing the right thing here. First, we want to check the actual app's reducers—not the root one. Second, we don't want this to be a _hard_ limitation—just something the _default_ utilities help you with. You're free to use your own `composeReducers` that allows `undefined` for individual reducers, or even not use `composeReducers` at all.

acdlite:
@gaearon Ah I see. It should go in both places, then, right?

acdlite:
Okay I can agree with that. That makes sense.

gaearon:
Why do that though? Redux doesn't really care what your root value is like. I'd say it should only be enforced at `composeReducers` level, as our built-in utilities follow “make it easier for a beginner” approach.

acdlite:
Yep, agreed. Carry on :)

taylorhakes:
Updated code to use invariant

gaearon:
Can you please rebase on master again? I think the linting problem is fixed there by your earlier PR.

test/composeReducers.spec.js
```diff
@@ -29,5 +29,28 @@ describe('Utils', () => {
         Object.keys(reducer({}, { type: 'push' }))
       ).toEqual(['stack']);
     });
+    it('should throw an error if undefined return from reducer', () => {
```
gaearon: missing newline before `it`

test/composeReducers.spec.js
```diff
@@ -29,5 +29,28 @@ describe('Utils', () => {
...
+        stack: (state = []) => state,
+        bad: (state = [], action) => {
+          if (action.type === 'something') {
+            return state;
+          }
+        }
+      });
+      expect(() => reducer({}, {type: '@@testType'})).toThrow();
+    });
+    it('should throw an error if undefined return not by default', () => {
```
gaearon: missing newline before `it`

taylorhakes:
I am a little confused. This change was made on top of breaking-changes-1.0. Do you want want me to merge master into breaking-changes-1.0? This change can't be rebased on master because it has to do with reducers.

gaearon:
Oh, I didn't mean on master, sorry :-(
I just updated `breaking changes` branch.

gaearon:
All right, this is in.
I slightly tweaked the tests and error messages: https://github.com/gaearon/redux/commit/de685de23289575f344638370912fea4989c0350.

PR #200, opened 2015-07-01
Modify test case ID generation for reducer changing

iamdustan:
While I was reading through the test cases in the #195 it caught me off guard that the IDs were being reused. This modifies the test case to not reuse IDs so you can follow the flow of where the reducer is swapped with where the next ID appears rather than only the task text string.

### DIFF
--- a/test/helpers/reducers.js
+++ b/test/helpers/reducers.js
@@ -1,10 +1,17 @@
 import { ADD_TODO } from './actionTypes';
 
+
+function id(state = []) {
+  return state.reduce((result, item) => (
+    item.id > result ? item.id : result
+  ), 0) + 1;
+}
+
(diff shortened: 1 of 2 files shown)


test/helpers/reducers.js
```diff
@@ -1,10 +1,16 @@
 import { ADD_TODO } from './actionTypes';
 
+
+var id = (state = []) =>
```
gaearon: Let's change this to be  `function nextId(state = [])`.
Also please run `npm run lint` to fix the errors.

Thanks!

iamdustan: but the I have to write all of `return` ;)

iamdustan: done

PR #204, opened 2015-07-02
Removed shallowEqualScalar, simplified connect update logic

taylorhakes:
I saw there was a lot of duplicated logic between shallowEqual and shallowEqualScalar. In the process of refactoring, I noticed createConnectDecorator has extra `shouldComponentUpdate` checks that don't exist in createConnector.

That means
```js
  render() {
    return (
      <Connector select={select}>
        {({ counter, dispatch }) =>
          /* Yes this is child as a function. */
          <Counter counter={counter}
                   {...bindActionCreators(CounterActions, dispatch)} />
        }
      </Connector>
    );
  }
}
```
behaves differently than this 
```js
@connect(state => ({
  counter: state.counter
}))
  render() {
    const { counter, dispatch } = this.props;
    // Instead of `bindActionCreators`, you may also pass `dispatch` as a prop
    // to your component and call `dispatch(CounterActions.increment())`
    return (
      <Counter counter={counter}
               {...bindActionCreators(CounterActions, dispatch)} />
    );
  }
}
```
when props are specified.

 I wanted to create this pull request to start a discussion and see if we feel that extra check is necessary. If it isn't, we can remove shallowEqualScalar completely.


gaearon:
IIRC it behaves differently because the `select` function passed to the `@connect` decorator also accepts the current `props` of the component. This is unnecessary for `<Connector>` because you can just read them from `this.props`.

taylorhakes:
@gaearon The `@connect` decorator accepts the current props of the component and redux is testing for shallowEqualScalar of the props (similar to adding the `PureRenderMixin` for them). In the case of `<Connector>`, the props of the child component aren't accessible to the `<Connector>` because it is only given a child function, not the actual props.

I would suggest that we don't do the `shallowEqualScalar` check in the `createConnectDocorator`. The user of redux can do it themselves (add PureRenderMixin or some other way) in the `DecoratedComponent` (the component with the `@connect` decorator attached).

Not only would the change make the `@connect` decorator and the `<Connector>` behave the same, it would make the library more flexible. I think that having a `shallowEqualScalar` check doesn't add much value. There will be very few people that add props to their decorated components (outside of the slice from the stores). If they do add additional props, we should give them full control of how the components updates with those props. We are already doing a shallowEqual check on the slice data for them in the `<Connector>`.

gaearon:
We're moving React wrappers to a separate project: https://github.com/gaearon/react-redux.
I'm closing this PR but I'll come back to it in https://github.com/gaearon/react-redux/issues/2.

gaearon:
>similar to adding the PureRenderMixin for them

To clarify, it is not similar. `shallowEqualScalar` doesn't break people's code if they do deep updates. `shallowEqual` does. `shallowEqualScalar` is more “relaxed” but still more performant than doing nothing at all.

We want good perf by default. Relay also uses `shallowEqualScalar` for its containers.

PR #209, opened 2015-07-03
Remove dispatcher state copy.

quazzie:
Why does the dispatcher have a copy of the state ?
I ran into a problem where the dispatcher state was out of sync (old) and my store was overwritten to a previous state.

I couldn't really figure out why, but only having state in redux solved the problem.


gaearon:
Thanks!
This is already fixed in https://github.com/gaearon/redux/pull/195 so I'm closing.

PR #220, opened 2015-07-06
Added another example project

dax0:
Browsed through listed examples and pulled together the best from each to create a base starting point


gaearon:
Hey, sorry for long time no reply!

I think this example is a bit confusing because names like `test`, `testAsync` don't really explain how to use in real project. We already have a simple built-in `counter` example; we'd rather link to more complex examples.

Thanks for the effort though! [Awesome Redux](https://github.com/xgrommx/awesome-redux) is a better place for examples now, because it has grouping and is much more extensive. In the future, we will only link to a few examples we personally reviewed.

PR #223, opened 2015-07-06
Correct path to react entries

Kureev:
- Changed `./lib/` to `./`


gaearon:
Entry point are copied to the root folder in prepublish script, so this is intentional.

johanneslumpe:
The paths are not incorrect. The files get moved during the build step: https://github.com/gaearon/redux/blob/master/scripts/build#L6

Kureev:
Ah, got it

PR #232, opened 2015-07-08
feat(examples): angular2-counter

gdi2290:
I need to replace TypeScript for Babel to keep the same build and remove all unneeded types. `CounterApp` has a lot more code because I didn't feel like recreating the `connect` decorator. I also didn't create Provider. 

Todo: 
 - [ ] Replace TypeScript for Babel with angular 2 plugins


gaearon:
Great job!
How would you pass state further down?

I definitely think this needs to be in a separate repo, like `redux-angular-counter-example`.

gdi2290:
`redux-angular2` might be better with the examples and abstractions like Angular's version of `Provider` and `Connector`. At the moment the logic for `Connector` is exposed in in `CounterApp`. To pass state down you do the same thing as react with props (in Angular2 you don't need the props namespace). If we have a data structure like this
```
var state = {
  "person": {
     "address": {
        "city": sf,
        "zipcode": 94102
      }
   }
}
```
the components would look like this
```html
<root></root>
  // inside <root>
  <app [state]="state"></app>
    //  inside <app>
    <person [person]="person"></person>
      // inside <person>
      <address [address]="address"></address>
        // inside <address>
        <span>{{ address.city }} {{ address.zipcode }}</span>
```
the `ON_PUSH` means that the component will only be updated if the prop references changed
```coffee
changeDetection: ON_PUSH
```
we don't have this on `CounterApp` because it's prop is `redux` and since redux doesn't change each time there is a change then it's ChangeDetection is default to always check (will be fixed with `Provider`).

gaearon:
Thanks for the effort! There's [ng-redux](https://github.com/wbuchwalter/ng-redux) now so I don't think we need it as a full-featured example inside Redux repo.

PR #236, opened 2015-07-09
Added test for shallowEqualScalar in connect decorator

taylorhakes:
Added an additional test to connect decorator to check for shallowEqualScalar logic and make sure props are passed through the connect decorator.


gaearon:
This looks good. Thanks!

PR #240, opened 2015-07-09
Improve code in examples

voter101:
In my opinion it's quite an anti-pattern of non-extending component classes with `React.Component`. It's an easy detail to miss when working on such code. I think we should extend those classes despite the fact React handles that case.

Another thing I propose to improve (unfortunately in single commit) is to not leave unused variables such as `React`, when we use destructing feature from ES6.

Have a nice day and thanks for the library! It's a great usage in everyday development. :)

### DIFF
--- a/examples/counter/components/Counter.js
+++ b/examples/counter/components/Counter.js
@@ -1,6 +1,6 @@
-import React, { PropTypes } from 'react';
+import { Component, PropTypes } from 'react';
 
-export default class Counter {
+export default class Counter extends Component {
   static propTypes = {
     increment: PropTypes.func.isRequired,
     incrementIfOdd: PropTypes.func.isRequired,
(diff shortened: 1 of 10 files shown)


acdlite:
I agree with extending `Component` (since it doesn't play well with TestUtils) but we need React to be in scope whenever we use JSX.

acdlite:
By the way, this would have been caught be the linter if you had run it :)

See the Travis build https://travis-ci.org/gaearon/redux/builds/70258318

voter101:
You are right, I wasn't aware of that, despite the fact I knew how the JSX de-sugar process looks like. :)

Allright. Got that fixed. I attached additional commits into my PR. I made double-check to see if nothing is broken.

Unfortunately, tests are not passing, but my changes didn't affect them. They were failing before my commits.

gaearon:
Thanks!

Issue #241, opened 2015-07-09
Cannot build on Windows


xgrommx:
When I cloned repo I cannot run npm install on Windows.

gaearon:
Yes, we should fix this. One way is https://github.com/gaearon/redux/pull/94. Another way is to go back to writing commands inline in `package.json`. I've been busy with other stuff so I'm not sure which option is better. What do you think?

xgrommx:
@gaearon Maybe this article can help you http://shapeshed.com/writing-cross-platform-node

xgrommx:
Also interesting this repo https://github.com/kimmobrunfeldt/concurrently and example how can be usable https://github.com/RickWong/react-isomorphic-starterkit/blob/master/package.json#L25

gaearon:
This will be fixed by #267. Let's track it instead.

gaearon:
#267 is merged into `breaking-changes-1.0` branch, feel free to give it a go.

urivasas:
this did not seem to fix the issue. removing prepublish from package.json seems to fix the problem. please consider re-opening. thanks.

gaearon:
@recfish Can you give details? This is a very vague error report. Without the exact error there's nothing I can do to fix it.

PR #248, opened 2015-07-11
Lintify before running tests

emmenko:
Particularly useful for travis and for linting a PR.

/cc @gaearon

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -1,3 +1,4 @@
-#!/bin/sh
+#!/bin/sh -e
 
+sh scripts/lint
 NODE_ENV=test `npm bin`/mocha --compilers js:babel/register --recursive
--- a/scripts/test-cov
+++ b/scripts/test-cov
@@ -1,3 +1,4 @@
-#!/bin/sh
+#!/bin/sh -e
 
+sh scripts/lint
 `npm bin`/istanbul cover `npm bin`/_mocha -- --compilers js:babel/register --recursive
--- a/scripts/test-watch
+++ b/scripts/test-watch
@@ -1,3 +1,4 @@
-#!/bin/sh
+#!/bin/sh -e
 
+sh scripts/lint
 NODE_ENV=test `npm bin`/mocha --compilers js:babel/register --recursive --watch


gaearon:
I think Travis already lints because it runs prepublish during install, doesn't it?

acdlite:
@gaearon Yes that's right

acdlite:
Here's an example of a PR that fails linting on Travis https://travis-ci.org/gaearon/redux/builds/70258318

emmenko:
Oh, right. Actually this kind of sucks about `npm`. 
I usually **run the tests** before publishing a package, which includes linting of course. So I'm sure I'm publishing a tested code. But using the `prepublish` script it runs also on `npm install`.

Personally I would rename `prepublish -> release`, and it has to be run manually with `npm run release` (and you can even setup automatic bumping of version, e.g.: https://www.npmjs.com/package/mversion).

gaearon:
What's the right course of action here? Any more opinions? I'm an npm noob.

emmenko:
Regardless of this PR, I think running tests + lint before publishing is important. But I don't think the `prepublish` task is appropriate in this case, since it executes on `npm install`.
I also personally find useful to have a task that automatically bumps the version, creates a tag and publish it.

But let's here what the others have to say :)

valscion:
I was very surprised to see `redux` running some sort of build task (the `prepublish` one) when I installed it. If it isn't necessary to run when installing, then I'd suggest using `npm run release` (or the like) for release build specific tasks.

gaearon:
@emmenko 

Can you update this PR to move build from `prepublish` to `release`?

And another thing: since you're changing build anyway, and it's simpler now, do you think you could fix #241? Basically this would mean moving the build back into NPM scripts like #9 attempted to do early on. I don't want to use shelljs (https://github.com/gaearon/redux/pull/94) as moving everything to the scripts seems simpler.

emmenko:
@gaearon sure, will do that after work

gaearon:
>Regardless of this PR, I think running tests + lint before publishing is important. But I don't think the prepublish task is appropriate in this case, since it executes on npm install.

I'm not sure this is entirely correct. It doesn't execute on *every* `npm install`.
See https://medium.com/@boennemann/what-is-npm-s-prepublish-and-why-is-it-so-confusing-a948373e6be1

emmenko:
You're right, thanks for the link! :+1:

emmenko:
Closed in favour of #267

Issue #230, opened 2015-07-07
Proposal: split redux-react into a separate package


gaearon:
I'm not a fan of over-splitting.

However we already export React-specific parts separately (`redux/react` and `redux/react-native`). React also embraces some splitting in React 0.14 (`react` and `react-dom`). It's becoming harder to buy the argument that people mind splitting so much.

I've heard complaints that people don't understand what is “core” to Redux and what is not because of `Provider` and `Connector` baked into Redux. I've also heard people want to use Redux with Ember, Angular 2 and Deku. There is a work in progress on an Angular 2 example project.

Things like `Provider` and `Connector` are React-specific and there also may be better ways to write them. It's harder to focus on developing them (and maybe [bringing in additional functionality](https://github.com/gaearon/redux/pull/86)) when they're right inside the core. At this point there's so little API that binds them together (basically `{ getState, dispatch, subscribe }` is their only common point) that I think splitting might bring more benefits than inertia and/or management overhead.

Here's my proposal:

* Move Redux to `reduxjs` org
* Create “blessed” `redux-react` and `redux-react-native`
* When React 0.14 is released, we'll be able to drop `redux-react-native` altogether

We'll also see if there is interest in “blessed” Angular/Ember bindings, and people willing to maintain them.

Thoughts?

gaearon:
One thing I want to be able to do is to make breaking changes to `redux-react` without affecting the core.

gaearon:
It's worth noting it's possible to do it the other way: move core to `redux-core` and have `redux` be a Swiss knife project. But I'm not sure it's right to give React preference given that people are starting to use it outside React (and indeed it's something I want to encourage).

acdlite:
:+1:

acdlite:
Perhaps it's just personal preference, but I'm more often annoyed by too much coupling versus too much splitting, and would prefer to err on the modular side.

nkt:
This not makes sense for me because it makes harder to maintain projects written using redux. At least we'll have to track versions for both packages, instead of one.
Maybe this will be the correct step if the redux will contains helpers for other frameworks.

acdlite:
@nkt It's only two packages :)

acdlite:
I've never understood that complaint about Babel. The version numbers for `babel-core` and `babel` are the same, and it takes like 2 seconds to update package.json.

jprichardson:
[...]

PR #249, opened 2015-07-11
Remove React-specific code in favor of gaearon/redux-react

gaearon:
Per #230, the plan for 1.0 is to split Redux into two (I promise, just two!) packages:

* Redux core (this package)
* React bindings (http://github.com/gaearon/redux-react, not on NPM yet)

We'll have different UMD builds for each. It's slightly more pain for the common case, but I think it's justified because React bindings API needs more experimentation than the core API, and we don't want to bump core versions too fast, as all ecosystem (middleware, devtools, etc) will depend on it.


gaearon:
If you use NPM though, the split will be as easy as depending on two packages instead of one, and replacing `redux/react` with `redux-react` in the source code.

gaearon:
Also.. should that be redux-react or react-redux? I think I'm in favor of react-redux now, that's more natural for React ecosystem.

acdlite:
Hey, but this is the Redux ecosystem :D

gaearon:
Haha. I think Babel does the right thing: https://github.com/babel
Depending on the ecosystem, it chooses the project naming to “fit in”.

acdlite:
I don't know if that's quite the same. Babel is a build tool, whereas Redux is a library. I prefer redux-react but I'm having trouble articulating why...

Is this to be a Redux library for React users, or a React library for Redux users?

emmenko:
@gaearon do you plan to have just those bindings in the `*-react-*` repo or something else as well? If only bindings, would it make sense maybe to have a more specific naming? 
E.g.: `redux-react-bindings`...

gaearon:
Global name needs to look non-awkward. `ReduxReactBindings` is awkward IMO. Both `ReduxReact` and `ReactRedux` look fine to me, but I like that `ReactRedux` (or `react-redux`) matches the dominant convention of React-related packages. Also some semi-official bindings use `react-` prefix (e.g. `react-art`).

erikras:
In English, we generally put the adjective before the noun, and I think that in this case, `react` is modifying (naming a specific binding to a framework) `redux`.

It's a tight match, but I think I'm voting `react-redux`.

acdlite:
If we're going to read it like an English phrase, then `react-redux` sounds like a new and improved version of React...

I think this a library for Redux users who also happen to use React, not React users who happen to use Redux. The difference between other projects like react-router is that React Router doesn't exist outside of the React ecosystem, whereas Redux could have redux-ember, redux-cycle, etc.

But I totally see the other argument. Meh. I don't care, I guess.

gaearon:
>I think this a library for Redux users who also happen to use React, not React users who happen to use Redux.

That's the project maintainer speaking in you. For now, I expect the majority of users is likely to feel they're React users trying Redux, not the opposite. :-)

gaearon:
We can always change later. For now I feel confident that's what'll cause least confusion.

erikras:
Good arguments, @acdlite.

:-) "Well, we've got all our state in redux, but we want to switch rendering engines." LOL.

emmenko:
> Global name needs to look non-awkward

It kinda feels awkward yes :)

> Meh. I don't care, I guess.

:+1:  

> We can always change later.

Agree.

acdlite:
> :-) "Well, we've got all our state in redux, but we want to switch rendering engines." LOL.

Not much sillier than the inverse :)

mattapperson:
Will the idea be to one day merge back into one?

erikras:
Sounds like not, @mattapperson. The point is to abstract out the state-managing bits that are UI-agnostic.

@acdlite, touché. :-)

gaearon:
@mattapperson 

I'd rather not. It's easier to evolve them separately, and in fact I'm sure somebody's going to come up with better API for the React part. This also signals that Redux is welcoming to Ember, Angular communities. We want everybody to have a great experience, not just our React friends.

erikras:
I look forward to being able to introduce redux into my PrototypeJS app that I still manage. :sob: :gun:

mattapperson:
@gaearon that makes sense. Thanks for the clarification!

HenrikJoreteg:
For whatever its worth, huge plus 1 from me for splitting it up. Loving the patterns in redux (and its small filesize), but would love to be able to try using it with other renderers too as the number of options keep growing: https://twitter.com/HenrikJoreteg/status/619543358151634944

paramaggarwal:
I vote for `redux-react`. As it's a extension of `redux` for `react`.

dariocravero:
:+1: for `redux-react`, after all it's specific React functionality to work together with redux, so it kind of feel more natural to me.

gaearon:
The opinions are split: https://twitter.com/dan_abramov/status/619976807995932673

PR #250, opened 2015-07-12
Dispatch sends action through entire middleware chain

jquense:
dispatch method passed to middleware was skipping middleware for subsequent.


gaearon:
:+1:

gaearon:
:+1:

Issue #206, opened 2015-07-02
Typescript definition


Keats:
Just an issue to remind me to start doing it as soon as https://github.com/gaearon/redux/pull/195 is merged.

Normally that would live in https://github.com/borisyankov/DefinitelyTyped but it doesn't support typescript 1.5 yet so I'll put it here if that's ok, it's also a good way to quickly see the public api of a library imo

gaearon:
Would #254 help here, or is “definition” a separate thing?

chicoxyzzy:
TS definition files are different from Flow definition files and AFAIK they aren't compatible with each other

Keats:
They are a bit different.
I got the following for now that I'm editing/changing as I'm using it

```typescript
declare module "redux" {
    interface ActionCreator {
        (...args: Array<any>): Object;
    }

    interface ActionCreators {
        [key: string]: ActionCreator
    }

    interface Reducer {
        (state: any, action: Object): any;
    }

    interface StoreMethods {
        dispatch(action: Object): Object;
        getState(): Object;
    }

    class Store {
        getReducer(): Function;
        replaceReducer(nextReducer: Reducer): void;
        dispatch(action: Object): Object;
        getState(): Object;
        subscribe(listener: Function): Function;
    }

    function createStore(
        reducer: Reducer | Object,
        initialState?: any
    ): Store;

    function bindActionCreators<T>(
        actionCreators: ActionCreator | ActionCreators,
        dispatch: Function
    ): T;

    function composeMiddleware(...middlewares: Array<Function>): Function;
    function composeReducers(reducers: Array<Reducer>): Reducer;

}

declare module "redux/react" {
    import React = require("react");
    import redux = require("redux");

    interface ProviderProps {
        store: redux.Store;
        children: Function;
    }

    interface ProviderState {
        store: redux.Store;
    }

    class Provider extends React.Component<ProviderProps, ProviderState> {}

    interface ConnectorProps {
        children: Function;
        select: Function;
    }

    class Connector extends React.Component<ConnectorProps, any> {}
}
```
It's missing some stuff stilll but it's getting there

Keats:
Updated a bit (and will likely wait for redux/react-redux to hit 1.0):

```ts
declare module "redux" {
    interface ActionCreator {
        (...args: Array<any>): Object;
    }

    interface ActionCreators {
        [key: string]: ActionCreator
    }

    interface Reducer {
        (state: any, action: Object): any;
    }

    interface StoreMethods {
[...]

PR #254, opened 2015-07-12
Flow types, take 3 (actually 4)

acdlite:
It works!

(Did I git it right this time?)


src/types.js
```diff
@@ -0,0 +1,10 @@
+export type State = any;
+export type Action = Object;
+export type IntermediateAction = any;
+export type Dispatch = (a: Action | IntermediateAction) => any;
+export type Reducer<S, A> = (state: S, action: A) => S;
+export type ActionCreator = (...args: any) => Action | IntermediateAction
+export type Middleware = (methods: { dispatch: Dispatch, getState: () => State }) => (next: Dispatch) => Dispatch;
+export type Store = { dispatch: Dispatch, getState: State, subscribe: Function, getReducer: Reducer, replaceReducer: void };
+export type CreateStore = (reducer: Function, initialState: any) => Store;
+export type HigherOrderStore = (next: CreateStore) => CreateStore;
```
gaearon: Woah. That's pretty awesome.

src/types.js
```diff
@@ -0,0 +1,10 @@
+export type State = any;
+export type Action = Object;
+export type IntermediateAction = any;
+export type Dispatch = (a: Action | IntermediateAction) => any;
+export type Reducer<S, A> = (state: S, action: A) => S;
+export type ActionCreator = (...args: any) => Action | IntermediateAction
+export type Middleware = (methods: { dispatch: Dispatch, getState: () => State }) => (next: Dispatch) => Dispatch;
+export type Store = { dispatch: Dispatch, getState: State, subscribe: Function, getReducer: Reducer, replaceReducer: void };
```
gaearon: Can/should we clarify subscribe type? The fact that it invokes the listener with no arguments and returns a void function.

src/types.js
acdlite: Oops obviously the type of reducer should be `Reducer` :)

src/types.js
acdlite: Yeah we should

acdlite:
Oh shit just realized I forgot to add `/* @flow */` to the top of the types file. :D Surfaced a few errors. I'll fix and update.

gaearon:
Can we integrate this into the build pipeline? Check Flow before running tests?

acdlite:
I believe we can with https://github.com/sindresorhus/flow-bin. That's what graph-ql uses https://github.com/graphql/graphql-js

gaearon:
Can you please bring this up to date with the last changes I made to breaking changes branch?

gaearon:
Also, should we take this in now? After 1.0?

gaearon:
Also, do we want https://github.com/babel-plugins/babel-plugin-flow-comments for the non-minified builds?

chicoxyzzy:
Don't sure `flow-comments` plugin is necessary. It's not very readable and flow is useful for compile time not after.

chicoxyzzy:
http://flowtype.org/blog/2015/02/20/Flow-Comments.html
Flow comments are alternative to standard Flow annotations actually

gaearon:
>It's not very readable and flow is useful for compile time not after.

Still might be useful for people browsing the compiled source (e.g. when debugging some issue).
*Writing* them as comments gets cumbersome, but I'd welcome comments in the compiled NPM output.

chicoxyzzy:
Are you talking about devs of Redux (who should actually use uncompiled sources) or devs who use Redux? IMO not so much JS devs are familiar with Flow so it's more code pollution then code documentation.

gaearon:
I'm talking about users. I think our core audience is at least passingly familiar with Flow. Also, I'd say even a person unfamiliar with Flow wouldn't have a problem reading most Flow annotations. 

Anyway I don't have a strong opinion here. Let's see what others have to say.

taylorhakes:
I really like the type annotation comments in the code. It would help me find a type bug earlier. 

The only downsides are file size and confusion. I don't imagine too many people confused by the comments (they are just comments) and I think it adds a lot of value.

gaearon:
File size isn't a problem because comments are stripped by any modification tool.

chicoxyzzy:
>I think our core audience is at least passingly familiar with Flow.

It would be much better to have types in jsDoc for non-minified build rather then not working actual alternative Flow type annotation. BTW personally I don't care much about it.

>It would help me find a type bug earlier.

I don't think so. You will find a bug for same time as without these comments cause they have nothing to do with types and they can't say "Hey! I want String not Integer!". All they can say is "Hey! Hey look at me! I'm String! Really!" when you scrolling your code. And __all of them__ do it. Do you really think it can help?

chicoxyzzy:
Also Flow comments are created as a temporary solution which should help in evaluating migration to standard Flow type annotation not for code documentation. jsDoc can help by static analysis, autocompletion and parameters hinting. Flow comments can't do any of that

chicoxyzzy:
But maybe I'm wrong... For example can I use standard Flow annotations with Redux's non-minified version Flow comments? Does someone has any experience with such workflow?

chicoxyzzy:
> But maybe I'm wrong... For example can I use standard Flow annotations in my app together with Redux's non-minified version Flow comments? Does someone has any experience with such workflow?

This is the answer http://flowtype.org/docs/third-party.html

chicoxyzzy:
So I suggest supporting Flow's interface file, TypeScript's tsd file and cut of all Flow annotations from any builds

taylorhakes:
> File size isn't a problem because comments are stripped by any minification tool.

@gaearon Yes. Minified is the same if comments exist or not. I thought we talking about unminified? That's where the difference lies.

@chicoxyzzy Unfortunately, currently, the flow type annotations are stripped before publishing to NPM. You can't access the raw flow files via NPM. A flow interface file like you linked or correct comments would work to remedy a full flow app.

If you aren't using flow, the type information inline (via flow comments, correct if possible, or jsDoc) is useful when debugging library code. I can inspect the `'hello'` value and see `/*: number*/` next to it.

I am definitely for a correct flow workflow. I am just saying, unless someone is working on creating jsDoc processors or correct flow comments, I think there is value in the flow comments with very little downside.

hzoo:
If you see any issues with the babel plugin's comment output - def make an issue (if it's not the one already there) in the repo or ping me

gaearon:
I updated the pull request to match the 1.0 API.
This is ready to merge, but I wonder if there is anything we can (or should) expose to Flow users..

(e.g. #206?)

gaearon:
In particular I'm interested what we need to change to enjoy this kind of static analysis: https://github.com/facebook/flux/pull/243

Do we need to change more signatures to be generic to allow this? Should it be `createStore<MyState, MyAction>()`? How do we expose these definitions?

gaearon:
I've merged this in the current state, but I opened #290 to track improvements to our definitions to allow facebook/flux#243-like static analysis.

gaearon:
>Also Flow comments are created as a temporary solution which should help in evaluating migration to standard Flow type annotation not for code documentation. 

[This](http://flowtype.org/blog/2015/02/20/Flow-Comments.html) says otherwise: 

>We plan to update our Flow transformation to wrap Flow syntax with these special comments, rather than stripping it away completely. This will help people write Flow code but publish code that works with or without Flow.

It's not *just* a temporary measure as I read it; it's a solution they plan to use for publishing Flow code on NPM. You don't need separate interface definitions if the comments “just work”.

I'm going to investigate the Babel plugin. If it helps Flow users consume (and interpret!) Redux code with annotations, isn't that awesome?

volkanunsal:
For what it's worth, in the past, I've found flow-comments useful as a supplement to Flow types because there will be cases when Flow types are in conflict with the ES6 or ES7 notation, e.g. argument destructuring.

gaearon:
One thing I'm not sure of is how to handle the fact that our types are in a separate `types` file.
@hzoo Can the plugin handle `import type`... somehow?

hzoo:
@gaearon I see you already got a PR for `import type` :smile:

PR #255, opened 2015-07-12
First attempt at updating docs for Redux Store

ellbee:
Am I along the right lines here?


acdlite:
Great! Some of it is a little out of date (I think we're removing the ability to pass an object of reducers to `createStore()`) but it's much better than what we have now :)

Thanks!

ellbee:
Haha, it's hard to keep up!

PR #256, opened 2015-07-12
Remove thunk middleware from the core

gaearon:
We're moving it into a separate project because:

* It's strange to export it as a top level export here.
* Asking people to reach into the internals is not an option either.
* Having it as a default middleware when you *still* have to call `applyMiddleware` manually doesn't make much sense.

### DIFF
--- a/examples/counter/containers/App.js
+++ b/examples/counter/containers/App.js
@@ -4,7 +4,15 @@ import { createStore, applyMiddleware } from 'redux';
 import { Provider } from 'react-redux';
 import * as reducers from '../reducers';
 
-const createStoreWithMiddleware = applyMiddleware()(createStore);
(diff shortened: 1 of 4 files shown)


examples/counter/containers/App.js
```diff
@@ -4,7 +4,15 @@ import { createStore, applyMiddleware } from 'redux';
 import { Provider } from 'react-redux';
 import * as reducers from '../reducers';
 
-const createStoreWithMiddleware = applyMiddleware()(createStore);
+// TODO: move into a separate project
```
gaearon: @acdlite Would you like to create a project for it?

acdlite: Yes I will, when I get home in a bit

PR #258, opened 2015-07-13
Update store.md to use combineReducers explicit call

ellbee:
As per commit 650f002 Remove combineReducers shortcut in favor of an explicit call

### DIFF
--- a/docs/store.md
+++ b/docs/store.md
@@ -9,18 +9,19 @@ The store has the following responsibilities:
 
 ####Initialization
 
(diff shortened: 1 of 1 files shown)


gaearon:
I'd break into two subsections.

First subsection shows a basic example with two reducers and one master reducer that combines their results a la https://gist.github.com/gaearon/d77ca812015c0356654f.

Second subsection goes like “we already have a helper for that called `combineReducers`”.

ellbee:
Yes, I like that approach. I'll update with changes.

gaearon:
Thank you!

ellbee:
Re-organized along the lines suggested by @gaearon.

gaearon:
Good job, thank you!

PR #260, opened 2015-07-13
Reference react-redux within react.md

mcamac:
Switches imports in react example to reflect separating out the packages


gaearon:
:+1:

PR #262, opened 2015-07-14
Compose applyMiddleware dispatch once.

tappleby:
`compose(middleware(methods), store.dispatch)` was happening on each dispatch, this pr updates `applyMiddleware` to only compose the dispatch method once.


gaearon:
@acdlite Can you take a look?

gaearon:
This looks great to me but just to make sure.

gaearon:
Thanks again. Pretty sure this is good to go.

PR #266, opened 2015-07-14
fix a typo in getting-started.md

chentsulin:


acdlite:
Oh boy, this isn't actually a typo. It's an issue of style. Most American style guides dictate that commas should also go inside of quotation marks, even for titles and scare quotes. I believe the British style is more traditional. My programmer's mind has always preferred putting it on the outside, but in school and in formal writing situations — at least in America — the "correct" place is inside the quotation mark.

Thoughts, @gaearon?

acdlite:
More info http://www.grammarly.com/answers/questions/4726-quotation-marks-and-other-punctuation-inside-or-outside/

emmenko:
@acdlite interesting, didn't know that :)

gaearon:
Pedants!
I'm fine with comma inside quote if that's what Americans do. ;-)

acdlite:
Okay let's just close this then so we can be done talking about it :) Thanks for the PR, though, @chentsulin!

chentsulin:
Good to know this. : ) Thanks for explanation.

PR #267, opened 2015-07-14
npm tasks

emmenko:
This supersedes #248 

/cc @gaearon


gaearon:
How does making a Github release fit into this workflow? Should I just do a release via web interface separately based on the pushed tag?

emmenko:
When you do a release, a github tag is created. Then you need to manually create the release note for that existing tag (but that's always a manual process).

Or do you mean something else?

gaearon:
Yeah this seems right.
Some tools also automate creating release notes so I just wanted to clarify.

This looks great to me.

.mversionrc
```diff
@@ -0,0 +1,9 @@
+{
+  "commitMessage": "Bumps to version v%s",
```
gaearon: Can we just change it to `%s` please

package.json
```diff
@@ -4,14 +4,16 @@
   "description": "Atomic Flux with hot reloading",
   "main": "lib/index.js",
   "scripts": {
-    "browser": "scripts/browser",
```
gaearon: I think `browser` got lost. We still want to have browser builds.

package.json
```diff
@@ -4,14 +4,16 @@
...
-    "build": "scripts/build",
-    "clean": "scripts/clean",
-    "lint": "scripts/lint",
-    "prepublish": "scripts/prepublish",
```
gaearon: I'm still not convinced we need to remove `prepublish`.
I want installing from Git to “just work”. That's what `prepublish` does if I'm not mistaken.

emmenko:
Cool :+1: 

One thing though to remember: the patch/minor/major tasks work for normal semver numbers, so normal lifecycle releases. 
If you want to release alpha/beta/-rc/whatever you should directly use the `mversion` cli (see [examples](https://www.npmjs.com/package/mversion#examples)), or do it manually, just to be sure.

gaearon:
Oh, okay, thanks for explaining.

package.json
emmenko: Sure. Should it be `prepublish: npm run build` ?

package.json
emmenko: Oops, my bad sorry.

.mversionrc
emmenko: Sure

package.json
gaearon: Yes I think so. Could also skip linting. I think it's better to have granular “low-level” commands and make some higher level commands on top of them.

package.json
emmenko: Should I call it `browser` or `dist`?

package.json
emmenko: :+1:

package.json
gaearon: `umd` is probably even better

gaearon: or `standalone`. Whatever you like more

emmenko: `umd` it is ;)

emmenko:
@gaearon ok, I made the changes and force pushed. Should be ok now.

.mversionrc
```diff
@@ -0,0 +1,9 @@
+{
+  "commitMessage": "Bumps to version %s",
```
gaearon: just %s please

emmenko: Oh, I thought `vs%` without `v`. Will change it, sorry

.mversionrc
```diff
@@ -0,0 +1,9 @@
...
+  "tagName": "v%s",
+  "scripts": {
+    "preupdate": "npm test && npm run umd",
```
gaearon: Why not build it as part of `npm run build`?

package.json
```diff
@@ -4,14 +4,18 @@
...
+    "lint": "eslint src test examples",
+    "prepublish": "npm run build",
+    "release": "mversion patch -m",
+    "release:patch": "mversion patch -m",
+    "release:minor": "mversion minor -m",
+    "release:major": "mversion major -m",
+    "test": "npm run lint && NODE_ENV=test mocha --compilers js:babel/register --recursive",
+    "test:watch": "npm test -- --watch",
+    "test:cov": "npm run lint && babel-node ./node_modules/.bin/isparta cover ./node_modules/.bin/_mocha -- --recursive",
+    "umd": "npm run clean && npm run lint && webpack src/index.js dist/redux.js && NODE_ENV=production webpack src/index.js dist/redux.min.js"
```
gaearon: I don't like that we do `clean` and `lint` in most commands. Isn't it better to have a few low-level “focused” commands that do just one thing (e.g. call `webpack`) and then assemble higher-level commands from them?

For example right now linting and cleaning will happen twice if I build both lib & UMD.

.mversionrc
emmenko: The `umd` part? Makes sense!

package.json
gaearon: Actually let's call it `build:umd`.

package.json
emmenko: I've simplified now. Build will clean and test will lint.

emmenko:
@gaearon feel free to do the final tweaks the way you like it ;)

gaearon:
Sure, thanks! I'll actually use some of these for a boilerplate I'm preparing. (Tired of doing the same setup for every lib.)

package.json
emmenko: Done

emmenko:
> Tired of doing the same setup for every lib

I know! :)

PS: Glad to help.

gaearon:
Thanks again! I'll merge tomorrow.
In the meantime, I extracted some stuff into https://github.com/gaearon/library-boilerplate

emmenko:
Good idea! Thanks for the link.

valscion:
Wow this PR clarifies the build process a lot and makes it easier to start hacking on redux. :+1:

mikekidder:
(Window compatibility)
npm tasks that set environment variables for Node, it requires a slight change. 
For example (from/to):

    "build:umd": "webpack src/index.js dist/redux.js && NODE_ENV=production webpack src/index.js dist/redux.min.js"

    "build:umd": "webpack src/index.js dist/redux.js && set NODE_ENV=production&& webpack src/index.js dist/redux.min.js"

I mentioned it in https://github.com/gaearon/library-boilerplate/issues/2.  Should work for Mac as well (I hope) to make it all cross-platform.  The no space after setting variable is the hack discovery that makes it possible in Windows.

gaearon:
@mikekidder This doesn't pass `NODE_ENV=production` to Webpack for some reason.

gaearon:
I tried to fix Windows build in #379.

mikekidder:
Yes, "set" on OS X was questionable, I didn't have a Mac to test.  I'll respond further in #379

PR #276, opened 2015-07-15
[WIP] Warn if initial state shape doesn't match reducer name

emmenko:
Refs #272 

- [x] add test to reproduce problem
- [ ] add warning and fix test


test/createStore.spec.js
```diff
@@ -45,6 +46,35 @@ describe('createStore', () => {
     }]);
   });
 
+  it('should warn if passed initial state has an unknown shape', () => {
+    const bootstrapData = {
+      pool: 1000,
```
gaearon: I think `counter` and `todos` would be more inline with the tests we have so far :-)

emmenko:
Done in #302

PR #286, opened 2015-07-17
Update package.json

aesopwolf:


aesopwolf:
These are the packages I needed installed locally for all of the linting to work correctly without displaying errors (within atom and using atomlinter).

aesopwolf:
It seems like my issue was actually related to https://github.com/AtomLinter/linter-eslint/issues/128

But my opinion would be to keep these packages in dev dependencies (in the example projects) to keep things literally as simple as:

```
git clone https://github.com/gaearon/redux.git
cd redux/examples/counter
npm install
npm start
```

... without running into the same issues I did

gaearon:
They are in `package.json` for the parent project. The linter configuration is also there so it feels wrong to add them as dependencies without also copying `.eslintrc`.

That said, I don't want to do that. The examples are *not* standalone—in fact `redux` is aliased to `../../src` in Webpack config for the convenience of tweaking the code. There doesn't seem to be a point in making them _seem_ more standalone when they're just not.

aesopwolf:
Gotcha!

PR #288, opened 2015-07-18
Make devtools: source-map for examples

sairion:
While looking into how redux working in examples, I found it gets compiled as `eval` mode. I think enabling `source-map` is more useful for newcomers like me! :)


gaearon:
Thanks. I changed it to a slightly faster mode: 1a602a7.

PR #289, opened 2015-07-19
Rewrite the docs

gaearon:
In this branch I'll be rewriting the docs from scratch against the 1.0 branch.

I'm going to gradually pick JSDoc annotations and the new docs from #140 but I wanted to start with a clean slate because having a mix of the old and the new docs in the same place puts me into a creative block.

I'm going to start with README (I already have a few paragraphs there). We won't be able to make a static website ahead of the release, so I'll focus on the README + a `docs` folder with the docs in Markdown.


gaearon:
New README is up: https://github.com/gaearon/redux/tree/rewrite-docs

gaearon:
I drafted the API reference: https://github.com/gaearon/redux/blob/rewrite-docs/docs/Reference/API.md
Comments are as always appreciated.

ellbee:
These look really good so far. Couple of minor things:

* I think I like the term "state tree" used in the API docs instead of state, although having said that I suppose the state is not necessarily a tree. If the term "state tree" is going to be used, should it be used more consistently across the docs? In the API docs "state tree" is used frequently down to the subscribe section, but not after that. In the glossary the term "state tree" is used only once.
* Would the API docs benefit from having an overview section at the top with hyperlinked function signatures that would take you to the corresponding detailed section?

gaearon:
>Would the API docs benefit from having an overview section at the top with hyperlinked function signatures that would take you to the corresponding detailed section?

Yes, and a PR is welcome!

ellbee:
@gaearon Ok, coming up!

gaearon:
This is superseded by #349.

PR #292, opened 2015-07-20
use function bind operator for hasOwn

edge:
In our case, `a.call(b, ...args)` is the same as `b::a(...args)`


gaearon:
Hey, thanks for PR. :-)

These functions are actually no longer in Redux because that React connectors that depend on them are now in http://github.com/gaearon/react-redux. Also, we're not using `::` syntax in core code anymore because we started using Flow (#254) which can't parse them yet.

PR #295, opened 2015-07-22
Allow es6 symbols to be used as action types

jhewlett:
Symbols are a good way to prevent naming collisions of action types.

However, if you attempt to use a symbol as an action type, `getErrorMessage` inside `combineReducers` will throw an error on line 13 because symbols can't be implicitly converted to a string. I can reproduce this in at least Chrome and Firefox.

Strangely enough, I could not reproduce this in the tests -- it looks like babel is implicitly converting the symbol to a string, contrary to the es6 spec as I understand it. I added a test anyway to verify the use case.

I understand Redux probably wasn't designed to be used with symbols as action types, but this seems like an easy way to add support for them.

I'm also open to other solutions or workarounds.

### DIFF
--- a/src/utils/combineReducers.js
+++ b/src/utils/combineReducers.js
@@ -11,7 +11,7 @@ import { ActionTypes } from '../createStore';
 function getErrorMessage(key: String, action: Action): string {
   var actionType = action && action.type;
   var actionName = actionType && `"${actionType.toString()}"` || 'an action';
-  
+
   return (
     `Reducer "${key}" returned undefined handling ${actionName}. ` +
     `To ignore an action, you must explicitly return the previous state.`
--- a/src/utils/combineReducers.js
+++ b/src/utils/combineReducers.js
@@ -10,8 +10,8 @@ import { ActionTypes } from '../createStore';
 
 function getErrorMessage(key: String, action: Action): string {
   var actionType = action && action.type;
-  var actionName = actionType && `"${actionType}"` || 'an action';
-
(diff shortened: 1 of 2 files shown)


gaearon:
I'm not sure we want to allow this but you're right that, even if we throw, we shouldn't throw inside the helper method..

The reason I would rather enforce string action type is because you can't serialize Symbols. The whole bunch of devtools and plugins would be unavailable to you. It's so easy to make this mistake without realizing the consequences that I'd rather enforce strings.

Thoughts?

gaearon:
:+1: 

While we won't encourage people to use Symbols (and will even actively discourage doing so) I don't think we should be opinionated here, and definitely we shouldn't crash inside the warning method.

jhewlett:
Yeah, that makes perfect sense.

I wonder if it's even worth keeping the test. Like I said, the test never failed before I made the change, presumably because babel has different behavior than the browsers.

goatslacker:
You can serialize Symbols by calling toString() on them. You will need to use Symbol.keyFor though if you want actions to be replayable.

jhewlett:
@goatslacker The only problem with `toString` is that `Symbol('test').toString()` and `Symbol('test').toString() both serialize to `Symbol(test)` (at least in Chrome and FF). How would you distinguish between the two? Furthermore, I don't see how you would deserialize back into a symbol at that point.

goatslacker:
That's why I mentioned you have to use Symbol.keyFor.

KSuttle-cng:
So, are there docs for this? The comments seem to indicate it didn't get pulled in, but the PR did get merged.

PR #296, opened 2015-07-22
remove coverage, docs, test directories in npm package

chentsulin:


gaearon:
Good catch. We're working on top of https://github.com/gaearon/redux/pull/195 right now so I excluded `coverage` there. I want to keep the `docs` though.

Issue #290, opened 2015-07-19
Explore using Flow for statically analyzing state and actions


gaearon:
See the example from [this PR to Facebook's Flux](https://github.com/facebook/flux/pull/243):

``` js
type Action =
  {
    type: 'foo',
    foo: number,
  } |
  {
    type: 'bar',
    bar: string,
  };

var FooDispatcher: Dispatcher<Action> = new Dispatcher();

function acceptsNumber(x: number): void {}

FooDispatcher.register(function(action: Action) {
  switch (action.type) {
    case 'foo':
      acceptsNumber(action.foo);
      acceptsNumber(action.bar); // flow error, property not found in object
      break;

    case 'bar':
      acceptsNumber(action.bar); // flow error, string incompatible with number
      break;
  }
});

type NotAction = string;
FooDispatcher.register(function(action: NotAction) {}); // flow error

// flow error
FooDispatcher.dispatch({
  type: 'foo-typo',
  foo: 100,
});
```

Now that #254 is merged (into 1.0 branch), I encourage you to help me figure out how to have the same awesomeness in Redux. We probably need to make `createStore` definition generic so that you can say `createStore<MyState, MyAction>(reducer, initialState)` and have your whole reducer tree type checked.

I'm a Flow noob so any help is welcome. You can start by copying the Counter example, annotating it with Flow and trying to make it type check like in https://github.com/facebook/flux/pull/243.

adelnizamutdinov:
I don't know if there's a better way to represent algebraic data types in flow's type system
https://gist.github.com/hallettj/281e4080c7f4eaf98317

adelnizamutdinov:
oh, apparently there's a more idiomatic way of handling ADTs http://flowtype.org/blog/2015/07/03/Disjoint-Unions.html

gaearon:
I managed to use the Flow definitions from the Redux source code by using the following Flow config in my standalone example:

```
[ignore]
.*/node_modules/webpack/.*
.*/node_modules/webpack-dev-server/.*
.*/node_modules/redux/examples/.*
.*/node_modules/redux/lib/.*

[include]

[libs]

[options]
module.system=node
module.name_mapper='\(redux\)' -> '\1/src/index'
```

Note that this requires `src` to be in the NPM distribution (we'll be including it then).

It's pretty cool already! I'll play with it and see if any common Redux mistakes can be caught this way, and if I can tweak the type definitions to check the action parameter types.

leoasis:
[...]

PR #299, opened 2015-07-22
Remove Flow types for now

gaearon:
I decided in favor of *not* shipping Flow types in 1.0.
This commit just strips types from the current Flow-friendly codebase (initially worked on in #254).

We will not use features unsupported by Flow (such as `let` or `const`) from now on so it's easy to gradually add Flow types later.

The reason I'm removing them is **they're too naive** and I don't want them to be perceived as the public API (at least not yet). I am ready to add Flow when [cases like this](https://github.com/gaearon/redux/issues/290) are solved. Adding them now is confusing because, for #290 to work, different (more polymorphic) signatures are needed. For example, `Store` type should be parameterized by `<MyState, MyAction>` instead of `any` types in the current annotations.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -1,8 +1,3 @@
-/* @flow */
-/*eslint-disable */
-import type { State, Reducer, Action, IntermediateAction, Store } from './index';
-/*eslint-enable */
-
 import invariant from 'invariant';
 import isPlainObject from './utils/isPlainObject';
 
@@ -13,10 +8,7 @@ export var ActionTypes = {
   INIT: '@@redux/INIT'
 };
 
-export default function createStore(
-  reducer: Reducer,
-  initialState: State
-): Store {
+export default function createStore(reducer, initialState) {
   invariant(
     typeof reducer === 'function',
     'Expected the reducer to be a function.'
@@ -30,7 +22,7 @@ export default function createStore(
     return currentState;
   }
 
-  function subscribe(listener: Function) {
+  function subscribe(listener) {
     listeners.push(listener);
 
     return function unsubscribe() {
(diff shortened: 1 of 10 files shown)


gaearon:
cc @acdlite

yanivtal:
Aww I was looking forward to this. Btw let/const should be in soon https://github.com/facebook/flow/pull/431

gaearon:
Haha. Help make it happen with the correct signatures :-)
I got stuck to be honest, but I really want to get #290 to work!

leoasis:
@gaearon do you have any branch you could publish with your progress there? If you have done any significant progress it would be nice to start from there

gaearon:
@leoasis 

I had one but it descended into pointlessness so it would be easier for you to start fresh.
Some advice:

* Concentrate on `createStore`, don't add annotations anywhere else, this will only confuse you.
* Make it return `Store<State, Action>` where `State` and `Action` types are user specified.
* You'll have a problem with the dispatch of initial action because it won't match the user's `Action` type. Good luck there. :-P Maybe it's easiest to change signature to `createStore<State, Action>(reducer: Reducer<State, Action>, initialState: State, initialAction: Action)` just to get it to type check for now, and then figure out something that makes sense..

gaearon:
The goal is to get #290 working at least in the minimal way of type checking action/state type validity with `createStore`. Then figure out how to get the rest of the API working. Middleware and higher order stores might get super tricky (if not impossible) because `applyMiddleware(...)` changes `dispatch` type with each middleware. Don't go there until you figure out the basics..

leoasis:
Cool, will give it a good go tomorrow, thanks for the advice

Issue #89, opened 2015-06-14
Provide JSDoc


emmenko:
Just for keeping track of this.

gaearon:
I'm on board but let's wait for some conclusion on #87.
Should we annotate types both in Flow and in JSDoc? What's the usual approach here?

emmenko:
I haven't used Flow yet, so I can't really say what's the best approach. I think it also depends if you just want JSDoc in the source code or use it to actually generate a documentation website or something like it.

gaearon:
This was done via #300.

PR #300, opened 2015-07-22
Add JSDoc annotations to the public API

gaearon:
This adds JSDoc annotations to the public API as discussed in #140 and #89.
Comments are welcome!

(This is written against the 1.0 API which is not officially released yet.)

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -1,13 +1,34 @@
 import invariant from 'invariant';
 import isPlainObject from './utils/isPlainObject';
 
-// Don't ever try to handle these action types in your code. They are private.
-// For any unknown actions, you must return the current state.
-// If the current state is undefined, you must return the initial state.
+/**
+ * These are private action types reserved by Redux.
+ * For any unknown actions, you must return the current state.
+ * If the current state is undefined, you must return the initial state.
+ * Do not reference these action types directly in your code.
+ */
 export var ActionTypes = {
   INIT: '@@redux/INIT'
 };
 
+/**
+ * Creates a Redux store that holds the state tree.
+ * The only way to change the data in the store is to call `dispatch()` on it.
+ *
+ * There should only be a single store in your app. To specify how different
+ * parts of the state tree respond to actions, you may combine several reducers
+ * into a single reducer function by using `combineReducers`.
+ *
+ * @param {Function} reducer A function that returns the next state tree, given
+ * the current state tree and the action to handle.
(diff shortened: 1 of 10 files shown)


src/utils/applyMiddleware.js
```diff
@@ -1,11 +1,17 @@
 import compose from './compose';
 
 /**
- * Creates a higher-order store that applies middleware to a store's dispatch.
+ * Creates a higher-order store that applies middleware to the dispatch method
+ * of the Redux store. This is handy for a variety of tasks, such as expressing
+ * asynchronous actions in a concise manner, or logging every action payload.
+ *
+ * See `redux-thunk` package as an example of the Redux middleware.
```
nbostrom: You could use @link here
http://usejsdoc.org/tags-inline-link.html

gaearon:
I'll get this in, but comments are still very much appreciated.

The hardest part to understand is probably `applyMiddleware` but some terminology docs from #140 will help there. Other than that, please let me know if we can rephrase some parts of it for clarity!

src/utils/applyMiddleware.js
gaearon: Good point. For now I'd avoid hardcoding URL because we might move that package to a separate Github organization soon.

src/createStore.js
```diff
@@ -18,10 +39,23 @@ export default function createStore(reducer, initialState) {
...
   function getState() {
     return currentState;
   }
 
+  /**
+   * Adds a change listener. It will be called any time an action is dispatched,
+   * and some part of the state tree may potentially have changed. You may then
+   * call `getState()` to read the current state tree inside the callback.
+   *
+   * @param {Function} listener A callback to be invoked on every dispatch.
```
nbostrom: A good way to document callbacks is to use the @callback tag. This way you can document the @params of the callback function.

gaearon: Thanks, I didn't know this! Want to send a PR?

src/utils/applyMiddleware.js
nbostrom: That's probably wise

src/createStore.js
nbostrom: Sure! I don't really have time tonight though. Guess I can do it after the merge?

gaearon: Yeah feel free to make a PR against `breaking-changes-1.0` branch.

PR #302, opened 2015-07-22
Warn if the initial state shape doesn't match the reducer names

ellbee:
Ok, not super happy with this but it is a start.

* I have written tests for verifyStateShape, but I couldn't think of a neat way to test the code in createStore so there is currently a hole in the test coverage.
* I don't really like the function name verifyStateShape. Is stateShapeWarning better?
* I haven't used an external warning module. Would it be a good idea? I'm currently just calling console.warn.

Happy to take some direction on this.

### DIFF
--- a/src/utils/combineReducers.js
+++ b/src/utils/combineReducers.js
@@ -1,7 +1,9 @@
+import { ActionTypes } from '../createStore';
+import isPlainObject from '../utils/isPlainObject';
 import mapValues from '../utils/mapValues';
 import pick from '../utils/pick';
 import invariant from 'invariant';
-import { ActionTypes } from '../createStore';
+import warning from 'warning';
 
 function getErrorMessage(key, action) {
   var actionType = action && action.type;
@@ -10,6 +12,41 @@ function getErrorMessage(key, action) {
   return (
     `Reducer "${key}" returned undefined handling ${actionName}. ` +
     `To ignore an action, you must explicitly return the previous state.`
+  );
+}
+
+function verifyStateShape(initialState, currentState) {
+  var reducerKeys = Object.keys(currentState);
+
+  if (reducerKeys.length === 0) {
+    warning(
+      false,
+      'Store does not have a valid reducer. Make sure the argument passed ' +
+      'to combineReducers is an object whose values are reducers.'
+    );
+    return;
+  }
+
+  if (!isPlainObject(initialState)) {
+    warning(
+      false,
+      'initialState has unexpected type of "' +
+      ({}).toString.call(initialState).match(/\s([a-z|A-Z]+)/)[1] +
+      '". Expected initialState to be an object with the following ' +
+      `keys: "${reducerKeys.join('", "')}"`
+    );
+    return;
+  }
+
+  var unexpectedKeys = Object.keys(initialState).filter(
+    key => reducerKeys.indexOf(key) < 0
+  );
+
+  warning(
+    unexpectedKeys.length === 0,
+    `Unexpected ${unexpectedKeys.length > 1 ? 'keys' : 'key'} ` +
+    `"${unexpectedKeys.join('", "')}" in initialState will be ignored. ` +
+    `Expected to find one of the known reducer keys instead: "${reducerKeys.join('", "')}"`
   );
 }
 
@@ -29,6 +66,7 @@ function getErrorMessage(key, action) {
  * @returns {Function} A reducer function that invokes every reducer inside the
  * passed object, and builds a state object with the same shape.
  */
+
 export default function combineReducers(reducers) {
   var finalReducers = pick(reducers, (val) => typeof val === 'function');
 
@@ -54,8 +92,11 @@ export default function combineReducers(reducers) {
     );
   });
 
-  return function combination(state = {}, action) {
-    return mapValues(finalReducers, (reducer, key) => {
+  var defaultState = mapValues(finalReducers, () => undefined);
+  var stateShapeVerified;
+
+  return function combination(state = defaultState, action) {
+    var finalState = mapValues(finalReducers, (reducer, key) => {
       var newState = reducer(state[key], action);
       invariant(
         typeof newState !== 'undefined',
@@ -63,5 +104,12 @@ export default function combineReducers(reducers) {
       );
       return newState;
     });
+
+    if (process.env.NODE_ENV !== 'production' && !stateShapeVerified) {
+      verifyStateShape(state, finalState);
+      stateShapeVerified = true;
+    }
+
+    return finalState;
   };
 }
--- a/test/utils/combineReducers.spec.js
+++ b/test/utils/combineReducers.spec.js
@@ -97,7 +97,7 @@ describe('Utils', () => {
         }
       });
 
-      expect(reducer(0, { type: increment }).counter).toEqual(1);
+      expect(reducer({counter: 0}, { type: increment }).counter).toEqual(1);
     });
 
     it('should throw an error if a reducer attempts to handle a private action', () => {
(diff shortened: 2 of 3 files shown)


gaearon:
Let's do this in `combineReducers` instead?

michalkvasnicak:
I think it's enough to put it here before `return` https://github.com/gaearon/redux/blob/breaking-changes-1.0/src/utils/combineReducers.js#L63

gaearon:
I think we should change `state = {}` to `state = mapValues(finalReducers, () => undefined)` so if the state isn't rehydrated, the warning is skipped.

michalkvasnicak:
I think that it need extra check because `typeof []` or `typeof null` return `object` too.

https://github.com/ellbee/redux/commit/0fee440b57b2c021a2b793b6423ce237c75d167f#diff-bc52e10bb986be8700c6e212e9bc581cR22

gaearon:
We don't need to worry about this if the check is moved to `combineReducers` because the root state is then guaranteed to be an object.

michalkvasnicak:
@gaearon Yeah you are right, it needs to call verifyState after mapValues are done, otherwise it'll warn everytime. I did't get mapValues right at first look.

gaearon:
Hmm. The initial state can still be wrong. I guess we should just check `isPlainObject` on it, shouldn't we? If it's not a plain object something is already bad.

michalkvasnicak:
@gaearon I agree. Initial state should be checked to be a plain object.

emmenko:
PS: this is a duplicate of #276 . We can close mine, but maybe you can keep the failing test?

ellbee:
@gaearon Yes, no problem. I've moved the check to combineReducers.

src/utils/verifyStateShape.js
```diff
@@ -0,0 +1,24 @@
...
+import isPlainObject from './isPlainObject';
+
+export default function verifyStateShape(initialState, currentState) {
+  if (isPlainObject(initialState) &&
+      !Object.keys(initialState).every(key => key in currentState)) {
+    return 'Shape of initial state object given to createStore does not ' +
+      'match shape of state object returned by reducer';
+  }
+
+  return '';
```
gaearon: Better `return null` IMO.

src/utils/combineReducers.js
```diff
@@ -47,5 +50,16 @@ export default function combineReducers(reducers) {
       );
       return newState;
     });
+
+    if (process.env.NODE_ENV !== 'production') {
+      var warning = verifyStateShape(state, finalState);
+      /* eslint-disable no-console */
+      if (console && console.warn && warning !== '') {
```
gaearon: Let's use `warning` module for consistency, `if (warning)` is also enough.

src/utils/verifyStateShape.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param {any} currentState Current state of the store.
+ *
+ * @returns {string} Warning message if initial and current state do
+ * not have the same shape.
+ */
+
+import isPlainObject from './isPlainObject';
+
+export default function verifyStateShape(initialState, currentState) {
```
gaearon: Let's put it into `combineReducers.js`. I don't see the benefit of keeping it separately.

test/utils/verifyStateShape.spec.js
```diff
@@ -0,0 +1,13 @@
+import expect from 'expect';
+import verifyStateShape from '../../src/utils/verifyStateShape';
+
+describe('verifyStateShape', () => {
+  it('should return warning message if object shapes differ', () => {
```
gaearon: I'd rather see the unit test of `combineReducers` having correct behavior, to make sure it's actually being used correctly.

ellbee:
@gaearon I've hopefully addressed your comments in the latest commit

@emmenko Do you think we still need the failing test in createStore? I'll certainly add it if so.

src/utils/combineReducers.js
```diff
@@ -47,5 +56,13 @@ export default function combineReducers(reducers) {
       );
       return newState;
     });
+
+    warning(
+      verifyStateShape(state, finalState),
+      'Shape of initial state object given to createStore does not ' +
+      'match shape of state object returned by reducer'
```
gaearon: By this point the user does not really define a reducer. All they do is pass `reducers` object (which they might get from `import * as something` ES6 import). Therefore they won't understand which reducer the warning message is referring to.

The warning should instead say something like `Unexpected key "<key>" in the initialState will be ignored. Expected to find one of the known reducer keys instead: "oneReducer", "otherReducer", ...`.

test/utils/combineReducers.spec.js
```diff
@@ -119,5 +119,17 @@ describe('Utils', () => {
...
+      const spy = expect.spyOn(console, 'error');
+      const reducer = combineReducers({
+        foo(state = {bar: 1}) {
+          return state;
+        }
+      });
+      reducer({foo: {bar: 2}});
+      reducer({bar: 2});
+      reducer(1);
+      expect(spy.calls.length).toEqual(2);
```
gaearon: And let's check for a sensible warning message here using a regexp

ellbee:
Updated with better warning message

gaearon:
Looks really good. I'll test this in a day or two and get this into 1.0 release.

:+1:

gaearon:
Please rebase on `breaking-changes-1.0` and squash this into a single commit.

emmenko:
@ellbee I think it's ok like this, just it's not really the complete flow. Usually you do `createStore(reducer, initialState)` and `store.getState()`.
But I'm not sure if it's really necessary...

gaearon:
@emmenko Do you mean the tests? I don't think it's a problem because this is behavior completely specific to `combineReducers`.

ellbee:
Ok, thanks @gaearon! Rebased and squashed.

emmenko:
:+1:

src/utils/combineReducers.js
```diff
@@ -1,7 +1,9 @@
+import { ActionTypes } from '../createStore';
+import invariant from 'invariant';
+import isPlainObject from '../utils/isPlainObject';
 import mapValues from '../utils/mapValues';
 import pick from '../utils/pick';
-import invariant from 'invariant';
-import { ActionTypes } from '../createStore';
+import warning from 'warning';
```
gaearon: Nitpick: let's move `invariant` here so their imports are together.

ellbee: Move invariant next to warning in the import order? Can do, I just put them into alphabetical order.

ellbee: Ok, done

gaearon:
I played with this a little bit more, and I think we could still improve this a little bit:

* When the initial state is plain object, there is no warning. What I meant is, we *should always* warn if the initial state is not a plain object: that's a symptom of a mistake. There is no way for `combineReducers` to do anything useful with something other than a plain object.

* If I remove `export default` from `todos.js` in `todomvc` example, I get a cryptic error message: `Unexpected key "todos" in the initialState will be ignored. Expected to find one of the known reducer keys instead: ""`. What does `""` mean? It means `Object.keys(finalReducers).length === 0`. But it's definitely a symptom of a mistake! Evidently the user to forgot to export any reducer, so we should probably have a better warning in this situation.

* The check is being performed at every invocation, not just at the initial one. This feels like a waste even in development environment. Can we only emit the warning the first time the function is invoked? I understand it makes it impure, but it's no big deal in development environment: we just want to warn (and check) only once. Does this make sense?

emmenko:
Good points! 

> Can we only emit the warning the first time the function is invoked?

emmenko:
Sorry, my finger slipped...

I wanted to ask if it would be a problem with hot reloading?

> Can we only emit the warning the first time the function is invoked?

gaearon:
I don't think this would be a problem? On hot reloading the function will be re-generated so the code will run again.

emmenko:
Ok, just wanted to make sure :)

ellbee:
Updated pull request to add extra warnings and only call verifyStateShape once as [suggested](#issuecomment-124712744). Ideas for improvements to the warning messages are welcome!

I have been wondering though if these new warnings should be invariants instead?

ellbee:
@gaearon I'll have some spare time later today, is there anything else you would like me to do with this PR?

gaearon:
@ellbee 

I don't have time to look at it right now, but please do what I'd do:

* Take a sample app
* Turn your mind off and make stupid mistakes with `initialState`, exporting reducers, `export default`, etc
* See if the warnings help you or not

ellbee:
@gaearon OK, will do. Thanks!

gaearon:
Hey, awesome job. Thanks.

nuragic:
Hi! I just get this warning and I don't understand what's the problem with this one..

I mean, I have an initial state from the server that looks like this for example:
```js
{
  foo: [/* ... */], // ok, I've a reducer called 'foo'
  bar: true,        // no reducer
  baz: false,       // same here

}
```
Should I have one reducer for every single state property in order to access it from my components?

gaearon:
Yes, if you use combineReducers, because that's exactly what it does. It lets you assign a single reducer to each state property.

You can also write the root reducer by hand, without combineReducers. Then it's up to you how to structure it.

https://github.com/rackt/redux/blob/master/docs/api/combineReducers.md

https://github.com/rackt/redux/blob/master/docs/basics/Reducers.md

nuragic:
Thanks for the quick response! 

Ok, point taken... Maybe I need to think a bit more about my app state in order to know if I need the 1:1 relation between state properties and reducers... Another option could be to have 1 reducer to manage the entire state right?

gaearon:
Yes. In fact that's what combineReducers generates — a single reducer. (And instead of using combineReducers you can write it by hand.)

nuragic:
Ok, I'll evaluate if I need to write my own or simply use one "master" reducer (I'm doing a small app btw eheh). 

Many thanks for the support! :)

PR #305, opened 2015-07-23
Is plain object fix

### LINKED ISSUE #304: Dispatching action from iframe to store defined in parent window always throws
Dispatching an action from an iframe to store defined in a parent window always throws `Invariant Violation: Actions must be plain objects. Use custom middleware for async actions.`

It fails on the check `Object.getPrototypeOf(obj) === Object.prototype` in `isPlainObject`.

Only way to fix this is to expose actions in the parent window.

michalkvasnicak:
When `isPlainObject` is called between two different contexts then `Object.prototype` is not the same. Now it is comparing only stringified `constructors`. Fixes #304


michalkvasnicak:
Ah, wrong branch, sorry :(

gaearon:
The PR should probably be done against `breaking-changes-1.0` branch

gaearon:
This was merged as #306.

PR #306, opened 2015-07-23
fix isPlainObject

### LINKED ISSUE #304: Dispatching action from iframe to store defined in parent window always throws

michalkvasnicak:
When isPlainObject is called between two different contexts then Object.prototype is not the same. Now it is comparing only stringified constructors. Fixes #304

### DIFF
--- a/src/utils/isPlainObject.js
+++ b/src/utils/isPlainObject.js
@@ -1,4 +1,4 @@
-const fnToString = (fn) => Function.prototype.toString.call(fn);
+var fnToString = (fn) => Function.prototype.toString.call(fn);
 
 /**
  * @param {any} obj The object to inspect.
@@ -9,7 +9,7 @@ export default function isPlainObject(obj) {
     return false;
   }
 
-  const proto = typeof obj.constructor === 'function' ? Object.getPrototypeOf(obj) : Object.prototype;
+  var proto = typeof obj.constructor === 'function' ? Object.getPrototypeOf(obj) : Object.prototype;
 
   if (proto === null) {
     return true;
--- a/src/utils/isPlainObject.js
+++ b/src/utils/isPlainObject.js
@@ -1,12 +1,23 @@
+const fnToString = (fn) => Function.prototype.toString.call(fn);
+
 /**
  * @param {any} obj The object to inspect.
  * @returns {boolean} True if the argument appears to be a plain object.
  */
 export default function isPlainObject(obj) {
-  if (!obj) {
+  if (!obj || typeof obj !== 'object') {
     return false;
   }
 
-  return typeof obj === 'object' &&
-         Object.getPrototypeOf(obj) === Object.prototype;
+  const proto = typeof obj.constructor === 'function' ? Object.getPrototypeOf(obj) : Object.prototype;
+
+  if (proto === null) {
+    return true;
+  }
+
+  var constructor = proto.constructor;
+
+  return typeof constructor === 'function'
+    && constructor instanceof constructor
+    && fnToString(constructor) === fnToString(Object);
 }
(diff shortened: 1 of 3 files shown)


gaearon:
Can you add a test that would have been failing with the previous version?

michalkvasnicak:
I will try but don't know how to reproduce it inside the same context.

gaearon:
Look into lodash tests? :-P

phated:
Cc @jdalton

michalkvasnicak:
It seems that lodash is creating its own context.

michalkvasnicak:
OK I can reproduce it but had to install contextify for it. Can I push it with contextify as devDependency?

gaearon:
Yes, sure.

src/utils/isPlainObject.js
```diff
@@ -1,12 +1,23 @@
...
  */
 export default function isPlainObject(obj) {
-  if (!obj) {
+  if (!obj || typeof obj !== 'object') {
     return false;
   }
 
-  return typeof obj === 'object' &&
-         Object.getPrototypeOf(obj) === Object.prototype;
+  const proto = typeof obj.constructor === 'function' ? Object.getPrototypeOf(obj) : Object.prototype;
```
gaearon: Please change `const` to `var`, we're maintaining Flow-friendliness for now. (I'll alter ESLint rule later.)

test/utils/isPlainObject.spec.js
```diff
@@ -7,11 +8,17 @@ describe('isPlainObject', () => {
       this.prop = 1;
     }
 
+    const sandbox = contextify();
```
gaearon: (It's fine to keep using `const` in tests)

gaearon:
Looks good, thank you!

michalkvasnicak:
I thank you for this awesome lib. :)

jdalton:
fwiw lodash v4 drops IE8 support and has revised it's `isPlainObject` to be considerably more [light weight](https://github.com/lodash/lodash/blob/e7ad50d2434ff907ef266c20d870386a6c1a5ecf/lodash.js#L8250-L8264). You caught us at a time when we pulled our modern method fork of `isPlainObject` for a single larger implementation.

jdalton:
It looks like this PR barrows from our WIP v4 implementation so you'll dig this method is designed to work with the es5-sham fallback for `Object.getPrototypeOf` too by punting if `constructor` is not a function.

michalkvasnicak:
@jdalton Yes it is, I was pointed to look at the lodash implementation. I hope it's OK with you.

gaearon:
@jdalton Appreciate you chiming in!

jdalton:
@michalkvasnicak
>  I hope it's OK with you.

In the long term copy pasting is a bummer as it shifts the implementation testing burden to you all. It also means you all have to spend time mucking with and becoming experts in utils in stead of making cool stuff on top of it. It's on me to make the existing lodash tools more consumer friendly and file size is part of that. I hope this can be revisited in v4 where file size is a major focal point.

gaearon:
@jdalton 

I agree totally. In our case we wanted to get some micro file size optimizations because the whole lib is 2KB minified+gzipped so we kind of sacrifice util maintainability. OTOH we only need three tiny utils so in my book it's worth the effort.

jdalton:
Yap. As the old saying goes you can lead a horse to water but you sure as hell can't make them bust out a sweet jazz solo

PR #320, opened 2015-07-25
added react native example

arnif:
Hi, I implemented the counter example for react native and added it to the example list.


gaearon:
Thank you!

PR #326, opened 2015-07-25
JSDoc improvements

emmenko:
**do not merge yet**

### DIFF
--- a/docs/Reference/API.md
+++ b/docs/Reference/API.md
@@ -50,7 +50,7 @@ the current state tree and the action to handle.
 
 * `initialState: any`: The initial [state](./Glossary.md#state). You may optionally specify it
 to hydrate the state from the server in universal apps, or to restore a
-previously serialized user session.
+previously serialized user session. If you use `combineReducers` to produce the root reducer function, this must be an object with the same shape as `combineReducers` keys.
 
 ##### Example
 
@@ -255,6 +255,8 @@ asynchronous actions in a concise manner, or logging every action payload.
 Because middleware is potentially asynchronous, this should be the first store enhancer in the composition chain.
 
 Returns a store enhancer function that needs to be applied to `createStore` to add any middleware to it.
+
+Note that each middleware will be given the `dispatch` and `getState` functions as named arguments.
 
 ##### Parameters
 
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -22,9 +22,11 @@ export var ActionTypes = {
  * @param {Function} reducer A function that returns the next state tree, given
  * the current state tree and the action to handle.
  *
- * @param {any} initialState The initial state. You may optionally specify it
+ * @param {any} [initialState] The initial state. You may optionally specify it
  * to hydrate the state from the server in universal apps, or to restore a
  * previously serialized user session.
+ * If you use `combineReducers` to produce the root reducer function, this must be
+ * an object with the same shape as `combineReducers` keys.
  *
  * @returns {Store} A Redux store that lets you read the state, dispatch actions
  * and subscribe to changes.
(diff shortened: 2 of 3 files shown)


emmenko:
@gaearon I've had a look at the JSDocs and improved them a little (hopefully). Otherwise LGTM :+1:

docs/Reference/API.md
```diff
@@ -48,9 +48,9 @@ There should only be a single store in your app.
 * `reducer: Reducer`: Required. A [reducer](./Glossary.md#reducer) function that returns the next state tree, given
 the current state tree and the action to handle.
 
-* `initialState: any`: The initial [state](./Glossary.md#state). You may optionally specify it
+* `initialState: Object`: The initial [state](./Glossary.md#state). You may optionally specify it
 to hydrate the state from the server in universal apps, or to restore a
-previously serialized user session.
+previously serialized user session. The object keys must match the reducer names managed by `combineReducers`.
```
gaearon: This is a tricky one. Please don't *have to* use `combinedReducers`. They may be passing a single reducer function that operates on an Immutable tree. Maybe rephrasing it as “If you use `combineReducers` to produce the root reducer function, this must be an object with the same shape as `combineReducers` keys.” would help.

src/createStore.js
```diff
@@ -125,6 +127,11 @@ export default function createStore(reducer, initialState) {
     dispatch({ type: ActionTypes.INIT });
   }
 
+  /**
```
gaearon: This is not really JSDoc, is it? It's describing a line of code, not a variable or export. Should it have `/**`?

docs/Reference/API.md
emmenko: Makes sense, will rephrase it :+1:

src/createStore.js
emmenko: Hmm good question. No it's not meant to be JSDoc, just a comment to explain why we are dispatching. Should I use `//` line comments or should I just remove it? What do you suggest?

gaearon: I think `//` is fine in these cases.

emmenko:
@gaearon ok, done the changes.

docs/Reference/API.md
```diff
@@ -48,9 +48,9 @@ There should only be a single store in your app.
 * `reducer: Reducer`: Required. A [reducer](./Glossary.md#reducer) function that returns the next state tree, given
 the current state tree and the action to handle.
 
-* `initialState: any`: The initial [state](./Glossary.md#state). You may optionally specify it
+* `initialState: Object`: The initial [state](./Glossary.md#state). You may optionally specify it
```
gaearon: It still doesn't have to be an object. `any` is correct.

src/createStore.js
```diff
@@ -22,9 +22,11 @@ export var ActionTypes = {
  * @param {Function} reducer A function that returns the next state tree, given
  * the current state tree and the action to handle.
  *
- * @param {any} initialState The initial state. You may optionally specify it
+ * @param {Object} [initialState] The initial state. You may optionally specify it
```
gaearon: Same: `any` is correct.

src/createStore.js
```diff
@@ -125,6 +128,10 @@ export default function createStore(reducer, initialState) {
     dispatch({ type: ActionTypes.INIT });
   }
 
+
+  // When a store is created, an "INIT" action is dispatched so that every
+  // reducer will return their initial state. This effectively populates
+  // the initial store tree.
```
gaearon: "store tree" -> "state tree".

src/createStore.js
```diff
@@ -125,6 +128,10 @@ export default function createStore(reducer, initialState) {
     dispatch({ type: ActionTypes.INIT });
   }
 
+
+  // When a store is created, an "INIT" action is dispatched so that every
+  // reducer will return their initial state. This effectively populates
```
gaearon: "will return" -> "returns" (grammar :-)

src/utils/isPlainObject.js
```diff
@@ -1,6 +1,6 @@
 /**
- * @param {any} obj The object to inspect.
- * @returns {boolean} True if the argument appears to be a plain object.
+ * @param {*} obj The object to inspect.
```
gaearon: I don't mind changing `any` to `*` but then this needs to be done everywhere (including API reference doc).
I'm also happy with keeping it called `any` :-)

src/createStore.js
emmenko: But shouldn't we expect an `Object` here?

src/utils/isPlainObject.js
```diff
@@ -1,6 +1,6 @@
...
+ * @returns {Boolean} True if the argument appears to be a plain object.
```
gaearon: Why change this? `boolean` is lowercase in Google Closure types reference (one of JSDoc type “providers”), it is also lowercase in Flow.

src/createStore.js
emmenko: :+1:

src/createStore.js
emmenko: :+1:

src/createStore.js
gaearon: Nope. Only if you use `combineReducers`. Otherwise it can be anything, an array, an Immutable object, etc.

gaearon: Also it can be `undefined`. It's optional

src/utils/isPlainObject.js
emmenko: Hmm ok, I just taken it from http://usejsdoc.org/tags-param.html

screen shot 2015-07-27 at 16 10 25

src/utils/isPlainObject.js
emmenko: My bad, I thought it was always uppercase...will revert.

src/createStore.js
emmenko: > it can be anything, an array, an Immutable object, etc.

Riiiight, did thought of that. Thanks! :+1:

emmenko:
@gaearon I've squashed the commits. I think it should be ok now. 

Regarding `any` vs `*` I'm not sure, I guess we can keep `any` :+1:

gaearon:
Thanks!

PR #334, opened 2015-07-27
Break down into smaller concepts and rewording.

anyong:


gaearon:
My mind is so infused with these concepts right now so it's really hard for me to say what is clearer.
What do you folks say?

cc @acdlite

gaearon:
Thank you! I decided I don't want bullet points in Core Ideas, but I tried to make it more reader-friendly:
https://github.com/gaearon/redux/blob/rewrite-docs-again/docs/basics/ThreePrinciples.md

I _will_ use your Redux Flow though! I'm closing PR as I work from another branch now, but I'll copy parts of it. Really appreciate your work there. :+1:

PR #335, opened 2015-07-27
Update The Redux Flow.md

wmertens:


gaearon:
I shortened the third paragraph in the next commit, can you please do the same? We don't need to get too far into the details here, we just want to provide a quick reference.

wmertens:
@gaearon done...

bendyorke:
FWIW, on line 10, "Although" has a "t" at the end

gaearon:
Hey, thank you for the effort! I want to go with this version from @anyong in #334 instead: https://github.com/anyong/redux/commit/c1ef3a24e8f5a59b08620907b5c688b9af507206.

Sorry for duplicate work: getting 1.0 out is a meat grinder, and I'm not very good at managing the community effort right now. :-(

Is there something you feel is missing in @anyong's version?

PR #336, opened 2015-07-27
Update ordering or things

quicksnap:
I find this ordering to be more clear IMO. I'm fine with this PR getting rejected =P


gaearon:
:+1: this is good, thx

PR #337, opened 2015-07-27
fix id assignment in example

Jon-Biz:
The todo example assumes that the earliest item in the item array is the most recent, which is perfectly fine when the example is used as presented.  However, once you start experimenting (retrieving items from elsewhere, adding sync from a server etc), this may no longer be the case. Since React uses the `id` to assign item keys, this becomes another thing to debug as one is familiarizing oneself with redux's behaviour. The small change proposed below makes `id` assignment simpler and consequently easier to understand.

### DIFF
--- a/examples/todomvc/stores/todos.js
+++ b/examples/todomvc/stores/todos.js
@@ -10,7 +10,7 @@ export default function todos(state = initialState, action) {
   switch (action.type) {
   case ADD_TODO:
     return [{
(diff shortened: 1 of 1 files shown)


knowbody:
this is just the example, you don't have to have the `id` there, you could have timestamp and sort them based on timestamp for example, or the `id` of the item which would come from some API

gaearon:
This is cleaner indeed, but can you please do this against `breaking-changes-1.0` branch?
Master is out of date.

PR #352, opened 2015-07-29
Allow `bindActionCreators` to be used with function as actionCreator

michaelcontento:
With this change you can use `bindActionCreators` to either bind a object of actionCreator functions to dispatch or only a single actionCreator directly.

```js
import { bindActionCreators } from 'redux';
import { addTodo } from './actions/todoActions';

const boundAddTodo = bindActionCreators(addTodo, dispatch);
boundAddTodo('Hello');
```

### DIFF
--- a/src/utils/bindActionCreators.js
+++ b/src/utils/bindActionCreators.js
@@ -1,13 +1,18 @@
 import mapValues from '../utils/mapValues';
 
+function bindActionCreator(actionCreator, dispatch) {
+  return (...args) => dispatch(actionCreator(...args));
+}
+
 /**
  * Turns an object whose values are action creators, into an object with the
  * same keys, but with every function wrapped into a `dispatch` call so they
  * may be invoked directly. This is just a convenience method, as you can call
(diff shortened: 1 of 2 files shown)


src/utils/bindActionCreators.js
```diff
@@ -1,13 +1,18 @@
 import mapValues from '../utils/mapValues';
 
+function bindActionCreator(actionCreator, dispatch) {
+    return (...args) => dispatch(actionCreator(...args));
```
gaearon: Should be two spaces

gaearon:
Please squash in a single commit?

michaelcontento:
Done :)

src/utils/bindActionCreators.js
michaelcontento: Oh, sorry! Maybe a [`.editorconfig`](http://editorconfig.org/) file would be nice? This way most editors will recognize the proper indent / lineending style :)

gaearon:
Awesome, thanks!

PR #356, opened 2015-07-29
[WIP] Polymorphic Flow Types

leoasis:
Adding Flow types again, this time polymorphic so that we can define custom action types in the apps and make them type check correctly when using them with redux. Addresses https://github.com/gaearon/redux/issues/290

So far this only has the `createStore` typed, will work on the other util functions. Just wanted to show the progress so far and if anyone wants to chime in and give feedback, more than welcome!

I have a temporary file in the counter example that is the actual one being type checked, instead of the real entry point (since I haven't yet typed the other util functions), which is `sample.js`. If you want to play around, uncommenting the lines there and running `flow` will correctly show the type errors.

You may need to change the counter's `package.json` dependencies to make `redux` point to `file:../../` so that it grabs the code from the real source instead of npm (which doesn't have flow types yet).

Caveat so far: There is a bug in Flow (https://github.com/facebook/flow/issues/582) which forces us to explicitly specify the init action type inside the union list of the app's action types. Until that gets fixed, I'm afraid we'll need to keep doing that instead of having that type provided by `redux` itself, which is defined in the `flow/types.js` file but not used.


examples/counter/reducers/counter.js
```diff
@@ -1,10 +1,12 @@
...
+/* @flow */
 
-export default function counter(state = 0, action) {
+import type {State, Action} from '../types';
+
+export default function counter(state: State, action: Action) : State {
   switch (action.type) {
-  case INCREMENT_COUNTER:
+  case 'INCREMENT_COUNTER':
     return state + 1;
```
gaearon: Will returning non-number fail the type check here?

examples/counter/reducers/counter.js
```diff
@@ -1,10 +1,12 @@
...
-  case DECREMENT_COUNTER:
+  case 'DECREMENT_COUNTER':
     return state - 1;
   default:
     return state;
```
gaearon: Will forgetting a `default` case fail the type check?

examples/counter/reducers/counter.js
leoasis: yes!

examples/counter/reducers/counter.js
leoasis: yes!

gaearon:
This is really really good so far! Thanks for working on it. The next (in terms of importance) after `createStore` is `combineReducers` and `bindActionCreators` I think.

leoasis:
@gaearon yes, been trying to think how to type those in a polymorphic way. Will push updates into this PR

examples/counter/reducers/counter.js
gaearon: :dancer: :dancers:

quicksnap:
:dancer: :dancers: Would love types! Thanks for working on this.

gaearon:
Closing, so I can delete `breaking-changes-1.0`.
Please resubmit against `master`.

PR #357, opened 2015-07-29
fix: remove badges from title

theaqua:
[Docs](http://gaearon.github.io/redux/) have a little bit unnecessary information in header :)

image


gaearon:
Haha.

PR #359, opened 2015-07-30
build-step for flow types

nmn:
In case you do decide to add flow types, this adds the build steps needed to keep in the flow types in the built files.

Using a babel plug-in flow types are turned in to flow comments, which are still parsed by flow.

The one downside to that currently, is that some type information from ES6 classes is lost. There may be a way around this problem, however.

Leaving this PR here for a future decision.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,6 @@
 {
   "stage": 0,
-  "loose": "all"
+  "loose": "all",
+  "plugins": ["flow-comments"],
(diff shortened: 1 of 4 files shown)


gaearon:
Flow comments seem really problematic because Babel output confuses Flow a lot. I'll close because we don't have soon-ish plans to integrate Flow again, at least not until it gets easier to use for libraries.

Issue #368, opened 2015-07-30
Forbid dispatch from inside a reducer


gaearon:
We should throw if you attempt to dispatch while the reducer is being called.

Note that this will *not* cause the dreaded `Cannot dispatch in a middle of dispatch` if you `dispatch` from component's `componentDidMount` or `componentWillReceiveProps`. Unlike Flux, we don't have this problem, because **we call all subscribers *after* the reducer is called**.

This means we can set a flag immediately before `currentState = reducer(currentState, action)` in `createStore` and reset it on the next line, so everything happens before the subscribers are called. The subscribers should be able to call `dispatch` just fine.

I'd very much like to see this as a PR as I don't have much time. It needs to:

* Introduce the error if you dispatch while the reducer is evaluating.
* Include a test for this.
* Should not fail the existing “should handle nested dispatches gracefully” test

Please build it on top of `breaking-changes-1.0` branch.

gaearon:
Another note: need another test to check that an error inside a reducer resets the flag back, i.e. there should be `try/finally` in the implementation to reset the flag.

knowbody:
I'm on it @gaearon

phated:
@gaearon what is the reason for adding this? I've disliked it in every Flux implementation that has it, and not just because the lifecycle issues.

gaearon:
@phated Simply because this is _never_ a valid use case. Reducers must be pure.

gaearon:
In Flux there are valid-ish situations when you'd rather wish this check didn't exist, but in Redux there's just no excuse for dispatching inside reducer. It's never the right solution, and it only indicates a problem in the code.

phated:
Can't it be implemented as middleware and be opt-in or opt-out. Having explicit throws that you can't turn off in library code is a pain.

jide:
For complicated edge cases there are middlewares.

phated:
Maybe it should just be a warning then?

gaearon:
If it's a middleware, I doubt anybody who'd be affected by the error (beginners) would know to use it.
Why would you want to dispatch inside a reducer? It's grossly misusing the library.
It's exactly the same as React doesn't allow you to `setState` inside `render`.

gaearon:
In fact a custom middleware can opt you “out” of this behavior by queueing the “bad” dispatch on a next tick. But I still can't envision a scenario where you'd need it.

phated:
[...]

PR #370, opened 2015-07-30
[WIP] initial work

knowbody:
#368

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -95,18 +95,19 @@ export default function createStore(reducer, initialState) {
     );
     invariant(
       !isDispatching,
-      'Cannot dispatch in the middle of a dispatch.'
+      'Cannot dispatch in the middle of reducer call.'
     );
 
     isDispatching = true;
 
     try {
       currentState = currentReducer(currentState, action);
-      listeners.forEach(listener => listener());
     } finally {
       isDispatching = false;
-      return action;
     }
+
+    listeners.forEach(listener => listener());
+    return action;
   }
 
   /**
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -38,6 +38,7 @@ export default function createStore(reducer, initialState) {
   var currentReducer = reducer;
   var currentState = initialState;
   var listeners = [];
+  var isDispatching = false;
 
   /**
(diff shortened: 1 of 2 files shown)


src/createStore.js
```diff
@@ -92,10 +93,20 @@ export default function createStore(reducer, initialState) {
...
+    );
+
+    isDispatching = true;
 
-    currentState = currentReducer(currentState, action);
-    listeners.forEach(listener => listener());
-    return action;
+    try {
+      currentState = currentReducer(currentState, action);
+      listeners.forEach(listener => listener());
```
gaearon: We want to move this line out. It's OK to call the listeners. The only line we want to guard is `currentState` assignment.

src/createStore.js
```diff
@@ -92,10 +93,20 @@ export default function createStore(reducer, initialState) {
       isPlainObject(action),
       'Actions must be plain objects. Use custom middleware for async actions.'
     );
+    invariant(
+      !isDispatching,
+      'Cannot dispatch in the middle of a dispatch.'
```
gaearon: Maybe change the wording `Cannot dispatch in the middle of a reducer call.`? To clarify this isn't exactly the same thing as what Flux has.

knowbody: sure, adding those changes and testing. thanks @gaearon

src/createStore.js
knowbody: so you would notify listeners before `try` of in the `finally`?

knowbody:
@gaearon this can be closed as it's done here #372

src/createStore.js
```diff
@@ -92,8 +93,19 @@ export default function createStore(reducer, initialState) {
...
       'Actions must be plain objects. Use custom middleware for async actions.'
     );
+    invariant(
+      !isDispatching,
+      'Cannot dispatch in the middle of reducer call.'
+    );
+
+    isDispatching = true;
+
+    try {
```
taylorhakes: We can't add a `try {} catch {}` here. This will cause way more problems then it solves. All reducer errors will be swallowed.

gaearon:
Closing in favor of #372, thanks for your effort!

PR #371, opened 2015-07-30
Throw error when trying to dispatch from action

quicksnap:


quicksnap:
This is bad--I forgot to take out that `only`

PR #372, opened 2015-07-30
Forbid dispatch from inside a reducer

quicksnap:
#368 -- This one actually works and doesn't disable the entire test suite =)

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -38,6 +38,7 @@ export default function createStore(reducer, initialState) {
   var currentReducer = reducer;
   var currentState = initialState;
   var listeners = [];
+  var isDispatching = false;
 
   /**
    * Reads the state tree managed by the store.
@@ -93,7 +94,18 @@ export default function createStore(reducer, initialState) {
       'Actions must be plain objects. Use custom middleware for async actions.'
     );
 
-    currentState = currentReducer(currentState, action);
+    invariant(
+      isDispatching === false,
+      'Reducers may not dispatch actions.'
+    );
+
+    try {
+      isDispatching = true;
+      currentState = currentReducer(currentState, action);
+    } finally {
+      isDispatching = false;
+    }
+
     listeners.forEach(listener => listener());
     return action;
   }
--- a/test/helpers/reducers.js
+++ b/test/helpers/reducers.js
@@ -1,4 +1,4 @@
-import { ADD_TODO } from './actionTypes';
+import { ADD_TODO, DISPATCH_IN_MIDDLE, THROW_ERROR } from './actionTypes';
 
 
 function id(state = []) {
@@ -26,6 +26,25 @@ export function todosReverse(state = [], action) {
       id: id(state),
       text: action.text
     }, ...state];
+  default:
+    return state;
+  }
+}
+
+export function dispatchInTheMiddleOfReducer(state = [], action) {
+  switch (action.type) {
+  case DISPATCH_IN_MIDDLE:
+    action.boundDispatchFn();
+    return state;
+  default:
+    return state;
+  }
+}
+
+export function errorThrowingReducer(state = [], action) {
+  switch (action.type) {
+  case THROW_ERROR:
+    throw new Error();
   default:
     return state;
   }
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -1,6 +1,6 @@
 import expect from 'expect';
 import { createStore, combineReducers } from '../src/index';
-import { addTodo } from './helpers/actionCreators';
+import { addTodo, dispatchInMiddle, throwError } from './helpers/actionCreators';
 import * as reducers from './helpers/reducers';
 
 describe('createStore', () => {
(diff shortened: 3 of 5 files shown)


src/createStore.js
```diff
@@ -93,7 +94,17 @@ export default function createStore(reducer, initialState) {
       'Actions must be plain objects. Use custom middleware for async actions.'
     );
 
-    currentState = currentReducer(currentState, action);
+    if (isDispatching === true) {
+      throw new Error('Reducers may not dispatch actions.');
+    }
+
+    try {
```
taylorhakes: We can't add a try {} catch {} here. This will cause way more problems than it solves. All reducer errors will be swallowed.

taylorhakes:
Edit: Accidental comment. Didn't realize catch was removed.

src/createStore.js
```diff
@@ -93,7 +94,17 @@ export default function createStore(reducer, initialState) {
       'Actions must be plain objects. Use custom middleware for async actions.'
     );
 
-    currentState = currentReducer(currentState, action);
+    if (isDispatching === true) {
+      throw new Error('Reducers may not dispatch actions.');
```
gaearon: Let's use `invariant` here for consistency.

test/helpers/reducers.js
```diff
@@ -30,3 +30,22 @@ export function todosReverse(state = [], action) {
...
+  switch (action.type) {
+  case BAD_TODO:
+    action.boundDispatchFn();
+    return state;
+  default:
+    return state;
+  }
+}
+
+export function antipatternTodoWithError(state = [], action) {
```
gaearon: I guess this one is just `throwingTodo`? Not really fair to call it `antipattern*` because it doesn't `dispatch` itself.

test/helpers/reducers.js
```diff
@@ -30,3 +30,22 @@ export function todosReverse(state = [], action) {
     return state;
   }
 }
+
+export function antipatternTodo(state = [], action) {
```
gaearon: Maybe call it `dispatchInTheMiddleOfReducer`? It isn't much related to todos anyway.

quicksnap:
Updated with feedback. LMK if names make sense

gaearon:
Love it. Thanks.

hnordt:
@quicksnap I was digging though Redux source code and found it:

```
    try {
      isDispatching = true;
      currentState = currentReducer(currentState, action);
    } finally {
      isDispatching = false;
    }
```

Just for curiosity, why you've used try/finally in that case?

quicksnap:
Thread resurrected! Haha.

 @hnordt if a user's reducer throws an exception, redux's internal state could get locked into `isDispatching === true` which would not be good.

hnordt:
@quicksnap thank you for the clarification 👍

vanvoorden:
> @hnordt if a user's reducer throws an exception, redux's internal state could get locked into isDispatching === true which would not be good.

@quicksnap Ahh… just discovered this one. I'm not a JS specialist… but I'm trying to learn some more history of the evolution of Redux specifically around error handling patterns.

If a user dispatches an action to a `Redux.Store` it looks like we have some defensive "soundness" checks taking place that can throw an error. One of those soundness checks is here:

```js
    try {
      isDispatching = true;
      currentState = currentReducer(currentState, action);
    } finally {
      isDispatching = false;
    }
```

If the `currentReducer` throws an error… we set `isDispatching` to `false`. We know that `currentReducer` might throw an error because of the defensive soundness checks we wrote here in `dispatch`. My question more then is to what potential extent would this `try-finally` have *also* been to handle Reducers that throw to indicate some kind of scenario where a State could not be transformed from an action.

From searching around it looks like I didn't find much historical consensus on whether or not throwing an error from a JS Reducer was ever considered "canon" or not. Reducers should be "pure" functions and it looks like there was debate whether or not throwing an error meant a function could no longer be pure.[^1]

Some of these questions might be legacies and artifacts of the JS runtime… it's possible that the way JS structures error handling and throwing makes it not a pure function anymore. It's less a question at that point of whether or not "Reducers" (as an abstract idea agnostic of any specific language requirements) can or cannot throw. It's more a question of whether or not "JS Reducers" can or cannot throw.

Another way of look at this is if there was ever a discussion about attempting to "ban" Reducers that throw from *inside* the Reducer as an attempt to enforce the "purity" of the function. I'm not sure what tools JS would make available for that or if it's even possible.

[^1]: https://github.com/reduxjs/redux-toolkit/issues/532#issuecomment-621516764

gaearon:
In React itself (for components), we use throwing to "suspend" execution if asynchronous dependencies haven't yet loaded. That's one example of an interesting pattern with throws.

But in Redux, I don't think throwing from reducers generally makes sense. Who's going to handle the throw? It seems like the knowledge to handle this specific error would need to be placed at every action that can lead to the throwing codepath. Unless we're happy with just blowing up the entire call stack e.g. to prevent some kind of state corruption. So I think because it's not very compositional and doesn't simplify authoring UI logic, there isn't much use for this pattern.

PR #373, opened 2015-07-30
Remove ES7 from examples

rwillrich:
Here is some features I have removed:

- Decorators;
- Bind syntax;
- Class properties;
- Object spread.


quicksnap:
These examples don't have automated tests AFAIK--did you test them out? =)

rwillrich:
I've tested them manually. Maybe we should open an issue to add test coverage to them.

quicksnap:
I haven't worked much with Travis--how well would it deal with a nested project?

rwillrich:
Don't know. What do you think @gaearon?

quicksnap:
I kind of see how it works.. I'll open an issue on it.

gaearon:
Awesome, thank you.

rwillrich:
You're welcome! :wink:

ms88privat:
Why no ES7 anymore?

quicksnap:
I believe they're being removed from the examples because not everyone is
comfortable using non-standard language features--doesn't make them
suitable for general purpose examples.

mindjuice:
You could have have ES5, ES6 and ES7 versions of the examples like @gaearon did for React DnD. https://gaearon.github.io/react-dnd/

Pick your poison.

gaearon:
I think at this point everybody who knows about ES7 features will use them anyway, so no big deal.

Maintaining three versions of examples is too big a burden for me at this point.

examples/counter/components/Counter.js
```diff
@@ -23,3 +16,12 @@ export default class Counter extends Component {
...
 }
+
+Counter.propTypes = {
+  increment: PropTypes.func.isRequired,
+  incrementIfOdd: PropTypes.func.isRequired,
+  decrement: PropTypes.func.isRequired,
+  counter: PropTypes.number.isRequired
+};
+
+export default Counter;
```
baddox: I thought it works fine to `export default class Foo` at the top and still add propTypes after the class declaration.

(Forgive me if I'm wrong, I can't test it out right now.)

knowbody: @baddox I do agree. looks weird but it's correct. maybe send a PR

PR #389, opened 2015-08-03
Correct link to redux-devtools.

ngbrown:


gaearon:
Sorry for the confusion, `rewrite-docs-again` is the current branch. Can you check there?

PR #394, opened 2015-08-04
Dispatch data from the Call API Middleware example

merk:
Just a small change, but going over this example, the data returned from the API call isnt being dispatched :)


gaearon:
You're right, thank you.

PR #399, opened 2015-08-05
explain a little more what a single reducer in combination is

constb:
I don't know if its my bad luck, but coming from facebook flux and seeing that every reducer in examples have parameter ``state``, I assumed that every reducer manipulates complete state of the application. It took me quite a few hours before I found out that when you use ``combineReducers`` things are different. Maybe docs could be a bit more explicit about that.

### DIFF
--- a/docs/api/combineReducers.md
+++ b/docs/api/combineReducers.md
@@ -7,7 +7,7 @@ reducing function you can pass to [`createStore`](createStore.md).
 
 The resulting reducer calls every child reducer, and gather their results into a single state object. The shape of the state object matches the keys of the passed `reducers`.
 
-Reducer function names are important. Don't be confused by the fact that reducer parameter is named `state`. *Single reducer in the combination does not update compete state of the store.* Its name is the key of the value in the store object it gets and updates. For example, reducer named `todos` receives and changes only `state.todos`.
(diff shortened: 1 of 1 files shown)


docs/api/combineReducers.md
```diff
@@ -7,6 +7,8 @@ reducing function you can pass to [`createStore`](createStore.md).
 
 The resulting reducer calls every child reducer, and gather their results into a single state object. The shape of the state object matches the keys of the passed `reducers`.
 
+Reducer function names are important. Don't be confused by the fact that reducer parameter is named `state`. *Single reducer in the combination does not update compete state of the store.* Its name is the key of the value in the store object it gets and updates. For example, reducer named `todos` receives and changes only `state.todos`.
```
gaearon: I'd say it's not exactly function names but keys in object you pass to combineReducers. They match function names when you use export syntax.

ms88privat:
I was surprised too. It is good, but it should be in the docs

gaearon:
Thanks. This is on my list to revisit. I added quite a bit about it into [reducers doc](http://gaearon.github.io/redux/docs/basics/Reducers.html).  
I'm closing because:

* I need to delete the branch now that docs are in master;
* This is not exactly how I'd like to describe it, because it's not accurate. Reducer function names are **not** important. See more clarifications in #428.

Issue #350, opened 2015-07-28
Find a way to cut getReducer and replaceReducer from the API


gaearon:
They are used for hot reloading by [react-redux](https://github.com/gaearon/react-redux) and code splitting. However the solution feels wrong and ad-hoc to me.

Something like a store enhancer that would wrap the reducer a la [devTools](https://github.com/gaearon/redux-devtools/blob/master/src/devTools.js) seems like the way forward. This way hot reloading would be opt in, but much more reliable.

Whatever the solution may be, it has to solve the following issues to be considered:

#37 #301 #340 #346 #348 https://github.com/gaearon/redux-devtools/issues/68

Ideas welcome!

taylorhakes:
Here is my proposal. I am sure it has some issues.

Expose reducer as a simple function to Higher Order Stores.

Higher order stores would be created like this:
```js
createStore(middleware, devtools, persistState);
```

Higher Order Stores would have the following signature:
```js
function higherOrderStore({ reducer, dispatch, getState, subscribe }) {
  // wrap components

  return {
    reducer: newReducer,
    dispatch: newDispatch,
    // ...
  };
}
```

The create store function would have the following signature
```js
createStore(...funcs) {
  const store = new Store();
  const finalStore = compose(...funcs, store);
  finalStore.dispatch({ action: '@@init' });
  return store;
}
```

This allows the store to be very flexible. `@@init` can be called by other components or not at all.

This is just off the top of my head. Haven't thought of all the implications of this change. It definitely would solve #376 though.

gaearon:
Interesting, thanks. On the first sight I like it, but I haven't thought through the implications.
I'd say we'll release 1.0 as is, but this will be on our radar as the next important breaking change.

gaearon:
@taylorhakes Can you hack up a proof of concept of this in a branch?

wbecker:
+1

taylorhakes:
Sure, I can do it this weekend.

acstll:
I have one more proposal. The idea is that `createStore` returns itself, so you can create a copy of a store updating `reducer` and/or `state`. It's a function but of course keeps the 3 main methods `getState`, `dispatch` and `subscribe`.

Here's how I'd implement it: https://github.com/acstll/chopped-redux/blob/next/index.js#L32-L40
You'd also have to take `listeners` as third argument, so you can keep them as well in new copies.

So the API would be this:

```js
let store = createStore(reducer, initialState)

[...]

PR #400, opened 2015-08-05
TodoMVC: Basic provider explanation and link

peteruithoven:
Also see: https://github.com/gaearon/redux/issues/397

### DIFF
--- a/examples/todomvc/containers/App.js
+++ b/examples/todomvc/containers/App.js
@@ -10,6 +10,8 @@ const store = createStore(reducer);
 export default class App extends Component {
   render() {
     return (
+      // Provider makes our store instance available to the components below. Also see:
+      // http://gaearon.github.io/redux/docs/basics/UsageWithReact.html#connecting-to-redux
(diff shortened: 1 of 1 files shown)


srph:
Isn't it preferred to make the Provider the top-most component?

```es6
React.render(
  <Provider store={store}>
    {() => <App /> }
  </Provider>,
  mountNode
);
```

danmartinez101:
I believe that in order for hot reloading to work, the Provider component needs to be exported from a module. I'll try to find a reference.

gaearon:
Don't worry about hot reloading yet. Right now to get hot reloading working you need an extra component, a la

```
const store = ...

export default class Root {
  render() {
    return (
      <Provider store={store}>
        {() => <App />}
      </Provider>
    );
  }
}
```

Later, we'll remove the need for extra component, when we solve #350.

gaearon:
This is out of date, so closing. We need to go through examples to comment in a single take later.

PR #402, opened 2015-08-05
Use const over let

wbecker:
These variables shouldn't be altered.


gaearon:
I agree but it's better to do this consistently all over the docs. Otherwise code doesn't match.

wbecker:
It should - I presume you are going to update the [examples](https://github.com/gaearon/redux/blob/rewrite-docs-again/examples/todomvc/index.js) as well to go along with it. 

Just noticed it, and with the inline editor it makes these changes trivial to perform, so thought I might as well. If it's just noise I'll butt out!

gaearon:
If we do this, I want to do this everywhere, so closing.

Issue #407, opened 2015-08-06
bindActionCreators should provide a better error


jarsbe:
Often I do this:

`import MyActions from '../actions/MyActions';`

And when I try to bind actions `{...bindActionCreators(MyActions, dispatch)}` I get an error from `mapValues.js`:`Cannot convert undefined or null to object`.

This is because my import should be:

`import * as MyActions from '../actions/MyActions';`

It would be nice if the error informed me that `MyActions` is undefined since I seem to do this ALL the time.

gaearon:
I'd accept a PR (to `breaking-changes-1.0` branch) that verifies that, if the argument isn't a function, then at least it's `!= null`.

We could change the error message to `bindActionCreators expected an object or a function, instead received %s. Did you write "import ActionCreators from" instead of "import * as ActionCreators from"?`

jarsbe:
Great, I'd be more than happy to.

gaearon:
Fixed by #411.

PR #411, opened 2015-08-06
Bind warning

jarsbe:
This adds invariant warning support to `bindActionCreators` as per issue #407.

### DIFF
--- a/src/utils/bindActionCreators.js
+++ b/src/utils/bindActionCreators.js
@@ -1,3 +1,4 @@
+import invariant from 'invariant';
 import mapValues from '../utils/mapValues';
 
 function bindActionCreator(actionCreator, dispatch) {
@@ -25,7 +26,15 @@ export default function bindActionCreators(actionCreators, dispatch) {
     return bindActionCreator(actionCreators, dispatch);
   }
 
+  invariant(
+    typeof actionCreators === 'object' && actionCreators != null,
+    'bindActionCreators expected an object or a function, instead received %s. ' +
+    'Did you write "import ActionCreators from" instead of "import * as ActionCreators from"?',
+    typeof actionCreators
+  );
+
   return mapValues(actionCreators, actionCreator =>
     bindActionCreator(actionCreator, dispatch)
   );
 }
+
--- a/test/utils/bindActionCreators.spec.js
+++ b/test/utils/bindActionCreators.spec.js
@@ -37,4 +37,31 @@ describe('bindActionCreators', () => {
       { id: 1, text: 'Hello' }
     ]);
   });
+
+  it('should throw an invariant violation for an undefined actionCreator', () => {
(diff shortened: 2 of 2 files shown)


src/utils/bindActionCreators.js
```diff
@@ -21,6 +22,13 @@ function bindActionCreator(actionCreator, dispatch) {
  * action creator wrapped into the `dispatch` call.
  */
 export default function bindActionCreators(actionCreators, dispatch) {
+  invariant(typeof actionCreators === 'function' ||
```
gaearon: Let's break line after the opening paren?

src/utils/bindActionCreators.js
```diff
@@ -21,6 +22,13 @@ function bindActionCreator(actionCreator, dispatch) {
...
+            typeof actionCreators === 'object',
```
gaearon: This will give false positive for `null`. You really want to check `!= null` instead.

src/utils/bindActionCreators.js
gaearon: Let's move this whole check after `if (typeof actionCreators === 'function') {` condition below. Then you only need to check `!= null` because if it's a function, we won't get there anyway.

test/utils/bindActionCreators.spec.js
```diff
@@ -37,4 +37,18 @@ describe('bindActionCreators', () => {
       { id: 1, text: 'Hello' }
     ]);
   });
+
+  it('should throw an invariant violation for an undefined actionCreator', () => {
+    expect(() => {
+      bindActionCreators(undefined, store.dispatch);
+    }).toThrow('bindActionCreators expected an object or a function, instead received undefined. ' +
```
gaearon: For consistency, let's test for `null` too (I'm not sure it's a real use case, but passing `null` never makes sense).

gaearon: Style nitpick: please put a line break between `.toThrow(` and the string literal.

src/utils/bindActionCreators.js
jarsbe: Good call moving it down. You could check if the actionCreators is NOT an object with `typeof actionCreators === 'object'`? That way undefined, null and primitives are all caught.

jarsbe: Ah silly, I always forget null is an object with typeof...

jarsbe:
@gaearon Thanks for the review (damn you are fast!) - have tightened everything up.

gaearon:
Can I please ask you to:

* Merge an update that fixes Travis from `breaking-changes-1.0`
* Squash this into a single commit

?

jarsbe:
Sure thing, that should do the trick.

gaearon:
:+1:

PR #418, opened 2015-08-07
Minor heading grammar fix in "MiddleWare Attempt #2" example

aarongifford:
I just added "insert" into the original heading "You can logging into a function:"


gaearon:
Oh, thanks. I meant “extract”. Fixed now.

agifford-agi:
Ah, yeah I wasn't exactly sure the wording you were going for so I took a stab with "insert".  Thanks for all the great work on Redux and it's documentation.

aarongifford:
Ah, yeah I wasn't exactly sure the wording you were going for so I took a stab with "insert". Thanks for all the great work on Redux and it's documentation.

PR #424, opened 2015-08-07
update counter example to use react-redux v0.5.1

hartzis:
Update Counter example with most recent react-redux.

Based on https://github.com/gaearon/react-redux/pull/29. 

Remove `CounterApp` component and instead connect to the `Counter` component itself.

I was also debating using the example with Object.assign, but didn't want to mess with all the different ways to bring that in. Destructuring off actions seems to work alright.

Let me know what you think.


hartzis:
I missed updating the tests, working on them now.

hartzis:
@gaearon Just saw you updated react-redux to [v0.6.0](https://github.com/gaearon/react-redux/releases/tag/v0.6.0). I'll update this request in the next couple of days.

examples/counter/components/Counter.js
```diff
@@ -1,9 +1,12 @@
 import React, { Component, PropTypes } from 'react';
+import { bindActionCreators } from 'redux';
+import { connect } from 'react-redux';
+import * as CounterActions from '../actions/CounterActions';
 
-class Counter extends Component {
+export class CounterComponent extends Component {
```
gaearon: Let's keep calling it `Counter`?

examples/counter/containers/CounterApp.js
```diff
@@ -1,23 +0,0 @@
-import React, { Component } from 'react';
```
gaearon: I'd do it differently. Let's encourage separation of “smart” and “dumb” components. See https://github.com/gaearon/react-redux#dumb-component-is-unaware-of-redux and https://github.com/gaearon/react-redux#smart-component-is-connect-ed-to-redux.

hartzis:
@gaearon Definitely agree with the "smart" and "dumb" component separation. I think this now appears to match your react-redux readme example. Let me know what you think.

gaearon:
Any specific reason you're putting them in `actions` instead of just spreading, like before?

hartzis:
It seemed more explicit assigning actions to an actions prop, [example](https://github.com/gaearon/react-redux#inject-todos-and-all-action-creators-addtodo-completetodo--as-actions). 

I also didn't want to require an 'Object.Assign' library, but now I realize we could do this instead:

```
function mapDispatchToProps(dispatch) {
  return {
    ...bindActionCreators(CounterActions, dispatch)
  }
}
```
vs
https://github.com/gaearon/react-redux#inject-todos-and-all-todoactioncreators-and-counteractioncreators-directly-as-props
```
function mapDispatchToProps(dispatch) {
  return bindActionCreators(Object.assign({}, todoActionCreators, counterActionCreators), dispatch);
}
```

I do like the spread operator for `mapDispatchToProps`.

gaearon:
This:

```js
function mapDispatchToProps(dispatch) {
  return {
    ...bindActionCreators(CounterActions, dispatch)
  }
}
```

is exactly equivalent to this:

```js
function mapDispatchToProps(dispatch) {
  return bindActionCreators(CounterActions, dispatch);
}
```

There is no need for spread here.

hartzis:
Thank you sir, nice catch.

Sounds good. Fixed spreading actions into `Counter`.

hartzis:
@gaearon Would you like me to resubmit/squash this pull request to a single commit? I realize only `CounterApp` and `package.json` should have been updated.

gaearon:
Yes, please!

gaearon:
Did you mean to open a new PR?

hartzis:
I did, https://github.com/gaearon/redux/pull/453. I thought it would be easier/cleaner.

gaearon:
Thanks!

PR #425, opened 2015-08-08
Breaking changes 1.0

dzannotti:
Added npm install to root path,
Someone on slack was having problems not being able to run the examples without. Readme wasn't clear if you are only running examples. Not even sure if after docs release this doc will make much sense tho


dzannotti:
Bah... git client gone mental, sorry, will pr properly again

PR #426, opened 2015-08-08
fixed readme examples

dzannotti:
Added npm install to root path,
Someone on slack was having problems not being able to run the examples without. Readme wasn't clear if you are only running examples. Not even sure if after docs release this doc will make much sense tho

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ cd redux
 Run the Counter example:
 
 ```
+cd redux/
+npm install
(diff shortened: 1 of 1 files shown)


gaearon:
The new README is in `rewrite-docs-again`, and instructions have moved to `docs/introduction/Examples.html` I think. Can you change there too?

gaearon:
Superseded by #427.

PR #435, opened 2015-08-09
Adds react-redux-starter-kit to examples

davezuko:
This boilerplate is still a work in progress but is currently functional. I'm actively working on it and am focused on keeping it up to date with the latest Redux and React Router API changes so people can get an idea of how to use both.


gaearon:
Great job.

PR #445, opened 2015-08-09
slight rewording in Dataflow docs

afram:
Hi Dan,

I found the original sentence a bit hard to read and make sense of, thought I'd submit a PR to update.


gaearon:
Thanks, this is indeed much better!

PR #446, opened 2015-08-09
Fix compose function example

sbugert:
applyMiddleware threw an error because the middleware array was not destructured.
I also removed an unused import.


gaearon:
Oh thanks. My bad copy pasting.

Issue #345, opened 2015-07-28
Need help polishing Gitbook


gaearon:
I have Gitbook working in [`rewrite-docs-again`](https://github.com/gaearon/redux/tree/rewrite-docs-again) branch but there are a few rough edges.

* `npm run docs:watch` (launching `gitbook serve`) is extremely slow on changes. Recompiling a book might take 7 seconds. Is that normal, or is something misconfigured?

* When navigating gitbook pages, I see flashes of unstyled content. This is a critical problem. Can we fix them? Does it only happen locally? It doesn't appear like anybody responds to Gitbook issues so I didn't bother filing.

I know Gitbook allows some customization. Maybe veterans can help me out?

cc @theaqua @xgrommx

- [x] Flashes of unstyled content (https://github.com/GitbookIO/gitbook/pull/847)
- [x] Syntax highlighting looks bad especially with JSX (we should use Prism instead of Highlight)
- [x] Add link to edit docs? (#449)
- [x] Cmd+click should open new tab (https://github.com/GitbookIO/gitbook/issues/846, https://github.com/GitbookIO/gitbook/pull/919)
- [x] Remove webfonts (#395, https://github.com/GitbookIO/gitbook/issues/850)
- [ ] Anchor links often don't work the first time (https://github.com/GitbookIO/gitbook/issues/920)
- [ ] Anchors should have permalinks like in React docs (https://github.com/rlmv/gitbook-plugin-anchors)
- [ ] /Dev/ Watches dies for 10 seconds on every second change

knowbody:
looking at it now

knowbody:
@gaearon I don't see the second issue happening on my local (the unstyled content). Everything works ok

gaearon:
@knowbody I see it when switching pages fast in Chrome. But maybe it's my Chrome being weird on beta OS or something.

knowbody:
checked both Chrome and Safari and works fine

gaearon:
That's good to hear!

knowbody:
@gaearon but the build takes a while after the update

gaearon:
If you can help figure out how to make it highlight ES6 syntax, that would help too. Right now the syntax highlighting seems to be ES5-only. For example `import` isn't highlighted.

knowbody:
cool, I'll look into it

theaqua:
* Not 7 seconds, just < 1s
[...]

PR #449, opened 2015-08-10
add edit-link plugin

knowbody:
RE: #345 (edit page)

to install the plugin you need to run:
`gitbook install`  

it runs whatever is in `book.json`, so now the Gitbook version can also be specified in there


knowbody:
closing in favour of #498

PR #451, opened 2015-08-10
Update Action defination

camsong:
Maybe `application` is not accurate here, since `store` is also part of application. According to the data flow graph of Flux/Redux, `View` maybe better.


gaearon:
One tricky part here is that network request results are also actions. So it's not just "views", it's input from user or network or any other source of input.

PR #459, opened 2015-08-11
Add async example

gaearon:
Work in progress. I'm porting https://github.com/gaearon/flux-react-router-example.


gaearon:
Bah, wrong branch.

PR #462, opened 2015-08-11
Fixes #461 - Copy listeners array, so subscribes can't affect the loop.

### LINKED ISSUE #461: Unsubscribe a listener inside a listener has side-effects on other listeners
Unsubscribing a listener inside a listener, will splice the `listeners` array in the store of `createStore`, while looping over that array using `forEach` in `dispatch`.

For example the following program fails:

```js
import assert from 'assert';
import {createStore} from 'redux';

var store = createStore(() => []);

var countA = 0;
var countB = 0;
var countC = 0;

var unsubA = store.subscribe(() => {
	countA++;
});

var unsubB = store.subscribe(() => {
	countB++;
	unsubB();
});

var unsubC = store.subscribe(() => {
	countC++;
});

store.dispatch({type: 'X'});
store.dispatch({type: 'Y'});

console.log(countA, countB, countC);

assert.equal(countA, 2, 'countA');
assert.equal(countB, 1, 'countB');
assert.equal(countC, 2, 'countC');
```

`countC` is **1**, but it should be **2**.

A fix for this problem would be to copy the array before doing the `forEach` in the `dispatch` function, for example:

```js
listeners.slice().forEach(listener => listener());
```

arian:
Unsubscribing during a dispatch would change the listeners array.
However that causes the next listener not to fire.

Fixes #461

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -106,7 +106,7 @@ export default function createStore(reducer, initialState) {
       isDispatching = false;
     }
 
-    listeners.forEach(listener => listener());
+    listeners.slice().forEach(listener => listener());
     return action;
   }
 
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -204,6 +204,28 @@ describe('createStore', () => {
     expect(listenerB.calls.length).toBe(2);
   });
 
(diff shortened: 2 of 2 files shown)


gaearon:
Awesome, thanks.

PR #466, opened 2015-08-12
Update real world example to remove ES7

danmartinez101:


danmartinez101:
@gaearon does standardizing on Object.assign vs _.merge matter?

gaearon:
This is fine, thank you!

PR #473, opened 2015-08-12
Tweak example conventions

### LINKED ISSUE #350: Find a way to cut getReducer and replaceReducer from the API

gaearon:
So... this is fairly opinionated but I think we need to do this.
Changes:

* `Root` component is the one that includes `Provider`, and in more complex cases, `Router`. We make it clear it's needed because that's what keeps hot reloading working until we fix #350.
* Make it clear it's fine to put constants together with `actions` if you have a tiny project. You'll recognize if it gets out of hand anyway.
* Make it a convention to have `configureStore()` method for easier testing (I'm just renaming it here).
* Bye bye, relying on `import *` for reducers. This is [one of the most commonly asked questions](https://github.com/gaearon/redux/issues/428). Now we'll recommend having `reducers/index.js` with `combineReducers` call so you can more clearly trace where the names are coming from.

Umm, that's about it. Everybody on board?


gaearon:
I'm impatient so I'll just merge it. Feel free to argue afterwards :-)

ghengeveld:
Don't forget to update the Arguments section in the combineReducers API docs. It still suggests using `import *`. So does the example.

gaearon:
@ghengeveld Would you like to make a PR?

ghengeveld:
Done: https://github.com/rackt/redux/pull/590

PR #481, opened 2015-08-14
Update Ecosystem.md

tappleby:
I have run into some issue with `redux-batched-updates`, depending where you have it in your middleware chain the return value of `dispatch` will be undefined.

I am proposing my redux-batched-subscribe store enhancer instead, it moves the batching to the edge (notifying subscribe handlers) and allows for any custom function to be used.


gaearon:
Can you file an issue on @acdlite's repo too? Is it documentation or implementation problem? It's fairly common that specific kinds of middleware or enhancers need to be close to start or end of the chain. If documentation is the only problem, let's improve it?

acdlite:
I agree that applying the batching strategy at the subscription-level rather than the dispatch-level is better conceptually, although in a practical sense there's not much difference. I'm fine with this PR, or we could list both.

tappleby:
The main difference is you can never access the dispatch result of a non thunk action. not sure how common this would be.

As far as documentation, as long as you have it near the end of the chain (after thunk, promise, etc) then you should be fine.

We could list both, although they are very similar.

Will open a ticket on @acdlite's repo to document this.

tappleby:
@gaearon is there any more actions to take for this?

gaearon:
:+1:

PR #488, opened 2015-08-14
Minor help text improvements

mindjuice:


docs/basics/Actions.md
```diff
@@ -47,7 +47,7 @@ Finally, we’ll add one more action type for changing the currently visible tod
 
 ## Action Creators
 
-**Action creators** are exactly that—functions that create actions. It's easy to conflate the terms “action” and “action creator,” so do your best to use the proper term.
+**Action creators** are exactly that — functions that create actions. It's easy to conflate the terms “action” and “action creator,” so do your best to use the proper term.
```
srph: I don't there should be a space before and after a hyphen

gaearon:
There are different styles. See http://www.thepunctuationguide.com/em-dash.html

I want to keep it without spaces in these docs.

mindjuice:
Sorry!  I realized afterwards that you were doing it intentionally and using an em dash. :)

docs/basics/Actions.md
mindjuice: I thought Dan was using a hyphen in place of an em dash, which I often lazily do.  In that case I would use spaces to avoid confusion with a hyphenated word.  However, Dan was being typically correct and using an em dash, which I didn't notice until later. :smile:

gaearon: Yeah, monospace fonts don't give it justice :-P

PR #500, opened 2015-08-14
Minor help text improvements

mindjuice:


docs/advanced/Middleware.md
```diff
@@ -326,7 +326,7 @@ let todoApp = combineReducers(reducers);
 let store = createStoreWithMiddleware(todoApp);
 ```
 
-This is it! Now any actions dispatched to the store instance will flow through `logger` and `crashReporter`:
+That's it! Now any actions dispatched to the store instance will flow through `logger` and `crashReporter`:
```
gaearon: Bad apostrophe :-)
Please use ’ instead of '.

PR #508, opened 2015-08-14
Add note about combineReducers in AsyncActions documentation

salehe:
It is somehow hard to grasp what's going on at first sight, specially for someone not much familiar with ES6.
I tried to explain the combineReducers part and let the reader figure out the ES6 part :)

### DIFF
--- a/docs/advanced/AsyncActions.md
+++ b/docs/advanced/AsyncActions.md
@@ -201,6 +201,10 @@ There are a few important bits here:
 
 Before going into the details of dispatching actions together with network requests, we will write the reducers for the actions we defined above.
 
+>##### Note
+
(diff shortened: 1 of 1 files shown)


gaearon:
I think that section is a bit too heavy now because the code snippet is followed by a note, and then the list. How about we move the note *before* the snippet, just below the section title, and put something like 

>#### Note

> This section assumes you understand reducer composition with [`combineReducers()`](/docs/api/combineReducers.md), as described in the [introduction to reducers](/docs/basics/Reducers.md#splitting-reducers). If you don’t, please read it first.

What do you say?

salehe:
I agree. I've updated the commit to reflect this.

gaearon:
Can you please make `combineReducers()` a link to its API reference?

salehe:
Ofcourse :)

PR #513, opened 2015-08-14
Add redux-requests to Ecosystem.md

idolize:
[redux-requests](https://github.com/idolize/redux-requests) is a small Redux middleware and reducer I wrote to help avoid the problem of issuing duplicate HTTP requests.

#### Inspiration

I was inspired by using [Marty's fetch API](http://martyjs.org/guides/fetching-state/index.html) (in that it avoids duplicate HTTP requests), although I was not very satisfied with Marty's approach. Mainly: the Marty fetch API conflates to separate issues:

1. Avoiding duplicate HTTP requests
2. Checking local cache

Mart's fetch API uses the "`id`" to avoid duplicate requests (similar to the "`meta.httpRequest.url`" here), but it also has all that `locally()` `remotely()` stuff to try and manage the local cache, and unfortunately [does not give a lot of control over invalidating the cache](https://github.com/martyjs/marty/issues/354).

#### Why add it to the Ecosystem section?

I want this library to be very simple, and do one thing and one thing well (local cache management should be left up to the developer). And I think managing HTTP requests is a very common requirement for Redux developers.

### DIFF
--- a/docs/introduction/Ecosystem.md
+++ b/docs/introduction/Ecosystem.md
@@ -18,6 +18,7 @@ On this page we will only feature a few of them that the Redux maintainers have 
 * [redux-rx](https://github.com/acdlite/redux-rx) — RxJS utilities for Redux, including a middleware for Observable
 * [redux-batched-updates](https://github.com/acdlite/redux-batched-updates) — Batch React updates that occur as a result of Redux dispatches
 * [redux-logger](https://github.com/fcomb/redux-logger) — Log every Redux action and the next state
+* [redux-requests](https://github.com/idolize/redux-requests) — Avoid issuing duplicate HTTP requests
 
 ## Utilities
 


taylorhakes:
@gaearon How are you deciding what gets on these lists? I know there are hundreds of 3rd party libraries that aren't listed.

It would be great if there were some sort of standards. Like high quality unit tests for all code and number of github stars? It gets crazy (and useless) if everything under the sun is listed.

idolize:
@taylorhakes Yeah, some guidance here would be appreciated. Perhaps this package is too specific for what is intended with the Ecosystem section, but I think it would be great place to point users in the right direction for common use cases (which I personally believe includes managing HTTP requests).

(Also, I just added unit tests to the project :))

gaearon:
Hey, sorry for long time no response. Basically I want `Ecosystem` to be a bunch of stuff we use in the examples, and with fairly stable API. (I'd say `redux-form` is an exception because it's so popular already and solves a very important need.) I'm very busy now and have no time to play with `redux-requests` now. If it attracts a large following, we'll definitely add it.

idolize:
@gaearon No problem! I actually closed this issue myself because I realized the real-world example already use an `isFetching` concept that largely eliminates the need for this library. I totally understand your desire to keep the Ecosystem section small and focused, and this library was already added to [awesome-redux](https://github.com/xgrommx/awesome-redux) anyway.

Keep up the good work!

PR #519, opened 2015-08-14
compatible run scripts for Win/NodeJS, and separate build/test examples

mikekidder:
This allows building examples with **Node 12.x** in windows.  
Also separates into "test:examples", and "build:examples" scripts

This leaves just running the tests exclusive to **io.js** due to **jsdom 4** dependency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,4 @@ node_js:
 script:
   - npm test
   - npm run build:examples
-
+  - npm run test:examples
--- a/examples/testAll.js
+++ b/examples/testAll.js
@@ -12,8 +12,7 @@ var exampleDirs = fs.readdirSync(__dirname).filter((file) => {
 
 // Ordering is important here. `npm install` must come first.
 var cmdArgs = [
-  { cmd: 'npm', args: ['install'] },
-  { cmd: 'webpack', args: ['index.js'] }
+  { cmd: 'npm', args: ['test'] }
 ];
 
 for (let dir of exampleDirs) {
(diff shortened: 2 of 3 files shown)


examples/buildAll.js
```diff
@@ -13,20 +13,23 @@ var exampleDirs = fs.readdirSync(__dirname).filter((file) => {
 // Ordering is important here. `npm install` must come first.
 var cmdArgs = [
   { cmd: 'npm', args: ['install'] },
-  { cmd: 'webpack', args: ['index.js'] },
-  { cmd: 'npm', args: ['test'] }
```
gaearon: Why is this removed?

gaearon:
I think, if you move out `test` from `buildAll`, you need to add `test:examples` explicitly into [`.travis.yml`](https://github.com/rackt/redux/blob/master/.travis.yml#L6).

mikekidder:
Yes, added.  Sorry about that, haven't played with Travis CI much yet.
Once Node v4 is out, some this compatible stuff will be resolved (hoping).

gaearon:
Should we also include `"0.12"` in `node_js` in `.travis.yml` now?
https://github.com/mikekidder/redux/blob/buildAll.forWinNodeJS/.travis.yml#L2

examples/buildAll.js
mikekidder: webpack still there.  separate file for testAll.js.  Otherwise, the test exception exits the script early under Node 12.x, and all examples will not be built.  Temporary until Node v4 released.

gaearon: Got it.

mikekidder:
Test scripts under Node 0.12 will still doesnt support the ES6 testing, so fails.  Node v4 will see the merging of io.js 3.x and node 12.x.  yea!

gaearon:
Got it.

Issue #525, opened 2015-08-15
process is not defined


alinz:
Hello,
First of all I would like to give you a big thank.

So, I had no problem with 1.0.0-rc until I removed the node_modules and install everything again. what I found was red death screen which complains about `process is not defined`. At the first, I thought I missed something and I started debugging the code. 

Here's the code which causes the issue.

```
const createStoreWithMiddleware = applyMiddleware(thunk)(createStore);
const reducer = combineReducers(reducers);
const store = createStoreWithMiddleware(reducer);
```

`combineReducers` method cause the whole thing to crash. I have not changed any code on my side.

So has anyone have this problem? Have you guys pushed code into 1.0.0-rc branch? It used to work.

chentsulin:
I believe this is because react-native's JavaScriptCore env doesn't have v8's `process` object. So we can use `__DEV__` instead.

And after some moment if this project will only need to support react-native version which include https://github.com/facebook/react-native/pull/2279 polyfill. Then, we can drop all of `__DEV__`.

gaearon:
Argh, this is really frustrating. Second time I'm burned by this.

gaearon:
Should be fixed in 1.0.1.

alinz:
Awesome, Thank you @gaearon!

PR #526, opened 2015-08-15
fix `process is not defined` in react-native

chentsulin:
`process` doesn't exist in react-native's JavaScriptCore environment, so we should do something like react-redux connect does now.

https://github.com/gaearon/react-redux/blob/master/src/components/createConnect.js#L192

### DIFF
--- a/src/utils/combineReducers.js
+++ b/src/utils/combineReducers.js
@@ -105,9 +105,20 @@ export default function combineReducers(reducers) {
       return newState;
     });
 
-    if (process.env.NODE_ENV !== 'production' && !stateShapeVerified) {
-      verifyStateShape(state, finalState);
-      stateShapeVerified = true;
+    if ((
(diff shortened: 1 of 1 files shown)


chentsulin:
#525

gaearon:
Thanks a lot! Out in 1.0.1.

vincentfretin:
Just for reference, I guess it's fixed in react-native too:
https://github.com/facebook/react-native/releases/tag/v0.10.0-rc lists the following:
process.env.NODE_ENV: This value is defined to be "production" or "development" based on your current environment. Now you can use npm packages that rely on NODE_ENV!

grabbou:
Yep, I think this can be simplified now back to the `NODE_ENV`

gaearon:
@grabbou 

It already is: https://github.com/rackt/redux/blob/09cd16b4ff2117a9bce84d81a7d5ed010bba9540/src/utils/combineReducers.js#L128

PR #529, opened 2015-08-15
Find next todo ID correctly. Fixes #524

### LINKED ISSUE #524: Error in count of TodoMVC example
I'd love to send a PR for this one since I'm just getting my feet wet but it might take me a bit so I'll defer to people smarter :)

To reproduce:
1. Load TodoMVC example
2. Add a few items to your todo list. Note that the counter properly increments (mine is at 4)
3. Mark one item as complete. Note that the counter properly decrements (mine is at 3)
4. Toggle the view to show only completed tasks. Note the counter still correctly shows the number of items left on your list. (mine is at 3)
5. Click "clear completed." You should still have the correct number of items displaying. (mine is at 3)
6. Go back to the "all" view. Add a new item to the list.

Expected result:
1. The list should grow in length with the item I just added at the top. The counter of "items left" should increment by one

Actual result
1. My item is added to the top but the last item on the list disappears. The counter of "items left" has incremented by one but only three items are visible. image
2. Adding new items after this will grow the list as expected but the Todo that disappeared will remain gone. The counter of "items left" still accounts for the missing Todo.

There's some additional funk on the counter that occurs if you remove an item from the list while it's in this incremented state. I can file more of a bug report if needed but I have a hunch that resolving this problem will resolve the latter issue.

gaearon:


docs/recipes/WritingTests.md
```diff
@@ -79,7 +79,7 @@ export default function todos(state = initialState, action) {
   switch (action.type) {
   case ADD_TODO:
     return [{
-      id: state.length,
+      id: state.reduce((maxId, todo) => Math.max(todo.id, maxId), -1) + 1,
```
vedvrat13: Hi @gaearon, Hi @acdlite,

I would like to know that how this particular line of code works.
I am struggling to understand that from where **maxId** has come and how the ids are getting generated.

Please help me with this. I am new to redux and wish to use it in my current project.

Thanks in advance.

vedvrat13: When maxId is not declared then it should have the value as **undefined**
however its getting the value as **null** ?

gaearon: Initially `maxId` comes from the seed argument to the `reduce(fn, seed)`. In this case the seed is `-1`. Then, inside the `reduce

PR #530, opened 2015-08-15
Changed below to above in docs/api/README.md

edvinerikson:
Not quite sure that this is correct but it do make more sense to me.

Changed `below` to `above` in `Every function described below is a top-level export. You can import any of them like this:`


gaearon:
:+1:

PR #531, opened 2015-08-15
add posts to combineReducers call

mjw56:
Should the posts reducer also be passed to combineReducers?


mjw56:
Ok nevermind, I now see now how it's being used privately :+1:

PR #536, opened 2015-08-16
Add browserify examples

ForbesLindesay:
I've created a copy of each example using browserify instead of webpack.  I think many people will find these useful.

TODO: implement hot reloading as part of browserify-middleware


gaearon:
Hmm, do we want to support two copies of each example just because of build system? Maybe it's better to publish them separately?

gaearon:
I'm closing because it's too painful to maintain two versions of everything.
Please feel free to create a separate repo!

Issue #301, opened 2015-07-22
Subscribing to changes from the store doesn't always play nice with hot-reloading


mvestergaard:
When using something like Rx observables to subscribe to changes in the store, it appears that hot-reloading can sometimes cause the subscriptions to be hooked up again, which can cause duplicate events.

gaearon:
Yeah good point. Need to figure this out before shipping 1.0.

gaearon:
Meh, this won't fit into 1.0.
Let's consolidate this with similar issues in #350.

gaearon:
Now we encourage people to use hot module replacement API directly, so that you have full control: https://github.com/rackt/redux/pull/667

PR #537, opened 2015-08-16
Pass store as prop to Root component in real-world example

ForbesLindesay:
The advantage of passing the store in from the top level is essentially the same as the advantage of passing the History object in from the top level:

It makes it easier to re-use the Root component in a universal app.


gaearon:
Unfortunately this breaks hot reloading reducers until we fix the underlying problem. #301 #350

gaearon:
This is superseded by https://github.com/rackt/redux/pull/667, where I removed `<Root>`.
Thanks!

PR #554, opened 2015-08-17
added pdf generation and link in readme

kolodny:
A quick note, `gitbook pdf` needs to have [calibre installed](http://calibre-ebook.com/)


paulwittmann:
seems we had the same idea :) (https://github.com/rackt/redux/pull/555)
let the maintainers decide what solution we should go for. should definitely add ePub and MOBI as well for mobile devices.

gaearon:
Thanks! For now, I don't want to bother installing Calibre etc, so I'll delegate PDF to @paulwittmann: #555. I'll keep this open for posterity and maybe will merge later.

gaearon:
I'll close for now because I'm too lazy to install calibre, lol. [redux-offline-docs](https://github.com/paulwittmann/redux-offline-docs) are well maintained so we'll keep referring to them.

PR #559, opened 2015-08-17
Add section on unit testing middleware to docs

idolize:
Updated `WritingTests.md` with a section on how to test middleware.

The function signature for middlware is a little confusing, so I think this could add a lot of value.

### DIFF
--- a/docs/recipes/WritingTests.md
+++ b/docs/recipes/WritingTests.md
@@ -140,6 +140,55 @@ describe('todos reducer', () => {
   });
 ```
 
+### Middleware
+
+Middleware functions wrap behavior of `dispatch` calls in Redux, so to test this modified behavior we need to mock the behavior of the `dispatch` call.
+
+#### Example
+
+```js
+import expect from 'expect';
+import * as types from '../../constants/ActionTypes';
+import singleDispatch from '../../middleware/singleDispatch';
+
+const fakeStore = fakeData => ({
+  getState() {
+    return fakeData;
+  }
+});
+
+const dispatchWithStoreOf = (storeData, action) => {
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks

gsklee:
Can we expand upon this part a little bit more? Like what's inside `singleDispatch.js` and an example `actions.js` showing how the `singleDispatch` middleware is being used in the code.

tristanls:
What is `singleDispatch.js`? This example is extremely confusing. Is this fictional middleware? Am I supposed to know how to implement this module? The string `singleDispatch` only occurs in the doc file and nowhere else in this repo.

markerikson:
Yeah, it appears to be a fictional middleware that is referenced in the unit test examples, but not actually shown anywhere.

tristanls:
I ended up using [redux-mock-store](https://github.com/arnaudbenard/redux-mock-store) for testing dispatch.

IndependentContractor:
Has anyone figured out how to implement `singleDispatch` or what it might look like?

brigand:
@IndependentContractor my guess is that it'd either be one dispatch ever, or one dispatch per action type.

```js
let dispatched = true;
singleDispatch = store)=> next => action => {
  if (dispatched) return;
  dispatched = true;
  next(action);
}
```

```js
let dispatchedTypes = new Map();
singleDispatch = store)=> next => action => {
  if (dispatchedTypes.get(action.type)) return;
  dispatchedTypes.set(action.type, true);
  next(action);
}
```

brigand:
I'm going to take a stab at improving these docs.

brigand:
#2496

Issue #552, opened 2015-08-17
Elm and pattern matching


joaomilho:
I'm not sure what is stated here http://rackt.github.io/redux/docs/introduction/PriorArt.html is correct. It says "Unlike Redux, Elm is a language, so it is able to benefit from static typing for actions, and pattern matching.". But Elm doesn't have pattern matching – like in Haskell – apart from the "case" construct. To be specific it supports touples matching on functions.

Do you think it's the case to keep this assertion there?

gaearon:
What I was trying to convey, is that in Elm, actions are typed. Can you edit to better reflect that?

gaearon:
I'm closing but feel free to submit a PR to make this sentence better.

PR #561, opened 2015-08-17
Improves explanation about Elm

joaomilho:
So here's my humble suggestion to #552 .


joaomilho:
[LGTM](http://lgtm.in/i/u9V7Rzipq)

PR #564, opened 2015-08-18
Require `type` to be specified on actions

### LINKED ISSUE #541: suggestion, don't allow store to dispatch action.type of undefined or null
I've haven't given this a lot of deep though so I'll probably someday want this post erased from the internet, but a few times already I've screwed up my 'constants', 

You can see in the below example, if I misspell ACCRUE either here or in the constants file (I'm not even sure if its spelled correctly here), it becomes undefined and then the reducer happily uses the default switch, and because store.dispatch is inclined to put whatever it pleases (@@init/etc) through the default path, I can't use that to catch screw ups.  Why not throw an exception when a falsey action.type is dispatched, since its so easy to do and sometimes difficult to catch, even in tests.

```
     switch ( action.type ) {                                                                                                                        
       case constants.AUTH_TOKEN:                                                                                                                    
         return authenticate(state, action)                                                                                                          
       case constants.ACCRUE:                                                                                                                        
         return accrue(state, action)                                                                                                                
       default:                                                                                                                                      
         return state                                                                                                                                
     }   

```

danmartinez101:
Fixes #541

### DIFF
--- a/docs/basics/Actions.md
+++ b/docs/basics/Actions.md
@@ -13,7 +13,7 @@ Here’s an example action which represents adding a new todo item:
 }
 ```
 
-Actions are plain JavaScript objects. By convention, actions should have a string `type` field that indicates the type of action being performed. Types should typically be defined as string constants. Once your app is large enough, you may want to move them into a separate module.
+Actions are plain JavaScript objects. Actions must have a `type` property that indicates the type of action being performed. Types should typically be defined as string constants. Once your app is large enough, you may want to move them into a separate module.
 
 ```js
 import { ADD_TODO, REMOVE_TODO } from '../actionTypes';
@@ -21,7 +21,7 @@ import { ADD_TODO, REMOVE_TODO } from '../actionTypes';
 
 >##### Note on Boilerplate
 
->You don’t have to define action type constants in a separate file, or even to define them at all. For a small project, it might be easier to just use string literals for action types. However, there are some benefits to explicitly declaring constants in larger codebases. Read [Reducing Boilerplate](../recipes/ReducingBoilerplate.md) for more practical tips on keeping your codebase clean. 
+>You don’t have to define action type constants in a separate file, or even to define them at all. For a small project, it might be easier to just use string literals for action types. However, there are some benefits to explicitly declaring constants in larger codebases. Read [Reducing Boilerplate](../recipes/ReducingBoilerplate.md) for more practical tips on keeping your codebase clean.
 
 Other than `type`, the structure of an action object is really up to you. If you’re interested, check out [Flux Standard Action](https://github.com/acdlite/flux-standard-action) for recommendations on how actions should be constructed.
 
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -94,6 +94,10 @@ export default function createStore(reducer, initialState) {
       throw new Error('Actions must be plain objects. Use custom middleware for async actions.');
     }
 
+    if (!action.type) {
+      throw new Error('Actions must specify a `type`.');
+    }
+
     if (isDispatching) {
       throw new Error('Reducers may not dispatch actions.');
     }
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -1,6 +1,6 @@
 import expect from 'expect';
 import { createStore, combineReducers } from '../src/index';
-import { addTodo, dispatchInMiddle, throwError } from './helpers/actionCreators';
+import { addTodo, dispatchInMiddle, throwError, unknownAction } from './helpers/actionCreators';
 import * as reducers from './helpers/reducers';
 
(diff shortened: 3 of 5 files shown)


gaearon:
@acdlite @johanneslumpe @goatslacker 

What do you think? Should we do this?

goatslacker:
I like the requirement, the error message is probably confusing though if you want to let the dev know that they are probably not passing in the correct action.

danmartinez101:
@goatslacker Can you tell me what about the message is confusing? 
I'll update it to be more clear

edge:
It could introduce an unnecessary bar in the future when we might develop use cases for Reducers that don't rely on type.

acdlite:
I could go either way on this. The only semi-reasonable situation I can think of where this might get in the way is when using non-plain-object actions, like Immutable.js Maps, in which case `action.type` would be undefined (but perhaps `action.get('type')` is not).

johanneslumpe:
I feel that this would allow us to further standardize the actions we accept and their structure. @acdlite how feasible is using immutable for actions over plain objects? Would it be possible for those people to just tack on a `type` property into their immutable object? A normal prop, not something that needs to be accessed with `get()`.

This change definitely catches errors where one might destructure the wrong action type from a constants file and else would have a hard time noticing why things break (typos etc.). I' learning towards this.

edge:
@johanneslumpe Couldn't that be implemented as middleware? If you want to throw errors if the action doesn't have a type, you could write a middleware that does that instead of putting in the core.

`type` might also be a reserved keyword in some language we haven't yet considered. In the core, it's only used for `'@@INIT'` and an debugging function.

It's also already on @acdlite's FSA spec, which a lot of people seem to be using.

johanneslumpe:
@edge That spec is exactly why I think having it in the core is a good thing - it furthers the standardization of the action shape. It is definitely possible to write a middleware which does the checking, no arguing there. But why force the user to do this type of check on their own, when it can happily live in the core and provide a benefit for all?

In terms of compile-to languages: I'd assume that you can put anything into a hash. If something would prevent you from adding a certain string as a key that would be kind of weird, no?

gaearon:
>It could introduce an unnecessary bar in the future when we might develop use cases for Reducers that don't rely on type.

Why? Even if reducers don't look at type (e.g. "entities" reducer in "real-world" example folder), the actions still need types for logging and cases when *later* you might want to add a reducer handling this specific type.

gaearon:
>The only semi-reasonable situation I can think of where this might get in the way is when using non-plain-object actions, like Immutable.js Maps

We don't allow them anyway. Currently we force actions to be plain objects.

gaearon:
>Couldn't that be implemented as middleware?

The most common mistake we want to catch is typo in constant name (and thus undefined `type`). I think we'll catch many more such mistakes if this was included in core instead of being opt-in.

acdlite:
> We don't allow them anyway. Currently we force actions to be plain objects.

@gaearon Forgot about that. If we're already making that requirement, then it this additional requirement seems only natural.

conundrumer:
@gajus would have complaints, given [his standard of using 'name' instead of 'type'](https://github.com/gajus/canonical-reducer-composition#action)

gajus:
I agree with @edge that this belongs be in the middleware.

@conundrumer Yes, but there is a way to use Flux Standard Action (i.e. "type" property) with Canonical Reducer Composition, https://github.com/gajus/canonical-reducer-composition#flux-standard-action. [redux-convention](https://github.com/gajus/redux-convention) is used to convert actions from FSA to CCA and vice-versa. In theory, if such a requirement were imposed, you could use the same middleware to workaround it, i.e. convert the action back to FSA at the end of the middleware store.

On the other hard, I don't agree that "type" is the right name for the property to begin with.

gaearon:
>On the other hard, I don't agree that "type" is the right name for the property to begin with.

Maybe it wasn't, but since this is pretty much a convention at this point, there needs to be a *very* compelling reason to break everyone by changing the property name. However, some property name *needs* to be enforced because there is much debugging benefit to preventing “bad constant import” errors early. Therefore we converge on `type` and enforce it.

gaearon:
Rebased and merged as https://github.com/rackt/redux/commit/71393b7036e26c8bb6cd6f1fc5b2a46df5560058 and https://github.com/rackt/redux/commit/87a9b26aebc43afa379615ac238c2f95d863d432.

silvenon:
What would be the recommended type of API actions in the [real-world example](https://github.com/rackt/redux/blob/master/examples/real-world/actions/index.js) then? Since now it throws an error.

gaearon:
What error? I will double check but examples have worked fine for a long period of time since this merge. The types are specified by the API middleware that interprets these actions.

silvenon:
My bad, my API middleware wasn't plugged in :sweat:

PR #574, opened 2015-08-18
removed resolve to src

jstrimpel:
Looks like someone was reading the redux libs from a local `src` dir, which does not exist. `npm install` and running the app now works.

### DIFF
--- a/examples/todomvc/webpack.config.js
+++ b/examples/todomvc/webpack.config.js
@@ -17,12 +17,6 @@ module.exports = {
     new webpack.HotModuleReplacementPlugin(),
     new webpack.NoErrorsPlugin()
   ],
-  resolve: {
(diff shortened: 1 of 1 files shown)


gaearon:
http://rackt.github.io/redux/docs/introduction/Examples.html explicitly says you need to run `npm install` *both* in the root folder and in the examples folders. This is for easier contributions and bugfixes to Redux itself.

We might want to reconsider this now that 1.0 is out and API is stable-ish.

jstrimpel:
Ah. I missed that. I didn't come into the example from the docs - actually can't remember how I stumbled across it.

gaearon:
Maybe you could copy these sections and create READMEs for each individual example explaining how to run it? That would be really helpful.

PR #578, opened 2015-08-19
Small error in _App.js_

tcoopman:


gaearon:
Did this instead: https://github.com/rackt/redux/commit/68545444b36bada650b7ad70416eea7e4a476f79

PR #583, opened 2015-08-19
Change store designation

Bernardstanislas:
Surely looks like useless, but I got confused about how many `stores` redux involves, that's why I suggest the change, to be clearer.


gaearon:
Fair!

PR #590, opened 2015-08-20
Don't recommend the use of `import * as reducers`.

ghengeveld:
Replaced the suggestion to use `import * as reducers` with the recommendation to use `reducers/index.js` and call `combineReducers` there. Updated the code examples accordingly.

As discussed here: https://github.com/rackt/redux/pull/473

### DIFF
--- a/docs/api/combineReducers.md
+++ b/docs/api/combineReducers.md
@@ -82,13 +82,7 @@ export default combineReducers({
 
 ```js
 import { createStore, combineReducers } from 'redux';
-
 import reducer from './reducers/index';
-console.log(reducers);
-// {
-//   todos: Function,
-//   counter: Function
-// }
 
 let store = createStore(reducer);
 console.log(store.getState());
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks!

I suggest two tweaks:

1. 

>we now recommend the use of `reducers/index.js` with a call to combineReducers.

I'd change “the use of `reducers/index.js`” to something like “exporting a single reducer obtained using `combineReducers()` from `reducers/index.js`”.

2. 

Let's move `index.js` *after* `todos` and `counter` are already declared so it's easier to trace what's going on.

ghengeveld:
Done.

docs/api/combineReducers.md
```diff
@@ -59,19 +65,31 @@ export function counter(state = 0, action) {
...
+```
+
 #### `App.js`
 
 ```js
 import { createStore, combineReducers } from 'redux';
 
-import * as reducers from './reducers';
+import reducer from './reducers/index';
 console.log(reducers);
```
gaearon: This `console.log` now can be removed

PR #593, opened 2015-08-20
Minor typo fix.

vasanthk:
Accidental semicolon. Use fullstop instead.


gaearon:
This is [intentional](https://www.noslangues-ourlanguages.gc.ca/bien-well/fra-eng/ponctuation-punctuation/puces-bulleted-eng.html). The correct fix would be to change these sentence fragments to start lowercase.

vasanthk:
Good to know. Never knew these rules.

PR #600, opened 2015-08-21
Minor docs fix: add UglifyJS link and style fix

camsong:


docs/api/compose.md
```diff
@@ -29,7 +29,7 @@ let finalCreateStore;
 
 // In production, we want to use just the middleware.
 // In development, we want to use some store enhancers from redux-devtools.
-// UglifyJS will eliminate the dead code depending on the build environment.
+// [UglifyJS](https://github.com/mishoo/UglifyJS2) will eliminate the dead code depending on the build environment.
```
camsong: Here may also need to describe what is UglifyJS and why need it.

gaearon: I don't think this is necessary. This makes comment harder to read, and we don't really care which version of Uglify they'd be using.

gaearon:
As I wrote above, I don't really see the need for this.
Thanks!

PR #607, opened 2015-08-22
Explain unit testing decorated components.

ghengeveld:
Also improved test setup a bit using a babelhook file rather than a command line option, for more flexibility.

### DIFF
--- a/docs/recipes/WritingTests.md
+++ b/docs/recipes/WritingTests.md
@@ -18,14 +18,22 @@ To use it together with [Babel](http://babeljs.io), add this to `scripts` in you
   ...
   "scripts": {
     ...
-    "test": "mocha --compilers js:babel/register --recursive",
+    "test": "mocha --require babelhook --recursive",
     "test:watch": "npm test -- --watch",
   },
   ...
 }
 ```
 
-and run `npm test` to run it once, or `npm run test:watch` to test on every file change.
+Then create a file named `babelhook.js` beside it:
+
+```js
+require('babel/register')({
+  // for options see https://babeljs.io/docs/usage/require/
+});
+```
+
+Run `npm test` to run it once, or `npm run test:watch` to test on every file change.
 
 ### Action Creators
 
--- a/docs/recipes/WritingTests.md
+++ b/docs/recipes/WritingTests.md
@@ -267,6 +267,40 @@ export default function jsdomReact() {
 
 Call it before running any component tests. Note this is a dirty workaround, and it can be removed once [facebook/react#4019](https://github.com/facebook/react/issues/4019) is fixed.
 
+#### Testing decorated React components
+
+In order to achieve separation of concerns and create reusable components, we often wrap one component inside another using decorators. For example, consider the `App` component:
+
+```js
+class App extends Component { /* ... */ }
+export default connect(select)(App);
+```
+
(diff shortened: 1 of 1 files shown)


gaearon:
Wouldn't `babel/register` use your `.babelrc`? I think it would, and IMO you should keep all your options in `.babelrc` anyway so they're centralized.

eugene1g:
Exporting a `default` and a named module is a great approach, however there is a caveat that it only works if the consuming side loads the Redux component via ES6 `import` and not via CommonJS `require`.

Babel handles the interchangeable use of `import` and `require` through its [interop](http://babeljs.io/docs/usage/modules/#interop) capability to run two module formats side-by-side, but the behaviour is [slightly different](https://github.com/babel/babel/issues/2047). Essentially, if all you have is `export default connect(...)` then babel overrides `module.exports` for you so that CJS `var App = require('./App.js')` gets you the default decorated component. As soon as you add a named export, the `require` fails and you need to change the consuming CJS side to be `var app = require("./app.js").default`. See a [sample example](https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=true&spec=false&code=function%20impressiveComponent()%7B%7D%3B%0Afunction%20cleanSelector()%7B%7D%3B%0A%0Aexport%20default%20connect(cleanSelector)(impressiveComponent)%3B%0A%0A%2F%2F%20Uncomment%20next%20line%20to%20add%20a%20named%20export%0A%2F%2F%20export%20var%20rawComponent%20%3D%20impressiveComponent%3B) In the upcoming Babel V6, handling of the `default` export [might change](https://github.com/babel/babel/issues/2212) so that CJS code will *always* have to use `require("./App.js").default` so that the behaviour is normalised, but for now it remains a gotcha.

The reason why it's relevant is that people might be using ES6 modules + Babel to build their front-end code, but still use CommonJS format in testing or server-side code/rendering. So the same component might be consumed in several places through different module systems. Right now in their CJS side `var App = require("./App.js")` works fine, but if they add a named export inside App.js then consumers will break and will need to be updated to `var App = require("./App.js").default`

ES6 is the clear winner long term and the suggested testing technique is very valuable, however there should be a note warning people that if they use CommonJS to require ES6 Redux smart components, then they should explicitely access it via `require('SmartApp.js').default` even if all they have is a default module. This is not idiomatic CommonJS, but it should continue working in the future even if Babel changes its behaviour or the module starts exporting other named piece of code.

ghengeveld:
@gaearon Fair point. I've reverted this change.
@eugene1g Thanks for your feedback. I have no experience with mixing module systems, so I'm happy you provided such clear explanation. I'll include a note on this in the docs.

gaearon:
This is in, thanks

ghengeveld:
Yesterday I had a chat with someone suggesting to simply unwrap the decorated component instead. He suggested using something like this to recursively unwrap it and use that in unit tests:

```js
function unwrap(component) {
  return component.wrapped ? unwrap(component.wrapped) : component;
}
```

I haven't tried this approach myself yet, but it seems a bit cleaner than having multiple exports, especially since you don't have to change your application sources to support your unit tests. How do you guys feel about this?

gaearon:
That would need every HOC to converge on the same property name. In reality I think it works worse than just exporting vanilla stuff.

PR #608, opened 2015-08-23
applyMiddleware.js can be better

Lucifier129:
The `next` parameter of the function that applyMiddleware return, should be `createStore`.


timdorr:
It doesn't have to be `createStore`:

```js
  applyMiddleware(thunk),
  devTools(),
)(createStore);
```

timdorr:
Other than aesthetic changes, this could result in someone misinterpreting the flexibility of `applyMiddleware`.

PR #611, opened 2015-08-24
Use lodash v4

gaearon:
Sending a PR on behalf of @jdalton :-)
Is there anything else that needs to be done here?


salehe:
We are **actually waiting** for Lodash 4 to be released, making second task done! :stuck_out_tongue_winking_eye:

jdalton:
Thanks for getting this rolling @gaearon!

I'll add some comments inline to talk through some of the changes.

package.json
```diff
@@ -66,6 +66,9 @@
     "webpack": "^1.9.6",
     "webpack-dev-server": "^1.8.2"
   },
+  "dependencies": {
+    "lodash": "~4.0.0"
```
jdalton: Using the `~` because we're referencing internal modules. We try to avoid issues with internal modules for the `~` range.

webpack.config.base.js
```diff
@@ -1,18 +1,35 @@
...
 
 module.exports = {
   module: {
-    loaders: [
-      { test: /\.js$/, loaders: ['babel-loader'], exclude: /node_modules/ }
-    ]
+    loaders: [{
+      test: /\.js$/,
+      loaders: ['babel-loader'],
+      include: [
```
jdalton: webpack recommends using `include` over `exclude` when possible.

webpack.config.base.js
```diff
@@ -1,18 +1,35 @@
...
   },
   output: {
     library: 'Redux',
     libraryTarget: 'umd'
   },
-  resolve: {
-    extensions: ['', '.js']
-  }
+  plugins: [
+    ['lodash/internal/baseIteratee.js', './toFunction.js'],
```
jdalton: The biggest reduction in size is from replacing `baseInteratee` references with `toFunction` to cut out the callback shorthand helpers (e.g. `_.mapValues(objects, 'propName')`).

webpack.config.base.js
```diff
@@ -1,18 +1,35 @@
...
+    ['lodash/object/keys.js',           '../internal/baseKeys.js'],
```
jdalton: Replacing `keys` and `keysIn` with their base forms since yall aren't concerned with treating sparse arrays as dense or compat paths for older enviros.

webpack.config.development.js
```diff
@@ -1,14 +1,14 @@
...
-var config = Object.create(baseConfig);
-config.plugins = [
-  new webpack.optimize.OccurenceOrderPlugin(),
-  new webpack.DefinePlugin({
-    'process.env.NODE_ENV': JSON.stringify('development')
-  })
-];
-
-module.exports = config;
+module.exports = _.merge({}, baseConfig, {
```
jdalton: Using `_.merge` here instead of `Object.create(baseConfig)` because the `Object.create` route doesn't clone object references so changes to them change the `baseConfig` properties.

webpack.config.production.js
```diff
@@ -1,20 +1,23 @@
...
-
-module.exports = config;
+module.exports = _.merge({}, baseConfig, {
+  plugins: _.union(baseConfig.plugins, [
+    new webpack.DefinePlugin({
+      '__DEV__': 'false',
+      'process.env.NODE_ENV': '"production"'
+    }),
+    new webpack.optimize.UglifyJsPlugin({
+      compressor: {
```
jdalton: Added more uglify options to crunch it more :)

webpack.config.base.js
```diff
@@ -1,18 +1,35 @@
...
+    ['lodash/object/keysIn.js',         '../internal/baseKeysIn.js']
+  ].map(function(pair) {
+    var resourceRegExp = RegExp('\\b' + _.escapeRegExp(pair[0]) + '$');
+    return new webpack.NormalModuleReplacementPlugin(resourceRegExp, pair[1]);
```
jdalton: For info on why `NormalModuleReplacementPlugin` instead of `resolve.alias` see this thead https://github.com/jdalton/redux/commit/7b50f3b7af340ab18663de4ca240a912c6d0a303#commitcomment-12833016

webpack.config.base.js
```diff
@@ -1,18 +1,35 @@
...
   module: {
-    loaders: [
-      { test: /\.js$/, loaders: ['babel-loader'], exclude: /node_modules/ }
-    ]
+    loaders: [{
+      test: /\.js$/,
+      loaders: ['babel-loader'],
+      include: [
+        path.resolve(__dirname, 'src'),
+        path.resolve(__dirname, 'node_modules', 'lodash')
```
aaronjensen: :question: Is this necessary because of `NormalModuleReplacementPlugin` below?

webpack.config.development.js
gaearon: :+1:

webpack.config.production.js
gaearon: Oh good. I was always curious what you're referred to in some older tweet about not going all the way with Uglify. :-)

webpack.config.base.js
jdalton: > :question: Is this necessary because of NormalModuleReplacementPlugin below?

I believe so. We can try removing lodash's path from it to see if it still works.

jdalton: Tested. It can safely be reverted to:
```js
loaders: [
  { test: /\.js$/, loaders: ['babel-loader'], exclude: /node_modules/ }
]
```

webpack.config.base.js
```diff
@@ -1,18 +1,35 @@
...
+      ]
+    }]
+  },
+  node: {
+    process: false
   },
```
jdalton: `node: { process: false }`  prevents webpack from including the `process` shim because of the simple inference check in `lib/utils/combineReducers`.

webpack.config.production.js
```diff
@@ -1,20 +1,23 @@
...
-  })
-];
-
-module.exports = config;
+module.exports = _.merge({}, baseConfig, {
+  plugins: _.union(baseConfig.plugins, [
+    new webpack.DefinePlugin({
+      '__DEV__': 'false',
+      'process.env.NODE_ENV': '"production"'
+    }),
```
jdalton: Using `DefinePlugin` for `__DEV__` as well as `process.env.NODE_ENV` removes the use of `verifyStateShape` in `lib/utils/combineReducers` as dead code.

jdalton:
@gaearon
> - [ ] Update “no dependencies” notice in README with something sensible

There's no need to update the notice since lodash methods will be built-in so still no external deps required.

gaearon:
Fair enough! I guess technically it looks like an (inner) dependency, but I'll just point people to this thread. :-)

jdalton:
Instructions to build lodash v4.

* Clone `lodash-cli` master and install master deps
```bash
git clone --depth 1 https://github.com/lodash/lodash-cli
cd lodash-cli && mkdir node_modules && cd node_modules
git clone --depth 1 https://github.com/lodash/lodash
cd .. && npm i && cd ..
```
* Build lodash dep for Redux
```bash
cd redux
mkdir node_modules && cd node_modules
node ../../lodash-cli/bin/lodash modularize exports=node -o ./lodash && node ../../lodash-cli/bin/lodash -d -o ./lodash/index.js
cd lodash && npm init

<go through wizard and ensure version is 4.0.0>

cd ../../ && npm i
```

gaearon:
Just to confirm.. we're still waiting for Lodash 4 to land right?

jdalton:
Correct.

gaearon:
Got it, thanks!

c2h5oh:
Lodash v4 is out :-)

https://github.com/lodash/lodash/commit/9a2748c1ead0308557a35ac47ab0b0f2f62db74e

jdalton:
> Lodash v4 is out :-)

Not yet, but close. It'll be published and tagged on Tues (Jan 12th)

calvinf:
Lodash 4: https://github.com/lodash/lodash/releases/tag/4.0.0

gaearon:
Would somebody like to revive this?

jdalton:
I'll take it up this weekend for sure.

gaearon:
Oh man, thank you. No rush.

jdalton:
Sorry, haven' gotten to this. Though I think it should wait to 4.0.1 for the release bugs to be shaken out.

webpack.config.base.js
wmertens: Why not add these commit notes as inline comments?

jdalton: Sounds good.

timdorr:
4.0.1 is out. Are we good to continue, do you think?

jdalton:
Yep, nows a good time with v4.1 out the private packages are flat too (not hidden away in a directory).
So like `'lodash/_toFunction'`.

gaearon:
FYI I think we might not need anything else except `isPlainObject`: https://github.com/rackt/redux/pull/1329

jdalton:
Updated.

jdalton:
With this in place it'll be trivial to use another lodash method when needed.
If needed later the `webpack.NormalModuleReplacementPlugin` was something like:
```js
  plugins: [
    ['lodash/_baseIteratee.js', './_toFunction.js'],
    ['lodash/keys.js',           './_baseKeys.js'],
    ['lodash/keysIn.js',         './_baseKeysIn.js']
  ].map(function(pair) {
    var resourceRegExp = RegExp('\\b' + _.escapeRegExp(pair[0]) + '$');
    return new webpack.NormalModuleReplacementPlugin(resourceRegExp, pair[1]);
  }).concat(
    new webpack.optimize.OccurenceOrderPlugin()
  )
```

webpack.config.base.js
```diff
@@ -1,5 +1,9 @@
 'use strict';
 
+var _ = require('lodash');
+var path = require('path');
```
timdorr: These lines can go. They're not used in this context.

timdorr:
Just so it's known (no judgement)
```
master
redux.min.js  5.73 kB

lodash
redux.min.js  5.85 kB
```
I'd rather spend 130 bytes to not have to maintain a standard function not critical to Redux. And if someone is already bundling in Lodash, then the net result is actually going to be negative for them.

I don't see any big issues with this PR. LGTM.

gaearon:
What is the post gzip difference?
(Just curious. I agree it's not significant.)

timdorr:
2062 vs 2153 bytes, so 91 bytes.

jdalton:
```bash
dist git(master):
2.05 kB
dist git(lodash): gzip-size redux.min.js| pretty-bytes
2.14 kB
```

webpack.config.development.js
```diff
@@ -1,14 +1,13 @@
...
+  plugins: _.union(baseConfig.plugins, [
```
aaronjensen: Just curious, but why use 'union' here? The plugins are all new objects, seems like basic concat would suffice.

gaearon:
:+1:

gaearon:
Interestingly, I have trouble trying to run tests on the UMD builds. If I change 

```diff
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -1,5 +1,5 @@
 import expect from 'expect'
-import { createStore, combineReducers } from '../src/index'
+import { createStore, combineReducers } from '../dist/redux'
```

I get 

```
/Users/dan/p/redux/dist/redux.js:411
				var objectProto = global.Object.prototype;
				                        ^

TypeError: Cannot read property 'Object' of undefined
    at Object.<anonymous> (/Users/dan/p/redux/dist/redux.js:405:26)
    at Object.exports.__esModule (/Users/dan/p/redux/dist/redux.js:467:31)
```

This is new with this PR. However I'm also able to run `require('./redux')` from `node` in `dist` folder so I'm confused what actually broke.

gaearon:
In tests, Mocha transpiles everything that isn't in `node_modules`. Previously Redux worked even when transpiled twice so I could manually test dist builds with Mocha. However after this PR it doesn't work when transpiled twice. In particular, `'use strict';` inserted by Babel breaks `function () {return this;}` check that Lodash does to find the global.

In general we don't want to encourage people to transpile Redux UMD build, but I wouldn't be surprised if some people do that. Feels like breaking them isn't a very nice thing to do.

phated:
@gaearon do you think this belongs over in the lodash issue tracker?  Something like "lodash breaks when transpiled twice"?  If so, would you mind opening it over there for tracking?  I know JD is keeping an eye on this thread but I figured it would be helpful.  Thanks.

gaearon:
Filed as https://github.com/lodash/lodash/issues/1916

cht8687:
If we only need a few methods, and if those methods are working fine within Redux, why do we need depend on another library which may cause troubles?

gaearon:
Because if we can avoid maintaining them, we want to.

gaearon:
I just force-pushed to master to not include this PR. Let's figure out a solution that doesn't break double transpiled builds. If we can't, I'm happy to bump major and ship, but that would be a breaking change.

jdalton:
@cht8687
> if those methods are working fine within Redux, why do we need depend on another library which may cause troubles?

It turns out DIY solutions were a burden on the project and caused support issues. Problems that were solved by lodash since it does utilities very well. Since lodash is a widely used utility libs it has more eyes on it and resolves more issues making it a solid choice when choosing whether to DIY or use lodash. Redux using lodash further feeds lodash's support chain as any issues it has in the future will pipe directly to lodash improving it for all other users.

@gaearon
> I just force-pushed to master to not include this PR. Let's figure out a solution that doesn't break double transpiled builds. If we can't, I'm happy to bump major and ship, but that would be a breaking change.

Can you describe the issue a bit more. The unit tests passed in this PR. If you liked it, then you shoulda put a unit test on it – Beyonce (probably)

gaearon:
>Can you describe the issue a bit more. The unit tests passed in this PR. If you liked it, then you shoulda put a unit test on it – Beyonce (probably)

Yeah, so this is something I tend to test manually because I don't have a good solution for doing that automatically. (I agree I should create an issue for that.)

I just changed imports inside tests to point to `../dist/redux` instead of `../src` to make sure that the UMD build didn't regress. The UMD builds themselves are fine, but I found another issue I described in https://github.com/lodash/lodash/issues/1916. 

People are supposed to use UMD builds “as they are” but I know some people have bad transpiler configs that transpile third party modules. So we end up in a situation where the UMD build of Redux *might* be transpiled with Babel by somebody else, and it used to work until now. I'm not saying I *want to support* this (I'm fine to break it in a major and declare we don't want people to transpile our already transpiled UMD builds) but it used to work. However now, due to Babel inserting `use strict` into modules it transpiles, Lodash breaks because it depends on non-strict behavior. At least this is my understanding of the issue.

gaearon:
So, to reproduce, you can:

Build a UMD build in dist folder
Go to http://babeljs.io/repl and transpile it, paste it back into the file
Expected:
3. UMD build still works despite being transpiled “again” (behavior in master)

Actual:
3. UMD build is broken due to crash in Lodash (behavior in this PR)

gaearon:
Let's continue discussion in https://github.com/lodash/lodash/issues/1916

gaearon:
@jdalton 

Thank you very much for helping me understand this issue and how global detection works in Lodash!
I wonder if there is any way you can publish an official version of 

```js

PR #613, opened 2015-08-24
add reddit example 

timuric:
It might be useful to have setup of the Reddit API tutorial in examples http://rackt.github.io/redux/docs/advanced/ExampleRedditAPI.html


gaearon:
It's there: https://github.com/rackt/redux/tree/master/examples/async

PR #616, opened 2015-08-24
real-world example - Use find instead of filter.

dbrans:


gaearon:
Fair, we're using the polyfill anyway. Thanks!

PR #622, opened 2015-08-25
API change proposal: listener() --> listener(action)

dherault:
Hi,

As our Redux project grows at our company, we have found useful to get the last dispatched action at the beginning of a `store.subscribe` like this:
```javascript
store.subscribe(() => {
    const { records } = store.getState(); // a reducer that holds every action ever dispatched
    const lastAction = records[records.length - 1];
```
But that approach is not reliable every time: if we wait too long before accessing `records[records.length - 1]` we might end up with another action dispatched in the meanwhile.

So we thought, in `createStore.js`, what about calling `listener` with the dispatched action as argument ?
```javascript
listeners.slice().forEach(listener => listener(action));
```
That would allow developers to use `store.subscribe()` as a way to create action-based side-effects, thus making Redux even more flexible.

Would anyone else like this ?

Thanks for the lib.

### DIFF
--- a/docs/api/Store.md
+++ b/docs/api/Store.md
@@ -89,7 +89,7 @@ To unsubscribe the change listener, invoke the function returned by `subscribe`.
 
 #### Arguments
 
-1. `listener` (*Function*): The callback to be invoked any time an action has been dispatched, and the state tree might have changed. You may call [`getState()`](#getState) inside this callback to read the current state tree. It is reasonable to expect that the store’s reducer is a pure function, so you may compare references to some deep path in the state tree to learn whether its value has changed.
+1. `listener(action)` (*Function*): The callback to be invoked any time an action has been dispatched, and the state tree might have changed. You may call [`getState()`](#getState) inside this callback to read the current state tree. It is reasonable to expect that the store’s reducer is a pure function, so you may compare references to some deep path in the state tree to learn whether its value has changed.
 
 ##### Returns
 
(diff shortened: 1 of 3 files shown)


acdlite:
This seems to make Redux less flexible, not more. People often ask why we don't call listeners with the most recent *state*, not *action*. This is an example of why — different people have different opinions about how this stuff should work.

The beauty of Redux is that you can customize this type of behavior yourself. Using a store enhancer:

```js
function subscribeWithAction(createStore) {
  return (reducer, initialState) => {
    const store = createStore(reducer, initialState);
    let listeners = [];
    let lastAction;
    store.listen(() => listeners.forEach(l => l(lastAction))

    return {
      dispatch: action => {
        lastAction = action;
      },
      subscribe: listener => {
        listeners.push(listener);
          listeners = listeners.filter(l => l !== listener);
        };
      }
    };
  };
}


// Use it like this
const finalCreateStore = subscribeWithLastAction(createStore);
const store = finalCreateStore(reducer, initialState);

store.subscribe(lastAction => {
  // Yay!
});
```

gaearon:
>But that approach is not reliable every time: if we wait too long before accessing `records[records.length - 1]` we might end up with another action dispatched in the meanwhile.

Can you show with a failing test how this happens? I don't understand.

mikekidder:
@acdlite that example.  Hope it makes it's way into the documentation.

gaearon:
To be honest I wouldn't encourage people to write store enhancers when a reducer is enough :-).
I'd rather want to understand what's wrong with just using `subscribe`—I couldn't reproduce the issue.

dherault:
@acdlite Thanks for the store enhancer that's something!

@gaearon We can't seem to reproduce that issue either -_-, sorry for the inconvenience.

PR #624, opened 2015-08-25
(Do not merge!) Remove getReducer and replaceReducer from public API

gaearon:
As discussed in #350, we want to get rid of them to simplify Store API, and express alternatives to them in userland.


gaearon:
Let's take it slow.
#668 is the first step. We'll revisit whether we really want to do this after we get some feedback on it.

PR #625, opened 2015-08-25
(Do not merge!) Discussion: A userland replacement for replaceReducer

gaearon:
Building on #624, here is a proposed replacement for `replaceReducer` API that can be implemented in userland. Note how we no longer require `<Root>` component. Refs #350.


examples/counter/store/switchReducer.js
```diff
@@ -0,0 +1,19 @@
+// Note: this is meant to be provided in a separate package.
+// I'm only including it here for discussion.
+
+const SWITCH_REDUCER = { type: 'redux-switch/SWITCH_REDUCER' };
```
acdlite: Any reason you've left off the `@@` at the beginning?

quicksnap: Leaving the `@@` may be because this is intended to be implemented in userland.

examples/counter/store/configureStore.js
```diff
@@ -1,11 +1,21 @@
...
+import { createStore, applyMiddleware, compose } from 'redux';
 import thunk from 'redux-thunk';
-import rootReducer from '../reducers';
+import reducer from '../reducers';
+import switchReducer from './switchReducer';
 
-const createStoreWithMiddleware = applyMiddleware(
-  thunk
-)(createStore);
+function onHotReload(useReducer) {
```
acdlite: Could the param for `onHotReload()` be the path to the reducer module? Or does that not work because the require statement can't be inside `node_modules`?

examples/counter/store/switchReducer.js
acdlite: I interpreted "userland" to mean another npm package, a la redux-thunk et al.

quicksnap:
Are there any drawbacks to this approach that come to mind? I really like this!

acdlite:
@quicksnap Only downside I can think of is there's slightly more set-up for the developer

examples/counter/store/switchReducer.js
gaearon: Ah, forgot about it. Yeah, it's probably better to leave it in because I can't imagine a siutation where you'd want to handle it.

examples/counter/store/configureStore.js
gaearon: Not sure I understand what you're proposing.

The problem is Webpack needs `module.hot.accept(name` call to be verbatim and statically analyzable (also true for `require` call inside).

Also, `switchReducer` argument isn't specific to hot reloading. It's just a function that invokes a callback whenever there's a new reducer available. This can be a result of hot reloading, code splitting, or even both.

gaearon:
Let's take it slow.
#668 is the first step. We'll revisit whether we really want to do this after we get some feedback on it.

PR #627, opened 2015-08-26
latest eslint related packages; full linting for redux

mikekidder:
NOTE:  The biggest changes for these package bumps were:

1) comma-dangles rule -- I turned off in .eslintrc
2) react sort component rule (set to 2) via **eslint-config-airbnb**
```
'react/sort-comp': [2, {         // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md
    'order': [
    'lifecycle',
    '/^on.+$/',
    '/^(get|set)(?!(InitialState$|DefaultProps$|ChildContext$)).+$/',
    'everything-else',
    '/^render.+$/',
    'render' ]
```
3) fixed/removed a couple of outstanding rule exceptions in .eslintrc


examples/async/containers/AsyncApp.js
```diff
@@ -50,7 +50,7 @@ class AsyncApp extends Component {
             </span>
           }
           {!isFetching &&
-            <a href='#'
+            <a href="#"
```
gaearon: It's a bit weird, are we using `"` for JSX strings but `'` for other strings? Why?

aphillipo:
I always use " in HTML and ' in Javascript by convention because generating HTML in Javascript if you use the same type of quotes looks messy.

var myhtml = '<a href="#">Something</a>';
var otherhtml = "<a href=\"#\">Somthing</a>";

I'd say this is obviously less of a concern is you are rendering in react but it's good to stick to linting conventions?  Just a thought.

mikekidder:
Yes.  Its another linting rule from the Airbnb guide.  Much like when you do a view source, your HTML rendered attributes are double quoted.

gaearon:
Oh well, whatevs. I actually liked `'` in JSX because it reminds be that “it's just JavaScript” and has nothing to do with HTML. I guess I'll [rant here](https://github.com/airbnb/javascript/issues/269#issuecomment-134990455) but merge this.

gaearon:
Thanks a lot. I know it's the kind of work nobody likes to do so huge :+1: from me.

mikekidder:
I'm glad you appreciate the help!  For me, its great as I don't do javascript programming everyday.  Awesome way for me to learn, while reading through code at the same time.

And first time for me to actually add/test/debug the React property types validations on components.  Cool stuff

hzoo:
Nice job @mikekidder!

goatslacker:
I think you can override Airbnb's eslint styles to fit your own, so if you really prefer `'` for jsx attributes, which is totally valid, you can just override that config but use the rest of it.

mikekidder:
Yes we can, doing it for the comma-dangles, which personally, I think is even more opinionated.

PR #631, opened 2015-08-26
Make usernames case insensitive in real-world example

rmadsen:
GitHub usernames are case insensitive. If you search 'redux' in the real-world example, the screen displays 'Loading redux's profile...' despite the fact that github's api returned a valid object for the username "Redux." This change now keys usernames in their lowercase form.

I discovered this when trying to follow along with the documentation. I'm new to react, redux, etc. so let me know if there is a cleaner way of fixing this!


gaearon:
Good call, but I'd rather do it by introducing `parseLogin(params)` and `parseFullName(params)` functions that would be used in route handlers to retrieve `login` and `fullName` from router parameters.

gaearon:
Closing for inactivity. Please feel free to open another PR addressing my comment above.

PR #636, opened 2015-08-27
Correct publicPath in webpack config

frodereinertsen:
Hot reloading was not working due to a config error as described here:
https://github.com/gaearon/react-hot-loader/blob/master/docs/Troubleshooting.md#syntax-error-unexpected-token-


timdorr:
Ah, thanks for that!

PR #638, opened 2015-08-27
Fix universal rendering example (configureStore vs. createStore)

eiriklv:
The example is using `createStore` in both the server and client script, instead of `store/configureStore.js` with the thunk middleware. Fixed this and now the async and conditional action works.


timdorr:
Oh, *that's* why it didn't work :)

eiriklv:
@timdorr It had me seriously tripping for a while :p

gaearon:
OK this is dumb of me but I had no idea we have `universal` example now.

:-PPPP

PR #639, opened 2015-08-27
Update Reducers.md

raineroviir:


gaearon:
This was actually correct: we want both of these actions to be handled by `todos`.

raineroviir:
Gotcha.   It seemed a little confusing to leave the empty case there so I wasn't sure if it was intentional

PR #640, opened 2015-08-27
Add universal to examples and update webpack config

timdorr:
This is to match the changes in #626, which happened before the universal example was merged, but after I had finished coding it.

Also, it adds the example to the Examples doc.

### DIFF
--- a/docs/introduction/Examples.md
+++ b/docs/introduction/Examples.md
@@ -64,6 +64,26 @@ It covers:
 * Caching responses and showing a spinner while data is fetching;
 * Invalidating the cached data.
 
+## Universal
+
+Run the [Universal](https://github.com/rackt/redux/tree/master/examples/universal) example:
+
+```
+git clone https://github.com/rackt/redux.git
+
+cd redux/examples/universal
+npm install
+npm start & npm run client
+
+open http://localhost:8080/
+```
+
+It covers:
+
+* [Universal rendering](/docs/recipes/ServerRendering.md) with Redux and React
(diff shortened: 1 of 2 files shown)


gaearon:
I think it's somewhat misleading that we serve the app at `8080` in this single example.

timdorr:
This one is different because it's not being handled by webpack. Hence a separate URL for it. You wouldn't normally run webpack-dev-server in production, so I didn't want to direct people to that port and think it's a workable app server.

gaearon:
Oh okay. But we should clarify this somewhere. I opened `3000` by habit and saw nothing. My console didn't show a port either so I was dumbfounded for a minute until I saw it uses `8080` in the code.

gaearon:
That's all I see when it's running:

screen shot 2015-08-27 at 21 35 23

I assumed it was waiting for something initially.

timdorr:
The example includes a `open http://localhost:8080/` command, but I can also add something to the console output on server boot.

gaearon:
Oh, something broke btw. Maybe it's the recent PR I merged.
I see "Uncaught SyntaxError: Unexpected token <" every time..

timdorr:
With `npm start`? Or something with the docs?

gaearon:
Forget it, I didn't realize you need to run both `npm start` and `npm run client`.
I understand why, but it's very non-obvious after other examples.

I'll file an issue.

gaearon:
https://github.com/rackt/redux/issues/643

PR #642, opened 2015-08-27
Add a console message to universal boot.

timdorr:
Sorry about so many PRs :neckbeard:

This addresses the comments in #640.

### DIFF
--- a/docs/introduction/Examples.md
+++ b/docs/introduction/Examples.md
@@ -75,7 +75,7 @@ cd redux/examples/universal
 npm install
 npm start & npm run client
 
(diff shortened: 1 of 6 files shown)


timdorr:
Maybe add a check that webpack was started up too and show a warning if it's not? 

A more sophisticated example pulls the webpack-dev-server bundle URL from a stats JSON blob dumped to a file. But frankly, that's a bit much for a simple example like this.

gaearon:
Let's discuss in https://github.com/rackt/redux/issues/643?

timdorr:
Will do.

PR #648, opened 2015-08-28
Add default param for combination()

DenisIzmaylov:


gaearon:
Why?

gaearon:
I don't think this is necessary.

DenisIzmaylov:
It's just semantic thing. When you start define default values for arguments you have to do it for all remained arguments.

PR #649, opened 2015-08-28
Don't depend on --harmony flag when running 'npm run client'

frodereinertsen:
Removed use of ES6 string interpolation

### DIFF
--- a/examples/universal/webpack.server.js
+++ b/examples/universal/webpack.server.js
@@ -18,5 +18,5 @@ new WebpackDevServer(webpack(config), {
     console.error(err);
   }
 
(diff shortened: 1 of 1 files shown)


timdorr:
Technically, the project is [built on iojs](https://github.com/rackt/redux/blob/master/.travis.yml#L3), but I can totally see someone hitting up the examples with node 0.12 and expecting them to work. Thanks!

frodereinertsen:
Thanks! I'm relatively new to this ecosystem and appreciate any aspect making it easier to take on, like things coincidentally working with node.

PR #658, opened 2015-08-30
Add store.resetState for reverting to initial state value

tgriesser:
I have a situation where I'd like to reset the entire app state to the original value (after `ActionTypes.INIT` is run), but couldn't find a clean way to write it as a plugin without replacing the majority of `createStore`.

I can also imagine situations where you'd want to replace the entire `currentState` with a new value, for example if you wanted to re-hydrate the entire app state from a different serialized session.

So I wanted to see if this `resetState` or an alternate more generic `replaceState` seemed like a good enough use case to add to the core API. The `replaceState` would look similar to this PR, but would pass in the state value and then run the init dispatch:

```js
/**
 * Replaces the state tree with a new value and runs "INIT"
 *
 * @returns {void}
 */
function replaceState(newState) {
  currentState = newState;
  dispatch({ type: ActionTypes.INIT });
}


// When a store is created, an "INIT" action is dispatched so that every
// reducer returns their initial state. This effectively populates
// the initial state tree.
replaceState(initialState);
```

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -36,7 +36,7 @@ export default function createStore(reducer, initialState) {
   }
 
   var currentReducer = reducer;
-  var currentState = initialState;
+  var currentState;
   var listeners = [];
   var isDispatching = false;
 
@@ -136,17 +136,27 @@ export default function createStore(reducer, initialState) {
     dispatch({ type: ActionTypes.INIT });
   }
 
+  /**
+   * Resets the state tree to the value after running "INIT"
+   *
+   * @returns {void}
+   */
+  function resetState() {
+    currentState = initialState;
+    dispatch({ type: ActionTypes.INIT });
+  }
 
   // When a store is created, an "INIT" action is dispatched so that every
   // reducer returns their initial state. This effectively populates
   // the initial state tree.
-  dispatch({ type: ActionTypes.INIT });
+  resetState();
 
   return {
     dispatch,
     subscribe,
     getState,
     getReducer,
-    replaceReducer
+    replaceReducer,
+    resetState
   };
 }
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -8,12 +8,13 @@ describe('createStore', () => {
     const store = createStore(combineReducers(reducers));
     const methods = Object.keys(store);
 
-    expect(methods.length).toBe(5);
+    expect(methods.length).toBe(6);
     expect(methods).toContain('subscribe');
     expect(methods).toContain('dispatch');
     expect(methods).toContain('getState');
     expect(methods).toContain('getReducer');
     expect(methods).toContain('replaceReducer');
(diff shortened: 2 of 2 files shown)


edge:
We do [this](https://github.com/edge/fluxette/blob/0.17.x/src/flux.js#L13-L17) at fluxette.

gaearon:
I don't think there is a need for this in core.
Store API must be as simple as possible so it's easy to write extensions like Redux DevTools.
Even one method is a huge addition.

See how to implement this in userland:

```
function makeHydratable(reducer, hydrateActionType) {
  return function (state, action) {
    switch (action.type) {
    case hydrateActionType:
      return action.state;
    default:
      return reducer(state, action);
    } 
  }
}
```

Usage:

```js
const HYDRATE_STATE = 'HYDRATE_STATE';
const reducer = combineReducers(myReducers);
const hydratableReducer = makeHydratable(reducer, HYDRATE_STATE);
const store = createStore(hydratableReducer, initialState);

// later
store.dispatch({
  type: HYDRATE_STATE,
  state: initialState
});
```

tgriesser:
I assume you'd need to also `dispatch({type: ActionTypes.INIT})` to re-init any of the reducers which aren't covered in the hydrate state?

gaearon:
Right!

gaearon:
Oh wait, you wouldn't need to. Just pass `action.state` to root reducer.

gaearon:
Here's the fixed version:

```js
function makeHydratable(reducer, hydrateActionType) {
  return function (state, action) {
    switch (action.type) {
    case hydrateActionType:
      return reducer(action.state, action);
    default:
      return reducer(state, action);
    } 
  }
}
```

tgriesser:
Awesome, so obvious now!

vicentedealencar:
@gaearon I don`t see a reason to call the reducer when the action is hydrateActionType, I don`t think any reducer should handle this action. I understand that the reducers would all ignore it and return state as is, so why not short-circuit the reducer returning action.state?

I am probrably missing something, but I have it implemented this way..

gaearon:
>I understand that the reducers would all ignore it and return state as is, so why not short-circuit the reducer returning action.state?

The author wanted to be able to specify non-full state, and have reducers return initial state in this case: https://github.com/rackt/redux/pull/658#issuecomment-136484887.

halt-hammerzeit:
@tgriesser Why would you do this:
```js
function replaceState(newState) {
  currentState = newState;
  dispatch({ type: ActionTypes.INIT });
}
```

Instead of this:

```js
function resetState() {
  currentState = {};
  dispatch({ type: ActionTypes.INIT });
}
```

What's the point of `currentState = newState` when it's gonna be overridden with the initial reducer state leaves anyway, so the new state is just gonna be the combined initial state.
Perhaps to store some non-reduceable data in the state tree but I don't see a motivation for that.

PR #660, opened 2015-08-30
Alternative to #658

tgriesser:
Still named `resetState`, but takes an optional `newState` value, defaulting to `initialState`.


gaearon:
See https://github.com/rackt/redux/pull/658#issuecomment-136478950.

PR #661, opened 2015-08-30
Skip triggering listeners if currentState === reducer(currentState, action)

tgriesser:
So this is probably too big of a change to have in the 1.x series, but I was thinking it'd be convenient if the listeners only fired when a change to the `currentState` value actually occurs.

In most cases when using the default `combineReducers` this won't be the case, but providing a reducer which wraps and does a shallow memoization it could save some extra overhead of calling every listener on every dispatch.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -98,6 +98,8 @@ export default function createStore(reducer, initialState) {
       throw new Error('Reducers may not dispatch actions.');
     }
 
+    var lastState = currentState;
+
     try {
       isDispatching = true;
       currentState = currentReducer(currentState, action);
@@ -105,7 +107,9 @@ export default function createStore(reducer, initialState) {
       isDispatching = false;
     }
 
-    listeners.slice().forEach(listener => listener());
+    if (lastState !== currentState) {
+      listeners.slice().forEach(listener => listener());
+    }
     return action;
   }
 
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -163,11 +163,11 @@ describe('createStore', () => {
     const listenerB = expect.createSpy(() => {});
 
     let unsubscribeA = store.subscribe(listenerA);
-    store.dispatch({});
+    store.dispatch(addTodo('Hello'));
     expect(listenerA.calls.length).toBe(1);
     expect(listenerB.calls.length).toBe(0);
 
-    store.dispatch({});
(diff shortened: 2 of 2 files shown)


clintwood:
Yes! Nice!  I'm currently wrapping the subscribe to get this behavior.

edge:
This is nice, but means that the equality check in react-redux would be redundant. So you'd have to change that too.

tgriesser:
@edge can you link to which equality check you're referring to?

timdorr:
@tgriesser: He means these `shallowEquals` calls: https://github.com/rackt/react-redux/blob/master/src/components/createConnect.js#L123

tgriesser:
Those would still be useful though, because if the listeners are triggered you'd re-compute the state props and you'd want to check those.

This change just prevents needing to slice/iterate the listeners if `dispatch` is a no-op.

timdorr:
And this would definitely trigger a version bump.  The current `subscribe()` API is described as follows: 
```
It will be called any time an action is dispatched, and some part of the state tree may potentially have changed
```
I'm not sure I like this change. Making decisions about event emissions on behalf of the user restricts what we can do. But I'm struggling to think of an example where that's harmful, so maybe I'm just being unreasonable :wink:

tgriesser:
> I'm not sure I like this change. 

Yeah on second thought, the triggering of events is almost certainly never the bottleneck of a real application, though it could reduce the reliance on dispatch acting as a general EventEmitter.

I could go either way, just thought it was worth taking a look at.

edge:
It also adds a subtle implicitness that I feel could mess up someone's application.

Shallow equality checks are cheap enough for this order of complexity not to be a problem. Honestly, a strict equality check (encouraging use of something like reselect) would be both faster and more intuitive, but it doesn't matter that much.

gaearon:
Some people might rely on mutability. Even though it's discouraged, there are valid use cases for mutating root objects if you're creating thousand of actions per second or something. While `react-redux` makes a hard assumption you never mutate, I don't want to bake it into Redux core. When doing things by hand, you should have more freedom.

gaearon:
You can also implement this in userland by creating a store enhancer that would provide its own `subscribe`. This is similar to how [redux-batched-subscribe](https://github.com/tappleby/redux-batched-subscribe/blob/master/src/index.js) does it.

PR #663, opened 2015-08-31
Switch from mutating DOM node to controlling form state in real world example.

dbunker:


dbunker:
Came across this [issue](https://github.com/rackt/redux/issues/662) and thought it might be illustrative to show how this could be done. Let me know what you think.

gaearon:
Thank you for the PR. I'd rather not complicate the example—at least, not until we can express React Router transitions as actions too. (See https://github.com/acdlite/redux-react-router/issues/1#issuecomment-136428844.)

PR #668, opened 2015-08-31
Remove getReducer() from Store API

gaearon:
It was only needed because React Redux used it to magically allow reducer hot reloading.
This magic used to cause problems (https://github.com/rackt/redux/issues/301, https://github.com/rackt/redux/issues/340), so we made the usage of `replaceReducer()` for hot reloading explicit (https://github.com/rackt/redux/pull/667). Now that it is explicit, there is no need for `getReducer()` so it can safely be removed from the public API.

This is a breaking change, so it goes into `wip-2.0` branch.

### DIFF
--- a/docs/Glossary.md
+++ b/docs/Glossary.md
@@ -97,7 +97,6 @@ type Store = {
   dispatch: Dispatch;
   getState: () => State;
   subscribe: (listener: () => void) => () => void;
-  getReducer: () => Reducer;
   replaceReducer: (reducer: Reducer) => void;
 };
 ```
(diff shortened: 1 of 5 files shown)


quicksnap:
Can we issue a deprecation warning in development builds for `getReducer`?

quicksnap:
(In the 1.0 branch that is)

gaearon:
We can, although I expect it to be called by React Redux in 99% cases. I don't think anybody else used it..

We need to release React Redux 2.0 that wouldn't use it.

quicksnap:
Given Redux's small API surface, it's probably overkill if there's no affordance for it already.

PR #674, opened 2015-09-02
Reorganize universal example

timdorr:
Splits it so we have this tree:

- `examples/universal/`
 - `client`
 - `server`
 - `common`

Should make it very clear what is common code and what is client or server specific.


gaearon:
:+1:

PR #675, opened 2015-09-02
Docs: Minor tweaks to advanced/ExampleRedditAPI.md

dmin:
Three changes:

1. Removed ```combineReducers``` from ```import``` statement under configureStore.js (combineReducers is not used there).

2. The file names as listed on this page seem to imply that reducers.js and actions.js are in the same directory, so I changed the relative path of the import statement in reducers.js to match this. (from ```"../actions"``` to ```"./actions"```).

3. Same issue as number 2 above: configureStore.js and reducers.js seem to be in the same directory, so I changed the relative path of the import statement in configureStore.js to match this (from ```"../reducers"``` to ```"./reducers"```).


timdorr:
Looks good. Thanks!

PR #676, opened 2015-09-02
Invert rows and columns in table of components

pedrottimark:
1. Easier to read: less wrapping of lines with 2 columns and 4 rows
2. Easier to read: heading column aligned at the right, component columns aligned at the left
3. Easier to understand: smart versus dumb components as columns are "dependent variables" in analogy with the x-axis of a graph, or like month columns in a spreadsheet; characteristics as rows are like line items in a spreadsheet.
4. Easier to modify: you probably won't add another type of component, but you might add another contrasting characteristic (for example, from the bulleted lists in https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0). Adding another row to the inverted table would be easy, adding another column to the original table would be harder to do and harder to read.

### DIFF
--- a/docs/basics/UsageWithReact.md
+++ b/docs/basics/UsageWithReact.md
@@ -20,35 +20,37 @@ React bindings for Redux embrace the idea of [separating “smart” and “dumb
 
 It is advisable that only top-level components of your app (such as route handlers) are aware of Redux. Components below them should be “dumb” and receive all data via props.
 
-<center>
 <table>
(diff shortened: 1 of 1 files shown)


gaearon:
:+1:

pedrottimark:
Keep up the good work! It is an honor for redux to where I made my first pull request in open source :)

By the way, this change also applies to README.md in react-redux

PR #679, opened 2015-09-03
Remove link to "intents"

moroshko:
Looks like there is no such thing as "intent" in the Glossary anymore.


timdorr:
Actually, they were renamed Async Actions in 1709fc9. Could you update the link to that section instead?

moroshko:
Closed in favor of #680

PR #690, opened 2015-09-05
Updates examples to use react-transform-webpack-hmr

davezuko:
Saw that you wanted all these done at the same time so figured I'd help out. Let me know if there's anything else that needs to be done. Thanks for all your hard work!


gaearon:
This is sweet, thank you! I'll review on Monday.

nodu:
Could one of you point me to the discussion about the switch to
react-transform-webpack-hmr
from react-hot-loader?

davezuko:
I'm sure @gaearon will be able to do a better job, being the author and all, but maybe this will help: https://medium.com/@dan_abramov/the-death-of-react-hot-loader-765fa791d7c4

There was another post somewhere about some issues specific to react-hot-loader that this is designed to fix, but I can't seem to find it right now. Overall, I think the impetus is that RHL, for all its awesomeness, was more of a first draft and this is supposed to be something a bit more robust.

timdorr:
The big "problem" with RHL is that it [patches in raw strings of code](https://github.com/gaearon/react-hot-loader/blob/master/index.js#L26-L72), which can be prone to breakage and makes it really hard to test. 

Because react-transform is a babel plugin, it gets to tap into the AST that babel builds and intelligently reason about the code structure. And it means [there are tests](https://github.com/gaearon/react-proxy/tree/master/test)! You'll also notice there are no ugly code patches in there. It's more complex (spread amongst 3 separate projects now), but much more durable and supports more features.

Amorino:
Thanks for this pull, it will be awesome checking redux on react-transform-webpack-hmr.

timdorr:
Works great for me. I'm going to merge this in and then I can get started on using `webpack-hot-middleware` in the universal example. And hopefully I can also get to a HMRTT (hot module reloaded time travel) example.

gaearon:
:+1:

examples/async/package.json
```diff
@@ -36,12 +36,13 @@
...
     "babel-core": "^5.6.18",
     "babel-loader": "^5.1.4",
+    "babel-plugin-react-transform": "^1.0.3",
     "expect": "^1.6.0",
     "jsdom": "^5.6.1",
     "mocha": "^2.2.5",
     "mocha-jsdom": "^1.0.0",
     "node-libs-browser": "^0.5.2",
-    "react-hot-loader": "^1.3.0",
+    "react-transform-webpack-hmr": "^0.1.4",
```
vincentfretin: now react-transform-hmr

PR #698, opened 2015-09-09
Update travis Node.js version & remove io.js

demohi:


timdorr:
Looks like there are issues with contextify. Unfortunately, @brianmcd hasn't update the project in a while and it's not been working with iojs 3.0 either. Anyone know of any alternatives?

gaearon:
We can wait for Lodash 4 (#611) and then we won't need that test relying on `contextify`.

timdorr:
@demohi Can you merge in master to your branch? It should now run correctly under 4.0.0. I've tested locally and want to double-check against Travis.

demohi:
@timdorr    I merged tha.

timdorr:
Looks like there's [an upstream fix in webpack-dev-server](https://github.com/webpack/webpack-dev-server/commit/4fe2cf88d8201c42e3e2fa033d741f953b273518) (and socket.io) that isn't been released yet. It should fix all those npm build errors in Travis. But it runs, so it's good for now.

PR #703, opened 2015-09-10
Fix unhighlighted JavaScript code.

f:
This part was unhighlighted in documentation. I basically pushed the code into hl.js and wrote the output as HTML.

Keep up the good work!


timdorr:
Since the HTML is generated, won't this be overridden by the next gitbook run?

Also, can you merge in #704 and #705 to this branch so we can track under one PR? Thanks!

gaearon:
Unfortunately this indeed will be overridden next time we update the docs.

gaearon:
All but one seem fixed with the new version of Highlight.js.
I filed an issue for the React part which is still not highlighted: https://github.com/isagalaev/highlight.js/issues/931

gaearon:
Meh. I just ported the docs to use [Prism](http://prismjs.com) instead.
Looks much better now.

f:
:+1: Better! Keep up good work guys!

PR #705, opened 2015-09-10
Fix unhighlighted JavaScript code on React documentation page.

f:


gaearon:
Please see https://github.com/rackt/redux/pull/703#issuecomment-139775055.

gaearon:
Fixed now that we use Prism instead of Highlight.js.

PR #708, opened 2015-09-11
Remove semi-colon from sample code

quamen:
Just a simple diff to the docs where I removed a semi-colon that causes a syntax error when running the code.


timdorr:
Thanks!

PR #714, opened 2015-09-13
Migrate real-world example to redux-router

gaearon:
Starting with Redux React Router 1.0 Beta, we're recommending it as the official routing solution.
You'll need React Router 1.0RC, <s>React 0.14RC (or newer, of course)</s>* to go with it.

<i>* - technically you can use RRR with React 0.13, but you'll have to wrap `<ReduxReactRouter>` into a function inside `<Provider>`.. come on, it's time to migrate!</i>

Highlights:

* No more using `context` to fire transitions
* No need for `ownProps` to read router state—it's in `state.router`
* Redux actions show up in log
* Time travel with Redux DevTools works


acdlite:
Cool! Technically you don't need React 0.14, though.

gaearon:
Oh, okay! But I guess I'll use it anyway because 0.14 is imminent and I don't want to bother with owner context nonsense in `<Provider>`.

acdlite:
Of course :)

gaearon:
By the way I'm extremely grateful for your efforts with this.

DJCordhose:
I am wondering how this is supposed to work in a universal approach on the server side. Tried it out, but did not know what to provide as createHistory in

reduxReactRouter({
    routes,
    createHistory
  }),

on the server side. createBrowserHistory won't work of course. 

Or should I just have two versions of configureStore, one for server one for client? Server side without redux-react-router as it does not make any sense there at all?

gaearon:
Good question, personally I don't know. @acdlite?

acdlite:
Still trying to figure out the best way to handle this. Right now I'm thinking that we'll provide a special version of the store enhancer for server-side rendering, similar to how there's a separate `render()` and `renderToString()`. Instead of `history` or `createHistory()`, this special version takes a `location` and an `onMatch()` function

```js
import { reduxReactRouter } from 'redux-react-router/server';

const location = createLocation(req.url);
const store = compose(
  reduxReactRouter({
    location,
    onMatch(error, redirectLocation, renderProps) => {
      // Copied from React Router docs
      if (redirectLocation)
        res.redirect(301, redirectLocation.pathname + redirectLocation.search)
      else if (error)
        res.send(500, error.message)
      else if (renderProps == null)
        res.send(404, 'Not found')
      else
        res.send(renderToString(<RoutingContext {...renderProps}/>))
    })
  });
)(createStore)(reducer);
```

acdlite:
How about this?

```js
import { createLocation } from 'history';
import { reduxReactRouter, match } from 'redux-react-router/server';

const store = compose(
  reduxReactRouter({ routes });
)(createStore)(reducer);

const location = createLocation(req.url);
const onMatch = store.dispatch(match(location)); // Returns callback

onMatch((error, redirectLocation) => {
  if (error) {
    // handle error
  } else if (redirectLocation) {
    // handle redirect
  } else {
    // Everything is fine, render like normal
    res.send(
      renderToString(
        <Provider store={store}>
          <ReduxRouter />
        </Provider>
      )
    );
  }
});
```

doctyper:
@DJCordhose You can use createBrowserHistory on the client and [createMemoryHistory](http://rackt.github.io/history/stable/GettingStarted.html) on the server.

acdlite:
@doctyper That won't work if there's a redirect or a 404, though. See this doc https://github.com/rackt/react-router/blob/master/docs/advanced/ServerRendering.md

examples/real-world/reducers/index.js
```diff
@@ -1,6 +1,7 @@
 import * as ActionTypes from '../actions';
 import merge from 'lodash/object/merge';
 import paginate from './paginate';
+import { routerStateReducer as router } from 'redux-react-router';
```
ooflorent: Is it really needed?

timdorr:
Fixed up some build and merge blockers for you. Looks good!

DJCordhose:
As someone over here https://github.com/rackt/redux-router/issues/43#issuecomment-141776533 asked how to wait until all necessary actions have been executed on server side, I created this gist (which really works): https://gist.github.com/DJCordhose/68fe2ddca6d9ba595534

Would you think it might make sense to add this server side rendering to the example? Or another example? Does my code make sense at all?

gaearon:
Two minor issues left here:

- [ ] The DevTools open to the whole screen size instead of a smaller window
- [ ] The dev server doesn't handle non-default URLs. For example I get 404 if I go to `localhost:3000/gaearon` and hit Refresh

Would somebody like to help with these two?

timdorr:
That second one is simple. Just change `app.get` to `app.use`. I can try to get to that today.

timdorr:
Got the second one for you. 

I'm not sure I understand the first request. If you don't specify a window size, which is what the example does, then it should size the window to the same size as your current one. That's pretty big for me, since it's coming from my primary browser window. Should we instead shoot for a standard width to best display the vertically-oriented UI of the dev tools and let the browser choose the height?

gaearon:
>Should we instead shoot for a standard width to best display the vertically-oriented UI of the dev tools and let the browser choose the height?

Yeah, if possible, I'd do that.

timdorr:
Finally, got around to doing that. Work has been crazy lately. Does everything look good?

gaearon:
Thank you! I squashed the commits for less noise.

Issue #715, opened 2015-09-13
Be less opinionated in combineReducers


ashaffer:
I know I can just make my own, and I will do that for now, but it seems like combineReducers should not be quite so opinionated about the state shape.  For instance, in my app, I want to do something like this:

```javascript
reduceReducers(
  handleActions(someGlobalActions),
  combineReducers({
    entities,
    collections: combineReducers({
      explore,
      following
    })
  })
})
```

But this does not work in a very non-obvious way.  It doesn't work because combineReducers will drop any keys that are not in the state shape passed to it, so all of the things done in my global actions will not be recorded.

gaearon:
If we allow (and pass through) arbitrary keys, we wouldn't be able to guard against a very common beginner mistake when the shape of the initial state doesn't match the reducer shape. Utilities that ship with Redux just illustrate patterns and provide guidance to beginners. It's indeed often better to write your own when you know what you need. In this case, I believe your pattern of usage deviates from the simple case `combineReducers()` was designed for.

Still, I agree silently omitting keys isn't nice. We should at least warn this is happening. There is code for this, but it only executes once—perhaps we should always run it? (In development only, of course.)

gaearon:
cc @ellbee

ellbee:
I agree that silently omitting the keys isn't nice. The state shape check is only executed once because it felt unnecessary to be calling it every time. I don't think calling it every time would have any bad side effects and it would warn for cases like this.

gaearon:
I'm happy to accept a pull request amending this, given that the warning wording makes sense in both situations.

gaearon:
Fixed by https://github.com/rackt/redux/pull/720, will be released in the next minor release.

gaearon:
The new warning is out in 3.0.1.
Please verify!

PR #720, opened 2015-09-13
Always call verifyStateShape for combined reducers

ellbee:
As discussed in #715

### DIFF
--- a/src/utils/combineReducers.js
+++ b/src/utils/combineReducers.js
@@ -15,8 +15,11 @@ function getErrorMessage(key, action) {
   );
 }
 
-function verifyStateShape(initialState, currentState) {
-  var reducerKeys = Object.keys(currentState);
+function verifyStateShape(inputState, outputState, action) {
+  var reducerKeys = Object.keys(outputState);
+  var argumentName = action && action.type === ActionTypes.INIT ?
+    'initialState argument passed to createStore' :
+    'previous state received by the reducer';
 
   if (reducerKeys.length === 0) {
     console.error(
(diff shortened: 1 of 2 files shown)


gaearon:
It's still saying “initial state” in the error message.

ellbee:
Ah, sorry. I thought you meant it could be changed because the message already made sense. Of course, it does not. I'll update.

ellbee:
I think it is clearer if the messages are slightly different depending on whether the warning has been triggered via the INIT action in createStore or by calling the the reducer directly. Here is how I propose the messages read for these two cases:

**Passing mismatching initialState to createStore (existing case)**

The initialState argument passed to createStore has unexpected type of "Number". Expected argument to be an object with the following keys: "foo", "baz"

Unexpected keys "fdfds", "dsfsdf" in initialState argument passed to createStore will be ignored. Expected to find one of the known reducer keys instead: "todos", "todosReverse",
"dispatchInTheMiddleOfReducer", "errorThrowingReducer"

**Passing mismatching state to reducer**

Unexpected key "bar" in state argument passed to reducer will be ignored. Expected to find one of the known reducer keys instead: "foo", "baz"

Unexpected keys "bar", "qux" in state argument passed to reducer will be ignored. Expected to find one of the known reducer keys instead: "foo", "baz"

The state argument passed to reducer has unexpected type of "Number". Expected argument to be an object with the following keys: "foo", "baz"

gaearon:
>in state argument passed to reducer will be ignored

This might be a bit misleading because most likely the user didn't *pass* the argument to reducer themselves. They just *returned* it the last time. Any idea how to change phrasing to suggest this?

ellbee:
Yeah, I think you are right. 

As reducers are shown in the docs to have signature of (previousState, action) => newState, how about something like this?

The previous state received by the reducer has unexpected type of "Number". Expected argument to be an object with the following keys: "foo", "baz"

Unexpected key "bar" in previous state received by the reducer will be ignored. Expected to find one of the known reducer keys instead: "foo", "baz"

I think this is better, but the first sentence of the second warning is starting to read a bit awkwardly.

ellbee:
Slight variations:

Unexpected keys "bar", "qux" found in previous state received by the reducer. Expected to find one of the known reducer keys instead: "foo", "baz". Unexpected keys will be ignored.

Unexpected keys "bar", "qux" found in reducer's previous state. Expected to find one of the known reducer keys instead: "foo", "baz". Unexpected keys will be ignored.

gaearon:
I like this:

>Unexpected keys "bar", "qux" found in previous state received by the reducer. Expected to find one of the known reducer keys instead: "foo", "baz". Unexpected keys will be ignored.

ellbee:
PR updated with the preferred wording

gaearon:
Great job! :+1:

PR #722, opened 2015-09-13
Catch errors in listeners

bkniffler:
Hi, 
I noticed that errors are silently swallowed in createStore.js

```javascript
listeners.slice().forEach(function (listener) {
    listener();
});
```

something like this might be more helpful:

```javascript
listeners.slice().forEach(function (listener) {
    try{
        listener();
    }
    catch(err){
        console.error(err);
    }
});
```

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -118,10 +118,9 @@ export default function createStore(reducer, initialState) {
     }
 
     listeners.slice().forEach(listener => {
-      try{
+      try {
         listener();
-      }
-      catch(err){
+      } catch (err) {
         console.error(err);
       }
     });
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -117,7 +117,14 @@ export default function createStore(reducer, initialState) {
       isDispatching = false;
     }
 
-    listeners.slice().forEach(listener => listener());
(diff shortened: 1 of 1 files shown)


gaearon:
I believe this is wrong. If error occurs inside `listener`, it's going to keep propagating until caught, and in this case, `dispatch` will throw. It's somewhere in your app that the error is caught; not in Redux.

gaearon:
For example, a common source of problem is dispatching from inside a Promise. If your Promise implementation doesn't report unhandled rejected Promises, it's going to seem that the error was swallowed by Redux, whereas in fact it resulted in a silently rejected Promise.

If you don't believe this to be true, please provide a failing test case.

bkniffler:
Ah, precisely, kudos for your presumption. I've been using @erikras https://github.com/erikras/react-redux-universal-hot-example/blob/master/src/redux/middleware/clientMiddleware.js and errors were not properly catched here.

Thanks for taking the time to help find the solution!

gaearon:
Can you please raise an issue there? It's a matter of using the right Promise polyfill I think..

bkniffler:
Yep, I was about to do that.

PR #725, opened 2015-09-14
docs: rename fakeStore to createFakeStore

chentsulin:
Use `createFakeStore` instead of `fakeStore`, because it's a function.


gaearon:
Strictly saying “fake” can be used as a verb, but I agree that's easier to follow.

PR #742, opened 2015-09-16
Fix typo in docs

liubko:


gaearon:
Thanks, but the two spaces are intentional. They create a newline in Markdown.

liubko:
@gaearon You are right. :sweat: Fixed and push with `--force`

acdlite:
> Note that such utilities add magic to your code. Are magic and indirection really worth it to avoid a few extra lines of code?

This is my first time reading that line and I have to say, I object to the idea that using a util function to create reducers, in lieu of using switch statements, is indirect or *magical*. We should reserve the word "magical" for code that is implicit or unpredictable, which redux-actions is not. It has a well-defined contract and a straightforward implementation. Also, it promotes the micro-reducerization (yes, I just made that word up) which is fundamental to the Redux architecture: composing a single reducer out of many smaller reducers.

In fact, Redux itself provides a util function `combineReducers()` which isn't all that different:

```js
const reducer = combineActions({
  stuff: stuffReducer,
  things: thingsReducer
});

// Compared to
function reducer(state = {}, action) {
  return {
    stuff: stuffReducer(state.stuff, action),
    things: stuffReducer(state.things, action)
  };
}
```

Is that really saving you so many lines? No, but it's repetitive and fragile and easy to accidentally mess up. So imagine if we had the disclaimer:

> Note that `combineReducers() adds magic to your code. Are magic and indirection really worth it to avoid a few extra few lines of code?

Yeah, that would be silly.

Sorry if my tone is a bit strong — the word "magic" is a pet peeve of mine. :)

timdorr:
But yeah, I tend to agree with @acdlite. Moreover, I think it's expressing an opinion without much substance to back it up. I mean, this article *is* about reducing boilerplate, after all :smile: 

The primarily confounding thing is it starts talking about the disadvantages auto-generation of action type constants, but doesn't proceed to address that issue. It should read something more like: "Don't autogenerate action type constants because that leads to indirection that's hard to keep track of. However, you can still reduce your boilerplate when generating action creators by writing a small function to automate the process or using tools like redux-actions and redux-action-utils"

I'm happy to re-write around that phrasing.

gaearon:
@acdlite Feel free to propose changes, I don't feel strongly about it :-).

timdorr:
I'll start something in a PR in this repo so @acdlite can build on (or overwrite!) my proposal. Thanks, @gaearon!

PR #745, opened 2015-09-17
Docs: `applyMiddleware` API

dminkovsky:
Moved logging example to the top.

Logging is the canonical example of a cross-cutting concern. It is
(or should always be) the first example of implementing a cross
cutting concern, followed by more exotic things like async thunks.


gaearon:
:+1:

PR #747, opened 2015-09-17
Docs: Move async thunk middleware example to own section

dminkovsky:


timdorr:
I think if this is going to warrant its own section in the docs, it should be more than a copypasta. There should be some supporting documentation and a clear breakdown of which files are to be created in this example.

gaearon:
I agree. Right now it doesn't fit with the docs flow (every section leading onto “Example” with code from previous chapters).

A better idea perhaps would be to create “Asynchronous Control Flow” in Recipes and rework this example into a tiny tutorial.

dminkovsky:
Yeah.

This PR, if anything, is more about showing how perhaps this doesn't belong
in the API section. Not a great PR, but illustrates what I think.

timdorr:
@dminkovsky Did you want to take on writing that up?

dminkovsky:
@timdorr I would have as part of the PR but I've never used that module.

I would be down to edit other sections of the docs, where I am familiar with the material. Let me know I can provide a list of suggestions, or PRs.

timdorr:
I'm going to close this out, but feel free to continue to make suggestions. I'll try to get to a full recipe in another PR soon, so we can all discuss the changes.

PR #751, opened 2015-09-18
Update Store.md

EcutDavid:
The unsubscribe not really unsubscribe the state change event.


EcutDavid:
Sorry I make a mistake, it really work.

EcutDavid:
Why subscribe again can unsubscribe...

PR #757, opened 2015-09-19
Doc: Don’t call non-pure functions in a reducer

dtinth:
The first time I use Redux, I did some silly things like this:

```js
  case ADD_TODO:
    return Object.assign({}, state, {
      todos: [...state.todos, {
        id: uuid.v4(), // ｉｍｐｕｒｅ！！！　๐·°(৹˃̵﹏˂̵৹)°·๐
        text: action.text,
        completed: false
      }]
    });
```

Here, `uuid.v4()` is not pure. This causes the state to be unpredictable. I had to come back and move this impure code to the action creator later. So I hope this docs change can prevent future mistakes like this.


gaearon:
Good catch. :+1:

PR #759, opened 2015-09-20
fix style of todomvc filter links

jas-chen:
Cursor value should be 'pointer'.


timdorr:
What about IE5?? :wink: 

Thanks!

jas-chen:
😇

Issue #717, opened 2015-09-13
Move safety checks inside combineReducer to first invocation


gaearon:
Currently we make a bunch of invariant tests inside `combineReducers()` to make sure it's not being misused in common ways. These tests currently occur at the `combineReducers()` call time.

This makes Redux DevTools unable to catch the errors inside reducers, as they often trigger those invariants to blow up during module definition, and abort hot reloading. Instead, I propose to move the tests to the *first real invocation* of the combined reducer. In most cases that would be `createStore` call because it causes an implicit initial dispatch. This way Redux DevTools can catch and display such errors.

Shall we do that? The code would change to be like

```js
  var reducerSanityVerified;
  var stateShapeVerified;

  // don't do it here

  return function combination(state = defaultState, action) {
    if (!reducerSanityVerified) {
      Object.keys(finalReducers).forEach(key => {
        // do that stuff
```

<img width="729" alt="screen shot 2015-09-13 at 07 09 31" src="https://cloud.githubusercontent.com/assets/810438/9835189/6aa2c77c-59e6-11e5-984f-7fbb19b73692.png">

gaearon:
Another possible alternative is to create errors and `console.error()` them ASAP during `combineReducers()` call, but throw during first dispatch. This way we both have a nice stack and keep DevTools happy.

gaearon:
Would anybody like to work on this?
@ellbee, do you think you might have the time?

ellbee:
Sure, I'll be able to take a look in the next couple of days.

ellbee:
@gaearon I have been working on this, but I am new to the dark arts of HMR and I have a question.

In the following module taken from the counter example, I think delaying the sanity check until the dispatch of the @@INIT action will still blow up the module definition because the init action is getting dispatched either inside `createStore` or `replaceReducer`. This module is going to get re-evaluated too when a reducer changes I think?

```
import React, { Component } from 'react';
import CounterApp from './CounterApp';
import { createStore, applyMiddleware, combineReducers, compose } from 'redux';
import { devTools, persistState } from 'redux-devtools';
import { DevTools, DebugPanel, LogMonitor } from 'redux-devtools/lib/react';
import thunk from 'redux-thunk';
import { Provider } from 'react-redux';
import * as reducers from '../reducers';

const finalCreateStore = compose(
  applyMiddleware(thunk),
  devTools(),
  persistState(window.location.href.match(/[?&]debug_session=([^&]+)\b/))
)(createStore);

[...]

PR #761, opened 2015-09-20
move combine reducers sanity check #717

ellbee:
Creates and logs sanity check error during `combineReducers`, but only throws on first invocation. Addresses #717.

### DIFF
--- a/src/utils/combineReducers.js
+++ b/src/utils/combineReducers.js
@@ -72,11 +72,12 @@ function verifyStateShape(inputState, outputState, action) {
 
 export default function combineReducers(reducers) {
   var finalReducers = pick(reducers, (val) => typeof val === 'function');
+  var sanityError;
 
   Object.keys(finalReducers).forEach(key => {
     var reducer = finalReducers[key];
-    if (typeof reducer(undefined, { type: ActionTypes.INIT }) === 'undefined') {
-      throw new Error(
+    if (!sanityError && typeof reducer(undefined, { type: ActionTypes.INIT }) === 'undefined') {
+      sanityError = new Error(
         `Reducer "${key}" returned undefined during initialization. ` +
         `If the state passed to the reducer is undefined, you must ` +
         `explicitly return the initial state. The initial state may ` +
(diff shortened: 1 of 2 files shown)


src/utils/combineReducers.js
```diff
@@ -97,9 +98,16 @@ export default function combineReducers(reducers) {
     }
   });
 
+  if (sanityError) {
+    console.error(sanityError.message);
+  }
+
```
ellbee: Not sure that this console.error is necessary, the error is thrown very shortly afterwards. And if it is useful, it has just occurred to me that it should probably log error.stack if available.

timdorr: `console.logs/errors` are a no-no in the library because they are normally uncatchable and noisy (and `console` sometimes doesn't exist). A thrown `Error` is best because that follows normal JS patterns.

ellbee: The idea is to delay throwing the error but log as soon as possible. The console.error was suggested by @gaearon in the second comment of issue #717. I do agree that this is noisy, as the error is thrown shortly afterwards and you end up with the same error logged twice in succession.

gaearon: I don't think `console.error` is inherently bad if it is a sign of an error. React also uses `console.error` for `propTypes` warnings because they're not really warnings: if you have them, you need to fix them. The nice thing about `console.error` is that it prints the stack trace, unlike its friends.

timdorr: Doesn't it disable logging in the production build? Maybe the same thing should happen here. A log isn't bad for devs, but shouldn't make its way to production.

gaearon: Yes, we previously `console.error`'d only in development, too.
I think we should hide this `console.error` behind the dev flag as well.

Of course, if it's not helpful, it's better to remove it. I'm not sure whether the confusion from error printed twice is bigger than the help from a proper stack trace. Thoughts?

ellbee: I've played with this a bit more using devtools and I don't think the `console.error` is helpful. Throwing `sanityError` logs the stack trace from where the error was created in `combineReducers`, so the two stack traces in the log are actually the same.

gaearon: Oh, okay. Then let's remove it! Thanks for checking.

ellbee: Cool, will do.

ellbee:
Updated to create but not log the error during `combineReducers`. The error is thrown on first invocation.

ellbee:
Added production check around sanity check (#771)

gaearon:
It's only OK to check for production env for errors we don't throw. Both environments should throw in the same conditions so error stack traces from production are reproducible in development. Please see https://github.com/rackt/redux/issues/771#issuecomment-142259143.

ellbee:
Got it. I'll revert back to how it was previously.

gaearon:
Sure, thanks!

PR #766, opened 2015-09-21
Use webpack-dev-middleware in universal example

### LINKED ISSUE #643: Tweak Universal example
Here's a few issues I'd like to see addressed:

* Unlike all examples, you need to open `localhost:8080` in it instead of `localhost:3000`. I'd rather change it to be the other way: put app server at `3000` and use something like `3001` for WebpackDevServer. This way you don't have to memorize which example you are running.

* You need to `npm start & npm run client` in universal example. I struggled with non-working example for 15 minutes before figuring that out. This is obvious to someone experienced with server rendering, but wasn't obvious to me. I missed that part in the doc, and I'm sure a lot of people will `npm start` and expect things to happen.

* I'd prefer to see `client`, `server` and `universal` folder structure at the very top, which signals which modules are related to which part. Lots of universal examples follow that structure and we should do it too.

timdorr:
Also updates to react-transform-hmr and handles some issues with it's throwing behavior.

Closes #643 & addresses comments on #690


timdorr:
I'm going to see how much work is converting the webpack-dev-server usage in the other examples too. I don't think it will be a lot, but will defer if there's some roadblock that I run into.

timdorr:
Cool, that wasn't too bad. I also made the universal example have a single start script, so it definitely cleans things up a bit.

examples/async/webpack.config.js
```diff
@@ -4,8 +4,7 @@ var webpack = require('webpack');
 module.exports = {
   devtool: 'cheap-module-eval-source-map',
   entry: [
-    'webpack-dev-server/client?http://localhost:3000',
-    'webpack/hot/only-dev-server',
+    'webpack-hot-middleware/client?path=/__webpack_hmr&timeout=20000',
```
ronag: There is no reason to add `?path=/__webpack_hmr&timeout=20000` as the defaults are good enough and adding it here just confuses beginners.

examples/counter/.babelrc
```diff
@@ -1,13 +1,17 @@
...
+      ],
+      "extra": {
+        "react-transform": [{
+          "target":  "react-transform-hmr",
+          "imports": ["react"],
+          "locals":  ["module"]
+        }]
+      }
+    }
   }
```
ronag: `"react-transform"` needs to go into webpack config as a loader query. As of `react-transform-hmr@1.0.0` this will throw during server side rendering.

examples/async/webpack.config.js
timdorr: [Ah, you're right](https://github.com/glenjamin/webpack-hot-middleware/blob/master/client.js#L4-L9). I'll rip that out.

examples/counter/.babelrc
timdorr: There is no server-side rendering in these examples, so the `.babelrc` usage is fine. The universal example has been updated, which *does* render server side.

Issue #542, opened 2015-08-16
Idea: Batching actions


FredyC:
What about being able to dispatch more actions at once? Currently the dispatch accepts single object as a result from action creator. Could it perhaps accept an array? It would then process all the actions sequentially before invoking change.

I've come to the situation in my editor app, that on press of a key I have to split the text (updating text of a current line), create a new line (with remaining text) and set focus on that new line. As of now I have this logic within single reducer, but that makes reducers rather complex and not really reusable. To keep a logic out off reducers, it means I would have to dispatch **three** consecutive actions. That means the re-render after each one. Perhaps I am worried too much and it might be just fine like that?

I guess that another option could be having a single action that would be basically processed by three different reducers, but I would have to basically call some external functions from reducer to provide necessary changes. Some other time I just want to update text of line or just create a new line. That would  mean a different set of actions.

ForbesLindesay:
That last option sounds best,  it's generally best to fire actions that describe what happened (e.g. User typed character) rather than what mutations you want (e.g. Update text of line)

quirinpa:
I'm using this middleware i made:
```
export default function nextMiddleware({ dispatch }) {
	return (next) => (action) => {

		let ret = next(action);
		const { then } = action;

		if (then) {
			// console.log('MIDDLEWARE NEXT ', action.type);
			// console.log(action)
			let nextActions = then;
			if (!Array.isArray(nextActions))
				nextActions = [nextActions];

			nextActions.map(item => dispatch((typeof item === 'function') ? item(dispatch) : item));
		}
		return ret;
	};
}
```
it's available on my profile as a repo idk if it's updated though.

FredyC:
@ForbesLindesay I guess you are right, but it means I would need merge result from all interested action creators. That means the logic about "what is happening" have to be combined at the time of dispatch and then split again in reducers. It's kinda tied together and it would be also harder to examine such dispatch seeing data from all action creators there.

@quirinpa I haven't though about middleware, that might be good way to go. Thanks !

taylorhakes:
@quirinpa You can't do it with just middleware. Each call to dispatch triggers all the subscribers. 

You need a higher order store that wraps the dispatcher and only calls the subscribers once.

quirinpa:
@FredyC No problem man :)
@taylorhakes It is working for me. What do you mean the subscribers are fired again?

taylorhakes:
[...]

PR #769, opened 2015-09-21
Transaction behaviour of dispatch()

arusakov:
Hi!
Each store.dispatch(/* some action */) will invoke all listeners callbacks. This can lead to a lot of unnecessary calls.
I propose to add some transaction-like behaviour.
```javascript
store.begin();
store.dispatch({type: 1});
store.dispatch({type: 2});
// ...
store.dispatch({type: 100});
store.commit(); // and invoke listeners callbacks one time at the end
```
It seems to me, that the idea of transaction would be very useful in redux. But this PR has big changes in API and need some work. 
Maybe another API seem more successful. Something like this easier to implement:
```javascript
function dispatch(action, notNotify=false) {
// ...
}

// in app
store.dispatch({type: 1}, true);
store.dispatch({type: 2}, true);
// ...
store.dispatch({type: 100}); // and invoke listeners callbacks one time at the end
```

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -39,6 +39,7 @@ export default function createStore(reducer, initialState) {
   var currentState = initialState;
   var listeners = [];
   var isDispatching = false;
+  var inTransaction = false;
 
   /**
    * Reads the state tree managed by the store.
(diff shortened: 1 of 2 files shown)


FredyC:
Transaction seems like bad term for this because there is no _rollback_ which is expected from such approach. Instead I would probably call it _batch_. And I am not sure if batching is really necessary. I needed something like this [some time ago](#542), but I managed to implement it with simple store enhancer.

arusakov:
@FredyC batch is a good term for this. Thanks. Maybe [redux-batched-subscribe] (https://github.com/tappleby/redux-batched-subscribe) is what I need.
But maybe batching inside redux will be more holistic solution. I don't know.

gaearon:
We don't want to expand Store API surface. (See https://github.com/rackt/redux/issues/303#issuecomment-125184409 for rationale.) Indeed, [redux-batched-subscribe](https://github.com/tappleby/redux-batched-subscribe) solves this in userland.

PR #773, opened 2015-09-22
update doc Glossary to `must` have type field

chentsulin:


gaearon:
Missed that one, thanks.

PR #794, opened 2015-09-24
Universal example inconsistencies

### LINKED ISSUE #785: 2 inconsistencies in docs for Universal example
1. Should the bulleted list have a point for Universal following Async and preceding Real World? in https://github.com/rackt/redux#examples

2. Second part of npm command refers to script which does not exist and is not needed? in http://rackt.github.io/redux/docs/introduction/Examples.html#universal

```
npm start & npm run client
```

but the scripts object in package.json for the Universal example has only start but not client (and not sure what that would mean) so should this line be just like the other examples? 

```
npm start
```

pedrottimark:
This pull request fixes #785

As in my previous pull request, there was an unexpected change to delete a space at the end of lines 10 and 11. When I saved the file from Atom editor? I hope that is not a problem.

Thank you for inviting me to make a pull request for the issue. The practice is very helpful to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@ It helps you write applications that behave consistently, run in different envir
 You can use Redux together with [React](https://facebook.github.io/react/), or with any other view library.  
 It is tiny (2kB) and has no dependencies.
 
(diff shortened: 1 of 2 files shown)


gaearon:
Thank you!

PR #795, opened 2015-09-24
Fix bug caused by inconsistent Todo state across components in the Basic chapter of the Docs

bottledsmoke:
The tutorial for the TodoMVC app in documentation, the visibleTodos prop is passed to TodoList, which is mapped into a new array -- which I will call visibleTodos -- and rendered into Todo components. Upon doing so, the visibleTodos array now is at risk of becoming unlinked with the state.todos array as soon as a single item is completed. 

Because the completeTodo reducer relies on parity between the indexes of the visibleTodos array and the state.todos array, bugs can occur when in trying to dispatch a completeTodo action in SHOW_ACTIVE view mode. The bug occurs because the visibleTodos array can be shorter than the state.todos array, and any index in the list can be filtered out once completed. 

In the screen shot below, "third" has already been completed. When clicking on "fourth" todo  -- the fourth item in the state.todos array and third item in the visibleTodos array, the payload informs the app that it should complete the third item in the array, so nothing happens. 
screen shot 2015-09-24 at 10 20 41 am

Likewise, when clicking on the "fifth" todo --- fifth in state.todos but fourth in visibleTodos, it completes the "fourth" item by dispatching a payload targeting the index: 3.
screen shot 2015-09-24 at 10 21 03 am

I have crafted two solutions. The first is to map a uuid to each todo and have the completeTodo action splice the array using the index returned by Array.prototype.findIndex(todo.id) instead of the current position of the Todo in the visibleTodos array. This solution fixes the inaccuracies of the reducer, but it also requires actions and the reducer to be rewritten and another module to be installed.  I'm not a fan.

Instead, why not pass the filtering mechanisms themselves to TodoList component via props by altering the selectTodos helper to return functions instead of filtered arrays? Now we are free to pass the entire state.todos array to TodoList and simply not render any todo within it that doesn't return true when passed to the filtering function. This keeps our state consistent across components without altering any reducers or actions. I believe this to be the elegant solution. It is included in this PR. 

My apologies for such a big switch.

### DIFF
--- a/docs/basics/ExampleTodoList.md
+++ b/docs/basics/ExampleTodoList.md
@@ -127,7 +127,7 @@ import Footer from '../components/Footer';
 class App extends Component {
   render() {
     // Injected by connect() call:
-    const { dispatch, visibleTodos, visibilityFilter } = this.props;
+    const { dispatch, todos, filteringCriteria, visibilityFilter } = this.props;
     return (
       <div>
         <AddTodo
@@ -135,7 +135,8 @@ class App extends Component {
             dispatch(addTodo(text))
           } />
         <TodoList
-          todos={visibleTodos}
+          todos={todos}
+          filteringCriteria={filteringCriteria}
           onTodoClick={index =>
             dispatch(completeTodo(index))
           } />
@@ -150,6 +151,7 @@ class App extends Component {
 }
 
 App.propTypes = {
+  filteringCriteria: PropTypes.func.isRequired,
   visibleTodos: PropTypes.arrayOf(PropTypes.shape({
     text: PropTypes.string.isRequired,
     completed: PropTypes.bool.isRequired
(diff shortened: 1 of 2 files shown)


gaearon:
Woah, that's a really good catch.

However I'd rather opt for generating unique IDs when creating todos.
This sidesteps the problem completely without the need for passing functions, and also corresponds to how you'd do it in a real app.

What do you think?

bottledsmoke:
Yes, I got the id-driven approach to work by tweaking the todo reducer and the createTodo action in [this branch](https://github.com/bottledsmoke/redux/tree/id-solution), and this sort of approach is much more likely to pay off as the app scales. It doesn't solve the issue of their being two variants the todos array floating around that can possibly interfere with each other, but it does teach an id-based approach to newcomers. 

My preference for the function-passing solution as opposed to the id-passing solution is based on looking at what the fleshed-out TodoMVC example was doing in the /examples directory and making an attempt to bring the basic example in line with it -- all the while keeping the dumb components as dumb as they possibly could be. I'm uncertain how performant of a solution it would be, but I like knowing that the state is pure, and I like knowing that it is possible using simple methods and operations. 

And yet in the end, I'm uncertain which way would be best to use as a tutorial for someone who is unfamiliar with React + Flux-ish things. On one hand, the approach of the tutorial right now emphasizes the functional basis of Redux in a way that plucking items out of a tree by their ID doesn't. As the examples grow more complex, introducing Ids is inevitable; however, for the first exercise, it may be a bit overkill; on the other hand, maybe it's a great spot to introduce the concept? You'll make the right call. You've already done such an amazing job with all of this. One of the main reasons I'm all for Redux is how damn well designed and composed the documentation is in the first place.

gaearon:
Closing as this particular PR is outdated.
If you feel like sending a PR for ID solution, please let us know in https://github.com/rackt/redux/issues/1129!

gaearon:
This is fixed in all examples now.

Issue #792, opened 2015-09-24
Add the @@redux/ prefix to randomly generated action type in combineReducers


gcanti:
Hi, I'm trying to add type safety to actions and state with the [redux-tcomb](https://github.com/gcanti/redux-tcomb) project.
For what concerns the actions I need to detect the redux specific internal actions in order to skip the type checking (for example the redux's `ActionTypes.INIT` action). Currently I have an issue with the `combineReducers` function which probes all reducers to be combined with a randomly generated action type to ensure each reducer returns undefined for action types that should be unrecognized by reducers (relevant LOC: https://github.com/rackt/redux/blob/master/src/utils/combineReducers.js#L88).

Is it possible to add the `@@redux/` prefix to those randomly generated action types?

gaearon:
Can you give a relevant snippet of code? What gets wrapped into type checking, at which point does checking occur, etc?

The original motivation was to discourage people from trying to handle `@@redux/` actions explicitly even by checking `substring`. We can add the prefix to the random action but then our check becomes weaker.

gcanti:
> Can you give a relevant snippet of code

Sure, I think it's a general issue for each library which wants to implement type-checked actions, so I'll try to keep the discussion as general as possible.

Let's start with a normal reducer (no type checking)

```js
function todoReducer(state, action) {
  return state || [];
}

var reducer = combineReducers({
  todos: todoReducer
});

assert.deepEqual(reducer(undefined, {type: 'ADD_TODO', text: 1}), {todos: []}); // ok
```

Sweet! Everything's fine. Now say I want to add type checking to the `todoReducer`:

```js
function todoReducer(state, action) {
  // I want to type check the action argument here
  return state || [];
}
```

In redux-tcomb you define a type `Action` which is the union of all the user defined actions:

```js
function todoReducer(state, action) {
  Action(action); // <= this must throw if the action argument contains an unknown action
  return state || [];
}
```

The first problem I encountered is that the `todoReducer` will throw when the `@@redux/INIT` action will be passed into, so I must add a check, say a `isReduxAction` predicate, which will skip the type checking in presence of an internal redux action:

```js
function isReduxAction(action) {
  // rule out the @@redux/INIT action
  return action.type.indexOf('@@redux/') === 0;
}
```

If I add this check to my reducer, again everything will be fine:

```js
function todoReducer(state, action) {
  if (!isReduxAction(action)) {
    Action(action); // <= type check only if the action is not an internal one 
  }
  return state || [];
}
[...]

PR #796, opened 2015-09-24
Add prefix to randomly generated action type in combineReducers #792

gcanti:
Addresses #792


timdorr:
:+1: I'll let @gaearon decide on the merge since he's closer to the original discussion.

PR #801, opened 2015-09-25
Migrating docs: point to `createStore()` source file

julen:
Feel free to close the PR if you disagree, although since it suggests to check the `createStore` implementation, I think it makes more sense to link to it, rather than to the API docs.


theaqua:
I think more useful have both links. Docs explains idea, but source shows code behind the idea.

gaearon:
Yes, I would rather have both. You can add " (source)" after the API link where "source" itself is a link.

julen:
Thank you for the feedback, updated accordingly.

PR #812, opened 2015-09-27
wording: combineReducers.md

chentsulin:
may -> must


gaearon:
AFAIK "may never" is a correct form...
Summoning English native speakers to this thread. :gb:

chentsulin:
I am not sure for this, but just want to correspond to `the initial state must not be undefined either`. If any native speakers could give a lesson here is also great, haha

ronag:
`must` or `shall` are the usual terms in requirements specification.

ronag:
`If the `state` given to it is `undefined` it must return the initial state for this specific reducer which, according to the previous rule, also must not be `undefined`.`

PR #817, opened 2015-09-28
New Take on adding Flow Types

leoasis:
Alright, another take, but not yet complete :(

Here I mostly made the counter example work with flow types (even types for redux-thunk and react-redux, with some caveats).

I've opted for just moving on with type declaration modules instead of types in the redux src directly since I've had problems with the `name_mapper` flow option (more details on that particular line). When we solve that, and a couple more problems that hopefully we can all solve together, we can be ready to add Flow types to redux!

And if not, then it will be here for posterity to perhaps try again in a future Flow version.

Pending stuff here:

* [ ] Type src files with Flow as much as possible
* [ ] https://github.com/facebook/flow/issues/844 (fixed, but not yet released)
* [x] ~~https://github.com/facebook/flow/issues/582 (supporting union of unions, not yet fixed) This requires us to include the redux INIT action as part of the action types in userland.~~
* [ ] https://github.com/facebook/flow/issues/880 (export default functions alongside with types). Currently worked around it by defining global types similar to what they are doing here: https://github.com/facebook/flow/blob/master/lib/node.js#L903. Doing this until we have an answer.
* [x] ~~https://github.com/facebook/flow/issues/879 (type signature to signify an object type being a merge from two other object types). No workaround for this for now~~

### DIFF
--- a/examples/flow-counter/.flowconfig
+++ b/examples/flow-counter/.flowconfig
@@ -1,0 +1,16 @@
+[ignore]
+.*/node_modules/
+# .*/node_modules/webpack/.*
+# .*/node_modules/redux/examples/.*
+# .*/node_modules/redux/lib/.*
+# Ignoring analyzing redux for now since I can't find a way for it to not also select redux-thunk (need a better regex)
+# module.name_mapper='^\(redux\)$' -> '\1/src/index'
+
+[include]
+
+[libs]
+interfaces/
+
+
+[options]
+module.system=node
--- a/examples/flow-counter/.flowconfig
+++ b/examples/flow-counter/.flowconfig
@@ -1,10 +1,5 @@
 [ignore]
 .*/node_modules/
-# .*/node_modules/webpack/.*
-# .*/node_modules/redux/examples/.*
-# .*/node_modules/redux/lib/.*
-# Ignoring analyzing redux for now since I can't find a way for it to not also select redux-thunk (need a better regex)
-# module.name_mapper='^\(redux\)$' -> '\1/src/index'
 
 [include]
 
--- a/examples/flow-counter/actions/counter.js
+++ b/examples/flow-counter/actions/counter.js
@@ -1,7 +1,8 @@
 /* @flow */
 
 import type { Action, AppState } from '../types';
-import type { Thunk } from 'redux-thunk';
+
+type Thunk<S, D> = reduxThunk$Thunk<S, D>;
 
 export const INCREMENT_COUNTER = 'INCREMENT_COUNTER';
 export const DECREMENT_COUNTER = 'DECREMENT_COUNTER';
--- a/examples/flow-counter/actions/counter.js
+++ b/examples/flow-counter/actions/counter.js
@@ -1,0 +1,39 @@
+/* @flow */
+
+import type { Action, AppState } from '../types';
+import type { Thunk } from 'redux-thunk';
+
+export const INCREMENT_COUNTER = 'INCREMENT_COUNTER';
+export const DECREMENT_COUNTER = 'DECREMENT_COUNTER';
+
+export function increment() : Action {
+  return {
+    type: INCREMENT_COUNTER
+  };
+}
+
+export function decrement() : Action {
+  return {
+    type: DECREMENT_COUNTER
+  };
+}
+
+export function incrementIfOdd() : Thunk<AppState, Action> {
+  return (dispatch, getState) => {
+    let { counter } = getState();
+
+    if (counter % 2 === 0) {
+      return;
+    }
+
+    dispatch(increment());
+  };
+}
+
+export function incrementAsync(delay: number = 1000) : Thunk<AppState, Action> {
+  return dispatch => {
+    setTimeout(() => {
+      dispatch(increment());
+    }, delay);
+  };
+}
--- a/examples/flow-counter/components/Counter.js
+++ b/examples/flow-counter/components/Counter.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import React, { Component, PropTypes } from 'react';
+import React, { Component } from 'react';
 
 export type CounterProps = {
   counter: number,
--- a/examples/flow-counter/components/Counter.js
+++ b/examples/flow-counter/components/Counter.js
@@ -1,0 +1,32 @@
+/* @flow */
+
+import React, { Component, PropTypes } from 'react';
+
+export type CounterProps = {
+  counter: number,
+  increment: () => any,
+  decrement: () => any,
+  incrementIfOdd: () => any,
+  incrementAsync: () => any,
+};
+
+class Counter extends Component<any, CounterProps, any> {
+  render() : ReactElement<any, any, any> {
+    let { increment, incrementIfOdd, incrementAsync, decrement, counter } = this.props;
+    return (
+      <p>
+        Clicked: {counter} times
+        {' '}
+        <button onClick={increment}>+</button>
+        {' '}
+        <button onClick={decrement}>-</button>
+        {' '}
+        <button onClick={incrementIfOdd}>Increment if odd</button>
+        {' '}
+        <button onClick={() => incrementAsync()}>Increment async</button>
+      </p>
+    );
+  }
+}
+
+export default Counter;
--- a/examples/flow-counter/components/Counter.js
+++ b/examples/flow-counter/components/Counter.js
@@ -10,7 +10,7 @@ export type CounterProps = {
   incrementAsync: () => any,
 };
 
-class Counter extends Component<any, CounterProps, any> {
+class Counter extends Component<{}, CounterProps, {}> {
   render() : ReactElement<any, any, any> {
     let { increment, incrementIfOdd, incrementAsync, decrement, counter } = this.props;
     return (
--- a/examples/flow-counter/containers/App.js
+++ b/examples/flow-counter/containers/App.js
@@ -1,0 +1,36 @@
+/* @flow */
+
+import { connect } from 'react-redux';
+import Counter from '../components/Counter';
+import type { CounterProps } from '../components/Counter';
+import { increment, decrement, incrementIfOdd, incrementAsync } from '../actions/counter';
+import { AppState, Dispatchable } from '../types';
+import type { Dispatch } from 'redux';
+import type { Component } from 'react';
+
+type StateProps = AppState;
+
+function mapStateToProps(state : AppState) : StateProps {
+  return {
+    counter: state.counter
+  };
+}
+
+type DispatchProps = {
+  increment: () => any,
+  decrement: () => any,
+  incrementIfOdd: () => any,
+  incrementAsync: () => any
+}
+
+function mapDispatchToProps(dispatch: Dispatch<Dispatchable>) : DispatchProps {
+  return {
+    increment: (...args) => dispatch(increment(...args)),
+    decrement: (...args) => dispatch(decrement(...args)),
+    incrementIfOdd: (...args) => dispatch(incrementIfOdd(...args)),
+    incrementAsync: (...args) => dispatch(incrementAsync(...args)),
+  };
+}
+// TODO: Cannot make this to correctly type check that the resulting props match
+// the Counter prop types
+export default connect(mapStateToProps, mapDispatchToProps)(Counter);
--- a/examples/flow-counter/containers/App.js
+++ b/examples/flow-counter/containers/App.js
@@ -2,11 +2,9 @@
 
 import { connect } from 'react-redux';
 import Counter from '../components/Counter';
-import type { CounterProps } from '../components/Counter';
 import { increment, decrement, incrementIfOdd, incrementAsync } from '../actions/counter';
-import { AppState, Dispatchable } from '../types';
+import type { AppState, Dispatchable } from '../types';
 import type { Dispatch } from 'redux';
-import type { Component } from 'react';
 
 type StateProps = AppState;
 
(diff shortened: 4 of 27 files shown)


examples/flow-counter/.flowconfig
```diff
@@ -0,0 +1,16 @@
...
+# module.name_mapper='^\(redux\)$' -> '\1/src/index'
+
+[include]
+
+[libs]
+interfaces/
+
+
+[options]
+module.system=node
```
leoasis: As you can see, I removed the line `module.name_mapper='^\(redux\)$' -> '\1/src/index'` and opted to continue with Flow declaration modules (the `[libs]` option) since that regex was conflicting with `redux-thunk`, and I was not able to create a regex that wouldn't stop that problem. If anyone can give this a try, or discuss approaches, let's start a thread here.

examples/flow-counter/containers/App.js
```diff
@@ -0,0 +1,36 @@
...
+  return {
+    increment: (...args) => dispatch(increment(...args)),
+    decrement: (...args) => dispatch(decrement(...args)),
+    incrementIfOdd: (...args) => dispatch(incrementIfOdd(...args)),
+    incrementAsync: (...args) => dispatch(incrementAsync(...args)),
+  };
+}
+// TODO: Cannot make this to correctly type check that the resulting props match
+// the Counter prop types
+export default connect(mapStateToProps, mapDispatchToProps)(Counter);
```
leoasis: I still need to make this type check correctly, I cannot find a way to specify in the type signature that the resulting props should be the merge of `StateProps` and `DispatchProps` from above (both should result in `CounterProps` that is the props type for the `Counter` component, but for some reason if I change any of the types from here, it still passes with no errors.

examples/flow-counter/interfaces/react-redux.js
```diff
@@ -0,0 +1,12 @@
+import type { Dispatch } from 'redux';
+import type { Component } from 'react';
+
+declare module 'react-redux' {
+  declare type MapStateToProps<State, Props> = (state: State) => Props;
+  declare type MapDispatchToProps<Dispatchable, Props> = (dispatch: Dispatch<Dispatchable>) => Props;
+
+  declare function connect<State, Dispatchable, ComponentProps, DefaultProps, ComponentState, StateProps, DispatchProps: $Diff<ComponentProps, StateProps>>(
```
leoasis: Here I tried to tell flow that `DispatchProps + StateProps == ComponentProps`, using an internal feature I found here in Flow's React type declaration: https://github.com/facebook/flow/blob/master/lib/react.js#L131, but with no luck.

examples/flow-counter/interfaces/redux-thunk.js
```diff
@@ -0,0 +1,13 @@
+import type { Middleware, MiddlewareAPI, Dispatch, Store } from 'redux';
+
+declare module 'redux-thunk' {
+  declare type Thunk<State, Dispatchable> = (dispatch: Dispatch<Dispatchable>, getState: () => State) => ?Dispatchable;
+  declare type ThunkMiddlewareDispatchable<State, Dispatchable> = Thunk<State, Dispatchable> | Dispatchable;
+
+  // Cannot find a way to export the default function with the correct type signature
+  // declare var exports: {
```
leoasis: I couldn't find a way make the default module export be a polymorphic function in Flow. Does someone know how to do this? If not, probably is a bug or something yet unsupported in Flow.

examples/flow-counter/store/configureStore.js
```diff
@@ -0,0 +1,39 @@
...
+import { createStore, applyMiddleware } from 'redux';
+import type { Middleware, MiddlewareAPI, Dispatch, Store, ReduxAction } from 'redux';
+// import thunk from 'redux-thunk';
+import type { Thunk, ThunkMiddlewareDispatchable } from 'redux-thunk';
+import reducer from '../reducers';
+import type { AppState, Action } from '../types';
+
+// Re-defining redux-thunk here first to show how to type it, and also
+// because I don't know how to put the type in the declaration module.
+function thunk<State, Dispatchable: Object>(_ref: MiddlewareAPI<State, ThunkMiddlewareDispatchable<State, Dispatchable>>) : (next: Dispatch<Dispatchable>) => Dispatch<ThunkMiddlewareDispatchable<State, Dispatchable>> {
```
leoasis: Since I couldn't define the type for the default export of `redux-thunk`, I copy-pasted the code and _typed_ it here. The good thing is that we now have types for redux-thunk that we can then move to the repo.

examples/flow-counter/types.js
```diff
@@ -0,0 +1,18 @@
...
+
+export type CounterState = number;
+
+export type AppState = {
+  counter: CounterState
+}
+
+// Need to explicitly put this here due to a bug in Flow: https://github.com/facebook/flow/issues/582
+export type Action =
+  { type: '@@redux/INIT'} |
```
leoasis: As noted here, I still need to specify this internal type because of the bug noted above that makes Flow not recognize nested union types correctly.

examples/flow-counter/actions/counter.js
```diff
@@ -0,0 +1,39 @@
...
+
+export function decrement() : Action {
+  return {
+    type: DECREMENT_COUNTER
+  };
+}
+
+export function incrementIfOdd() : Thunk<AppState, Action> {
+  return (dispatch, getState) => {
+    let { counter } = getState();
```
leoasis: I changed this `const` to a `let` because of this bug in Flow, that is now fixed in master: https://github.com/facebook/flow/issues/844

examples/flow-counter/containers/App.js
gaearon: Shall we ask somebody from Flow team?

examples/flow-counter/interfaces/redux-thunk.js
gaearon: Shall we file an issue?

examples/flow-counter/interfaces/redux.js
```diff
@@ -0,0 +1,27 @@
...
+  declare function createStore<State, Action>(reducer: Reducer<State, Action>, initialState?: State) : Store<State, Action, ReduxAction<Action>>;
+
+  declare type MiddlewareAPI<State, Dispatchable> = {
+    dispatch: Dispatch<Dispatchable>,
+    getState: () => State,
+  };
+
+  declare type Middleware<State, Dispatchable, DispatchableNext> = (api: MiddlewareAPI<State, Dispatchable>) => (next: Dispatch<DispatchableNext>) => Dispatch<Dispatchable>;
+
+  declare function applyMiddleware<State, Action, Dispatchable, DispatchableNext>(middleware: Middleware<State, Dispatchable, DispatchableNext>)
```
gaearon: You'd need to call `applyMiddleware` many times to have several middleware while staying typed, right? Seems like a good enough compromise to me.

gaearon:
Hey, thanks for your work on this again. I feel like we can get it right with three or four more iterations :wink: Can I ask you to write up a summary of the blocker issues in Flow that currently prevent us from finalizing this and recommending this setup to Flow users?

examples/flow-counter/containers/App.js
leoasis: Yeah I think so, I'll create an issue in their repo so that we can get feedback both here and there.

examples/flow-counter/interfaces/redux-thunk.js
leoasis: Yup! will do that as well

examples/flow-counter/interfaces/redux.js
leoasis: I can define a couple of overrides, was just testing with one (since the counter example just uses one middleware), but it should be possible to define multiple, as we discussed in the other thread (perhaps up to 5, 6 middleware functions)

examples/flow-counter/containers/App.js
leoasis: Just brainstorming a possible solution to work around this problem, but with a big BUT which involves changing the connect API, would be to just have a single function be passed as an argument to `connect`, which receives the `state`, `dispatch`, and the parent `props`, pretty much like the async functions in `redux-thunk`.

Just thinking out loud, not necessarily meaning it is the only solution

examples/flow-counter/interfaces/redux-thunk.js
leoasis: Filed here: https://github.com/facebook/flow/issues/880

examples/flow-counter/containers/App.js
gaearon: I see. We can't do this because it's performance worst case though. We don't want to bind on every prop or state change unless completely necessary.

examples/flow-counter/components/Counter.js
```diff
@@ -0,0 +1,32 @@
+/* @flow */
+
+import React, { Component, PropTypes } from 'react';
```
tcoopman: Why do you import `PropTypes`?

leoasis:
Currently the issues that are blocking us in Flow are:

* https://github.com/facebook/flow/issues/844 (fixed, but not yet released)
* ~~https://github.com/facebook/flow/issues/582 (supporting union of unions, not yet fixed) This requires us to include the redux INIT action as part of the action types in userland.~~ NOTE: Just tried this one locally and it seems to be working fine now, will push an update
* https://github.com/facebook/flow/issues/880 (export default functions alongside with types). Currently worked around it by defining global types similar to what they are doing here: https://github.com/facebook/flow/blob/master/lib/node.js#L903. Doing this until we have an answer.
* https://github.com/facebook/flow/issues/879 (type signature to signify an object type being a merge from two other object types). No workaround for this for now

The recommended setup for Flow users would be to include the module declaration files as part of their `.flowconfig` files (under the `[libs]` option). Those module declaration files should be included in each library (`redux`, `react-redux`, `redux-thunk`, etc)

In order to not forget to maintain these, it would be nice to type check the source code of the libraries internally and generate that declaration module, but that is currently not possible (https://github.com/facebook/flow/issues/714). So I'm afraid this would be something to pay attention when changing the API of any lib :(.

examples/flow-counter/components/Counter.js
leoasis: Ah forgot to remove this from the original example, thanks!

examples/flow-counter/containers/App.js
leoasis: Ah you're right, forgot about those perf optimizations. Will think a bit more about a possible workaround then, or we can just wait for the Flow team to show us the way

examples/flow-counter/index.js
```diff
@@ -0,0 +1,15 @@
...
+import React from 'react';
+import { Provider } from 'react-redux';
+import App from './containers/App';
+import configureStore from './store/configureStore';
+
+const store = configureStore();
+
+React.render(
+  // Flow limitation: Does not detect children type if we don't explicitly
+  // set the `children` prop
```
leoasis: This is being tracked here: https://github.com/facebook/flow/issues/892

examples/flow-counter/interfaces/react-redux.js
```diff
@@ -0,0 +1,22 @@
...
+    mapStateToProps: MapStateToProps<State, StateProps>,
+    mapDispatchToProps: MapDispatchToProps<Dispatchable, DispatchProps>
+  ) : (component: Component<DefaultProps, ComponentProps, ComponentState>) => Component<DefaultProps, ComponentProps, ComponentState>;
+
+  declare type ProviderProps<S, D> = {
+    // TODO: For some super weird reason I need to use this global declaration of the store,
+    // otherwise this doesn't run type checks (never throws an error).
+    // If I use the Store type imported from 'redux' here, it will treat it as "any"
+    store: redux$Store<S, D>,
```
leoasis: I'm not completely sure why this is happening, and not even sure how to phrase the issue to create in the Flow repo. Seems like a very specific edge case. I'll see if I can find a smaller example to reproduce, and if not file an issue just with this

examples/flow-counter/containers/App.js
```diff
@@ -0,0 +1,33 @@
...
+function mapDispatchToProps(dispatch: Dispatch<Dispatchable>) : DispatchProps {
+  return {
+    increment: () => dispatch(increment()),
+    decrement: () => dispatch(decrement()),
+    incrementIfOdd: () => dispatch(incrementIfOdd()),
+    incrementAsync: () => dispatch(incrementAsync()),
+  };
+}
+
+export default connect(mapStateToProps, mapDispatchToProps)(Counter);
```
leoasis: Finally made this to type check!!! The `$Diff` type was working, the problem was that the correct flow type for classes that extend `React.Component` is `ReactClass`, so that it knows how to infer the polymorphic props and state types

examples/flow-counter/interfaces/react-redux.js
```diff
@@ -0,0 +1,23 @@
...
+
+
+declare module 'react-redux' {
+  declare type MapStateToProps<State, Props> = (state: State) => Props;
+  declare type MapDispatchToProps<Dispatchable, Props> = (dispatch: Dispatch<Dispatchable>) => Props;
+
+  declare function connect<State, Dispatchable, StateProps, ComponentProps, DispatchProps: $Diff<ComponentProps, StateProps>, ComponentDefaultProps, ComponentState>(
+    mapStateToProps: MapStateToProps<State, StateProps>,
+    mapDispatchToProps: MapDispatchToProps<Dispatchable, DispatchProps>
+  ) : (component: ReactClass<ComponentDefaultProps, ComponentProps, ComponentState>) => ReactClass<ComponentDefaultProps, ComponentProps, ComponentState>;
```
leoasis: Here's the fix to make `connect` type check. Using `ReactClass` instead of `Component` or `ReactComponent`. The `$Diff` internal type works, though it is still internal, so we should pay attention to the evolution of it in Flow

examples/flow-counter/interfaces/react-redux.js
leoasis: Seems that `import type` in the module declarations never worked at all: https://github.com/facebook/flow/issues/905. So for now working with globally declared types

iamdustan:
@leoasis have you looked at this more recently? I believe that the majority of those issues in flow has been fixed, in addition to the ability for flow to type check files in node_modules (files with `.flow` extensions)

leoasis:
@iamdustan no, haven't been able to, but will definitely try to take a look into it sometime soon. If interested, let me know and we can discuss what needs to be done to get this

iamdustan:
i definitely am interested, just not sure on current availability :(

leoasis:
If I could get back to this (or someone else), would you be willing to have the flow types directly in code? Or would you prefer to have them as external declaration files? What would you guys prefer?

gaearon:
I don’t care either way as long as we can meaningfully export them for consumers.

iamdustan:
If the source files are copied to the npm distro with .flow extensions flow will automatically pick them up. :)

gaearon:
I am closing for inactivity and because this hasn’t been updated for quite a while.
If somebody wants to pick this up, please do, and open another pull request!

jfrolich:
Interested in this as well!

PR #820, opened 2015-09-30
Update Reducers.md

ch3lo:
Link to an issue with a clear explanation of why is necessary to return a clone or new entity

### DIFF
--- a/docs/basics/Reducers.md
+++ b/docs/basics/Reducers.md
@@ -151,7 +151,7 @@ case COMPLETE_TODO:
   });
 ```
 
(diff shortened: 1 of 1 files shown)


timdorr:
Unfortunately, there is a typo and grammar problem. 

But rather than patching those up, perhaps it's a best if we don't hide something relatively important like this in a link out of the documentation. Instead of referring to a random comment on a random issue, I'd link back to [the part of the docs that cover it best](https://github.com/rackt/redux/blob/master/docs/introduction/ThreePrinciples.md#mutations-are-written-as-pure-functions). That keeps people from diverging away from the docs and ensures we're being consistent with our explanations.

gaearon:
I agree it's better to link to docs than to issues.

gaearon:
I'm closing because we'd rather explain things in docs than link to issues.

PR #825, opened 2015-09-30
Remove all unnecessary files from npm package

nkbt:
Current version of Redux being installed has lots of unnecessary files:

With this update it looks like:

Tested by installing locally via `npm install ~/nkbt/redux`


.npmignore
```diff
@@ -1,8 +1,14 @@
-.DS_Store
 *.log
-examples
-test
-coverage
-_book
-book.json
-docs
+/.*/
```
nkbt: `/.*/` ignores some config folders like `.idea`

.npmignore
```diff
@@ -1,8 +1,14 @@
-.DS_Store
```
nkbt: `.DS_Store` is removed because it is ignored by default:
https://docs.npmjs.com/misc/developers

```
By default, the following paths and files are ignored, so there's no need to add them to .npmignore explicitly:

    .*.swp
    ._*
    .DS_Store
    .git
    .hg
    .npmrc
    .lock-wscript
    .svn
    .wafpickle-*
    config.gypi
    CVS
    npm-debug.log
```

.npmignore
chicoxyzzy: I am strictly disagree. This is developers own folder and it is not in project. Any application can create any folders in project and that doesn't mean we need to ignore all possible folder names. You should add `.idea` to your global gitignore

nkbt: This is .npmignore not .gitignore. Otherwise I left the same comment as you for elementalUI last week ;)

nkbt: @chicoxyzzy https://github.com/elementalui/elemental/pull/65/files#r40164567

chicoxyzzy: @nkbt `npm` ignores stuff matched by `.gitignore`

chicoxyzzy: my bad. looks like it really should be there

nkbt: :+1:

gaearon:
:+1:

PR #828, opened 2015-10-01
Change name from index.js to configureStore.js

abalone0204:
In line 384 and line 475, `index.js` means nothing.
I think we should put the middleware setting codes in `configureStore.js` ,
instead of `index.js`. 
:D


gaearon:
In examples, `configureStore.js` is a file exporting a function. There's no function export in the inline examples in docs, and I don't think we should complicate them with it.

`index.js` just means an entry point, which serves to explain that there's just one store and it's created in entry point (or file referenced from it).

I'd keep `index.js` for this reason.

abalone0204:
Thanks for your explanation for this.
I can understand why you use `index.js` here now.

PR #836, opened 2015-10-02
Added posts to rootReducer

sholden:
Also changed the order of the reducers to match the definitions above.


gaearon:
`posts` is not included by purpose.
It is *used* by `postsByReddit`—it's not meant to be a second-level reducer.

Think of `combineReducers()` as simple utility: you can use it to combine a few reducers in a very specific way, but you can also forgo it and compose them in a different way, e.g. how `postsByReddit` delegates individual lists to `posts`.

PR #840, opened 2015-10-03
Update Actions.md

npverni:
Using a constant here seemed more consistent with further recommendations and examples.


gaearon:
Let's put

```js
const ADD_TODO = 'ADD_TODO'
```

just before it to make it super clear we're talking about strings.

npverni:
Good call, done.

PR #845, opened 2015-10-04
Update babel config for new transform format

bvasko:
## Issue

When using v1.1.0 of the babel-plugin-react-transform, it throws warnings that the format for transforms is outdated and needs to be updated.

## Job Story

N/A

## Solution

Update the babel-plugin-react-transform config format to the new, non-deprecated format from upstream. (v1.1.0+)

### DIFF
--- a/examples/async/.babelrc
+++ b/examples/async/.babelrc
@@ -6,11 +6,13 @@
         "react-transform"
       ],
       "extra": {
-        "react-transform": [{
-          "target":  "react-transform-hmr",
+        "react-transform": {
+          "transforms": [{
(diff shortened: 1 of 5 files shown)


bvasko:
This breaks hot module reloading...

gaearon:
Hi, thanks for the PR!
`imports` and `locals` are related to `react-transform-hmr` in particular, so they should be together:

```
        "react-transform": {
          "transforms": [{
            "transform":  "react-transform-hmr"
            "imports": ["react"],
            "locals":  ["module"]
          }],
```

Would you mind sending another PR amending that?

PR #847, opened 2015-10-05
Update babel config for new transform format

bvasko:
## Issue

When using v1.1.0 of the babel-plugin-react-transform, it throws warnings that the format for transforms is outdated and needs to be updated.

## Job Story

N/A

## Solution

Update the babel-plugin-react-transform config format to the new, non-deprecated format from upstream. (v1.1.0+)

### DIFF
--- a/examples/async/.babelrc
+++ b/examples/async/.babelrc
@@ -6,11 +6,13 @@
         "react-transform"
       ],
       "extra": {
-        "react-transform": [{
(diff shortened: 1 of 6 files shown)


gaearon:
Thank you!

bvasko:
Should I update the babel plugin to '^1.1.0' in the examples package.json files as well?

gaearon:
Good catch, yes please do!

PR #855, opened 2015-10-06
Add wallaby.js config

aaronjensen:
redux

http://wallabyjs.com/ is a great tool for TDD that I like to use. Up to you if you want it in your project, but i'd love to continue to spread awareness of it.

It's like hot reloader for your tests :smile:

### DIFF
--- a/wallaby.conf.js
+++ b/wallaby.conf.js
@@ -1,0 +1,28 @@
+var babel = require('babel');
+
+module.exports = function wallabyConfig(wallaby) {
+  var babelCompiler = wallaby.compilers.babel({ babel: babel, stage: 0 });
+
+  return {
+    debug: true,
+    files: [
+      'src/**/*.js',
+      'test/helpers/*.js',
+    ],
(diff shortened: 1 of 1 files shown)


gaearon:
This looks neat but I don't see how it's fundamentally different from Mocha we use as the test runner. It is also commercial, which isn't a problem per se, but which means I can't even maintain this file because I don't want to buy it.

aaronjensen:
In that it runs tests and tells you if they fail or pass it is the same as mocha watch. It's different in that it:  

* coverage in files
* updates as you type (no saving necessary)
* shows failures and console logs inline, in the js files

But yeah, I can see how you wouldn't want to buy it to maintain it if you don't see value in it. Plus, I think you use Sublime, yeah? There's no sublime version yet, just atom/VS/intellij, I think.

gaearon:
Yeah I use Sublime. In general I think one would add these tools if maintainers use them, but we currently don't.

PR #856, opened 2015-10-06
Have combineReducers return same object if nothing changes

### LINKED ISSUE #853: Should `combineReducers` return the same object if nothing changes?
Right now, the reducer returned by `combineReducers` returns a new object regardless of whether or not the reducers it is combining make any changes. This means breaking reference equal every action, which limits its usefulness when using nested combinations unless you always select all the way down to the piece of state that maintains reference equality integrity. It'd be a little bit more slow to introduce an equality check, after the `mapValues` but it seems like it'd be better, yeah?

aaronjensen:
Fixes #853

### DIFF
--- a/src/utils/combineReducers.js
+++ b/src/utils/combineReducers.js
@@ -113,13 +113,16 @@ export default function combineReducers(reducers) {
       throw sanityError;
     }
 
+    var hasChanged = false;
     var finalState = mapValues(finalReducers, (reducer, key) => {
-      var newState = reducer(state[key], action);
-      if (typeof newState === 'undefined') {
+      var previousStateForKey = state[key];
+      var nextStateForKey = reducer(previousStateForKey, action);
+      if (typeof nextStateForKey === 'undefined') {
         var errorMessage = getUndefinedStateErrorMessage(key, action);
         throw new Error(errorMessage);
       }
-      return newState;
+      hasChanged = hasChanged || nextStateForKey !== previousStateForKey;
+      return nextStateForKey;
     });
 
     if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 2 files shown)


test/utils/combineReducers.spec.js
```diff
@@ -112,6 +112,17 @@ describe('Utils', () => {
...
+      const reducer = combineReducers({
+        child(state = {}) {
+          return state;
+        }
+      });
+
+      const initialState = reducer(undefined, '@@INIT');
+      expect(reducer(initialState, { type: 'FOO' })).toBe(initialState);
+    });
+
```
gaearon: I'd like to see more tests.
We need to check that:

* With several reducers, having just one return something different is enough to cause inequality
* With several reducers, having all return the same is enough to maintain equality

src/utils/combineReducers.js
```diff
@@ -113,12 +113,16 @@ export default function combineReducers(reducers) {
       throw sanityError;
     }
 
+    var hasChanged = false;
+
```
gaearon: Let's remove this newline; this declaration is directly related to the logic just below.

src/utils/combineReducers.js
```diff
@@ -113,12 +113,16 @@ export default function combineReducers(reducers) {
...
     var finalState = mapValues(finalReducers, (reducer, key) => {
-      var newState = reducer(state[key], action);
+      var originalState = state[key];
+      var newState = reducer(originalState, action);
```
gaearon: I think it's a little confusing now that we have many `*State` variables, almost all of which refer to slices of state, but some refer to the state object itself.

Let's rename `originalState` and `newState` to `previousStateByKey` and `nextStateByKey`?

aaronjensen:
@gaearon I believe I've addressed your feedback, please let me know if there is anything else. Thanks!

gaearon:
This is good to go, thanks.

PR #858, opened 2015-10-06
Filter out undefined values from compose()

joshhunt:
Using redux-devtools, I would like to use this pattern to conditionally include the `devTools()` middleware:

```javascript
    applyMiddleware(createApiMiddleware(httpClient, apiClient)),
    __DEVTOOLS__ ? devTools() : undefined,
  )(createStore);
```

Unfortunately, this throws the error `TypeError: undefined is not a function` in compose.js.

In this PR, I've changed compose.js to filter out falsey values so you can use the inline ternery.

Tell me what you think, and if you would like any changes before accepting this.

P.S. I've never been all that enthused for TDD, but writing the (failing) test for this first, making the change and seeing it pass made creating this PR super simple :100:

### DIFF
--- a/src/utils/compose.js
+++ b/src/utils/compose.js
@@ -6,5 +6,7 @@
  * left. For example, compose(f, g, h) is identical to arg => f(g(h(arg))).
  */
 export default function compose(...funcs) {
-  return arg => funcs.reduceRight((composed, f) => f(composed), arg);
+  return arg => funcs
+    .filter(Boolean)
+    .reduceRight((composed, f) => f(composed), arg);
 }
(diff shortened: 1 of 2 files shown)


chentsulin:
`compose` is a common function and concept in functional programming, so this change will bring unexpected behavior to developers who is familiar with functional programming.

You can define your own `ignoreFalseyCompose`:
```js
function ignoreFalseyCompose(...args) {
  return compose(...args.filter(Boolean));
};
```

gaearon:
I think we discussed this before (although in regards to `applyMiddleware`).
We don't want special cases here.

They encourage you to write code you like you wrote above but there's a big problem: you're going to have `redux-devtools` code included in your production bundle this way. This is because you are not putting the import itself behind a conditional.

Instead you should do something like this:

```js
if (__DEVELOPMENT__ && __CLIENT__ && __DEVTOOLS__) {
    const { devTools, persistState } = require('redux-devtools');
    finalCreateStore = compose(
      applyMiddleware(middleware),
      devTools(),
    )(createStore);
  } else {
    finalCreateStore = applyMiddleware(middleware)(createStore);
  }
```

Note that `require` call is inside the condition so Uglify can cut it out.

Issue #862, opened 2015-10-07
Error in example app- redux/examples/real-world


eriknyk:
Hi I'm getting an error running real-world example, and it is the following:

![screen shot 2015-10-07 at 11 44 21 am](https://cloud.githubusercontent.com/assets/205721/10344434/dfb9968e-6ce8-11e5-8e9c-fc69770fd152.png)

and if I try:  /login

![screen shot 2015-10-07 at 11 48 11 am](https://cloud.githubusercontent.com/assets/205721/10344526/5bdfcab2-6ce9-11e5-9b38-825a21fbef2e.png)

gaearon:
1. Do you have a popup blocker? It tries to open a popup with DevTools.
2. What is the problem? Are you referring to `findDOMNode` warning?

eriknyk:
mainly was the error:
Cannot read property 'location' of undefined,
but when I unblock the popup, the error has gone.

thank you..

gaearon:
Would you like to send a PR to

1. Add a `try catch` when opening the window and do something like `console.error('Could not open Redux DevTools—are you running a popup blocker?');` in case of failure

2. Use ref as node directly instead of `React.findDOMNode()`?

eriknyk:
I just sent 1. https://github.com/rackt/redux/pull/869
for 2. the change should do it into redux-devtools repo, I tried into its repo, but I couldn't reproduce the error, I tried on master and next branches without luck. for latest oficial version, which branch should I use?

regards

PR #869, opened 2015-10-08
ISSUE #862, unhandled error when devTools popup window is blocked

eriknyk:
https://github.com/rackt/redux/issues/862

- adding try/catch when when devTools is throwing an error,
  usually when the popup is blocked


gaearon:
Can you add try-catch to `createDevToolsWindow` instead, on the particular line where it throws? This seems over-generic.

eriknyk:
DONE!

gaearon:
Thanks!

PR #870, opened 2015-10-08
Add permalink anchors to gitbook

jdlehman:
Solves part of Issue #345

### DIFF
--- a/book.json
+++ b/book.json
@@ -3,7 +3,7 @@
   "structure": {
     "summary": "docs/README.md"
   },
-  "plugins": ["edit-link", "prism", "-highlight"],
+  "plugins": ["edit-link", "prism", "-highlight", "anchors"],
   "pluginsConfig": {
     "edit-link": {
       "base": "https://github.com/rackt/redux/tree/master",


timdorr:
You actually don't need the plugin in the package.json file. Gitbook installs and maintains it for you. Can you update your commit with that removed? I think it should solve the merge conflict too.

gaearon:
There are two more issues here:

* Last time I checked, the plugin didn't generate nice looking links like in React docs. We'd want to style them to be nicer.
* This issue makes anchor links semi-useless and it's better to fix it first https://github.com/GitbookIO/gitbook/issues/920

jdlehman:
@timdorr cool didn't know that!

@gaearon Yeah... The anchors only work sometimes because of that issue.  I'll take a look at it.  I agree that the hover could use some custom styling.  Here is what it looks like by default:

anchors

gaearon:
Closing for inactivity.
Would be happy to take this if it looked like in React docs.

screen shot 2015-12-12 at 18 34 50
screen shot 2015-12-12 at 18 34 53

PR #873, opened 2015-10-09
Update UsageWithReact.md

camsong:
It should be `props` as we always called in React


gaearon:
“prop” is singular for “props”.
“as a props” doesn't seem like valid grammar.

Also, from the [official React docs](https://facebook.github.io/react/docs/thinking-in-react.html):

> the top of the hierarchy ( FilterableProductTable ) will take your data model **as a prop.**

camsong:
Thanks for the clarification.

PR #876, opened 2015-10-10
Add doc for testings async actions

takashi:


gaearon:
Great job. 
:+1:

takashi:
@gaearon thx for advising me and merging

phkavitha:
@takashi , I'm using the exact code for testing my async action creators. However, the following error is thrown when executing the `fetch(url)` inside the action creator.
```
- TypeError: Cannot read property 'then' of undefined
````
Can you please let me know how I can fix it.

gaearon:
Maybe fetch() doesn't return a Promise in your case? Please ask this on StackOverflow with a full example to reproduce the issue.

PR #878, opened 2015-10-10
Suggest to use applyMiddleware() for testing async action creators

gaearon:
This tweaks #876 to suggest the approach described in https://github.com/rackt/redux/issues/546#issuecomment-146755322 instead, as it works with any other middleware, and even with composed thunks.


takashi:
:+1:

PR #893, opened 2015-10-13
Update UsageWithReact.md

roylee0704:
amended comments to indicate that with React 0.14 and later version, there is not necessarily to wrap `<Provider>` child into a function,


gaearon:
We will soon update the docs to match 0.14 anyway, so I'd rather avoid this change for now.
Thanks!

PR #898, opened 2015-10-14
Remove useless code in applyMiddleware.js

zhou-yg:
The code is useless , the 'dispatch'  will override after compose(...chain) and in middlewareAPI.dispatch, the 'dispatch' is same as ‘dispatch’ which was overrided.


gaearon:
I did it this way to prevent Flow from complaining.
We didn't use Flow yet in the end, but we will eventually, so I'd like to keep this.

PR #901, opened 2015-10-15
Update examples and docs to React 0.14

gaearon:
Work in progress.

- [x] Release an update to React Redux @gaearon
- [x] Update all simple examples @gaearon
- [x] Update all docs @omnidan 
- [ ] Release an update to Redux DevTools @gaearon
- [ ] Update the real world example @gaearon
- [ ] Release an update to React Transform handling stateless components @gaearon
- [ ] Update examples to use the new React Transform @gaearon
- [ ] Convert components to be stateless where appropriate (we can do this as a separate pass later)

### DIFF
--- a/docs/advanced/ExampleRedditAPI.md
+++ b/docs/advanced/ExampleRedditAPI.md
@@ -10,9 +10,10 @@ This is the complete source code of the Reddit headline fetching example we buil
 import 'babel-core/polyfill';
 
 import React from 'react';
+import { render } from 'react-dom';
 import Root from './containers/Root';
 
-React.render(
+render(
   <Root />,
   document.getElementById('root')
 );
@@ -195,7 +196,7 @@ export default class Root extends Component {
   render() {
     return (
       <Provider store={store}>
-        {() => <AsyncApp />}
+        <AsyncApp />
       </Provider>
     );
   }
--- a/docs/api/applyMiddleware.md
+++ b/docs/api/applyMiddleware.md
@@ -173,10 +173,12 @@ function makeSandwichesForEverybody() {
 // This is very useful for server side rendering, because I can wait
 // until data is available, then synchronously render the app.
 
+import { renderToString } from 'react-dom/server';
+
 store.dispatch(
   makeSandwichesForEverybody()
 ).then(() =>
-  response.send(React.renderToString(<MyApp store={store} />))
+  response.send(renderToString(<MyApp store={store} />))
 );
 
 // I can also dispatch a thunk async action from a component
--- a/docs/basics/ExampleTodoList.md
+++ b/docs/basics/ExampleTodoList.md
@@ -8,6 +8,7 @@ This is the complete source code of the tiny todo app we built during the [basic
 
 ```js
 import React from 'react';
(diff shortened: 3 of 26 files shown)


gaearon:
I'd appreciate somebody making a PR *against this branch* to update all the `docs` to match the example updates I just pushed. Please keep the same code style and conventions. Where examples match the docs, it's best to use the exact code from examples.

If you intend to work on this, please write a comment in this PR so we don't duplicate the effort.

omnidan:
@gaearon I already updated the `Implementing Undo History` you can just revert your revert commit for that

omnidan:
I can work on updating the rest of the docs - the only things that need to be changed are `react-dom` and removing the function within `<Provider>`, right?

gaearon:
@omnidan

That's why I wrote

>Please keep the same code style and conventions

There are two issues in that doc commit:

1. It uses `ReactDOM.render` over importing named `render`. Let's say it's my personal whim. :wink:
2. It uses `findDOMNode` where the call needs to be removed completely: DOM refs *are* nodes in 0.14.

omnidan:
@gaearon aah okay - I'll do that then :smile:

gaearon:
>I can work on updating the rest of the docs - the only things that need to be changed are react-dom and removing the function within <Provider>, right?

These two, and removing `findDOMNode` whenever it is already a DOM node—that is, pretty much everywhere it's used right now. For testing docs, it involves using `react` instead of `react/addons`, and using `react-addons-test-utils`.

gaearon:
Thank you!

omnidan:
In `applyMiddleware.md`: `React.renderToString` is used, but `React` isn't even imported - is this a mistake in the current version of the docs?

I would just import `renderToString` from `react-dom/server` and use that instead.

gaearon:
@omnidan :+1: on what you suggested

omnidan:
@gaearon alright, I updated the docs - only thing missing now are the tests. I'm too unsure about what to change there, maybe you should look into that :sweat_smile:

omnidan:
Seems like the `React.addons.update` API changed too, so the link in `Troubleshooting.md` is broken right now (https://facebook.github.io/react/docs/update.html)

omnidan:
`React.addons.update` isn't mentioned on https://facebook.github.io/react/docs/update.html which is linked in `Reducers.md` - not sure if that changed or is fine.

gaearon:
It should be `react-addons-update` now.

gaearon:
You can remove

```js
  // The child must be wrapped in a function
  // to work around an issue in React 0.13.
```

from UsageWithReact. Let's not leave traces of old API.

omnidan:
@gaearon Done! I updated the other links in earlier commits (React -> ReactDOM, ReactDOMServer). Anything else, besides tests?

gaearon:
This looks good to me, thank you for your help!

gaearon:
This is good enough for now, so I'll merge and follow up in a second PR.
Everything but `real-world` is updated.

omnidan:
Awesome! You're welcome :grin:

PR #903, opened 2015-10-15
Edited syntax mistake in Reducers.md file.

CoderQuest:


CoderQuest:
My mistake, I realized that the state has become an array.

PR #909, opened 2015-10-18
Es7 migrartion

jmarceli:
Some info about using redux with ES7 decorators and static properties.


jmarceli:
Created PR #910 with resolved conflicts.

PR #910, opened 2015-10-18
Es7 migration

jmarceli:
Some info about using ES7 with Redux.


gaearon:
It's too early to call it ES7—the spec isn't stabilized, features come and go, etc.
We actually don't want to encourage people to use these features because they're subject to change.

jmarceli:
Ok. I understand this decision.

gaearon:
See also https://github.com/rackt/redux/issues/226 for contex.t

PR #918, opened 2015-10-19
Cleanup: removed dead code

julen:
Came across some unused variables while reading the source.


gaearon:
They're not unused.
It's ES6 destructuring—the variables being used are `stargazersByRepo`, `repos` and `users`.

julen:
D'oh, should have triple-checked :sweat:
For some reason the example was running fine on my end — sorry for the waste of time and thanks for clarifying!

PR #919, opened 2015-10-19
fix small typo

stoikerty:
- fixes small typo in `docs/introduction/PriorArt.md`


gaearon:
That would probably be "to using" for better grammar, no?

stoikerty:
Indeed, it wasn't quite right! I updated it to "in using" which I think sounds just about :ok_hand: 
Funnily enough I made a typo while commiting the typo for the fix of the typo...

Nice work on redux btw :)
If I find other typos I'll commit them as well (hopefully with better precision!)

PR #920, opened 2015-10-21
Update compose.md

gsklee:
Fix incorrect desugared example code of `compose()`.


gaearon:
Thanks. Forgot to update this one in 2.0

PR #923, opened 2015-10-21
[use Set instead of Array]

AntouanK:
- replace an Array with a Set ( ES2015 )

Since you use babel and all of the ES2015 goodies, I was wondering why not use things like `Set` and `Map`. 
They will be more perfomant on browsers that support them natively, and developer wise, I think they have a nicer API to work with, than plain old Arrays.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -37,7 +37,7 @@ export default function createStore(reducer, initialState) {
 
   var currentReducer = reducer;
   var currentState = initialState;
-  var listeners = [];
+  var listeners = new Set();
   var isDispatching = false;
 
   /**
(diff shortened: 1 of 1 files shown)


gaearon:
We don't want to force the consumer to use Polyfills.

AntouanK:
But you already use babel to "transpile" it. Doesn't it convert that to ES5?

dbrans:
babel transpiles syntax only.

gaearon:
Babel only converts the syntax. Something with runtime behavior (e.g. `Set`) needs to be polyfilled. Babel includes a polyfill, but it's fairly large, and not worth it for Redux.

gaearon:
To clarify—using a polyfill can be a valid choice for a larger project. However Redux is purposefully tiny (~2KB). So we don't want to grow beyond that just to use new features.

AntouanK:
Oh I see. Ok, that makes sense.
Including the polyfill makes it 19KB, it's a big difference.

Issue #938, opened 2015-10-23
Call unsubscribe many times will remove other listeners


jzlxiaohei:
I write a test

    const store = createStore(reducers.todos);
    const listenerA = expect.createSpy(() => {});
    const listenerB = expect.createSpy(() => {});

    let unsubscribeA = store.subscribe(listenerA);
    store.subscribe(listenerB);

    unsubscribeA();
    unsubscribeA();//invoke two times

    store.dispatch(unknownAction());
    expect(listenerA.calls.length).toBe(0);
    expect(listenerB.calls.length).toBe(1); // failed

reason:
   
   `arr.splice(-1,1)` still remove the item of the arr.

//createStore.js  
   
    function subscribe(listener) {
        listeners.push(listener);

        return function unsubscribe() {
          var index = listeners.indexOf(listener);
          // if index ===1 ,should return
          listeners.splice(index, 1);
        };
  }

gaearon:
Fixed in #939 and b7031ce3acb23b6ecadbd977b1cfa32486447904.
Out in 3.0.4.
Thanks.

PR #939, opened 2015-10-23
fixed unsubscribe bug for issue #938

jzlxiaohei:


src/createStore.js
```diff
@@ -62,6 +62,7 @@ export default function createStore(reducer, initialState) {
 
     return function unsubscribe() {
       var index = listeners.indexOf(listener);
+      if (index === -1)return;
```
gaearon: Please be consistent with the project style.
In this case I'd rather write it as

```js
if (index !== -1) {
  listeners.splice(index, 1);
}
```

gaearon:
Thank you for the PR! I left a nit, will be happy to merge after it's fixed.

src/createStore.js
johanneslumpe: Nitpick: `index > -1` - saved 2 chars!

gaearon: I don't mind either way. :stuck_out_tongue: 
I was kind of a fan of `> -1` for some time, but now I think `!== -1` makes the intention clearer.
But whatever, I'm fine with either.

gaearon:
I changed your fix to cover one more edge case: https://github.com/rackt/redux/commit/b7031ce3acb23b6ecadbd977b1cfa32486447904

jzlxiaohei:
nice  :metal:

PR #941, opened 2015-10-23
Readme tinyfix

Maximilianos:
I think this is more what was trying to be said.
`takes a few minutes` by itself is like saying it takes time to get started with.
`only takes a few minutes` is saying it doesn't take much time to get started with.


omnidan:
You're right - thanks :grin:

PR #942, opened 2015-10-23
Cached Object string representation

matthewtff:
Calling toString on Object may cost a lot. It's much better to
have cached value for it.

### DIFF
--- a/src/utils/isPlainObject.js
+++ b/src/utils/isPlainObject.js
@@ -1,4 +1,5 @@
 var fnToString = (fn) => Function.prototype.toString.call(fn);
+var objStringValue = fnToString(Object);
 
 /**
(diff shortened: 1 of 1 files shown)


gaearon:
In general we don't want to diverge from Lodash implementation because we're going to eventually use it (#611). Can you update the code to what Lodash uses in 4.0 instead?

matthewtff:
Not sure if it's a good idea. Looks like in lodash isPlainObject has internal dependencies.
So I'll have to bring here at least several core functions. Also it would make me to write own) or also copy-paste) some tests. What I'm doing here is a small, straightforward optimization. Just to live here until lodash gets shipped...

gaearon:
Seems straightforward enough.

PR #945, opened 2015-10-24
fix(docs): test recipe for async action creators contains inconsistencies

r-park:
The section `Writing Tests > Async Action Creators` contains inconsistent names for action-type constants and action-creator function names.


omnidan:
@r-park thanks a lot! you're totally right.

PR #952, opened 2015-10-25
Update ESLint configuration

taion:
Cleaning this up a bit in light of https://github.com/rackt/history/pull/112:

- Upgrade `eslint-config-airbnb` - the old version is missing some stuff like `no-undef` (the new one adds `id-length` which has to be disabled though)
- Remove redundant settings in `.eslintrc`
- Don't need to ignore `lib/` since `npm run lint` already specifies explicit included directories
- Move `mocha` environment settings into specific test/example directories rather than at top level


examples/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "env": {
+    "mocha": true
```
gaearon: Does this make sense here? Should each example have it in test folder instead?

taion: I can do that if you prefer. Thought this made the most sense w/r/t getting the mocha environment setting out of the top-level `.eslintrc` without adding too many sub-`.eslintrc`s.

Your call.

gaearon: If we split seems like we should split all the way, or it's confusing.

taion:
I updated the `.eslintrc` files under `examples/` to only be in the respective `test` directories.

omnidan:
@taion looks good now :ok_hand:

mjackson:
@taion Is the plan to eventually use the rackt ESLint config?

taion:
I don't have a plan - I just noticed that this repo was using an old version of `eslint-config-airbnb` that did not include the extremely useful `no-undef` rule and thought I'd fix it.

Probably easier to merge anyway given a smaller base of rules on top of e.g. the airbnb config.

PR #960, opened 2015-10-27
added reference to redux-mock-store in docs

arnaudbenard:
Hi,

I discussed with @gaearon yesterday on [Twitter](https://twitter.com/dan_abramov/status/658689636227792896) about extracting the mock store logic into a separate package. The benefits are that we can test the library and don't need to copy paste code around when the mock example is updated in the docs.

Here's the module: https://github.com/arnaudbenard/redux-mock-store

Let me know if you want to append changes/more tests to the current implementation and reference the package in the redux docs.

Thank you ;)


omnidan:
@arnaudbenard looks good to me - you can also add it to the Ecosystem section (I'll keep this PR open in case you want to do that :wink:)

arnaudbenard:
@omnidan Awesome. Should I create a `Testing` section or simply put it under `Utilities`?

omnidan:
@arnaudbenard I would put it under `Utilities`, a section for just one module doesn't make sense (so: maybe later :smile:)

arnaudbenard:
@omnidan done! Ready to merge :dancer:

omnidan:
Thanks a lot! :grin:

PR #961, opened 2015-10-27
update Shopping Cart example to 0.14

psimyn:


omnidan:
Thanks a lot! Good catch - I didn't even notice it's not 0.14 when it was added :blush:

Issue #928, opened 2015-10-21
Explore using Rollup for UMD builds


gaearon:
Our compiled code isn't looking nice because of all the junk code generated by Babel for ES6 modules:
https://cdnjs.cloudflare.com/ajax/libs/redux/3.0.3/redux.js

While this is not technically an issue, it would be nice to get a neater compiled output.
Maybe we can explore using [Rollup](http://rollupjs.org/) for this.

- [ ] Turn off ES6 module transformation in Babel
- [ ] Try to make both CJS (`./lib/`) and UMD (`./dist/`) builds with [Rollup](http://rollupjs.org/) instead
- [ ] Of course Babel should still be turned on for the actual transpiling

If somebody wants to explore this feel free to.

phated:
:-1:

gaearon:
More details on :-1: ?

phated:
The author of rollup is a huge NIH junkie that jumps on threads to promote his stuff and doesn't contribute ideas back anywhere.  By moving away from webpack, the community splits again instead of empowering them to help get webpack 2 out.  Jumping ship every few months doesn't enforce the perception of stability in the library.

Rich-Harris:
@phated Feel free to @ me in your personal attacks. Surely we should be discussing the technical merit of the two approaches?

phated:
@Rich-Harris @gaearon posted on twitter about a cool new thing in webpack, you jumped in pushing your own stuff.  According to https://github.com/webpack/webpack/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+rich-harris+ - you've never once contributed to the project.  Technology isn't the only thing that matters, as a whole community has built up around React/Rackt/Redux.  I've said my peace and have no more time to argue with you.

gaearon:
Different approaches have different merits. Some new approaches may be difficult to implement on top of an existing project, but easier to implement from the ground up. By the same argument, Redux shouldn't have been made, and Flux should have been improved instead. Whether something is different enough that it “deserves” to be a separate thing over splitting a community is always a subjective question, and especially so when the project is young.

Regardless of this, personal attacks are not allowed here. I redacted the part of @phated's comment that contained a personal attack, and we won't tolerate more of them. Even the community discussions can be respectful and polite, no need for extra emotions.

gaearon:
[...]

PR #964, opened 2015-10-27
Generate UMD build with Rollup

Rich-Harris:
This PR generates a UMD build with Rollup, per @gaearon's suggestion on #928. The goal is to produce more compact, readable output.

It *doesn't* change the `build:lib` task, as Rollup only bundles modules rather than doing 1:1 transpilation, but I could amend the PR to create a CommonJS bundle in `lib` if that's desirable.

Zipped size is 14.5% smaller than the current build:

                       | Unminified | Minified | Gzipped
---------------------- | ---------- | -------- | -------
[Current](https://gist.github.com/Rich-Harris/d47fcf788ce160fdefd8)                | 21643      | 5402      | 1904   
[With Rollup](https://gist.github.com/Rich-Harris/d9bcf812495c3fc45923)            | 16177      | 3978      | 1628


timdorr:
@gaearon What do you think about this? One of the nice side benefits is a more readable UMD output, and a removal of the ugly Webpack bootstrap code. The tradeoff being you now have two build tools in the same project. But for the UMD end user, it's a win.

gaearon:
@timdorr 

I want to wait for Lodash 4 to come out and then revisit, per https://github.com/rackt/redux/issues/928#issuecomment-153202978

build.js
```diff
@@ -0,0 +1,22 @@
+var rollup = require( 'rollup' );
+var babel = require( 'rollup-plugin-babel' );
+var replace = require( 'rollup-plugin-replace' );
+
+var env = process.argv[2] === '--prod' ? 'production' : 'development';
+
```
tracker1: how about `process.argv.contains('--prod') >= 0`, then it can be on any argument.

thisconnect:
I'd love to use redux in a project where we rollup everything. As far as I know, it is not possible to rollup an application that imports redux, neither with redux's main, jsnext:main nor the dist/redux.js. :(
umd or es6 would let us use redux. hope this gets merged.

gaearon:
Redux has UMD build in dist folder.

gaearon:
>I'd love to use redux in a project where we rollup everything. As far as I know, it is not possible to rollup an application that imports redux, neither with redux's main, jsnext:main nor the dist/redux.js. :(

Can you please explain why Redux currently cannot be used with Rollup?

thisconnect:
@gaearon Hi, sorry for my delay. I finally succeeded building a simple redux example (yes!!!) using some of @Rich-Harris PR's ideas. The main issue that it did not work in the first place was that I was missing `replace({ 'process.env.NODE_ENV': JSON.stringify('production') })`.
It did build but error'd in the browser `Uncaught ReferenceError: process is not defined`. With "replace process" stuff all works fine!

I setup a gist with 2 builds, one using es6 import and one commonjs require. Download `npm i && npm start` then open the html in a browser. Note: the minified version of es6 import has about half the size.

https://gist.github.com/thisconnect/af9c83ed56e9d554f63b

Thanks btw for the video's I love them! :)

simplesmiler:
@gaearon `jsnext:main` of redux points to the es6 code, which breaks rolling up (this can be worked around with plugins). I believe the consensus is to point `jsnext:main` to the transpiled code with only imports/exports left intact.
Reference: https://github.com/rollup/rollup/issues/337#issuecomment-161319114

gaearon:
@simplesmiler 

I'm happy to fix it but I can't figure out how to create a Rollup project reproducing the issue with Redux. Can you help me by pointing me at one?

TrySound:
@gaearon via cli `rollup --config rollup.config.js`
```js
import npm from 'rollup-plugin-npm';

export default {
  entry: 'src/index.js',
  dest: 'dist/index.js',
  format: 'iife',
  plugins: [
    npm({ jsnext: true })
  ]
};
```

Rich-Harris:
The best way I've found to use Redux with Rollup is by building it separately: https://github.com/Rich-Harris/rollup-redux-debug. Lil' bit hacky but it works, and is very space-efficient! https://github.com/rackt/redux/issues/1042 would render the custom build unnecessary.

thisconnect:
Thanks, looks great! [Updated gist](https://gist.github.com/thisconnect/af9c83ed56e9d554f63b/revisions) to use Redux 3.3.1 and removed commonjs plugin.

main.iife.min.js shrank from 2,972 bytes (with Redux  3.2.1) to 2,202 bytes (with Redux 3.3.1)

gaearon:
I’m still up for adding this, but it currently doesn’t merge cleanly.
Lodash 4 is out so seems like the only thing blocking this is now fixed.
If you’d like, please rebase and I’ll gladly reopen and merge.

PR #965, opened 2015-10-28
explore rollup builds

Sinewyk:
ref: https://github.com/rackt/redux/issues/928

I could use rollup-plugin-replace instead of a simple code.replace but ... it's just an exploration.

Results are pretty cool. Didn't know that even if webpack bundling is better than browserify, it's still quite a cost.

But we don't gain THAT much after min + gz.

```
Webpack               |   Rollup
                      |
 22K redux.js         |  16K  redux.rollup.js
5,9K redux.js.gz      |  4,9K redux.rollup.js.gz
5,3K redux.min.js     |  3,9K redux.rollup.min.js
1,9K redux.min.js.gz  |  1,6K redux.rollup.min.js.gz
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,11 @@
     "isparta": "^3.0.3",
     "mocha": "^2.2.5",
     "rimraf": "^2.3.4",
-    "webpack": "^1.9.6"
+    "rollup": "^0.20.2",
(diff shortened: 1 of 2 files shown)


Sinewyk:
nvm, https://github.com/rackt/redux/pull/964

PR #824, opened 2015-09-30
Shopping cart example

hedgerh:
For some reason the product images can't be found.  Guessing it is some webpack config issue, if anyone has any suggestions.

Other than that, it should be good to go.

As per issue https://github.com/rackt/redux/issues/823

PR #970, opened 2015-10-28
Shopping cart example fixes

### LINKED ISSUE #823: Shopping Cart example
I just rewrote [flux-comparison Redux example](https://github.com/voronianski/flux-comparison/tree/master/redux/js) to be idiomatic Redux: https://github.com/voronianski/flux-comparison/pull/64.

I want it to be in our repo too because it contains fairly interesting bits such as queries and denormalization—something a lot of “advanced” examples miss.

@voronianski would you be comfortable with us copying (and tweaking) that example to be in Redux repo too?

I'd imagine that moving it here would involve:

* porting to use Webpack since that's what other our examples use
* stripping down UI (no styles—a vanilla UI like in `todos-with-undo` example is fine for our purposes)
* moving the "common" components and API mocks to the example folder

If @voronianski doesn't mind, would somebody be up to do this?

justingreenberg:
minor fixes per @gaearon's request on PR #824; resolves #823

redux-shopping-cart-ui

### DIFF
--- a/examples/shopping-cart/components/Cart.js
+++ b/examples/shopping-cart/components/Cart.js
@@ -7,7 +7,7 @@ export default class Cart extends Component {
 
     const hasProducts = products.length > 0;
     const nodes = !hasProducts ?
-      <div>Please add some products to cart.</div> :
+      <em>Please add some products to cart.</em> :
       products.map(product =>
         <Product
           title={product.title}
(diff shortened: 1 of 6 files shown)


omnidan:
Looks good :+1:

gaearon:
Thank you!

PR #971, opened 2015-10-29
Add Shopping Cart to Docs (Examples.md)

justingreenberg:
@gaearon I referenced some patterns you described in your PR, can you take a quick look?

redux-shopping-cart-docs

With respect to the Examples in general, maybe we could replace "It covers:" with something like "Patterns:" and try to include links to relevant code in bullet points. Since many people use these examples as learning tools and references, the ability to scan and jump to patterns would be helpful.

PR #978, opened 2015-10-29
Add shopping cart example to README and docs

hedgerh:
Added shopping cart example to docs, but I'm not sure what is worth mentioning that the example covers on the docs page.  I added "denormalization," but if someone wants to offer up some bullet points I can toss them in.


gaearon:
Duplicate of #971?

hedgerh:
@gaearon Just caught that.  Closing now.

PR #984, opened 2015-10-30
be more explicit when importing files

6ewis:
It's minor but it's helpful when someone try to follow the code. In addition, there's no point importing files we do not use.

### DIFF
--- a/examples/shopping-cart/containers/CartContainer.js
+++ b/examples/shopping-cart/containers/CartContainer.js
@@ -1,7 +1,7 @@
 import React, { Component, PropTypes } from 'react'
 import { connect } from 'react-redux'
-import { checkout } from '../actions'
-import { getTotal, getCartProducts } from '../reducers'
+import { checkout } from '../actions/index'
+import { getTotal, getCartProducts } from '../reducers/index'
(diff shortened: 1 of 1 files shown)


johanneslumpe:
I feel that adding a `/` at the end should be enough to signal to a user that we are importing the index file of a directory.

6ewis:
@johanneslumpe In this instance what's the cost of being more explicit when we can avoid subtle confusion? nothing. It did confuse me while trying to grok redux.

johanneslumpe:
@6ewis To me it is just superfluous because a slash already makes the intention clear. But I will let others decide what the direction should be here. I agree though that having nothing at the end so that it looks like a file require is definitely not great.

omnidan:
I think it makes sense to explicitly state we import `index`, but we should at least add a `/` at the end.

gaearon:
We use this convention all over the place, including the docs.
I'd say we either do it one way, or the other way, but doing it only in one place is inconsistent.

gaearon:
I reverted—again, I don't mind the change, but only as a big commit fixing it all over the docs. :wink:

PR #985, opened 2015-10-30
Add missing reducer composition example to docs

sirbrillig:
The code for the first implementation of `todos` as an example of reducer composition is actually the same as the previous example, causing the following paragraph ("This is called *reducer composition*, and it’s the fundamental pattern of building Redux apps") to make no sense. 

This change adds the `todos` function from the following example to make this one more clear.

There are other ways the composition could be done, certainly, and I'm not sure that the one I chose is the most iconic, so feel free to suggest alternatives. :-)

### DIFF
--- a/docs/basics/Reducers.md
+++ b/docs/basics/Reducers.md
@@ -199,6 +199,29 @@ function todoApp(state = initialState, action) {
 Is there a way to make it easier to comprehend? It seems like `todos` and `visibilityFilter` are updated completely independently. Sometimes state fields depend on one another and more consideration is required, but in our case we can easily split updating `todos` into a separate function:
 
 ```js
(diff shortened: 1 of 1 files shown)


omnidan:
You're right - thanks for the PR :grin:

ellbee:
Argh, my bad. I'm glad you were paying attention @sirbrillig, I'm going to double check the docs to make sure I didn't make any other silly mistakes like this.

I made a minor update, removing a semi-colon that had snuck in, to keep the example following the new Rackt lint config.

sirbrillig:
:+1: :grinning:

ellbee:
Ok, went through the diffs and couldn't spot any more glaring errors.

PR #989, opened 2015-11-01
descriptive syntax in all tests

christian-fei:
hello,
this change unifies the test names to use descriptive syntax in all tests.
I found that in some places there were leftovers from the "should" syntax. :)

### DIFF
--- a/test/utils/bindActionCreators.spec.js
+++ b/test/utils/bindActionCreators.spec.js
@@ -47,7 +47,7 @@ describe('bindActionCreators', () => {
     )
   })
 
-  it('throwfor a null actionCreator', () => {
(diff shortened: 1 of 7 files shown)


test/utils/bindActionCreators.spec.js
```diff
@@ -47,7 +47,7 @@ describe('bindActionCreators', () => {
     )
   })
 
-  it('should throw for a null actionCreator', () => {
+  it('throwfor a null actionCreator', () => {
```
gaearon: Should be `throws for`?

christian-fei: :+1:

omnidan:
Thanks :ok_hand:

PR #814, opened 2015-09-27
adds documentation for exampleTodoList file structure and up & running

alanbsmith:
### OVERVIEW
This PR adds documentation to ExampleTodoList.md. Specifically it lists the file structure of the app and how to get it up and running.

### WHY?
While the documentation is really great for this example, it would be useful to have an overview of the file structure (what goes where) and also what other files are needed to get it up and running. This just makes that process more approachable.

### CLOSING
I tried to follow the Markdown conventions already present, but if something needs to be changed/improved, please let me know! I'll be happy to fix it.

PR #1005, opened 2015-11-04
Add example app from basic tutorial

naw:
This is just lifting the code from http://rackt.org/redux/docs/basics/ExampleTodoList.html and putting it into a readily executable example app.   Started with the basic infrastructure from the todomvc example and swapped in the source code from the basic tutorial example.

I believe it's helpful to be able to clone and run the tutorial example, not only for learning `redux` itself, but also because this is the example referenced in the `reselect` docs.

Did the work for myself in my fork, and just issuing this PR in case you think it would be helpful to have in the main repo.

Certainly no offense taken if you just want to close this.

### DIFF
--- a/examples/tutorial-todo/.babelrc
+++ b/examples/tutorial-todo/.babelrc
@@ -1,0 +1,19 @@
+{
+  "stage": 2,
+  "env": {
+    "development": {
+      "plugins": [
+        "react-transform"
+      ],
+      "extra": {
+        "react-transform": {
+          "transforms": [{
+            "transform": "react-transform-hmr",
+            "imports": ["react"],
+            "locals":  ["module"]
+          }]
+        }
+      }
+    }
+  }
+}
--- a/examples/tutorial-todo/actions.js
+++ b/examples/tutorial-todo/actions.js
@@ -1,0 +1,33 @@
+/*
+ * action types
+ */
+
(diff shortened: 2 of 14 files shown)


gaearon:
This is pretty much what `todos-with-undo` covers though, right?
Maybe we should highlight this in docs because it isn't the first related PR: #814

naw:
All I can do is offer my perspective as someone familiar with web applications, but a _newbie_ to react and redux, I believe there is great value in being able to clone and run the _exact_ code in the tutorial.  

Reading through the tutorial, I strove to understand each section, and upon reaching a basic understanding, it was time to dive in, run the example, and make tweaks to further develop that understanding.  Actually for me, the motivation had more to do with starting with the exact code in the `reselect` docs (which is presently identical to the `redux` basic tutorial, I believe).

While the `todo-with-undo` example may be highly similar, the presence of additional features (redo and undo) further complicates things and is an impediment to the progress of a newbie (in my opinion).

omnidan:
@naw I understand where you're coming from. IMO `todos-with-undo` doesn't complicate the example too much - it's just a few lines of code that got changed to add undoability. I'll leave it up to the other guys to decide whether to add this or not :sweat_smile:

gaearon:
Yeah, I don't think it's really worth it to have two examples that are so similar in the repo.
Appreciate the effort though!

PR #1007, opened 2015-11-05
Fix typo in docs

kolybasov:
naïve was changed to native


gaearon:
[naïve](https://en.wiktionary.org/wiki/na%C3%AFve) is the right word in this context.

PR #1008, opened 2015-11-05
updated instructions

mjunaid106:
added instruction to install the devDependencies as I came across thi…s issue while following the instructions.


gaearon:
Can you please clarify what you mean?
Any project needs `npm install` to be run in its root, but this has nothing to do with Redux itself.
We list instructions to install Redux, so we're assuming the user is already familiar with NPM.

PR #1013, opened 2015-11-06
Add .editorconfig

kolybasov:
I think it may be useful for people that working with different languages(like Golang or something else with tabs indentation) to switch between projects and easily contribute to redux. 

I know that eslint helps with it, but think that using .editorconfig is more convenient.

[Specs](http://editorconfig.org)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# EditorConfig helps developers define and maintain
+# consistent coding styles between different editors and IDEs.
+
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 2


gaearon:
It requires a plugin to work though, is this correct?
Would this “just work” with any editors, or are we assuming the user has some plugins installed?

kolybasov:
Without plugin:
- BBEdit
- Coda
- CLion
- GitHub
- IntellijIDEA
- RubyMine
- SourceLair
- WebStorm

Requires plugin:
- AppCode
- Atom
- Brackets
- Eclipse
- Emacs
- CodeBlocks
- Geany
- Gedit
- jEdit
- Komodo
- NetBeans
- Notepad++
- PhpStorm
- PyCharm
- Sublime Text
- TextMate
- Vim
- Visual studio
- Xcode

All plugins are accessible from main page http://editorconfig.org/#download

gaearon:
OK, I guess no harm merging it, and good if it helps some people.
:+1:

PR #1015, opened 2015-11-07
Docs fix: Initial state isn't a string, it's `undefined`

prewk:


omnidan:
@prewk but the `typeof` operator returns a string :wink:

prewk:
Wow, I'm tired. Sorry.

PR #1019, opened 2015-11-09
Clone listeners at subscription / unsubscription

rpominov:
This should improve performance a bit because dispatch is more frequent operation than subscription / unsubscription.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -58,7 +58,7 @@ export default function createStore(reducer, initialState) {
    * @returns {Function} A function to remove this change listener.
    */
   function subscribe(listener) {
-    listeners.push(listener)
+    listeners = listeners.concat([ listener ])
     var isSubscribed = true
 
     return function unsubscribe() {
@@ -68,7 +68,10 @@ export default function createStore(reducer, initialState) {
 
       isSubscribed = false
       var index = listeners.indexOf(listener)
-      listeners.splice(index, 1)
+      listeners = [
+        ...listeners.slice(0, index),
+        ...listeners.slice(index + 1)
+      ]
     }
   }
 
@@ -123,7 +126,7 @@ export default function createStore(reducer, initialState) {
       isDispatching = false
     }
 
-    listeners.slice().forEach(listener => listener())
+    listeners.forEach(listener => listener())
     return action
   }
 


src/createStore.js
```diff
@@ -66,9 +66,13 @@ export default function createStore(reducer, initialState) {
         return
       }
 
-      isSubscribed = false
-      var index = listeners.indexOf(listener)
-      listeners.splice(index, 1)
+      listeners = listeners.filter(x => {
```
gaearon: Any particular reason why you're filtering rather than concat()ing two slice()d arrays and using indexOf() like before?

rpominov: Filter just looks more clean to me. I can change it to concat + slice that also fine.

gaearon: I'd rather see a `slice` + array spread here.

rpominov: Yeah, this looks much nicer indeed.

src/createStore.js
```diff
@@ -68,7 +68,7 @@ export default function createStore(reducer, initialState) {
 
       isSubscribed = false
       var index = listeners.indexOf(listener)
-      listeners.splice(index, 1)
+      listeners = [ ...listeners.slice(0, index), ...listeners.slice(index + 1) ]
```
gaearon: Do you mind splitting this to two lines please, and then squashing your commits? Thanks.

rpominov: Done. No problem!

gaearon:
Sorry for leaving this hanging. Can you please run this in a real app and give us some numbers with regards to how many allocations occurred with either method? What are the worse conditions for either method? Basically, what kind of apps would we make slower as a result of this change?

rpominov:
To be honest I don't think it will give a noticeable difference with performance, I just wanted to justify this change somehow. The real reason is that this way we don't mutate the array, which just feels better :)

Feel free to close this if you think this doesn't matter, I completely understand.

Btw, since `forEach()` is used we can remove cloning anyway. The `.slice()` would be required if we would use `for` cycle.

rpominov:
> Btw, since forEach() is used we can remove cloning anyway. The .slice() would be required if we would use for loop.

Oh, sorry. Please ignore this, `.slice()` is needed when `.forEach()` is used too.

gaearon:
FYI, I implemented a fix for this that doesn't suffer from moving cloning elsewhere: c031c0a8d900e0e95a4915ecc0f96c6fe2d6e92b

rpominov:
Wow, that's smart :+1:

PR #1037, opened 2015-11-14
Rename smart/dumb to presentational/container

camsong:
fix https://github.com/rackt/react-redux/issues/170


omnidan:
Thank you :grin:

PR #1039, opened 2015-11-15
Use files in package.json

### LINKED ISSUE #1033: [React Native] Upgrading to Babel 6 and/or adding .babelrc to .npmignore
React Native will be shipping with Babel 6 in about three weeks. One consequence of this is that Babel finds .babelrc in `node_modules/redux` and tries to consume it -- since Redux is using Babel 5's configuration, this causes Babel 6 to crash. The current verdict is that packages will have to exclude .babelrc when distributed through npm: https://github.com/facebook/react-native/issues/4062.

SimenB:
Fixes #1033

Before:
image

After:
image

Difference is .editorconfig, .babelrc, CNAME, CODE_OF_CONDUCT and PATRONS. Can add back if you want. I'm also guessing `dist` is supposed to be included for npmcdn etc.

A whitelist is much easier to manage than a blacklist, as can be seen in the diff of this PR


gaearon:
Looks good to me, thank you.

PR #1050, opened 2015-11-18
composition can be seeded with multiple arguments

Bjvanminnen:
Related to https://github.com/rackt/redux/issues/632#issuecomment-157361578

This allows us to provide our composition with multiple arguments.

It does so by pulling off the last function from our list, calling it with the provided args, and using the result as the initialValue for reduceRight (which we then call on the remaining functions). Unfortunately, this does make our simple one liner a little more complex

Let me know if you see a cleaner approach.

### DIFF
--- a/src/utils/compose.js
+++ b/src/utils/compose.js
@@ -7,8 +7,14 @@
  */
 export default function compose(...funcs) {
   return (...args) => {
-    return funcs.length ?
-      funcs.slice(0, -1).reduceRight((composed, f) => f(composed), funcs[funcs.length - 1](...args)) :
-      args[0]
+    if (funcs.length === 0) {
+      // We weren't given any functions, just return the first passed in arg.
+      return args[0];
+    }
+
+    const last = funcs[funcs.length - 1];
+    const rest = funcs.slice(0, -1);
+
+    return rest.reduceRight((composed, f) => f(composed), last(...args));
   }
 }
--- a/src/utils/compose.js
+++ b/src/utils/compose.js
@@ -9,12 +9,12 @@ export default function compose(...funcs) {
   return (...args) => {
     if (funcs.length === 0) {
       // We weren't given any functions, just return the first passed in arg.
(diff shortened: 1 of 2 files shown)


dbrans:
Hey @Bjvanminnen, If you need a multiple-arg compose in your app could you not use [lodash's compose](https://lodash.com/docs#flowRight)?

Bjvanminnen:
I could, and am fine with doing so. However, it also seems reasonable to expect the compose provided by redux to support this functionality.

dbrans:
Watch out for an empty function list. The way compose is written in this PR, compose()(x) will throw an error.

Currently redux's `compose()(x)` returns `x`, lodash's, `_.compose()(x, y)` returns `x`.

Bjvanminnen:
Latest commit ensures that we don't change existing behavior for case where no funcs are passed.

src/utils/compose.js
```diff
@@ -6,5 +6,9 @@
  * left. For example, compose(f, g, h) is identical to arg => f(g(h(arg))).
  */
 export default function compose(...funcs) {
-  return arg => funcs.reduceRight((composed, f) => f(composed), arg)
+  return (...args) => {
+    return funcs.length ?
```
nettofarah: It is a little hard to grasp what this is doing at first.
Maybe re-writing this as an `if` statement would make this function a bit easier to follow.

Bjvanminnen:
Stopped trying to do this as a one-liner, and I think it makes it a lot easier to follow.

gaearon:
Would this be a breaking change?

Bjvanminnen:
It shouldn't be.

If given no funcs, we return the first argument, same as before.
If given funcs and a single argument, behavior should be identical to before.
Only new behavior is that if given multiple arguments, we pass all of them to the last func instead of ignoring the additional arguments.

gaearon:
Out in 3.0.6, thanks!

PR #1052, opened 2015-11-18
Simplify applyMiddleware

buunguyen:
The assignment `dispatch = store.dispatch` is essentially a no-op as `dispatch` is assigned to the result of `compose(..chain)(store.dispatch)`. This pull request removes the assignment to simplify the code and avoid confusion as to what the actual value of `dispatch` is by the time the middleware invokes it (I had that question myself when reading the source).

### DIFF
--- a/src/utils/applyMiddleware.js
+++ b/src/utils/applyMiddleware.js
@@ -19,8 +19,8 @@ import compose from './compose'
 export default function applyMiddleware(...middlewares) {
   return (next) => (reducer, initialState) => {
     var store = next(reducer, initialState)
-    var dispatch = store.dispatch
     var chain = []
+    var dispatch
 
     var middlewareAPI = {
       getState: store.getState,


just-boris:
dispatch is used at the [line 27] (https://github.com/rackt/redux/blob/master/src%2Futils%2FapplyMiddleware.js#L27).

So that is good point to have test against occasional remove.

buunguyen:
But it's overridden at [line 30](https://github.com/rackt/redux/blob/master/src%2Futils%2FapplyMiddleware.js#L30). Original point remained.

just-boris:
Interesting point. Middleware able to call original version of dispatch during initialization on the line 29. No one of my known middlewares does it, but may be somewhere it is.

buunguyen:
Good point. Not sure if that use case (i.e. middleware invokes `dispatch` during initialization and this `dispatch` goes to the store directly instead of through the middleware chain) is accidental or desired. If accidental, then this PR might make sense (or probably replace the first assignment to a function that shows an error?). If it is desired, I suppose it should be mentioned in the doc and/or come with a test. Curious what @gaearon thinks.

buunguyen:
Sorry, closed by mistake.

gaearon:
We had this change to make Flow happy and sure it's a function. We didn't ship Flow support in the end (it was too lacking) but we will eventually.

buunguyen:
If that was to make Flow happy instead of enabling a use case, would initially assign `dispatch` to a function that throws an error be better?  Basically replace line 22 with:

```js
var dispatch = function () {
  throw new Error('dispatch must not be invoked during middleware initialization')
}
```

gaearon:
Is middleware invoking dispatch during initialization a valid use case or not?
Do we currently allow it? If we do, should we break it?

buunguyen:
I personally don't see any reason why a middleware should dispatch directly to store during initialization. I think it will be quite confusing if the app hasn't triggered any action yet a bunch are dispatched to the store by some middleware. That being said, my experience is limited to a few middleware I wrote and those I have used so far. If there's valid reasons for this use case to exist, I'd love to hear about them and hope to see the behavior documented.

sompylasar:
@buunguyen As far as reducers are stateless, i.e. are pure functions, there should be no trouble for them to handle any actions in any order. They just change the state to what they think is appropriate, having the current state and the action, or ignore the action (i.e. do not change the state).

buunguyen:
@sompylasar and why is that make it necessary to allow middleware dispatch arbitrary events during initialization phase?

sompylasar:
@buunguyen Sorry, I just wanted to emphasize that this case (i.e. dispatching during initialization) should not cause troubles with the reducers. Personally I like your suggestion with throwing an error to explicitly disable this case.

Dispatching from the middleware could be used to initialize a part of state (e.g. from a redux-router middleware), to probe the reducers for unknown action (IIRC this is done by the DevTools).

Should middleware somehow handle this transition from initialization to working phase, so it is able to do some state initialization via actions?

gaearon:
I'm inclined not to change this because it might break some middleware, and it's not important enough to warrant a breaking change IMO. Thanks for the effort though!

PR #1054, opened 2015-11-18
Add check that reducer returns current state for unknown actions

msafi:
Hey folks,

First ever open source contribution here and a Redux newb, so I have no idea what I'm doing! Alright, with that out of the way...

I was reading the test cases for `combineReducers` and noticed that [this test](https://github.com/rackt/redux/blob/master/test/utils/combineReducers.spec.js#L154) doesn't really test what it intends to test because if you change the `default` case to return `state`, the reducer will not throw and the test will fail.

I changed the way Redux inits reducers by making it do an init and a random probe at the same time. This will make it impossible for users to handle private Redux actions, so there's no need for that test anymore. However, I altered the test to make it assert that Redux checks reducers to make sure they return the current state for an unknown action.


timdorr:
Ah, good call. I think this looks good to me. I'll let Dan chime in if he wants to.

gaearon:
Hey, thanks for the PR!
I'm a bit busy now with the screencasts.
I'll review when I'm less busy.

Let's hold this off for now. I need to understand this change better. I think DevTools currently actually rely on `INIT` having a stable name although I'm not sure.

timdorr:
FWIW, DevTools has [a different INIT signature](https://github.com/gaearon/redux-devtools/blob/master/src/devTools.js#L14) (and [in the next branch](https://github.com/gaearon/redux-devtools/blob/next/src/instrument.js#L51)), so this shouldn't cause problems with it.

nettofarah:
@msafi is there any particular reason for using `Math.random().toString(36).substring(7).split('').join('.')` as opposed to a shorter random string?

msafi:
@nettofarah `combineReducers.js` already had this code [here](https://github.com/rackt/redux/blob/07ce117b56ee4aa0f22f5d64a733c716836b1b48/src/utils/combineReducers.js#L68). That's where I copied it from.

test/utils/combineReducers.spec.js
```diff
@@ -159,16 +159,14 @@ describe('Utils', () => {
               return state + 1
             case 'decrement':
               return state - 1
-            // Never do this in your code:
-            case ActionTypes.INIT:
-              return 0
             default:
-              return undefined
```
gaearon: Can we keep the test that if you forget a `default` case you'll get this warning? It's an important thing to remember to check.

msafi: Okay, I added it [here](https://github.com/rackt/redux/pull/1054/files#diff-91d70a37da68f9dea7fa323102684774R87).

src/utils/combineReducers.js
```diff
@@ -59,19 +59,17 @@ function assertReducerSanity(reducers) {
     if (typeof initialState === 'undefined') {
       throw new Error(
         `Reducer "${key}" returned undefined during initialization. ` +
-        `If the state passed to the reducer is undefined, you must ` +
-        `explicitly return the initial state. The initial state may ` +
-        `not be undefined.`
+        `Reducers should never return undefined. Make sure this reducer ` +
+        `has a catch-all clause for unknown action types and that it returns a ` +
+        `default initial state if the state passed to it is undefined.`
       )
```
msafi: I changed the error message here because if the user doesn't have a `default` case in their code, this is the error that will be thrown.

timdorr:
@msafi Could you update this against master? We can get the ball rolling again and get this merged in, hopefully :smile:

msafi:
@timdorr Sorry about the delay. I updated the code against master. Please let me know if there's anything else I need to do!

gaearon:
Sorry for the long wait.

Alas, I just remembered why we can't rely on this. We previously rejected a similar change inhttps://github.com/rackt/redux/pull/174#issuecomment-114878235 because

>If someone has runtime type checking in their reducers, this will cause an error.

Potentially this might have problems for static type checkers as well since we're calling a function with a string `state` argument whereas they might have annotated their reducer with a specific type.

This is how https://github.com/rackt/redux/pull/174 got superseded by https://github.com/rackt/redux/pull/193 which is roughly what you see now.

msafi:
Ah, good catch!

So I guess there's no way to have a runtime check that a reducer returns given state for unknown action type.

Alright, but there's another issue that this PR tried to fix, which is that there's still a way for the user to handle `ActionTypes.INIT` and [this check](https://github.com/rackt/redux/blob/ccff48df5c068fda7788c76d24a9b613a0e309f6/src/combineReducers.js#L66-L78) would not catch it. Like, if we change the test [like this](https://github.com/rackt/redux/compare/ccff48df5c068fda7788c76d24a9b613a0e309f6...msafi:broken-check), the user can bypass the check.

I'll send another PR that only addresses this second issue.

PR #1061, opened 2015-11-21
Improving mockStore in WritingTests doc

MrOrz:
Currently `mockStore` in the doc [Writing Tests](http://rackt.org/redux/docs/recipes/WritingTests.html) would produce `done() called multiple times` error when assertion error occurs. It causes mocha to produce uninformative reports, instead of useful assertion errors:

```
1) async actions shall not pass:
     Error: done() called multiple times
      at Suite.<anonymous> (test.js:57:3)
      at Object.<anonymous> (test.js:56:1)
      at loader (node_modules/babel-core/node_modules/babel-register/lib/node.js:127:5)
      at Object.require.extensions.(anonymous function) [as .js] (node_modules/babel-core/node_modules/babel-register/lib/node.js:137:7)
      at Array.forEach (native)
      at node.js:961:3

  2) async actions shall not pass:
     Error: done() called multiple times
      at Suite.<anonymous> (test.js:57:3)
      at Object.<anonymous> (test.js:56:1)
      at loader (node_modules/babel-core/node_modules/babel-register/lib/node.js:127:5)
      at Object.require.extensions.(anonymous function) [as .js] (node_modules/babel-core/node_modules/babel-register/lib/node.js:137:7)
      at Array.forEach (native)
      at node.js:961:3
```

The detailed analysis and the minimal project to reproduce this bug is listed in this gist: https://gist.github.com/MrOrz/50bda5e23db2c39713e1

This pull request fixes the bug by turning `dispatch()` in `mockStore` into no-op after `done()` is called.

### DIFF
--- a/docs/recipes/WritingTests.md
+++ b/docs/recipes/WritingTests.md
@@ -122,6 +122,8 @@ function mockStore(getState, expectedActions, done) {
   }
 
   function mockStoreWithoutMiddleware() {
+    let hasDoneBefore = false
+  
     return {
       getState() {
         return typeof getState === 'function' ?
@@ -130,16 +132,22 @@ function mockStore(getState, expectedActions, done) {
       },
 
       dispatch(action) {
+        if (hasDoneBefore) {
+          return
+        }
+      
         const expectedAction = expectedActions.shift()
 
         try {
           expect(action).toEqual(expectedAction)
           if (done && !expectedActions.length) {
             done()
+            hasDoneBefore = true
           }
           return action
         } catch (e) {
           done(e)
+          hasDoneBefore = true
         }
       }
     }


MrOrz:
Just found that this patch will cause tests to erroneously pass.

If there are actually more actions dispatched than `expectedActions`, the test will be marked as "passed", which is not desired.

Haven't come up a fix yet.

docs/recipes/WritingTests.md
```diff
@@ -130,16 +132,22 @@ function mockStore(getState, expectedActions, done) {
...
+          return
+        }
+      
         const expectedAction = expectedActions.shift()
 
         try {
           expect(action).toEqual(expectedAction)
           if (done && !expectedActions.length) {
             done()
+            hasDoneBefore = true
```
bfink13: I *think* removing this will solve your issues. There are a few different scenarios to account for:
* We got an expected action and there are still actions left. Continue as normal.
* We got an expected action and there are no actions left.
  * Call `done()`
  * Do NOT no-op dispatch. We've made it through all of our expected actions and have a successful test (so far). Any future actions will fail the expect check, just as an invalid action would. See the next bullet
* We got an unexpected action (it doesn't matter if we have actions left, since we want to fail the test at this point)
  * Throw, call `done(e)`, and set the hasDoneBefore flag.

You will still end up calling `done` multiple times if all actions are successful and then you pass an additional action. However, mocha will report the error correctly despite done being called twice (since you're passing an error on the second `done` call). Subsequent actions will no-op since anytime you reach the catch block, no more actions will be processed.

I have not considered how other testing frameworks would handle this scenario.

gaearon:
Rather than maintain this code, I changed the doc to point to [redux-mock-store](https://github.com/arnaudbenard/redux-mock-store) source. Feel free to improve that package!

Closed via 02ed21cd278951722e9367d840db55c03f016adb.

PR #1067, opened 2015-11-24
Docs example: Use Object.assign also for updating state.todos

johannes-photobox:
The current example throws a syntax error (unexpected token). Not sure if we can even use "...-syntax" in object literals? Maybe this example just got never checked. But please let me know if I'm wrong.

### DIFF
--- a/docs/basics/Reducers.md
+++ b/docs/basics/Reducers.md
@@ -230,10 +230,9 @@ function todoApp(state = initialState, action) {
       })
     case ADD_TODO:
     case COMPLETE_TODO:
-      return {
-        ...state,
+      return Object.assign({}, state, {
         todos: todos(state.todos, action)
-      }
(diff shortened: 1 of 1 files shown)


just-boris:
Spread syntax is possible with ES7 extension.

That doc page already mentioned that you can use `Object.assign` instead, but generally spread looks more readable.

gaearon:
Good call. We shouldn't use ES7 proposals in docs.

just-boris:
@gaearon that's weird, because rest of the examples in this file still uses it.

Does that mean, that I can send a PR with replacement for them too?

gaearon:
The rest of the examples use ES6 array spread operator, not ES7 object spread operator proposal.

just-boris:
Ok, got it, that makes sense

PR #1076, opened 2015-11-27
Make gh-pages examples runnable

kolodny:
This will let people see the examples without having to have a dev environment set up

To see what it would look like:
* http://kolodny.github.io/redux/examples/async/
* http://kolodny.github.io/redux/examples/counter/
* http://kolodny.github.io/redux/examples/real-world/   <- this one needs to use a <Redirect /> :/
* http://kolodny.github.io/redux/examples/shopping-cart/
* http://kolodny.github.io/redux/examples/todomvc/
* http://kolodny.github.io/redux/examples/todos-with-undo/

universal won't work because it needs a node server to serve it

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,5 @@
 **/dist/*
+**/static/*
 **/node_modules/*
 **/server.js
 **/webpack.config*.js
--- a/docs/introduction/Examples.md
+++ b/docs/introduction/Examples.md
@@ -5,7 +5,7 @@ Redux is distributed with a few examples in its [source code](https://github.com
 >##### Note on Copying
 >If you copy Redux examples outside their folders, you can delete some lines at the end of their `webpack.config.js` files. They follow a “You can safely delete these lines in your project.” comment.
 
-## Counter
+## [Counter](examples/counter)
 
 Run the [Counter](https://github.com/rackt/redux/tree/master/examples/counter) example:
 
@@ -24,7 +24,7 @@ It covers:
 * Basic Redux flow
 * Testing
 
-## TodoMVC
+## [TodoMVC](examples/todomvc)
 
 Run the [TodoMVC](https://github.com/rackt/redux/tree/master/examples/todomvc) example:
 
@@ -44,7 +44,7 @@ It covers:
 * Updating nested data
 * Testing
 
-## Todos with Undo
+## [Todos with Undo](examples/todos-with-undo)
 
 Run the [todos-with-undo](https://github.com/rackt/redux/tree/master/examples/todos-with-undo) example:
 
@@ -63,7 +63,7 @@ It covers:
 * Redux flow with two reducers
 * Undo/Redo functionality in Redux with [redux-undo](https://github.com/omnidan/redux-undo)
 
-## Async
+## [Async](examples/async)
 
 Run the [Async](https://github.com/rackt/redux/tree/master/examples/async) example:
 
@@ -103,7 +103,7 @@ It covers:
 * Prefetching state based on input and via asynchronous fetches.
 * Passing state from the server to the client
 
-## Real World
+## [Real World](examples/real-world)
 
 Run the [Real World](https://github.com/rackt/redux/tree/master/examples/real-world) example:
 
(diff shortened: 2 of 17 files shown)


omnidan:
Awesome! Thanks a lot :grin:

gaearon:
@omnidan 

How did you verify that this works? Currently all links on http://redux.js.org/docs/introduction/Examples.html appear broken.

gaearon:
@omnidan 

I reverted. Please don't merge large-scale changes without making 100% sure that they work. :wink:

gaearon:
Before sending a PR, please make sure the links in docs are valid when viewed both Github interface and Gitbook compiled HTML. Currently links from `Examples.md` appear broken in both of them. From Github interface they link to nowhere, and from Gitbook they link to relative paths. Given that the links are not part of docs, the only solution I see that would work here is to make them absolute.

Another problem is `real-world` example “hijacking” the URL because it uses the client history. This looks weird and also breaks links because the result link is not copy-pastable. A better solution would be to switch to using `createHashHistory` when compiled for the GH pages, and keep `createBrowserHistory` for the localhost.

If we want to do this, we’d also want to link back to the source from every example, link back to the page with the list of them, etc. We’d also need to turn these into *real* production builds—with `DefinePlugin`, minification, no DevTools, etc. Otherwise it’s confusing people regarding best practices because people copy-paste our examples all the time without realizing they’re just examples. We don’t want to add a half-baked build.

I appreciate the effort, and if you can work on all these issues and turn this into a holistic picture, I’m all for it. However just publishing current examples as is is not enough.

omnidan:
@gaearon Sorry! It looked fine when I ran this locally. :disappointed:

I'll just ping you before merging changes like this again :sweat_smile: As this was a link issue and not related to the examples actually building, is there any way I could test this (publicly) without merging?

kolodny:
Ok, I think I'll create `redux-counter-example` and so on, and have those link to the github repo, and link the docs to the absolute url of those demos, will that work?

I'll poll the examples folder for changes and pull the changes into the relevant repos

kolodny:
@omnidan I'm try reforking my repo and having it testing it by pushing to my own gh-pages (gitbook and examples)

gaearon:
@omnidan 

There are issues with examples themselves, please see my above comment. Not compiling for production, using `createBrowserHistory` on GH pages, not having links to source are all issues with examples.

You can test GH links by viewing .md files from PR itself and trying to open links. As for links inside Gitbook, I believe you can `npm run docs:watch`, open the watcher and try opening files from Examples page?

gaearon:
@kolodny 

What problem are you looking to solve with making examples available online? This can be handy to convince people that “Redux works” but we've passed the stage where this is necessary. What else do we need them for? To show how to create production builds?

omnidan:
@gaearon I'll just leave ping you next time then :wink:

I think it's cool to be able to play with devtools (in the examples that have them) or just get a feel for what redux can do, but of course you could always just run them on your own.

gaearon:
If we provide an official example including DevTools in production build, that's what people will do. I'd rather be for a separate hosted DevTools demo unrelated to the `real-world` example.

omnidan:
I agree - maybe it makes more sense to have a live example for the devtools (in the devtools repo) and link to that?

gaearon:
Yes—this belongs to DevTools rather than Redux itself.

kolodny:
@gaearon I was thinking mostly demo-ability in situations without a dev setup like when sitting with a client or in middle of a meeting when you can only send someone a link.

gaearon:
A link to trivial examples without design? I’m afraid that would only freak out the client :stuck_out_tongue: 
We have amazing third-party examples like http://soundredux.io. Maybe we should emphasize them more.

kolodny:
yup, that's cooler, maybe we should consider having an *Apps Built With Redux* page

gaearon:
>maybe we should consider having an Apps Built With Redux page

:+1:

Want to assemble an initial list? That would involve finding these apps, splitting them into open source / closed source, making screenshots and creating a new doc page a la http://backbonejs.org/#examples.

kolodny:
Sure! I'd love to

gaearon:
https://github.com/rackt/redux/issues/1078

PR #1077, opened 2015-11-27
Revert "Make gh-pages examples runnable"

gaearon:
Reverts rackt/redux#1076 because it breaks the website.
None of the links on http://redux.js.org/docs/introduction/Examples.html work, and links like http://redux.js.org/examples/todomvc/ don't work either.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,4 @@
 **/dist/*
-**/static/*
 **/node_modules/*
 **/server.js
 **/webpack.config*.js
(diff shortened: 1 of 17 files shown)


kolodny:
hmm, it appears that the bundle.js files never got added to the git index, I thought it would with https://github.com/rackt/redux/blob/master/package.json#L33

gaearon:
I'll reply in original PR soon, just a sec

PR #1081, opened 2015-11-28
Remove unnecesary brackets and use single backticks for templating.

nschurmann:
Hello!, i removed unneeded brackets in the if statements and use single back ticks for the string templates, this reduce some lines of code.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -31,9 +31,8 @@ export var ActionTypes = {
  * and subscribe to changes.
  */
 export default function createStore(reducer, initialState) {
-  if (typeof reducer !== 'function') {
+  if (typeof reducer !== 'function')
     throw new Error('Expected the reducer to be a function.')
-  }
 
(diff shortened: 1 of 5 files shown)


ellbee:
The string will include the indentation if you use single backticks across multiple lines like that. That is why the test `Utils combineReducers warns if input state object does not match state object returned by reducer` is failing.

Regarding the brackets, I personally like to include them, but this change does match the style React Router uses. I like semi-colons too, but it hasn't bothered me since we got rid of them.

timdorr:
Also, changing the backticks makes the output from the error a little more messy. You have both the linebreak and some extra spacing on the line that follows.

Overall, @rackt is trying to keep a standard style, which is over in [`eslint-config-rackt`](https://github.com/rackt/eslint-config-rackt). If you have a style change proposal, it's probably better suited to start over there and work its way back here. That way we can enforce it with eslint.

PR #1082, opened 2015-11-29
mapValues could return source object when run with identity function

slorber:
Not a big deal, I thought I could simplify combineReducers with it but actually not... still doing the PR in case you want to merge...


gaearon:
I don't think we use it like this, so I'm closing.

PR #1084, opened 2015-11-29
Stay consistent for shouldCallAPI method

cyrilf:
Regarding the previous examples, I think that `shouldCallAPI` method isn't consistent.

```
if (posts[userId]) {
// There is cached data! Don't do anything.
    return
}
```

was translated to:

`shouldCallAPI: (state) => !state.users[userId]`

I think it's `shouldCallAPI: (state) => !state.posts[userId]`


ellbee:
Hm, I believe you are right. Thanks!

PR #1085, opened 2015-11-30
Update Troubleshooting.md

xcatliu:
The example which mutate state is using `switch case` without `return state` may misleading some people. So I modified the example to match the rewritten example.

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -26,12 +26,14 @@ function todos(state = [], action) {
       text: action.text,
       completed: false
     })
+    return state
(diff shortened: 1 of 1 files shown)


ellbee:
Oh, the code in the example isn't quite right because the `ADD_TODO` case falls through to the `COMPLETE_UNDO` case. We could put some `break`s in to fix this instead of the returns, but as you have prepared this pull request let's go with `return`s for now. Thanks for the help!

xcatliu:
@ellbee Yes `break`s could also work fine. Thanks for the awesome redux docs, it helps me a lot!

PR #1086, opened 2015-11-30
Make visibleTodos a required property

mjethani:
Since TodoList's todos property is required, visibleTodos should be required too.


ellbee:
Thanks

PR #1087, opened 2015-11-30
Fix docs/api/Store.md Example

xcatliu:
Fix a bug in `docs/api/Store.md`'s Example.
`currentValue` was not initialized.

### DIFF
--- a/docs/api/Store.md
+++ b/docs/api/Store.md
@@ -101,11 +101,11 @@ function select(state) {
   return state.some.deep.property
 }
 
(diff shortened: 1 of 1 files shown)


ellbee:
Hi @xcatliu, I don't think that `currentValue` being `undefined` is a bug here. It should work fine?

xcatliu:
@ellbee Oh, It's fine if `currentValue` being `undefined` at beginning. I'll close the pr.

PR #1089, opened 2015-11-30
Use ES6 syntax in fetchPosts()

mjethani:
The code snippet uses an ES5-style function expression. I've changed it to use an arrow function instead.


ellbee:
This change is consistent with the `fetchPosts` implementation further down the page and also on the `ExampleRedditAPI.md` so looks good to me. Thanks.

PR #1096, opened 2015-12-03
Add an optional 'callback' to dispatch

mhodgson:
Called when all listeners have resolved. 

This enables a dispatch caller to run a callback once all changes from an action have been flushed and rendered (in my case to the DOM via React). 

I didn't include a promise implementation here since I don't know what people's preferences are. 

Thoughts about this? I have a specific use case around DOM selection binding that this enables.

See https://github.com/rackt/react-redux/pull/208 for related change to react-redux.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -92,12 +92,17 @@ export default function createStore(reducer, initialState) {
    * a `type` property which may not be `undefined`. It is a good idea to use
    * string constants for action types.
    *
+   * @param {Function} callback An optional function to be called when any and all
+   * promises returned by listeners have been resolved. This can be used for cases
+   * when you need to perform a callback after asyncronous changes or rendering
+   * caused by an action have been completed.
+   *
    * @returns {Object} For convenience, the same action object you dispatched.
    *
    * Note that, if you use a custom middleware, it may wrap `dispatch()` to
    * return something else (for example, a Promise you can await).
    */
-  function dispatch(action) {
+  function dispatch(action, callback=null) {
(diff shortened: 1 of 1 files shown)


mhodgson:
@gaearon @ellbee Any thoughts here? This is a non-breaking change with no effect on the existing API and enables callbacks that need to occur after all updates are complete (i.e. inspecting selection state, taking screenshots, re-enabling listeners, etc.).

just-boris:
If you want to get this working, you should pass callback through all store decorations in [compose](https://github.com/rackt/redux/blob/master/src%2Futils%2Fcompose.js) method.

gaearon:
I don't think we want to get this in.
The store acts like a very simple event emitter.
We don't want to make it more complex.

I don't know what a better solution would look like but it shouldn't involve Redux.
It's similar to how you wouldn't modify Rx or EventEmitter to fix a similar problem.
It's just not their concern.

mhodgson:
@gaearon as far as I can tell there is no way to use redux for this use case then. Why is having a callback for when a dispatch is 'complete' outside of the scope of Redux?

gaearon:
Because `dispatch` is synchronous. It doesn't make sense to add an async callback to a synchronous method. Asynchrony is an implementation detail of your *subscribers* and should be dealt with on their level—whether in React Redux, React itself, or via some other means.

mhodgson:
Thanks for the explanation. I *think* I see a path forward by adding a decorator outside of @connect and overriding the listener trigger in Connect.

PR #1102, opened 2015-12-04
Fix typo in AsyncActions.md

Aaronik:
...and only render after the Promise it returns, completes.
->
...and only render after the Promise it returns completes.


ellbee:
Thanks!

PR #1103, opened 2015-12-04
Docs: Added clarity re return value of store.dispatch

tomatau:
Added some more information about the return value of store.dispatch.

It should be the dispatched action but could also be the return value of the middleware chain, such as a promise of the dispatched action

### DIFF
--- a/docs/api/Store.md
+++ b/docs/api/Store.md
@@ -47,7 +47,7 @@ The store’s reducing function will be called with the current [`getState()`](#
 
 #### Returns
 
-(Object<sup>†</sup>): The dispatched action.
+(Object<sup>†</sup>): The dispatched action or the return value of the middleware chain e.g. a Promise of the dispatch action object.
 
 #### Notes
 
--- a/docs/api/Store.md
+++ b/docs/api/Store.md
@@ -47,7 +47,7 @@ The store’s reducing function will be called with the current [`getState()`](#
 
 #### Returns
 
-(Object<sup>†</sup>): The dispatched action or the return value of the middleware chain e.g. a Promise of the dispatch action object.
+(Object<sup>†</sup>): The dispatched action (see notes).
 
 #### Notes
 


ellbee:
The reason that it is specified solely as a plain object is explained in the Notes section below. Is there anything there you would suggest to change to make it clearer?

tomatau:
True! Perhaps adding a "see notes"?

API references should be easy to lookup quickly - notes are easy to overlook.

gaearon:
"Promise of dispatch action object" sounds very specific as if we enforce something like that. I'd just say "a Promise" there.

tomatau:
I feel like the focus here should be on the end value being the action object. It gives middleware creators more direction and can improve consistency within the ecosystem.

gaearon:
I don't think this is necessarily true at all. Async action creators often dispatch multiple actions. I don't think it makes sense to suggest it's the action that the Promise should resolve to.

tomatau:
When middleware can return anything, it makes the return from store.dispatch more complicated and less type safe. Whereas the general pattern of `return next(action)` is very easy to interpret.

When multiple actions are dispatched, the middleware should be able to comfortably provide a promise or stream of actions.

I don't see the down side to encouraging this?

pburtchaell:
I agree that a standard return value of `store.dispatch` would be preferable. I would argue it ensures that middleware across the Redux ecosystem as a whole will work similarly in that aspect. 

I don't think a standard return value should be _required_, but I think encouraging a standard would be beneficial. I say this because I also think that _if it makes sense_ for a middleware to return a value other than the standard, e.g., a promise object, then that is acceptable. Maybe the update to the docs should express this?

pburtchaell:
It may also be worth noting that this is coming from discussions at https://github.com/pburtchaell/redux-promise-middleware/issues/39 and https://github.com/este/este/issues/633.

ellbee:
I think adding `see notes` is a improvement, so I'm going to merge this. Maybe for the standard return value it would be worth creating a standard like [FSA](https://github.com/acdlite/flux-standard-action) which we could link to from the ecosystem page in docs?

ellbee:
Oh, where are my manners? Thanks Tomatau!

tomatau:
@ellbee Hah thanks, was a very trivial PR :)

I very like the idea of a standard with link in the docs -- just a github readme page should be enough. I don't see how any libraries could really help here, perhaps a middleware to check against the dispatch return from following middleware.

PR #1107, opened 2015-12-05
Change MD to be more understandable on example.

Cplantijn:


omnidan:
Maybe `logs` or `prints` would be more accurate here, or alternatively remove the `console.log`.

Cplantijn:
Yes, you're right.

omnidan:
Thanks! :grin:

PR #1111, opened 2015-12-07
Update ThreePrinciples.md

urbanvikingr:
Change `Mutations` with `Actions`. I had a discussion with another React developer and it seemed to be some confusing and it turned that it was caused by a misunderstanding of the concept of functional programming, pure functions and the use of the word `mutation`. I hope that this helps.

### DIFF
--- a/docs/introduction/ThreePrinciples.md
+++ b/docs/introduction/ThreePrinciples.md
@@ -46,7 +46,7 @@ store.dispatch({
 })
 ```
 
-### Transformation is written as pure functions
+### State changes are made with pure functions
 
 **To specify how the state tree is transformed by actions, you write pure [reducers](../Glossary.md#reducer).**
 
--- a/docs/introduction/ThreePrinciples.md
+++ b/docs/introduction/ThreePrinciples.md
@@ -46,7 +46,7 @@ store.dispatch({
 })
 ```
 
-### Actions are written as pure functions
+### Transformation is written as pure functions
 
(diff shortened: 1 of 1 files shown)


sompylasar:
@urbanvikingr These are mutations, not actions -- the mutations to the state. You don't write actions as functions -- they are plain objects.

urbanvikingr:
@sompylasar Mutation mean that you change the existing state, but Redux is creating a new state with the changes. I think that the word mutation gives some confusing. The text underneath clearly states that the Redux state tree is immutable.

urbanvikingr:
I changed `Actions` to `Transformation`. I hope that this helps.

sompylasar:
That makes more sense. Let's see what @gaearon thinks.

gaearon:
Maybe just "changes are described by pure functions"?

urbanvikingr:
I updated to `Changes are made with pure functions`, because it would be more readable, but I like `Changes are done by pure functions` too. I would just like to avoid using the word mutation. I hope that this helps.

acdlite:
Action creators aren't necessarily pure

urbanvikingr:
@acdlite The section is about Reducers :)

acdlite:
Haha yes it is, was skimming through my notifications and got my wires crossed. Carry on :D

ellbee:
Close to what @gaearon suggested, and reads nicely, so I think this can be merged. Thanks @urbanvikingr!

urbanvikingr:
Thank you :)

PR #1118, opened 2015-12-09
Refactor combineReduer's Unexpected State Shape Warning

jridgewell:
The shape of the previous state is irrelevant, what we’re actually
testing is that state has the same shape as our reducer object
`finalReducers`.

Also fixes the algorithm to be O(n) by using `#hasOwnProperty` vs doing
an array search O(n<sup>2</sup>).

### DIFF
--- a/src/utils/combineReducers.js
+++ b/src/utils/combineReducers.js
@@ -15,8 +15,8 @@ function getUndefinedStateErrorMessage(key, action) {
   )
 }
 
-function getUnexpectedStateKeyWarningMessage(inputState, outputState, action) {
-  var reducerKeys = Object.keys(outputState)
+function getUnexpectedStateShapeWarningMessage(inputState, reducers, action) {
+  var reducerKeys = Object.keys(reducers)
   var argumentName = action && action.type === ActionTypes.INIT ?
     'initialState argument passed to createStore' :
     'previous state received by the reducer'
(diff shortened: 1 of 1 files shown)


acdlite:
This looks good to me. :+1: for using `hasOwnProperty` especially.

acdlite:
Someone else should review this too, though, since I'm not as familiar with this module.

src/utils/combineReducers.js
```diff
@@ -15,8 +15,8 @@ function getUndefinedStateErrorMessage(key, action) {
   )
 }
 
-function getUnexpectedStateKeyWarningMessage(inputState, outputState, action) {
-  var reducerKeys = Object.keys(outputState)
+function getUnexpectedStateShapeWarningMessage(inputState, reducer, action) {
```
gaearon: The argument should probably be called `reducers`, not `reducer`.

gaearon:
I think @ellbee should remember this better than me.
Basically we need to make sure we don't regress on the warning we introduced in https://github.com/rackt/redux/issues/715#issuecomment-139843090 and https://github.com/rackt/redux/pull/720.

ellbee:
I have just been playing with this, and I am pretty sure that this behaves the same as what is currently there.

gaearon:
Leaving this for you to review then :+1:

src/utils/combineReducers.js
jridgewell: Updated.

ellbee:
So, the non-cosmetic changes in this PR are the switch to `hasOwnProperty` and passing an empty object when initial state is not specified. I did some testing and they both look good. Thanks @jridgewell.

gaearon:
Out in 3.0.6, thanks!

PR #1120, opened 2015-12-09
Remove unnecessary arrow parens

jridgewell:
The codebase seems to not use parenthesis around a single variable arrow
function, except with the following. This forces the pattern.


gaearon:
We are using https://github.com/rackt/eslint-config-rackt, so any further changes to the rules should probably happen there, and we would adjust the code here to match them.

PR #1122, opened 2015-12-10
Reducer Assertions when using createStore

jridgewell:
This extracts the two generic reducer assertions (reducer sanity and never returning undefined) from `combineReducers`, so that they can be used with `createStore`'s reducer.

### DIFF
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -2,6 +2,7 @@ import expect from 'expect'
 import { createStore, combineReducers } from '../src/index'
 import { addTodo, dispatchInMiddle, throwError, unknownAction } from './helpers/actionCreators'
 import * as reducers from './helpers/reducers'
+import ActionTypes from '../src/utils/actionTypes'
 
 describe('createStore', () => {
   it('exposes the public API', () => {
@@ -29,7 +30,7 @@ describe('createStore', () => {
     ).toThrow()
 
     expect(() =>
-      createStore(() => {})
+      createStore(() => ({}))
     ).toNotThrow()
   })
 
@@ -90,7 +91,7 @@ describe('createStore', () => {
     ])
 
     store.dispatch(unknownAction())
-    expect(store.getState()).toEqual([ 
+    expect(store.getState()).toEqual([
       {
         id: 1,
         text: 'Hello'
@@ -140,11 +141,11 @@ describe('createStore', () => {
       {
         id: 3,
         text: 'Perhaps'
-      }, 
+      },
       {
         id: 1,
         text: 'Hello'
-      }, 
+      },
       {
         id: 2,
         text: 'World'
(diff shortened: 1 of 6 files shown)


test/createStore.spec.js
```diff
@@ -29,7 +30,7 @@ describe('createStore', () => {
     ).toThrow()
 
     expect(() =>
-      createStore(() => {})
+      createStore(() => ({}))
```
kjbekkelund: These two are not the same — any reason for the change?

acdlite:
Those checks are only in `combineReducers()` and not `createStore()` to allow for advanced users to opt-out of them, if they desire.

jridgewell:
> allow for advanced users to opt-out of them, if they desire.

Wouldn't that be better accomplished with `process.env.NODE_ENV` like https://github.com/rackt/redux/blob/master/src/utils/combineReducers.js#L128?

> only in combineReducers() and not createStore()

Treating them differently seems odd.

acdlite:
Are you suggesting that advanced users set their `NODE_ENV` to `'production'` as a way to opt-out? :)

Hmm, why does it seem odd? They're two different types of functions. `createStore()` creates your store, `combineReducers()` is a particular way of composing many reducers into one.

If the goal is to extract the sanity checks into a separate module, then perhaps we should export those modules separately, rather than force everyone to use them no matter what.

jridgewell:
> Are you suggesting that advanced users set their NODE_ENV to 'production' as a way to opt-out? :)

Yup.

> Hmm, why does it seem odd? 

Because they both handle reducers. I'd argue this is more important in `createStore` since it's the real API and `combineReducers` is just sugar to make making a reducer nicer. But `combineReducers` will always return an object, so it defeats the check in `createStore`, so it needs the checks as well. (I have a bad building analogy I'll spare you from. :wink:)

Really, anything that handles reducers should be guiding users to the "proper" way: never return `undefined`.

acdlite:
Setting `NODE_ENV` to `'production'` is not a viable option; now you're opt-ing out of all warnings entirely.

I would rather see these sanity checks exported as a separate, opt-in module that wraps a reducer. `createStore()` should be exclusively concerned with state management.

I can see where you're coming from, though. I do agree with the "never return `undefined`" rule — even if you want to return nothing, you can return `null` instead. But every time someone suggests that we add more convenience features to `createStore()` (in the beginning, middleware, `combineReducers()`, and the thunk middleware were all built-in) I feel a moral obligation to push back. Indeed, if it were up to me, `createStore()` wouldn't perform an initial dispatch, but I already lost that battle :D

gaearon:
We like to think `createStore` is a core API with minimum limitations and perf hitting checks, and `combineReducers` is a helper designed for beginners to get up to speed quickly. So `createStore` errs on the side of flexibility, and `combineReducers` errs on the side of safety.

>Wouldn't that be better accomplished with process.env.NODE_ENV like https://github.com/rackt/redux/blob/master/src/utils/combineReducers.js#L128?

Adding even a single option is a huge maintenance burden. For every issue, there are two possible code paths how it could've happened. We also never want to have different behavior for hard errors in development and production. If you turn an error assertion in production, it will surface later in code (because it's likely buggy) but stack traces will be useless because you won't be able to reproduce them in development.

PR #1125, opened 2015-12-10
Added a note concerning babel-loader v6 and polyfills

vnctaing:
Babel-loader seems to come with Promise polyfills

### DIFF
--- a/docs/advanced/AsyncActions.md
+++ b/docs/advanced/AsyncActions.md
@@ -384,6 +384,8 @@ export function fetchPosts(reddit) {
 >import 'babel-core/polyfill'
 >```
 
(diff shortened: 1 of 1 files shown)


gaearon:
Where can I read more about this change in Babel / babel-loader 6?

vnctaing:
You can read it here : https://github.com/babel/babel-loader, in the "Custom polyfills  (e.g. Promise library)". Actually Babel comes with core.js which includes itself the Promise polyfill.

timdorr:
This actually isn't correct. You still have to include the polyfill, either before your main script in `entry` in your webpack config, or via the `import`, as the docs say. The one that isn't correct in Babel 6 is that it's now a separate package and should be `import 'babel-polyfill'`.

PR #1133, opened 2015-12-12
Fix id-related bug in example ToDo app and update README

arjunblj:
Hey Dan — I believe this should fix https://github.com/rackt/redux/issues/1129 by generating a unique ID for every to do and updating correctly when a to do is completed. I went ahead and cleaned up the naming to reflect this as well.

### DIFF
--- a/docs/basics/ExampleTodoList.md
+++ b/docs/basics/ExampleTodoList.md
@@ -52,12 +52,18 @@ export const VisibilityFilters = {
  * action creators
  */
 
+let nextTodoId = 0;
+
 export function addTodo(text) {
-  return { type: ADD_TODO, text }
+  return {
+    type: ADD_TODO,
+    id: nextTodoId++,
+    text
+  };
 }
 
-export function completeTodo(index) {
-  return { type: COMPLETE_TODO, index }
+export function completeTodo(id) {
+  return { type: COMPLETE_TODO, id }
 }
 
 export function setVisibilityFilter(filter) {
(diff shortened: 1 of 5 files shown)


docs/basics/ExampleTodoList.md
```diff
@@ -89,17 +95,18 @@ function todos(state = [], action) {
...
+          id: action.id
           text: action.text,
           completed: false
         }
       ]
     case COMPLETE_TODO:
       return [
-        ...state.slice(0, action.index),
-        Object.assign({}, state[action.index], {
+        ...state.slice(0, action.id),
```
gaearon: This assumes that `id` is also an index, but this may not be true. Instead, I suggest you to use `map()` to iterate over the todos, and return either the original or the copy with different `completed` value depending on its ID. I also suggest you to separate the logic to update a single todo into a separate `todo` reducer like [in my Egghead course](https://egghead.io/lessons/javascript-redux-reducer-composition-with-objects?series=getting-started-with-redux): 

```js
const todo = (state, action) => {
  switch (action.type) {
    case 'ADD_TODO':
      return {
        id: action.id,
        text: action.text,
        completed: false
      };
    case 'TOGGLE_TODO':
      if (state.id !== action.id) {
        return state;
      }

      return {
        ...state,
        completed: !state.completed
      };
    default:
      return state;
  }
};

const todos = (state = [], action) => {
  switch (action.type) {
    case 'ADD_TODO':
      return [
        ...state,
        todo(undefined, action)
      ];
    case 'TOGGLE_TODO':
      return state.map(t =>
        todo(t, action)
      );
    default:
      return state;
  }
};
```

arjunblj:
@gaearon I've updated to reflect your comment + squashed my changes.

gaearon:
This is looking great! Can you please also update `todos-with-undo` example in the repo that uses a very similar code?

arjunblj:
@gaearon Updated the example. :smiley:

gaearon:
Thank you!

PR #1144, opened 2015-12-15
Fix nock response to return a body

TimurM:
-The thunk action 'fetchTodos' resolves the promise and then passes json.body to the fetchTodosSuccess. 
-The bug was returning 'undefined' for json.body because the mock response was not returning a body


gaearon:
Thank you!

PR #1151, opened 2015-12-16
perhaps it should be Monkeypatching

PuYart:


gaearon:
Why?
I think it reads fine as is.

PR #1157, opened 2015-12-18
Set listener scope

tiagofreres:
How about explicitly define a scope to listener instead implicitly using **store**? (anyways you could use both with this update).


gaearon:
I think any reasonable efficient listener would likely want to know the previous state too. At this point a custom wrapper like [redux-watch](https://github.com/jprichardson/redux-watch) works better than complicating core API.

PR #1173, opened 2015-12-23
remove insin/redux-action-utils (deprecated)

chentsulin:
'insin/redux-action-utils' has been deprecated, so should be removed from the document.


gaearon:
Let's put https://github.com/pauldijou/redux-act there?

chentsulin:
Sounds great~ Already updated the PR.

PR #1175, opened 2015-12-23
Make App container clearer on counter example

gabrielhpugliese:
* Instead of exporting the connected Counter directly, create a visible wrapper to improve readability
* Add some docstrings to make connect params clearer on first read.

The motivation was trying to read this code but failing on first try, because I (and a friend of mine). couldn't understand what was happening here without writing this down.

### DIFF
--- a/examples/counter/components/Counter.js
+++ b/examples/counter/components/Counter.js
@@ -4,9 +4,8 @@ class Counter extends Component {
   render() {
     const { increment, incrementIfOdd, incrementAsync, decrement, counter } = this.props
     return (
-      <p>
+      <span>
         Clicked: {counter} times
-        {' '}
         <button onClick={increment}>+</button>
         {' '}
         <button onClick={decrement}>-</button>
@@ -14,7 +13,7 @@ class Counter extends Component {
         <button onClick={incrementIfOdd}>Increment if odd</button>
         {' '}
         <button onClick={() => incrementAsync()}>Increment async</button>
-      </p>
+      </span>
     )
   }
 }
--- a/examples/counter/containers/App.js
+++ b/examples/counter/containers/App.js
@@ -1,16 +1,10 @@
-import { bindActionCreators } from 'redux'
-import { connect } from 'react-redux'
-import Counter from '../components/Counter'
-import * as CounterActions from '../actions/counter'
+import React, { Component } from 'react'
+import CounterContainer from './CounterContainer'
 
-function mapStateToProps(state) {
-  return {
-    counter: state.counter
+export default class App extends Component {
+  render() {
+    return (
+      <h1>Counter: <CounterContainer /></h1>
+    )
   }
 }
-
-function mapDispatchToProps(dispatch) {
-  return bindActionCreators(CounterActions, dispatch)
-}
-
-export default connect(mapStateToProps, mapDispatchToProps)(Counter)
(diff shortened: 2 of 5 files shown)


gabrielhpugliese:
Well, I think Travis didn't like it https://travis-ci.org/rackt/redux/jobs/98581632#L789
Think it needs to re-run.

gaearon:
I'd avoid creating a separate App because people will incorrectly copy this pattern and create unnecessary extra components.

gabrielhpugliese:
So wrapping the ConnectedComponent inside an App Component is disencouraged?

naw:
Typically, the `App` component is going to render multiple things, not just one component, so the example is actually an unusual case where you _can_ collapse the `App` and `Counter` together.

Of course there is a ton of flexibility, but I think there are two common patterns:

1. Make `App` a connected component that renders a dumb  `Counter` component, passing down props explicitly to `Counter`

2. Make `Counter` a connected component, with `mapStateToProps` living inside `Counter.js`, and `App` rendering `Counter`.

In the example counter app, due to the fact that you can collapse `App` and `Counter` together, neither pattern is followed.

I agree with @gabrielhpugliese that the example is a little confusing in this regard.

I agree with @gaearon that this PR introduces a different kind of confusion, even though it correctly recognizes that `App` doesn't have to be a connected version of `Counter`.

Just my opinion, but I think the example would be clearer if you moved `connect`, `mapStateToProps`, etc. into `Counter.js` (following pattern 2), and then made `app.js` contain a simple `App` component with a render method (like in this PR). You could add some extra DOM (like a "Hello World, this is the App component" string) so it's clear `App` can do something more than just wrap `ConnectedCounter`, which I would claim is more typical than just connecting a single dumb component.

gabrielhpugliese:
Yes, that is **exactly** what I'm thinking! :heart: 
And I've done that in my own example project (even better with @gaearon's hint): https://gist.github.com/gabrielhpugliese/3c3bf1fd42153ae24440
Then the `containers/App.js` import it and use like I've done in this PR.

gaearon:
>Just my opinion, but I think the example would be clearer if you moved connect, mapStateToProps, etc. into Counter.js (following pattern 2), and then made app.js contain a simple App component with a render method (like in this PR). You could add some extra DOM (like a "Hello World, this is the App component" string) so it's clear App can do something more than just wrap ConnectedCounter, which I would claim is more typical than just connecting a single dumb component.

:+1: on this.

Let's rename `containers/App` to `containers/CounterContainer` and add a separate `containers/App` that renders it. We can change `components/Counter` to render just a number with buttons inside a `span`, and add `<h1>Counter: <CounterContainer /></h1>` inside the `App`.

gabrielhpugliese:
:+1:

gabrielhpugliese:
I'm refactoring the components:

1. Counter that renders an `<h1>` with the {counter}
2. CounterButtons that renders all buttons inside a `<div>`

Then I will make App that renders both together.

Question: Should I connect the entire App and pass props down to children OR connect both those components and do not connect the App?

gaearon:
Why create CounterButtons? Seems like they can stay part of the Counter.

I would ConnectContainer only. Please see my previous comment.

gaearon:
Whoops I meant CounterContainer. It would be the result of calling connect() on Counter. Just like I wrote before.

gabrielhpugliese:
Ok, I will submit the PR the way you said and then I will do a gist with I think it's more natural for me and we can discuss?

gabrielhpugliese:
This is how it looked like - now I got your point, but I would change those `{' '}` and wrap buttons in a div and the title in another div so one would be below the other. But it would be a job for another PR if needed:

image

Thanks for your patience

gabrielhpugliese:
Woops, gotta fix the tests

timdorr:
Closing in favor of #1288

PR #1181, opened 2015-12-26
Clarify installation instructions

jmeas:
When I read the installation instructions, it didn't come across to me that the npm package could be used without a CJS module bundler. I tried to reorganize the instructions here to make it a bit more clear.

===

The rest of the commit msg:

> The previous instructions may have come across that the only way to use Redux with npm was with a module bundler, like Browserify or Webpack.

> This reorganizes the text in an effort to more explicitly explain that it's only the extensions require a CJS module bundler.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,16 +45,38 @@ To install the stable version:
 npm install --save redux
 ```
 
-Most likely, you’ll also need [the React bindings](https://github.com/rackt/react-redux) and [the developer tools](https://github.com/gaearon/redux-devtools).
+This requires using the [npm](https://www.npmjs.com/) package manager. It includes a [UMD](https://github.com/umdjs/umd) build, which is compatible with many popular JavaScript module loaders and environments.
+
+#### Other Installation Methods
+
+We **strongly** recommend using [npm](https://www.npmjs.com/) over these other options, as most of the [libraries complementary to Redux](#complementary-libraries) are only available on [npm](https://www.npmjs.com/).
+
+However, if you don’t yet use [npm](https://www.npmjs.com/) or a modern module bundler, you have a few options.
+
+##### Bower
+
+If you're using [Bower](http://bower.io/), this command will install the stable version of Redux:
+
+```
+bower install redux=https://npmcdn.com/redux/dist/redux.js
+```
+
+##### CDNs
+
+Redux is available via [cdnjs](https://cdnjs.com/libraries/redux) and [npmcdn](https://npmcdn.com/redux/dist/redux.min.js). The files on these CDNs are [UMD](https://github.com/umdjs/umd) builds, which means that they will export a browser global, `window.redux`, if you're not using a tool like [Webpack](http://webpack.github.io), [Browserify](http://browserify.org/), or [RequireJS](http://requirejs.org/) to build your library.
+
+An example of using the global Redux object can be seen in [the vanilla counter example](https://github.com/rackt/redux/blob/765a3a5742a591adba7ae8b61e0516c49ca594bd/examples/counter-vanilla/index.html).
+
+#### Complementary Libraries
+
+Most likely, you’ll also need [the React bindings](https://github.com/rackt/react-redux) and [the developer tools](https://github.com/gaearon/redux-devtools) while working with Redux.
 
 ```
 npm install --save react-redux
 npm install --save-dev redux-devtools
 ```
 
-This assumes that you’re using [npm](https://www.npmjs.com/) package manager with a module bundler like [Webpack](http://webpack.github.io) or [Browserify](http://browserify.org/) to consume [CommonJS modules](http://webpack.github.io/docs/commonjs.html).
-
-If you don’t yet use [npm](https://www.npmjs.com/) or a modern module bundler, and would rather prefer a single-file [UMD](https://github.com/umdjs/umd) build that makes `Redux` available as a global object, you can grab a pre-built version from [cdnjs](https://cdnjs.com/libraries/redux). We *don’t* recommend this approach for any serious application, as most of the libraries complementary to Redux are only available on [npm](https://www.npmjs.com/).
+Use of the extensions requires that you’re using [npm](https://www.npmjs.com/) in conjunction with a module bundler like [Webpack](http://webpack.github.io) or [Browserify](http://browserify.org/) to consume [CommonJS modules](http://webpack.github.io/docs/commonjs.html).
 
 ### The Gist
 


README.md
```diff
@@ -135,7 +135,7 @@ This architecture might seem like an overkill for a counter app, but the beauty
 
 So, what are you waiting for?
 
-#### [Watch the 30 Free Videos!](https://egghead.io/series/getting-started-with-redux) 
+#### [Watch the 30 Free Videos!](https://egghead.io/series/getting-started-with-redux)
```
jmeas: I can revert this change if ya want. My text editor stripped out the extra space.

gaearon:
Let's add an explicit paragraph explaining how to install and use the UMD build.
It's available both inside NPM package and standalone via `npmcdn`.
In fact you can even use it with Bower:

```
bower install redux=https://npmcdn.com/redux/dist/redux.js
```

timdorr:
@jmeas Can you add that npmcdn link? Other than that, this looks good! :smile:

timdorr:
@jmeas Any chance you can get to this? I can cherry pick in and add the npmcdn link for you, if not.

jmeas:
I'll add it to my to-do list, and try to get to it in a few days.

Since all of the builds export UMD, I think there's some good rephrasing that can be done to capture that important detail.

timdorr:
There's also the counter-vanilla example now that uses npmcdn, which you may way to point to.

jmeas:
Hmmm..."counter-vanilla"? Not sure if I know what ya mean.

timdorr:
https://github.com/rackt/redux/blob/master/examples/counter-vanilla/index.html

jmeas:
Ah, right, haha. Cool, and that's a great idea. I'll add a link to that!

jmeas:
Mmk, updated with some of the feedback y'all gave. Let me know what ya think, and if there are any other changes you think I should make.

README.md
```diff
@@ -1,10 +1,10 @@
 # [Redux](http://rackt.github.io/redux)
 
-Redux is a predictable state container for JavaScript apps.  
+Redux is a predictable state container for JavaScript apps.
 
 It helps you write applications that behave consistently, run in different environments (client, server, and native), and are easy to test. On top of that, it provides a great developer experience, such as [live code editing combined with a time traveling debugger](https://github.com/gaearon/redux-devtools).
 
-You can use Redux together with [React](https://facebook.github.io/react/), or with any other view library.  
+You can use Redux together with [React](https://facebook.github.io/react/), or with any other view library.
```
gaearon: These spaces are significant. They force Markdown to make a newline.

README.md
```diff
@@ -14,18 +14,18 @@ It is tiny (2kB) and has no dependencies.
 [![#rackt on freenode](https://img.shields.io/badge/irc-%23rackt%20%40%20freenode-61DAFB.svg?style=flat-square)](https://webchat.freenode.net/)
 [![Changelog #187](https://img.shields.io/badge/changelog-%23187-lightgrey.svg?style=flat-square)](https://changelog.com/187)
 
->**New! Learn Redux from its creator:  
+>**New! Learn Redux from its creator:
 >[Getting Started with Redux](https://egghead.io/series/getting-started-with-redux) (30 free videos)**
```
gaearon: Same here

README.md
gaearon: Please do :-)

jmeas:
I can fix the spaces. That aside, how are the text changes?

README.md
```diff
@@ -45,16 +45,38 @@ To install the stable version:
...
 
 ```
 npm install --save react-redux
 npm install --save-dev redux-devtools
 ```
 
-This assumes that you’re using [npm](https://www.npmjs.com/) package manager with a module bundler like [Webpack](http://webpack.github.io) or [Browserify](http://browserify.org/) to consume [CommonJS modules](http://webpack.github.io/docs/commonjs.html).
-
-If you don’t yet use [npm](https://www.npmjs.com/) or a modern module bundler, and would rather prefer a single-file [UMD](https://github.com/umdjs/umd) build that makes `Redux` available as a global object, you can grab a pre-built version from [cdnjs](https://cdnjs.com/libraries/redux). We *don’t* recommend this approach for any serious application, as most of the libraries complementary to Redux are only available on [npm](https://www.npmjs.com/).
+Use of the extensions requires that you’re using [npm](https://www.npmjs.com/) in conjunction with a module bundler like [Webpack](http://webpack.github.io) or [Browserify](http://browserify.org/) to consume [CommonJS modules](http://webpack.github.io/docs/commonjs.html).
```
timdorr: Sorry for the death by a thousand comments...

I'd change "Use of the extensions requires that you’re using npm" back to something like "This assumes you are using npm", so you avoid having "use" and "using" in the same sentence.

But good catch changing "npm package manager" to just "npm". That's like saying "ATM machine" :smile:

jmeas: > Sorry for the death by a thousand comments...

No problem at all! I want this to be good. I appreciate the thorough feedback.

> I'd change "Use of the extensions requires that you’re using npm" back to something like "This assumes you are using npm", so you avoid having "use" and "using" in the same sentence.

I agree that the current wording is awkward. At the same time, I want to say something a little stronger than "this assumes," because that leaves it open that there are other methods (perhaps the CDNs above, or Bower). The existing text (before I made any changes) made it out that these tools can only be used with npm, or only with CommonJS. I just pulled down react-redux, though, and it includes a UMD build all the same. I then checked cdnjs and saw it there, too.

Perhaps the recommendation to stick with npm isn't as strong as it was originally worded, and how I've worded it to be? It looks like the three install methods (npm, Bower, CDNs) may work for all of 3 of these libraries?

If that's the case, I might want to do another restructuring to make that more clear.

Another thought is that maybe the original wording was referring to popular React libraries outside of the mini-Redux ecosystem described here. Is that the case?

Sorry I'm not more knowledgeable, I haven't used Redux much yet :)

> But good catch changing "npm package manager" to just "npm". That's like saying "ATM machine"

Heh heh. [Maybe not...](https://www.quora.com/I-keep-hearing-NPM-doesnt-stand-for-Node-Package-Manager-what-does-it-stand-for)

README.md
```diff
@@ -45,16 +45,38 @@ To install the stable version:
...
-Most likely, you’ll also need [the React bindings](https://github.com/rackt/react-redux) and [the developer tools](https://github.com/gaearon/redux-devtools).
+This requires using the [npm](https://www.npmjs.com/) package manager. It includes a [UMD](https://github.com/umdjs/umd) build, which is compatible with many popular JavaScript module loaders and environments.
+
+#### Other Installation Methods
+
+We **strongly** recommend using [npm](https://www.npmjs.com/) over these other options, as most of the [libraries complementary to Redux](#complementary-libraries) are only available on [npm](https://www.npmjs.com/).
+
+However, if you don’t yet use [npm](https://www.npmjs.com/) or a modern module bundler, you have a few options.
+
+##### Bower
```
gaearon: I really don’t want to see us contributing to “Bower is still a good choice for front-end devs” meme by even dedicating it a header. Instead, I would suggest to merge the below two sections, remove cdnjs reference and keep only npmcdn (it’s a better project IMO), and add Bower commend as an afterthought (“If you use an alternative package manager such as Bower, it is possible that your package manager allows installing libraries from URLs, for example: <bower example>”). This way it doesn’t look like we encourage it.

README.md
```diff
@@ -45,16 +45,38 @@ To install the stable version:
...
+
+If you're using [Bower](http://bower.io/), this command will install the stable version of Redux:
+
+```
+bower install redux=https://npmcdn.com/redux/dist/redux.js
+```
+
+##### CDNs
+
+Redux is available via [cdnjs](https://cdnjs.com/libraries/redux) and [npmcdn](https://npmcdn.com/redux/dist/redux.min.js). The files on these CDNs are [UMD](https://github.com/umdjs/umd) builds, which means that they will export a browser global, `window.redux`, if you're not using a tool like [Webpack](http://webpack.github.io), [Browserify](http://browserify.org/), or [RequireJS](http://requirejs.org/) to build your library.
```
gaearon: It’s `window.Redux`, not `window.redux`.

gaearon:
This hasn’t been updated in a while so I made some changes inspired by this in 7b859625b727aa7f8e7fffd4481693d81b82a532.
Thank you!

jmeas:
Thanks @gaearon :v:

PR #1186, opened 2015-12-29
Update to Usage With React documention

urbanvikingr:
The documentation UsageWithReact.md has been updated with code samples according to Egghead.io video series. A complete sample app can be found here https://github.com/urbanvikingr/todo. Please tell me how you would like to bring it to Rackt.

### DIFF
--- a/docs/basics/UsageWithReact.md
+++ b/docs/basics/UsageWithReact.md
@@ -79,18 +79,18 @@ These are all presentational components. They don’t know *where* the data come
 
 If you migrate from Redux to something else, you’ll be able to keep all these components exactly the same. They have no dependency on Redux.
 
-We also have some container components that connects to Redux. The container components are responsible for getting and transforming state. Remember that changes result in a new state as state is immutable.
+We also have some container components that connect to Redux. Container components calculate the props to pass to the presentational components they wrap based on the current state of the Redux store.
 
 * **`AddTodo`** is an input field with a button.
   - `onAddClick(text: string)` is a callback to invoke when a button is pressed.
 
-* **`FilterLink`** gets active link from Redux and pass it as props to the presentational component, Link.
+* **`FilterLink`** gets the current visibility filter and passes it as a prop to `Link` component.
 
-* **`VisibleTodoList`** gets todos from Redux, filter the todos and pass them as props to the presentational component, TodoList.
+* **`VisibleTodoList`** gets todos from Redux, filters the todos and passes them as props to the presentational component, TodoList.
 
 Let’s write the components! We begin with the presentational components, so we don’t need to think about binding to Redux yet.
 
-## Presentational Components
+### Presentational Components
 
 These are all normal React components, so we'll not stop and examine them in detail. Here they are:
 
@@ -99,18 +99,11 @@ These are all normal React components, so we'll not stop and examine them in det
 ```js
 import React from "react";
 
-const Todo = ({
-  onClick,
-  completed,
-  text
-}) => (
+const Todo = ({ onClick, completed, text }) => (
   <li
     onClick={onClick}
     style={{
-      textDecoration:
-        completed ?
-          "line-through" :
-          "none"
+      textDecoration: completed ? "line-through" : "none"
     }}
   >
     {text}
@@ -126,10 +119,7 @@ export default Todo
 import React from "react";
 import Todo from "./Todo";
 
-const TodoList = ({
-  todos,
-  onTodoClick
-}) => (
+const TodoList = ({ todos, onTodoClick }) => (
   <ul>
     {todos.map(todo =>
       <Todo
@@ -175,11 +165,7 @@ export default Footer
 ```js
 import React from "react";
 
-const Link = ({
-  active,
-  children,
-  onClick
-}) => {
+const Link = ({ active, children, onClick }) => {
   if (active) {
     return <span>{children}</span>;
   }
@@ -199,7 +185,7 @@ const Link = ({
 export default Link
 ```
 
-We'll now write the container components.
+### Container Components
 
 #### `containers/AddTodo.js`
 
@@ -278,10 +264,7 @@ import { connect } from "react-redux";
 import { toggleTodo } from "../actions";
 import TodoList from "../components/TodoList";
 
-const getVisibleTodos = (
-  todos,
-  filter
-) => {
+const getVisibleTodos = (todos, filter) => {
   switch (filter) {
     case "SHOW_ALL":
       return todos;
@@ -307,9 +290,7 @@ const mapStateToProps = (
   };
 };
 
-const mapDispatchToProps = (
-  dispatch
-) => {
+const mapDispatchToProps = (dispatch) => {
   return {
     onTodoClick: (id) => {
       dispatch(toggleTodo(id));
--- a/docs/basics/UsageWithReact.md
+++ b/docs/basics/UsageWithReact.md
@@ -92,7 +92,7 @@ Let’s write the components! We begin with the presentational components, so we
 
 ### Presentational Components
 
-These are all normal React components, so we'll not stop and examine them in detail. Here they are:
+These are all normal React components, so we'll not stop and examine them in detail. We write functional stateless components unless we need to use either React state or the React life-cycle functions.
 
 #### `components/Todo.js`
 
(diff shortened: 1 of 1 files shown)


docs/basics/UsageWithReact.md
```diff
@@ -52,18 +52,16 @@ It is advisable that only top-level components of your app (such as route handle
     </tbody>
 </table>
 
-In this todo app, we will only have a single container component at the top of our view hierarchy. In more complex apps, you might have several of them. While you may nest container components, we suggest that you pass props down whenever possible.
+In this todo app, we will have three container components that pass props to presentational components. We suggest that you do not nest container components and that you pass props down whenever possible.
```
gaearon: I'd say "We suggest that you do not nest container components beyond necessity. Start by having a few top-level container components pass the props down to the presentational component tree. When you notice that components in the middle of that tree pass too many props down without actually using them, it is a good time to introduce some containers."

urbanvikingr: Thanks. I have updated PR.

docs/basics/UsageWithReact.md
```diff
@@ -74,196 +72,281 @@ I see the following components (and their props) emerge from this brief:
...
   - `onFilterChange(nextFilter: string)`: Callback to invoke when user chooses a different filter.
+* **`Link`** is a link with a callback.
+  - `onClick()` is a callback to invoke when link is clicked.
 
 These are all presentational components. They don’t know *where* the data comes from, or *how* to change it. They only render what’s given to them.
 
 If you migrate from Redux to something else, you’ll be able to keep all these components exactly the same. They have no dependency on Redux.
 
-Let’s write them! We don’t need to think about binding to Redux yet. You can just give them fake data while you experiment until they render correctly.
+We also have some container components that connects to Redux. The container components are responsible for getting and transforming state. Remember that changes result in a new state as state is immutable.
```
gaearon: "that connect"

docs/basics/UsageWithReact.md
```diff
@@ -74,196 +72,281 @@ I see the following components (and their props) emerge from this brief:
...
 
-## Presentational Components
+* **`AddTodo`** is an input field with a button.
+  - `onAddClick(text: string)` is a callback to invoke when a button is pressed.
 
-These are all normal React components, so we won’t stop to examine them in detail. Here they go:
+* **`FilterLink`** gets active link from Redux and pass it as props to the presentational component, Link.
```
gaearon: I think "gets active link from Redux" sound mysterious. It's better to say "gets the current visibility filter and passes it as a prop". Link in this case should be in backticks: `Link`, or it's not formatted otherwise.

docs/basics/UsageWithReact.md
gaearon: I think "getting and transforming state" is a bit confusing. Changes happen in reducer, so I'm not sure why the next sentence is relevant here.

Maybe something like "Container components calculate the props to pass to the presentational components they wrap based on the current state of the Redux store." could work better instead.

docs/basics/UsageWithReact.md
```diff
@@ -74,196 +72,281 @@ I see the following components (and their props) emerge from this brief:
...
 
-#### `components/AddTodo.js`
+* **`VisibleTodoList`** gets todos from Redux, filter the todos and pass them as props to the presentational component, TodoList.
```
gaearon: "filters", "passes"

docs/basics/UsageWithReact.md
```diff
@@ -74,196 +72,281 @@ I see the following components (and their props) emerge from this brief:
...
-    return (
-      <div>
-        <input type='text' ref='input' />
-        <button onClick={e => this.handleClick(e)}>
-          Add
-        </button>
-      </div>
-    )
-  }
+Let’s write the components! We begin with the presentational components, so we don’t need to think about binding to Redux yet.
```
gaearon: Overall it feels like this section lost its focus. It may help to add two subheaders: "Presentational Components" and "Container Components" (inside "Designing Component Hierarchy"). This would better separate these paragraphs and provide some visual rhythm.

docs/basics/UsageWithReact.md
```diff
@@ -74,196 +72,281 @@ I see the following components (and their props) emerge from this brief:
...
-    node.value = ''
-  }
-}
+## Presentational Components
 
-AddTodo.propTypes = {
-  onAddClick: PropTypes.func.isRequired
-}
-```
+These are all normal React components, so we'll not stop and examine them in detail. Here they are:
```
gaearon: I'd actually clarify here that we are going to use functions for presentational components because it's the simplest way of declaring components in React. I'd clarify that whenever we need them to be stateful, or react to lifecycle hooks, we can use classes instead.

docs/basics/UsageWithReact.md
```diff
@@ -74,196 +72,281 @@ I see the following components (and their props) emerge from this brief:
...
+
+const Todo = ({
+  onClick,
+  completed,
+  text
+}) => (
+  <li
+    onClick={onClick}
+    style={{
+      textDecoration:
```
gaearon: Let's fit this on the same line, there is more space here than in JSBin :-)

docs/basics/UsageWithReact.md
```diff
@@ -74,196 +72,281 @@ I see the following components (and their props) emerge from this brief:
...
+      {children}
+    </a>
+  );
+};
+
+export default Link
 ```
 
-#### `components/Footer.js`
+We'll now write the container components.
```
gaearon: Please use the correct apostrophe: ’ (you can copy paste it).
Not '.

docs/basics/UsageWithReact.md
```diff
@@ -74,196 +72,281 @@ I see the following components (and their props) emerge from this brief:
...
+import { setVisibilityFilter } from "../actions";
+import Link from "../components/Link";
+
+const mapStateToProps = (
+  state,
+  ownProps
+) => {
+  return {
+    active:
+      ownProps.filter ===
```
gaearon: Please align these on the same line, as there is plenty of space

docs/basics/UsageWithReact.md
```diff
@@ -74,196 +72,281 @@ I see the following components (and their props) emerge from this brief:
...
+      state.visibilityFilter
+  };
+};
+
+const mapDispatchToProps = (
+  dispatch,
+  ownProps
```
gaearon: Here, too: anything that fits in 80 chars should be on the same line.

docs/basics/UsageWithReact.md
```diff
@@ -74,196 +72,281 @@ I see the following components (and their props) emerge from this brief:
...
+const VisibleTodoList = connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(TodoList);
+
+export default VisibleTodoList
 ```
 
-That’s it! We can verify that they work correctly by writing a dummy `App` to render them:
+Then we'll write the `TodoApp` component that assembles the app.
```
gaearon: Apostrophe: ’
Rather than "assembles" I'd say "renders these components together".

docs/basics/UsageWithReact.md
```diff
@@ -271,111 +354,37 @@ By itself, it’s not very interesting. Let’s connect it to Redux!
...
+  <Provider store={todoStore}>
+    <TodoApp />
   </Provider>,
-  rootElement
-)
+  document.getElementById("app")
+);
 ```
 
+That’s it! Our todo app now functions correctly.
```
gaearon: But we said "we need to make two changes". The second change follows after "that's it" and this is confusing because the text is old. Should it be removed? I'd probably want to move and edit it before defining container components because we never actually get to explain `mapStateToProps` and `connect()`. With this PR, we just jump into defining them without preparing the reader, which is a problem.

gaearon:
Thanks for your work on this! I left some comments. Also, `todos-with-undo` example needs updating, as well as the Implementing Undo History and Computing Derived Data recipes, as they contain fragments of the same code.

urbanvikingr:
@gaearon Thanks, I'll update your comments this weekend. I also have the source code to match. Would you have time to review https://github.com/urbanvikingr/todo. Would you it to be transferred to this repo?

gaearon:
Thank you very much. I continued your work in #1285.

PR #1197, opened 2016-01-04
Changing Stateless components to default to function syntax instead of classes syntex

ajaybeniwal:
I have changed some examples folder to default stateless components to more cleaner new syntax introduced in react 0.14 , If you like you can merge the changes.

### DIFF
--- a/examples/async/components/Picker.js
+++ b/examples/async/components/Picker.js
@@ -1,7 +1,7 @@
-import React, { Component, PropTypes } from 'react'
+import React, { PropTypes } from 'react'
 /* Changing the syntax for Stateless function in more functional syntax */
-const Picker = ({value, onChange, options}) => {
-    return (
+const Picker = ({ value, onChange, options }) => {
+  return (
       <span>
         <h1>{value}</h1>
         <select onChange={e => onChange(e.target.value)}
@@ -23,5 +23,4 @@ Picker.propTypes = {
   value: PropTypes.string.isRequired,
   onChange: PropTypes.func.isRequired
 }
-
-export default Picker;
+export default Picker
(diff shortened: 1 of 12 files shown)


timdorr:
Hmm, I'm kind of split on this. On the one hand, SFC's are cool and I'd love to see them used more. On the other hand, this raises the lowest common denominator (since they're so new and not as well-known as ES6 components) and are a tad slower (but that would be pre-optimization on our part).

I dunno, @gaearon what are you thoughts on this?

gaearon:
I'm up for changing all examples to functional components where possible.

However I don't see why we need `react-stateless-wrapper`  I would like to avoid this dependency.

Also the comment is unnecessary, and there are some spacing inconsistencies.

ajaybeniwal:
HI @gaearon I added react-stateless-wrapper to make test cases pass because of the below mentioned issue 
https://github.com/facebook/react/issues/4972

timdorr:
I think I might take that as a ding against them right now. They should be testable without throwing in too many libraries.

SimenB:
The wrapper is 12 lines of code though (https://github.com/ngerritsen/react-stateless-wrapper/blob/master/react-stateless-wrapper.js), and stateless components are a recommended pattern from FB.

> In an ideal world, most of your components would be stateless functions because these stateless components can follow a faster code path within the React core. This is the recommended pattern, when possible.

sompylasar:
The react-stateless-wrapper misses this feature of React:
https://facebook.github.io/react/docs/context.html#referencing-context-in-stateless-functional-components

urbanvikingr:
Heads up! I've a complete repo with the todo example https://github.com/urbanvikingr/todo which I'm currently updating the docs with together @gaearon .
I'll transfer the source code to this repo soon.

timdorr:
Based on the work by the Dan's over in #1285, I'm going to close this in favor of the community effort going on over there. Feel free to join in!

PR #1198, opened 2016-01-04
Prepare for Internationalization

thejameskyle:
This is as far as I got with setting up internationalization via Crowdin in this repo. Once it's fully setup, we'll be able to crowdsource the translation of the Redux documentation.

The only thing that's being difficult here is what GitBook does for multiple languages. It would change the homepage to [this](https://cloud.githubusercontent.com/assets/952783/12086154/3405cd42-b27c-11e5-911c-e23ecf650f60.png) (which is bad for SEO, and a poor experience for switching between languages).


thejameskyle:
Tracking over in https://github.com/GitbookIO/gitbook/issues/1080

gaearon:
If there’s a movement on Gitbook side of things I will gladly reopen. For now I’m closing because it doesn’t appear actionable for us. Thanks for the effort though!

PR #1206, opened 2016-01-06
Add 'npm run build' to examples/universal

gabrielflorit:
This creates a minified `bundle.js`.


timdorr:
The universal example is to show how one might set up a development environment for a universally rendered app. However, I don't think we should be making the jump to building for production and implying any best practices there. That's really more of the job for a boilerplate tool/repo or something more webpack-specific.

We could use some documentation on getting to production, though. Would you be willing to write up something for that as a PR?

PR #1211, opened 2016-01-07
Hyperlinks and Time Travel

grahammendick:
Thank you for Redux. It’s a beautiful way to handle web state. The one missing piece is how Hyperlinks fit into the picture. This is why you had trouble deciding where to hold the filter parameter in the todomvc example. In this project you added it to the React component state but in your egghead tutorial you added it to the Redux store. In this PR, I show that neither location is correct. The filter is part of the URL and so should live with the router.

This PR adds routing to the todomvc example to show how Redux and Hyperlinks should fit together. I’ve used my [Navigation router](http://grahammendick.github.io/navigation/). Like Redux, the Navigation router can be used with any JavaScript application, with a separate npm package providing the React-specific Hyperlink components. You can see that the Navigation router has simplified the code quite a bit. Most importantly, the filter isn’t stored in React state anymore. Instead, it’s held in the Navigation router and passed in as props whenever the URL changes.

The filter isn’t part of the Redux store, it’s part of the Navigation router. To capture the state of a web page, you need the URL together with the store: URL + store = web state. The Navigation router owns the URL and Redux owns the store. So, there are two different types of time travelling. You can either time travel the URL or time travel the store. You use the Redux developer tools to time travel the store and the browser back button to time travel the URL.


examples/todomvc/components/Footer.js
```diff
@@ -22,14 +23,12 @@ class Footer extends Component {
...
 
     return (
-      <a className={classnames({ selected: filter === selectedFilter })}
-         style={{ cursor: 'pointer' }}
-         onClick={() => onShow(filter)}>
+      <RefreshLink toData={{ filter: filter }}
+                   activeCssClass='selected'>
         {title}
-      </a>
+      </RefreshLink>
```
grahammendick: The Hyperlink component provided by the Navigation React plugin. It builds the href and sets the active css class.

examples/todomvc/constants/TodoFilters.js
```diff
@@ -1,3 +1,3 @@
-export const SHOW_ALL = 'show_all'
-export const SHOW_COMPLETED = 'show_completed'
-export const SHOW_ACTIVE = 'show_active'
+export const SHOW_ALL = 'all'
+export const SHOW_COMPLETED = 'completed'
+export const SHOW_ACTIVE = 'active'
```
grahammendick: These strings appear in the URL as the value of filter route parameter

examples/todomvc/index.js
```diff
@@ -5,12 +5,26 @@ import { Provider } from 'react-redux'
 import App from './containers/App'
 import configureStore from './store/configureStore'
 import 'todomvc-app-css/index.css'
+import { start, StateInfoConfig } from 'navigation'
+
+StateInfoConfig.build([
+    { key: 'todomvc', initial: 'app', states: [
+        { key: 'app', route: '{filter?}', defaults: { filter: 'all' }, trackCrumbTrail: false }
+    ]}   
+]);
```
grahammendick: Builds the route configuration. There’s one route with an optional filter parameter that has a default value of 'all'

examples/todomvc/index.js
```diff
@@ -5,12 +5,26 @@ import { Provider } from 'react-redux'
...
-)
+var todomvc = StateInfoConfig.dialogs.todomvc;
+todomvc.states.app.navigated =  (data) => {
+    render(
+        <Provider store={store}>
+            <App filter={data.filter} />
+        </Provider>,
+        document.getElementById('root')
+    )
+}
```
grahammendick: Any time the URL changes the Navigation router calls this navigated function which re-renders the App with the new filter

examples/todomvc/routes.js
```diff
@@ -0,0 +1,10 @@
+import { StateInfoConfig } from 'navigation'
+
+function configureStateInfo() {
+  StateInfoConfig.build([
+    { key: 'todomvc', initial: 'app', states: [
+      { key: 'app', route: '{filter?}', defaults: { filter: 'all' }, trackCrumbTrail: false }
+    ] }   
+  ])
+}
+export default configureStateInfo
```
grahammendick: Builds the route configuration. There’s one route with an optional filter parameter that has a default value of 'all'

examples/todomvc/index.js
```diff
@@ -5,12 +5,22 @@ import { Provider } from 'react-redux'
...
-)
+var todomvc = StateInfoConfig.dialogs.todomvc
+todomvc.states.app.navigated =  (data) => {
+  render(
+    <Provider store={store}>
+      <App filter={data.filter} />
+    </Provider>,
+    document.getElementById('root')
+  )
+}
```
grahammendick: Any time the URL changes the Navigation router calls this navigated function which re-renders the App with the new filter

examples/todomvc/routes.js
```diff
@@ -0,0 +1,11 @@
+import { StateInfoConfig } from 'navigation'
+import { SHOW_ALL } from './constants/TodoFilters'
+
+function configureStateInfo() {
+  StateInfoConfig.build([
+    { key: 'todomvc', initial: 'app', states: [
+      { key: 'app', route: '{filter?}', defaults: { filter: SHOW_ALL }, trackCrumbTrail: false }
+    ] }   
+  ])
+}
```
grahammendick: Builds the route configuration. There’s one route with an optional filter parameter that has a default value of 'all'

timdorr:
We already have a routing example with `redux-router` in the `real-world` example. (I'd actually like to get that switched out to `redux-simple-router`, but that's another issue...). So, I'll close this one for that reason.

Nonetheless, this seems better suited as a separate library, rather than sneaking it into an example here and potentially having it overlooked. Have you worked up a binding library with redux for your Navigation project? That would be more beneficial to users, I would think.

Issue #1207, opened 2016-01-07
"Usage with React" has a rendering inefficiency


camjackson:
In the docs, this example is given for usage with react: 

```js
class App extends Component {
  render() {
    // Injected by connect() call:
    const { dispatch, visibleTodos, visibilityFilter } = this.props
    return (
      <div>
        <AddTodo
          onAddClick={text =>       //This line
            dispatch(addTodo(text)) //and this line
          } />
        <TodoList
          todos={visibleTodos}
          onTodoClick={index =>
            dispatch(completeTodo(index))
          } />
        <Footer
          filter={visibilityFilter}
          onFilterChange={nextFilter =>
            dispatch(setVisibilityFilter(nextFilter))
          } />
      </div>
    )
  }
}
```

There's a small problem on the lines I've commented above. Every render of the `App` component will create a new anonymous function to be given for the `onAddClick` prop. This means that any state change for the whole application (because that state is passed to `App` as props) will cause all three sub-components  (`AddTodo`, `TodoList`, and `Footer`) to be re-rendered.

What you want to happen is that a subcomponent only re-renders if a relevant prop has really changed, e.g. `TodoList` re-renders if `visibleTodos` has changed. 

Here is one solution to this problem, I'm not sure if it's the right one:

```js
class App extends Component {
  constructor(props) {
    // Injected by connect() call:
    const { dispatch } = props;
    this.onAddClick = text => dispatch(addTodo(text))
    this.onTodoClick = index => dispatch(completeTodo(index))
    this.onFilterChange = nextFilter => dispatch(setVisibilityFilter(nextFilter))
  }
  
  render() {
    // Injected by connect() call:
    const { visibleTodos, visibilityFilter } = this.props
    return (
      <div>
        <AddTodo onAddClick={this.onAddClick} />
        <TodoList
          todos={visibleTodos}
          onTodoClick={this.onTodoClick} 
        />
        <Footer
          filter={visibilityFilter}
          onFilterChange={this.onFilterChange}
        />
      </div>
    )
  }
}
```

I'm not sure whether the example in the docs needs to be changed (or whether the above would be the correct change). Obviously performance is not a big concern for such a simple todo app. However, I think at the very least the docs should mention the problem, otherwise people (like me :smiley:) have to discover the hard way that creating a fresh callback on every render call will cause child components to re-render unnecessarily.

winstonewert:
Since the subcomponents don't implement a shouldComponentUpdate, aren't they going to re-render regardless?

[...]

PR #1215, opened 2016-01-08
Add note on callback render efficiency to react usage docs

camjackson:
Refer to #1207 for more context.

### DIFF
--- a/docs/basics/UsageWithReact.md
+++ b/docs/basics/UsageWithReact.md
@@ -377,6 +377,9 @@ export default connect(select)(App)
 
 That’s it! The tiny todo app now functions correctly.
 
+>##### Note on callbacks and render efficiency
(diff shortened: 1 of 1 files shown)


gaearon:
It would be more correct to say:

1) There are two problems in this example: function allocations on every render, and the fact that props are shallowly unequal because of that.
2) Shallow inequality of props is not a problem by itself, but it makes optimizations harder because you can no longer use `PureRenderMixin` or `shallowCompare` to optimize components you know to be pure.
3) If you wrap component into `connect()` from React Redux and bind the action creators using its second argument, it will make sure to only bind them when necessary, and implement a strict `shouldComponentUpdate()` on the wrapper component.

Right now it's a bit unclear in this PR how these several factors come into play so I would appreciate a clarification.

timdorr:
Closing in favor of #1285

PR #1224, opened 2016-01-10
Move exported functions to the top-level src folder

### LINKED ISSUE #1223: Provide more minimal export
This isn't a huge deal because redux is pretty small, but i'm working on [vdux](https://github.com/ashaffer/vdux), which currently wraps redux and only uses a small portion of the bundled API. At the moment, redux is my single largest dependency, and i'm only really using `applyMiddleware` and `createStore`.

Requiring only those files, like this:

```javascript
import applyMiddleware from 'redux/lib/utils/applyMiddleware'
import createStore from 'redux/lib/createStore'
```

Reduces the size of the redux dep to about 1/4 of what it is normally. I can just do this for now, but it feels sort of unsupported and like something that may break in the future, unannounced. What i'd like is to be able to have an official thing like:

```javascript
import {applyMiddleware, createStore} from 'redux/core'
```

So that people like me who aren't using the other helper functions can avoid the bloat.

ashaffer:
Here you go. Closes #1223

Also, have you considered depending on external modules for the things left in utils (and possibly compose)? I understand the intuitive appeal of 'zero dependencies', but I think that only really means you get unnecessary code duplication. To plug my own thing as a solution to this problem, i'd be happy to implement all of the utilities in [micro-js](https://github.com/micro-js) and also add you to the organization if you'd like.

EDIT: I made a separate PR for the modularization thing. #1225


gaearon:
Can you please reorganize the `test` folder in a similar fashion?

ashaffer:
Done.

gaearon:
Thanks.

gaearon:
Out in 3.0.6, thanks!

PR #1225, opened 2016-01-10
Don't include standard utilities in redux

ashaffer:
This increases the minified bundle by only ~100 bytes, but means that people can re-use these same utilities in their application and save some space, and squeeze a bit of extra peformance out of JS engines by having fewer implementations of the same thing.

I'd also be happy to add you @gaearon to the organization/npm account if you want control over the things redux depends on.

Whether or not you use my specific deps, it seems like a good idea to externalize the common utilities, since almost every project needs a `pick/compose/map-values/is-object` anyway.


gaearon:
Hey, thanks.
We plan to use Lodash 4 for this: #611.

>but means that people can re-use these same utilities in their application and save some space, and squeeze a bit of extra peformance out of JS engines by having fewer implementations of the same thing.

It's very unrealistic to expect people to use these same libraries since they are very new. Relying on something already popular like Lodash seems like a better idea. If we want to shift support burden (which isn't a lot to be honest for these utilities), we should switch it to something more popular than Redux—not less.

Thanks for PR though!

ashaffer:
Ya, that makes sense. FYI it looks like lodash 4 [just got released](https://github.com/lodash/lodash).

PR #1230, opened 2016-01-13
docs(ExampleTodoList.md): Rename todo(s) to todo(s)Reducer

dmitriz:
Because of too many todos in town.

### DIFF
--- a/docs/basics/ExampleTodoList.md
+++ b/docs/basics/ExampleTodoList.md
@@ -89,7 +89,7 @@ function visibilityFilter(state = SHOW_ALL, action) {
   }
 }
 
-function todo(state, action) {
+function todoReducer(state, action) {
   switch (action.type) {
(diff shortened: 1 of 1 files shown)


gaearon:
This example is incorrect. Now it uses `todosReducer` key in state which is probably not the desired effect. You probably want to change `combineReducers` call to accept `todos: todosReducer` instead of using the same name shortcut notation.

That said we use this convention (reducer named after the state field) all over the docs. We find it good enough, and changing it in a single particular case doesn't seem to help much. Therefore I'm closing.

Thanks for contributing!

dmitriz:
Oh yes, forgot that -- you are totally right!

I felt, it would really help a newcomer to have names underlying the nature of the entities they represent.

I've thought that was the only place in that example where reducers are defined, but you are certainly right in that names should be consistent within the guide.

PR #1231, opened 2016-01-14
Replaced the word "reddit" with "subreddit" where subreddit is the correct term.

yarabarla:


CyrusRoshan:
+1

timdorr:
Haha, I noticed this a while ago, but didn't want go to through all the effort :smile: 

Can you squash that merge commit? We should be good to go then.

PR #1253, opened 2016-01-19
Counter example: replace mod with bitwise

avindra:
Also removes confusing early return for more clarity.

### DIFF
--- a/examples/counter/actions/counter.js
+++ b/examples/counter/actions/counter.js
@@ -17,11 +17,10 @@ export function incrementIfOdd() {
   return (dispatch, getState) => {
     const { counter } = getState()
 
-    if (counter % 2 === 0) {
-      return
+    if (counter & 1) {
+      dispatch(increment())
     }
 
-    dispatch(increment())
   }
 }
 


gaearon:
Thanks for the PR. I'm inclined to not take this. I don't know how bitwise operators work (count me stupid!) and lots of JavaScript users like me will scratch their heads at this.

I don't have strong opinion on whether to exit early. However we want to make it clear it's possible to return undefined from a thunk action creator (unlike a normal action creator that has to return an object). So I'd keep the early return to make this stand out.

avindra:
Np, thanks for the explanation. I was mostly looking for a way to keep up my contribution streak :-)

Bitwise operators are a little confusing, but I can explain how ```&``` works. Bitwise operators are low level operations that work in binary. AND just requires that a ```1``` bit is present in both numbers.

For example, take 1 and 2:

````
0 1 
1 0
-----
0 0
````

Here's 5 and 3:

```
1 0 1
0 1 1
------
0 0 1

```

This makes checking for odd numbers quite efficient using ```& 1```, because any odd number will have a 1 as the rightmost bit.

Even number (2): 

```
1 0
0 1
----
0 0
````

Odd number (3):

```
1 1
0 1
----
0 1
```` 

In most modern compilers, this optimization is done automatically for ```% 2```, but I'm not sure about JavaScript. I don't expect you to merge this, but I just thought I'd share how & works.

Btw thanks for redux, I'm going through your tutorial on egghead and excited to implement it.

PR #1257, opened 2016-01-21
Examples: todomvc: Removed precondition to clear completed todo's

DerekCuevas:
Perhaps this is an optimization to prevent against an unneeded re-render. However this may be adding more confusion to the example than benefit. Am I wrong by removing this?


gaearon:
Good point, I'm not sure why the check was there to begin with.
Thanks!

PR #1259, opened 2016-01-21
Update bindActionCreators.md

designeng:
will provide you with the [`dispatch`] -> will provide you store with the [`dispatch`]


gaearon:
The wording is actually correct. React Redux won't give you the store instance—just the `dispatch` function.

designeng:
@gaearon, thank you, there was some kind of misunderstanding for me.

PR #1263, opened 2016-01-22
Update ServerRendering.md

keithyong:
On the Server Rendering Redux recipe, the `<div id="root">` and `document.getElementById('app')` do not match.

Screenshots:

screen shot 2016-01-22 at 10 56 34 am
screen shot 2016-01-22 at 10 56 48 am

**Fix**
Use `app` for both id's.


dukex:
Nice! LGTM

PR #1269, opened 2016-01-25
Add tree view example

gaearon:
This adds a performant tree view example by a popular request.
Answers http://stackoverflow.com/questions/34981924/performance-issues-with-a-tree-structure-and-shouldcomponentupdate-in-react-re.

### DIFF
--- a/docs/introduction/Examples.md
+++ b/docs/introduction/Examples.md
@@ -153,6 +153,29 @@ It covers:
 * Using only [React Redux](https://github.com/rackt/react-redux) to bind action creators 
 * Conditional middleware (logging example)
 
+## Tree View
+
+Run the [Tree View](https://github.com/rackt/redux/tree/master/examples/tree-view) example:
+
+```
+git clone https://github.com/rackt/redux.git
+
+cd redux/examples/tree-view
+npm install
+npm start
+
+open http://localhost:3000/
+```
+
+This is an example of performant rendering.
+
+It covers:
+
+* Normalized state
+* Reducer composition
+* State representing a tree view
+* Granual re-rendering of a large subtree
+
 ## More Examples
 
 You can find more examples in [Awesome Redux](https://github.com/xgrommx/awesome-redux).
--- a/examples/tree-view/actions/index.js
+++ b/examples/tree-view/actions/index.js
@@ -1,0 +1,26 @@
+export const INCREMENT = 'INCREMENT'
+export const CREATE_NODE = 'CREATE_NODE'
+export const ADD_CHILD = 'ADD_CHILD'
+
+export function increment(nodeId) {
+  return {
+    type: INCREMENT,
+    nodeId
+  }
+}
+
+let nextId = 0
+export function createNode() {
+  return {
+    type: CREATE_NODE,
+    nodeId: `new_${nextId++}`
+  }
+}
+
+export function addChild(nodeId, childId) {
+  return {
+    type: ADD_CHILD,
+    nodeId,
+    childId
+  }
+}
--- a/examples/tree-view/containers/Node.js
+++ b/examples/tree-view/containers/Node.js
@@ -1,0 +1,61 @@
+import React from 'react'
+import { Component } from 'react'
+import { connect } from 'react-redux'
+import * as actions from '../actions'
+
+class Node extends Component {
+  constructor(props) {
+    super(props)
+    this.handleIncrementClick = this.handleIncrementClick.bind(this)
+    this.handleAddChildClick = this.handleAddChildClick.bind(this)
+  }
+
+  handleIncrementClick() {
+    const { increment, id } = this.props
+    increment(id)
+  }
+
+  handleAddChildClick(e) {
+    e.preventDefault()
+
+    const { addChild, createNode, id } = this.props
+    const childId = createNode().nodeId
+    addChild(id, childId)
+  }
+
+  renderChild(childId) {
+    return (
+      <li key={childId}>
+        <ConnectedNode id={childId} />
+      </li>
+    )
+  }
+
+  render() {
+    const { counter, childIds } = this.props
+    return (
+      <div>
+        Counter: {counter}
+        {' '}
+        <button onClick={this.handleIncrementClick}>
+          +
+        </button>
+        <ul>
+          {childIds.map(this.renderChild)}
+          <li key='add'>
+            <a href='#' onClick={this.handleAddChildClick}>
+              Add child
+            </a>
+          </li>
+        </ul>
+      </div>
+    )
+  }
+}
+
+function mapStateToProps(state, ownProps) {
+  return state[ownProps.id]
+}
+
+const ConnectedNode = connect(mapStateToProps, actions)(Node)
+export default ConnectedNode
--- a/README.md
+++ b/README.md
@@ -160,6 +160,7 @@ For PDF, ePub, and MOBI exports for offline reading, and instructions on how to 
 * [Universal](http://rackt.github.io/redux/docs/introduction/Examples.html#universal) ([source](https://github.com/rackt/redux/tree/master/examples/universal))
 * [Real World](http://rackt.github.io/redux/docs/introduction/Examples.html#real-world) ([source](https://github.com/rackt/redux/tree/master/examples/real-world))
 * [Shopping Cart](http://rackt.github.io/redux/docs/introduction/Examples.html#shopping-cart) ([source](https://github.com/rackt/redux/tree/master/examples/shopping-cart))
+* [Tree View](http://rackt.github.io/redux/docs/introduction/Examples.html#tree-view) ([source](https://github.com/rackt/redux/tree/master/examples/tree-view))
 
 If you’re new to the NPM ecosystem and have troubles getting a project up and running, or aren’t sure where to paste the gist above, check out [simplest-redux-example](https://github.com/jackielii/simplest-redux-example) that uses Redux together with React and Browserify.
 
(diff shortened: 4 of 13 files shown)


erikras:
This is fantastic. :+1:

CrocoDillon:
Agreed, this example covers so much :smile:

examples/tree-view/containers/Node.js
```diff
@@ -0,0 +1,61 @@
...
+class Node extends Component {
+  constructor(props) {
+    super(props)
+    this.handleIncrementClick = this.handleIncrementClick.bind(this)
+    this.handleAddChildClick = this.handleAddChildClick.bind(this)
+  }
+
+  handleIncrementClick() {
+    const { increment, id } = this.props
+    increment(id)
```
stephenwf: No dispatch needed?

erikras: The `connect()` call on line 60 binds the actions to `dispatch`.

stephenwf: Ah, that makes sense! Not seen that being used before.

docs/introduction/Examples.md
```diff
@@ -153,6 +153,29 @@ It covers:
...
+```
+
+This is an example of performant rendering.
+
+It covers:
+
+* Normalized state
+* Reducer composition
+* State representing a tree view
+* Granual re-rendering of a large subtree
```
glenjamin: nit: "Granular" :)

erikras: Probably not "annual". If you want UI that slow, use Angular! :laughing:

gaearon: Fix is welcome. That's just 5am me

examples/tree-view/containers/Node.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react'
+import { Component } from 'react'
+import { connect } from 'react-redux'
+import * as actions from '../actions'
```
rik: Since this is an exemple, I don't think this syntax is helping readability. Later you can't determine where the actions are coming from.

examples/tree-view/containers/Node.js
```diff
@@ -0,0 +1,61 @@
...
+    this.handleIncrementClick = this.handleIncrementClick.bind(this)
+    this.handleAddChildClick = this.handleAddChildClick.bind(this)
+  }
+
+  handleIncrementClick() {
+    const { increment, id } = this.props
+    increment(id)
+  }
+
+  handleAddChildClick(e) {
```
sompylasar: Shouldn't this be `handleAddChildClick = (e) => {` to bind `this` properly, or am I missing something?

gaearon: We bind these functions in the constructor. It's more efficient (which is one of the points of this example) than allocating a function on every render.

sompylasar: Ah, missed that, sorry. Property initializers should do the same, but I noticed you prefer not to use them.

falzhobel:
Nice example, I'm wondering tho how an action affecting the entire state would be handled? For example removing a node, which in turn affects the entire tree?

I need to get my head around this for optimization purpose, I did encounter performance issues with a somewhat big tree of data.

Say for example I have a `<Selected />` component that stores added items from a list, when added it shows up in the Selection component, but in the list shows as "added", and vice-versa. How would that need to be handled in the tree-view example?

Really appreciate you taking the time doing an example for this by the way, thanks!

tamagokun:
A great enhancement to this example would be removing node and moving nodes.

gaearon:
I'm happy to accept a PR for removing node. Would you like to work on this?

gaearon:
>Nice example, I'm wondering tho how an action affecting the entire state would be handled? For example removing a node, which in turn affects the entire tree?

Why does removing affect the entire tree? Seems like it would only affect the removed node's parent (and, of course, its children, but they would just get unmounted).

examples/tree-view/containers/Node.js
gaearon: We settled on only using ES6 features in examples a while ago. Mostly to stop alienating newcomers who are confused by the ecosystem already and assume Redux requires knowledge of ES2016+ proposals.

sompylasar: Yes, I remember that. Thanks for clarification!

erikras:
There are three possible ways, I see, for removing a node, assuming that you want to remove all of its children from the store...

1. Actually keep the data in a tree structure in the store, so removing a node object automatically removes all its children, which it contained.
2. Maintain the flat structure and give them string ids like `'1.5.23.1'` so that the reducer can just remove all ids that start with the removed node's id.
3. Maintain the flat structure and depth-first iterate on the removal through all the child nodes.

I did `#1` in my `redux-form` library, which added much complication. I think I'd recommend `#3` for this example.

bramdevries:
Here's a method I used to remove nodes on my own project (uses Immutable.js) but using the same flat structure:

```js
const parentIds = state.get('tree').reduce((result, item) => {
    if (item.get('children').includes(action.id)) {
        result.push(item.get('id'));
    }
    return result;
}, []);

parentIds.forEach(id => {
    const path = ['tree', String(id), 'children'];
    path.push(state.getIn(path).findIndex(i => i === action.id));
    state = state.removeIn(path);
});

return state.removeIn(['tree', String(action.id)]);
```

Basically what happens is that we run through all nodes and get the id's of where the removed node is a child, then we loop over all these nodes and remove the reference. Finally we remove the actual node.

falzhobel:
Maybe my data structure is what's confusing me here. I looked back at the example and noticed the multiple `connect()`ed components instead of only one HOC. I always thoughts this was best practice, so basically what I've been doing is connect the entire state to the top-most container, and then pass along state and dispatch to the children as props:

```javascript

class FilesDownload extends Component {

  render() {
    return (
      <div className="files-download">
        <div className="downloads">
          <div className="downloads__selection">
            <Selection {...this.props} />
          </div>
          <div className="downloads__list">
            <Filters {...this.props} />
            <Results {...this.props} />
          </div>
        </div>
      </div>
    );
  }

}

function select(state) {
  return state;
}

// Wrap the component to inject dispatch and state into it
export default connect(select)(FilesDownload);
```

Here's an example Gif of the actual app, you can see the delay when I first click on the "add to selection" button:

redux

erikras:
Node Removal PR #1274.

falzhobel:
Ah yes thanks @erikras I will look at the PR for reference. Merci :smile:

gaearon:
>I always thoughts this was best practice, so basically what I've been doing is connect the entire state to the top-most container, and then pass along state and dispatch to the children as props:

Yeah, we no longer recommend single top container in favor of a more gradual approach. See the updated https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0#.49376my94.

falzhobel:
Yes thanks for the reminder, as I tweeted earlier (tweet which you "liked") I'm slowing getting the hang of this. It's really starting to make sense now with Presentational and Container Components.

Perhaps, with my [example above](https://github.com/rackt/redux/pull/1269#issuecomment-174652023), I should have a `connect()` container for each selectable items (CAD Drawing, Install Drawings etc.), what's your thoughts on this?

gaearon:
>Perhaps, with my example above, I should have a connect() container for each selectable items (CAD Drawing, Install Drawings etc.), what's your thoughts on this?

Yes, I think this can work fine.

falzhobel:
:+1: Merci beaucoup mon ami !

examples/tree-view/actions/index.js
```diff
@@ -0,0 +1,26 @@
...
+
+let nextId = 0
+export function createNode() {
+  return {
+    type: CREATE_NODE,
+    nodeId: `new_${nextId++}`
+  }
+}
+
+export function addChild(nodeId, childId) {
```
fhelwanger: Just wondering... couldn't this action creator receive just `nodeId` and generate `childId` internally?

This way, the component doesn't need to call `createNode` just to get the new id.

gaearon: This is what I tried at first but it made the reducer a little bit awkward. It's hard to say which approach is cleaner until the app grows large enough and more use cases are implemented so I went with this version that keeps the reducers simpler.

fhelwanger: I don't get it, I was thinking about something like this:

```js
export function createNode() {
  return {
    type: CREATE_NODE,
    nodeId: generateId()
  }
}

export function addChild(nodeId) {
  return {
    type: ADD_CHILD,
    nodeId,
    childId: generateId()
  }
}

let nextId = 0
function generateId() {
  return `new_${nextId++}`
}
```

How does it make the reducers more complex?

Sorry if I'm missing something :smile:

gaearon: Please try and submit a PR if it works. It's hard to discuss hypothetical code. ;-)

fhelwanger: Nice, I'll do it :)

fhelwanger: Oh, I got it now :sweat: 

When you call `const childId = createNode().nodeId` you're actually dispatching the action because of `mapDispatchToProps`.

When reading the code, I assumed it was being called only to get the new id.

You were right, it's hard to discuss hypothetical code :-)

Sorry for asking without testing first, and thank you so much for all the good work you're doing :smile:

gaearon: No problem :wink: If you do find a way to make this code simpler let us know!

Issue #1270, opened 2016-01-25
Error resolving lodash methods in the real-world example


pooyaj:
After running ```npm start``` and heading to http://localhost:3000/ (or just running ```webpack```), I am getting the following errors (in the real-world example):  

```
ERROR in ./reducers/index.js
Module not found: Error: Cannot resolve module 'lodash/object/merge' in /redux/examples/redux-contrib/redux/examples/real-world/reducers
 @ ./reducers/index.js 13:25-55

ERROR in ./containers/UserPage.js
Module not found: Error: Cannot resolve module 'lodash/array/zip' in /redux/examples/redux-contrib/redux/examples/real-world/containers
 @ ./containers/UserPage.js 35:22-49

ERROR in ./reducers/paginate.js
Module not found: Error: Cannot resolve module 'lodash/object/merge' in /redux/examples/redux-contrib/redux/examples/real-world/reducers
 @ ./reducers/paginate.js 8:25-55

ERROR in ./reducers/paginate.js
Module not found: Error: Cannot resolve module 'lodash/array/union' in /redux/examples/redux-contrib/redux/examples/real-world/reducers
 @ ./reducers/paginate.js 12:24-53
```

PR #1271, opened 2016-01-25
fixed the issue with importing lodash methods in the real-world example

pooyaj:
Fixed the issue reported in #1270


gaearon:
Thanks! A better fix would be to use `import zip from 'lodash/zip'` etc. Can you change it?

gaearon:
Thanks again. Weird I couldn't repro this.
¯\_(ツ)_/¯

pooyaj:
Thanks Dan for the quick response and merge :) cheers

PR #1273, opened 2016-01-25
Change next to createStore for better readability

mayankchd:


gaearon:
Fair enough, let's do it.

PR #1274, opened 2016-01-25
Added tree-view node removal, as requested on #1269

erikras:
Also put the `id` in the `title` prop for node identification.

### DIFF
--- a/examples/tree-view/actions/index.js
+++ b/examples/tree-view/actions/index.js
@@ -1,6 +1,7 @@
 export const INCREMENT = 'INCREMENT'
 export const CREATE_NODE = 'CREATE_NODE'
 export const ADD_CHILD = 'ADD_CHILD'
+export const REMOVE_NODE = 'REMOVE_NODE'
 
 export function increment(nodeId) {
   return {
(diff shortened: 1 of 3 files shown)


CrocoDillon:
I was also giving this a shot. However it looks like you're not really removing the (child)nodes, just remove the node id from it parent's childIds list... right?

erikras:
No, I'm removing the node, too. It's maybe a little counterintuitive to follow in the code, but that happens [here](https://github.com/rackt/redux/pull/1274/files#diff-c0354bd716170b207154b564fb9a209aR32]). By just falling through to return `accumulator`, the node for that key is removed.

erikras:
Oh! Yes, you're right.

gaearon:
I think it would be easier to pass `{ childId, nodeId }` to the action where `nodeId` is the ID of the parent. The `Node` component can pass `parentId` to its child `Node` components as a prop, and children will use that when the fire the action. Then we don't need to search for the parent.

CrocoDillon:
My take on this in #1275, however I don't like the heavy binding... plus I'm also not removing the node. Anyway bedtime for me so maybe I'll continue tomorrow. Unless you finish it first which is more likely :wink:

erikras:
@gaearon Does the `Node` component have its parent's id, though? I guess it could be made to.

gaearon:
#1275 is more like what I had in mind. Thanks!

gaearon:
>I guess it could be made to.

Yeah, if we make `Node` pass it.

PR #1276, opened 2016-01-25
Update real-world example with react-router-redux

Pierrickouw:
As discussed here https://github.com/rackt/redux/pull/1152, this is a small proposition to update the `real-world` example with the latest version of `react-router-redux`
Feedback welcome !

### DIFF
--- a/examples/real-world/package.json
+++ b/examples/real-world/package.json
@@ -23,10 +23,10 @@
     "react": "^0.14.6",
     "react-dom": "^0.14.6",
     "react-redux": "^4.0.6",
-    "react-router": "^1.0.3",
+    "react-router": "2.0.0-rc5",
+    "react-router-redux": "^2.1.0",
     "redux": "^3.0.6",
     "redux-logger": "^2.4.0",
-    "redux-router": "^1.0.0-beta3",
     "redux-thunk": "^1.0.3"
   },
   "devDependencies": {
--- a/examples/real-world/package.json
+++ b/examples/real-world/package.json
@@ -15,7 +15,6 @@
   },
   "homepage": "http://rackt.github.io/redux",
   "dependencies": {
-    "history": "^1.17.0",
     "humps": "^0.6.0",
     "isomorphic-fetch": "^2.1.1",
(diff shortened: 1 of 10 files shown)


examples/real-world/reducers/index.js
```diff
@@ -50,7 +50,9 @@ const rootReducer = combineReducers({
   entities,
   pagination,
   errorMessage,
-  router
+  routeReducer,
+  routing: routeReducer
```
SpencerCDixon: shouldn't there just be one reducer added here?

Pierrickouw: True ! Nice catch, thanks !

examples/real-world/package.json
```diff
@@ -23,10 +22,10 @@
     "react": "^0.14.6",
     "react-dom": "^0.14.6",
     "react-redux": "^4.0.6",
-    "react-router": "^1.0.3",
+    "react-router": "2.0.0-rc5",
```
timdorr: https://www.youtube.com/watch?v=7nqcL0mjMjw

examples/real-world/routes.js
```diff
@@ -1,14 +1,16 @@
...
-)
+export default function Routes() {
+  return (<Router history={browserHistory}>
+            <Route path="/" component={App}>
+              <Route path="/:login/:name"
+                     component={RepoPage} />
+              <Route path="/:login"
+                     component={UserPage} />
+            </Route>
+          </Router>)
```
timdorr: Small nitpick here: Would you might keeping the original indentation? The less width in your code, the better. It's normal to do something like this with JSX:
```js
return (
   <SomeComponent />
)
```

timdorr:
Just that small nitpick and this LGTM :+1:

examples/real-world/routes.js
Pierrickouw: Yes, sure :smile: 
Thanks for the advice

timdorr:
Awesome! Thanks for beating me to the punch :godmode:

gaearon:
Thank you!

PR #1278, opened 2016-01-25
counter example: write actual test on incrementIfOdd action

just-boris:
Now the test on `incrementIfOdd` doesn't test it actually. `expectedActions` can contain anything, but it will not be checked because `store.dispatch` is never called. So, it would be good to call some assertion to be sure that all actions had been matched.

Also, the `verifyNoPendingActions` helper can be useful in other tests to get more descriptive message rather than `timeout of 2000ms exceeded`

### DIFF
--- a/examples/counter/test/actions/counter.spec.js
+++ b/examples/counter/test/actions/counter.spec.js
@@ -31,6 +31,12 @@ function mockStore(getState, expectedActions, onLastAction) {
           onLastAction()
         }
         return action
+      },
+
+      verifyNoPendingActions() {
+        if(expectedActions.length > 0) {
+          throw new Error('There is undispatched actions:\n  ' + JSON.stringify(expectedActions))
+        }
       }
     }
   }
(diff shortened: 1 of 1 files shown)


gaearon:
We should just use https://github.com/arnaudbenard/redux-mock-store.
Does it handle this case?

just-boris:
That's great, that it exists as a separate project, but it has almost the same code. 
Also, they have an issue about it: https://github.com/arnaudbenard/redux-mock-store/issues/17

@gaearon what is your point about enhancing store API? Is it ok to have extra method `verifyNoPendingActions` for the mocked store? If it is, I can send a pull-request there, and then use it here instead of the current one.

gaearon:
I don't like that it's a manual operation. It would be best if we could use `beforeEach()` and `afterEach()` for that. This would involve changing the `redux-mock-store` API but IMO it would be worth it.

gaearon:
Another alternative is to always pass `done` to the mock store even for sync cases. Then the test that didn't dispatch will hang.

just-boris:
> It would be best if we could use beforeEach() and afterEach() for that. 

Of course. But it doesn't seem difficult if you have this method

```js
var store;
it('is a test', () => {
})

afterEach(() => {
})
```

just-boris:
> It would be best if we could use beforeEach() and afterEach() for that. 

Of course. But it doesn't seem difficult if you have this method

```js
var store;
it('is a test', () => {
})

afterEach(() => {
})
```

just-boris:
> Then the test that didn't dispatch will hang.

That is the thing that I want to avoid. If you have a big project with many developers, you can encounter the error 

> It would be best if we could use beforeEach() and afterEach() for that. 

Of course. But it doesn't seem difficult if you have this method

```js
var store;
it('is a test', () => {
     store = mockStore()
     // ... testing ...
})

afterEach(() => {
    store.verifyNoPendingActions()
})
```

just-boris:
In the current master branch, these tests are gone. Closing this as not actual.

PR #1281, opened 2016-01-26
split code in chunks

tinybug:


.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea
```
timdorr: Can you remove this? This is an IDE thing that you should have in your machine's global gitignore. It'll save you some keystrokes in the future, trust me :smile:

timdorr:
I don't think this is really helpful. We're not trying to teach people about webpack chunks or code splitting. 

However, I do like that the routes are converted to remove the Router component from that export. Hot reloading might trigger a Router hot reload, which it doesn't deal well with right now. You should be isolating the routes, not the router. 

Can you switch it to just that part of the changes and keep the routes in JSX (to avoid beginner confusion)? That would be awesome! Thanks!

tinybug:
ok, thanks for your advise. I have updated the codes as you said.

timdorr:
Thank you! Looks great!

gaearon:
To be fair I do think we want to have an example with code splitting. However I'd rather see it as a separate example. It should also show how to code split reducers a la https://gist.github.com/gaearon/0a2213881b5d53973514.

tinybug:
@gaearon It's great!

PR #1288, opened 2016-01-27
Simplify counter example

gaearon:
For many people, Counter example is the intro to Redux. We used to keep examples uniform but I no longer think this is a good idea. The examples should show flexibility of Redux and how different usage patterns afford different project scopes.

With this change, we:

* No longer use React Redux in favor of `subscribe()` API
* Remove action constants, action creators, and middleware
* Remove `combineReducers()` in favor of a single reducer
* Make it clear that you can do conditions and async from the component if you want to

This could introduce confusion at early stages of the project but there are plenty complex examples now (both inside and outside) that we can afford to show something simple first.


timdorr:
Lots of red! I like it! :smile:

What about taking it further and keeping it all in one file. Our simpler examples on react-router do this. It serves to show you can use the library with something as simple as a little widget you embed on a page. Redux is so minimalistic, that's entirely feasible. 

You could even do it with plain DOM functions at the most extreme end of the spectrum.

gaearon:
I want to add another `counter-vanilla` that is just `index.html`.

rossipedia:
@gaearon if you're a bit too busy, I wouldn't mind putting that together. I could probably use by single file TodoMVC sample as a good starting point.

gaearon:
@rossipedia Missed your comment, I just did https://github.com/rackt/redux/pull/1289

rossipedia:
no worries! Just happy it's done :)

Issue #1180, opened 2015-12-25
How should (un)subscribing in the middle of dispatch work?


gaearon:
Here's something that came up a few times: https://github.com/gaearon/react-dnd/issues/307#issuecomment-148528999

We let people subscribe and unsubscribe during a dispatch. This is not a problem per se.

The current behavior is that subscribe during dispatch only takes effect the next time. So if you subscribe from during a dispatch, you will only be called on next dispatch. **This doesn't seem problematic** because you can just read the initial state and rely on the subscriber being called the next time. It doesn't violate your expectations.

The current behavior is also that *unsubscribing* during dispatch only takes effect *after* the dispatch. So if I unsubscribe, but I'm in the middle of dispatch *and* the subscriber hasn't been called yet, it *will* be called. Even though I unsubscribed. **This seems problematic** because unsubscribing carries an expectation that I won't ever be called again unless I subscribe. However we violate this expectation.

One thing we can do is to treat unsubscriptions *while dispatching* in a special way and put those listeners in a special array. When dispatching we can make sure to skip those listeners. However this seems like a perf problem for a rare case.

We might also choose to not fix this, and document this behavior. Then it'll be up to consumers like React Redux to protect themselves from it. (FWIW React Redux already handles this correctly.)

slorber:
Incriminated line: `listeners.slice().forEach(listener => listener())` 

So on every dispatch you copy the whole listener array :) I understand the benefit of this to avoid weird index issues when modifying the array while iterating it. However it's still already a copy on each dispatch and has a performance cost (that probably does not matter much).

IMO the solution does not need to have any perf problem and could even give better performances.

Instead of using a forEach maybe it would be possible to modify the listeners array in place without having any issue during the iteration process? 

I don't know what's the best solution but a solution could be to replace the listener by undefined in the array on unsubscribe, and skip undefined listeners while iterating. At some point the array should be cleaned from undefined listeners (when?) and it may affect performances but you removed the array copy on every dispatch so it might be ok.


How react redux handles this btw?

Aweary:
What if instead of just calling `slice` in the `ubsubscribe` function, it checked whether `isDispatching` was true, and just registered the index of the listener in a new array instead.

```js
...
function unsubscribe() {
      if (!isSubscribed) {
        return
      }

      var index = listeners.indexOf(listener)

     if (isDispatching) {
        indexList.push(index)
        return
     }

      isSubscribed = false
      listeners.splice(index, 1)
    }

```

[...]

PR #1291, opened 2016-01-27
Add tests for the behavior of unsubscribing affecting next dispatch a…

slorber:
…nd not current dispatch listeners (decision of gaearon in #1180)

### DIFF
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -286,36 +286,32 @@ describe('createStore', () => {
     expect(listenerC.calls.length).toBe(2)
   })
 
-  // See final decision here: https://github.com/rackt/redux/issues/1180
   it('removes listeners NOT immediately when unsubscribe is called', () => {
     const store = createStore(reducers.todos)
 
     const unsubscribeHandles = []
     const doUnsubscribeAll = () => unsubscribeHandles.forEach(unsubscribe => unsubscribe() )
-
+    
     const listener1 = expect.createSpy(() => {})
-    unsubscribeHandles.push(store.subscribe(listener1))
     const listener2 = expect.createSpy(() => {})
-    unsubscribeHandles.push(store.subscribe(listener2))
+    const listener3 = expect.createSpy(() => {})
 
-    const listener3 = () => doUnsubscribeAll()
-    unsubscribeHandles.push(store.subscribe(listener3))
-
-    const listener4 = expect.createSpy(() => {})
-    unsubscribeHandles.push(store.subscribe(listener4))
-    const listener5 = expect.createSpy(() => {})
-    unsubscribeHandles.push(store.subscribe(listener5))
+    unsubscribeHandles.push(store.subscribe(() => listener1())))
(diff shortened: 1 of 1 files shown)


test/createStore.spec.js
```diff
@@ -286,6 +286,66 @@ describe('createStore', () => {
...
+    const listener4 = expect.createSpy(() => {})
+    unsubscribeHandles.push(store.subscribe(listener4))
+    const listener5 = expect.createSpy(() => {})
+    unsubscribeHandles.push(store.subscribe(listener5))
+
+    store.dispatch(unknownAction())
+    store.dispatch(unknownAction())
+    expect(listener1.calls.length).toBe(1)
+    expect(listener2.calls.length).toBe(1)
+    // listener3 is the one removing the others so obviously it got executed
```
gaearon: I'd still prefer an explicit check to an “obvious” comment ;-)

test/createStore.spec.js
```diff
@@ -286,6 +286,66 @@ describe('createStore', () => {
...
+    expect(listener4.calls.length).toBe(1)
+    expect(listener5.calls.length).toBe(1)
+  })
+
+  it('not fire immediately if a listener is added inside another listener', () => {
```
gaearon: Nit: `it('does not`

test/createStore.spec.js
```diff
@@ -286,6 +286,66 @@ describe('createStore', () => {
...
+    const store = createStore(reducers.todos)
+
+    const listener1 = expect.createSpy(() => {})
+    const listener2 = expect.createSpy(() => {})
+    const listener3 = expect.createSpy(() => {})
+
+    let listener3Added = false
+    const maybeAddThirdListener = () => {
+      if ( !listener3Added ) {
```
gaearon: Nit: we don't put spaces around conditions. (Why doesn't ESLint catch that?)

test/createStore.spec.js
```diff
@@ -286,6 +286,66 @@ describe('createStore', () => {
...
+        listener3Added = true
+        store.subscribe(() => listener3())
+      }
+    }
+
+    // Listener 1 is normal
+    store.subscribe(() => listener1())
+
+    // Listener 2 adds the 3rd listener another listener
```
gaearon: I think we can remove these comments, the code is pretty descriptive

gaearon:
Thanks! This is a valuable PR. Always good to ensure we don't break promises.
I'll be happy to merge after you address a few nits.

test/createStore.spec.js
slorber: oooh ok i hate this rule :)

gaearon: FWIW I hate no-semicolons but other folks from `rackt` use this style. Consistency :100:

slorber:
@gaearon I've updated the code it should be fine now

np that will be my first merged PR on redux ecosystem :)

gaearon:
Appreciated. Congrats!

PR #1293, opened 2016-01-27
Update object spread suggestion in Troubleshooting for babel@6.

### LINKED ISSUE #1292: Object rest spread suggestion in Troubleshooting is outdated for babel@6
This suggestion / link in [Troubleshooting](http://redux.js.org/docs/Troubleshooting.html) is outdated with respect to Babel 6:

> You can also enable [ES7 object spread proposal](https://github.com/sebmarkbage/ecmascript-rest-spread) with [Babel stage 1](http://babeljs.io/docs/usage/experimental/):

PR incoming.

jmm:
Fix #1292. You could continue to describe how to do it for Babel 5, but no one is maintaining that and people using it have bigger fish to fry than enabling stage 2 syntax, so IMO it's better not mentioned.

I also trimmed the example. IMO this more focused example makes is easier to understand without sorting through the visual noise of the adjacent array spread syntax. The key takeaway should be how object spread contrasts with `Object.assign()` -- I don't think people will lose sight of the big picture because of not reproducing the whole previous example, but it will be easier to see the key difference. If it's not clear to them how the syntax can be applied to any object literal, they're probably better off not using it yet.

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -9,11 +9,7 @@ Sometimes, you are trying to dispatch an action, but your view does not update. 
 
 #### Never mutate reducer arguments
 
-It is tempting to modify the `state` or `action` passed to you by Redux. Don’t do this!
-
-Redux assumes that you never mutate the objects it gives to you in the reducer. **Every single time, you must return the new state object.** Even if you don’t use a library like [Immutable](https://facebook.github.io/immutable-js/), you need to completely avoid mutation.
-
-Immutability is what lets [react-redux](https://github.com/gaearon/react-redux) efficiently subscribe to fine-grained updates of your state. It also enables great developer experience features such as time travel with [redux-devtools](http://github.com/gaearon/redux-devtools).
+It’s tempting to modify the `state` (or objects nested within it) or `action` passed to you by Redux. Don’t do this!
 
 For example, a reducer like this is wrong because it mutates the state:
 
@@ -37,7 +33,31 @@ function todos(state = [], action) {
 }
 ```
 
-It needs to be rewritten like this:
+Redux depends on you never mutating the objects it passes to your reducer, or objects nested within them. **Every single time you change state, you must return a new state object.** Even if you don’t use a library like [Immutable](https://facebook.github.io/immutable-js/), you need to completely avoid mutation.
+
+Immutability is what lets [react-redux](https://github.com/gaearon/react-redux) efficiently subscribe to fine-grained updates of your state. It also enables great developer experience features such as time travel with [redux-devtools](http://github.com/gaearon/redux-devtools).
+
+To avoid mutating arrays you can use things like `.slice()`, `.concat()`, and spread (`...state`) to extract the elements you want to retain and combine them with any new elements you may want to add, while omitting elements you don’t want.
+
+To avoid mutating objects you’ll generally want to use a utility like `extend()` from Underscore, `assign()` from Lodash, or [`Object.assign()`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) or a polyfill for it.
+
+`Object.assign()` mutates its first argument, so make sure that you use it correctly. For example, instead of returning something like `Object.assign(state, { completed: true })` from your reducers, return `Object.assign({}, state, { completed: true })`. This way you don’t mutate the previous `state`.
+
+If you’re using Babel 6 you can alternatively use the [proposed object spread syntax](https://github.com/sebmarkbage/ecmascript-rest-spread) by enabling [`babel-plugin-transform-object-rest-spread`](http://babeljs.io/docs/plugins/transform-object-rest-spread/) or a preset that includes it. (At the time of this writing the proposal is stage 2 and therefore the plugin is included directly in [`babel-preset-stage-2`](http://babeljs.io/docs/plugins/preset-stage-2/) and indirectly in lower numbered stage presets.) Example:
+
+```js
+// Before:
+state[action.index] = Object.assign({}, state[action.index], {
+  completed: true
+})
+
+// After:
+state[action.index] = { ...state[action.index], completed: true }
+```
+
+Note that non-standard language features are subject to change, and it’s unwise to rely on them in large codebases.
+
+The earlier example needs to be rewritten to not mutate state. Here’s a corrected version using the techniques we’ve discussed:
 
 ```js
 function todos(state = [], action) {
@@ -52,15 +72,13 @@ function todos(state = [], action) {
         }
       ]
     case 'COMPLETE_TODO':
-      // Return a new array
-      return [
-        ...state.slice(0, action.index),
-        // Copy the object before mutating
-        Object.assign({}, state[action.index], {
-          completed: true
-        }),
-        ...state.slice(action.index + 1)
-      ]
+      // Copy elements to a new array
+      state = [...state]
+      // Copy the object before mutating
+      state[action.index] = Object.assign({}, state[action.index], {
+        completed: true
+      })
+      return state
     default:
       return state
   }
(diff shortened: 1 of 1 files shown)


docs/Troubleshooting.md
```diff
@@ -93,27 +93,17 @@ Finally, to update objects, you’ll need something like `_.extend` from Undersc
 
 Make sure that you use `Object.assign` correctly. For example, instead of returning something like `Object.assign(state, newData)` from your reducers, return `Object.assign({}, state, newData)`. This way you don’t override the previous `state`.
 
-You can also enable [ES7 object spread proposal](https://github.com/sebmarkbage/ecmascript-rest-spread) with [Babel stage 1](http://babeljs.io/docs/usage/experimental/):
+If using Babel 6 you can alternatively use the [proposed object spread syntax](https://github.com/sebmarkbage/ecmascript-rest-spread) by enabling [`babel-plugin-transform-object-rest-spread`](http://babeljs.io/docs/plugins/transform-object-rest-spread/) or a preset that includes it. (At the time of this writing the proposal is stage 2 and therefore the plugin is included directly in [`babel-preset-stage-2`](http://babeljs.io/docs/plugins/preset-stage-2/) and indirectly in lower numbered stage presets.) Example:
```
gaearon: "If using" => "If you are using"?

gaearon:
I agree with your goal here but I don't like that this example doesn't match the 3 other examples on this page. What we can do instead is to split `todo` from `todos` reducer, like in my Egghead series. `todo` would update a single todo and use `Object.assign()` whereas `todos` would delegate to it. Then we can take different parts of this example (e.g. `todo` reducer) and show how different libraries or syntaxes simplify these parts.

docs/Troubleshooting.md
jmm: Either works for me, so no objection.

jmm:
@gaearon Thanks for the feedback. I haven't watched the video, but I believe I know what you mean. I'll make some updates. I need to check exactly how `React.addons.update` works and how splitting the reducers up affects the context for that example.

jmm:
@gaearon I pushed some new commits. I tried reorganizing and rewriting this somewhat. Please see what you think.

1. I originally did modify this to break out a `todo` reducer, but it pretty much ruins the context for the `React.addons.update` example. So instead I just reworked the gnarliest bit of the code:

  ```js
  // Return a new array
  return [
    ...state.slice(0, action.index),
    // Copy the object before mutating
    Object.assign({}, state[action.index], {
      completed: true
    }),
    ...state.slice(action.index + 1)
  ]
  ```

  To this:

  ```js
  // Copy elements to a new array
  state = [...state]
  // Copy the object before mutating
  state[action.index] = Object.assign({}, state[action.index], {
    completed: true
  })
  return state
  ```

  That way a fragment of that same *after* example can be used in the *before* of the object spread example while preserving the context of a deeper state tree for the `update()` example.

2. Since it didn't make much sense to me to show examples updated to use `Object.assign()` before getting to:

  > Finally, to update objects

  I reorganized this to discuss general techniques for working with arrays and objects before showing the original example reworked to use those techniques. This leaves the object spread example a bit out of place, but the *before* part of it is now a direct excerpt from the full rewritten example.

3. Some general copy changes, e.g. clarifying and strengthening the language around not mutating:

  Before:

  > Redux assumes that you never mutate the objects it gives to you in the reducer.

  After:

  > Redux depends on you never mutating the objects it passes to your reducer, or objects nested within them.

FYI, I haven't run the updated code examples as of yet.

docs/Troubleshooting.md
```diff
@@ -52,15 +72,13 @@ function todos(state = [], action) {
...
-      return [
-        ...state.slice(0, action.index),
-        // Copy the object before mutating
-        Object.assign({}, state[action.index], {
-          completed: true
-        }),
-        ...state.slice(action.index + 1)
-      ]
+      // Copy elements to a new array
+      state = [...state]
```
gaearon: This is problematic. We’ve seen people doing this and then mutating something deep inside the objects in array, thinking `state = [...state]` saved them. I know you have `Object.assign()` below but people will skip that and remember just `state = [...state]`.

docs/Troubleshooting.md
```diff
@@ -37,7 +33,31 @@ function todos(state = [], action) {
...
 ```
 
-It needs to be rewritten like this:
+Redux depends on you never mutating the objects it passes to your reducer, or objects nested within them. **Every single time you change state, you must return a new state object.** Even if you don’t use a library like [Immutable](https://facebook.github.io/immutable-js/), you need to completely avoid mutation.
+
+Immutability is what lets [react-redux](https://github.com/gaearon/react-redux) efficiently subscribe to fine-grained updates of your state. It also enables great developer experience features such as time travel with [redux-devtools](http://github.com/gaearon/redux-devtools).
+
+To avoid mutating arrays you can use things like `.slice()`, `.concat()`, and spread (`...state`) to extract the elements you want to retain and combine them with any new elements you may want to add, while omitting elements you don’t want.
+
+To avoid mutating objects you’ll generally want to use a utility like `extend()` from Underscore, `assign()` from Lodash, or [`Object.assign()`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) or a polyfill for it.
```
gaearon: `Object.assign()` should be the first option here. We should not encourage people to use Lodash/Underscore just for the sake of assigning objects.

docs/Troubleshooting.md
```diff
@@ -37,7 +33,31 @@ function todos(state = [], action) {
 }
 ```
 
-It needs to be rewritten like this:
+Redux depends on you never mutating the objects it passes to your reducer, or objects nested within them. **Every single time you change state, you must return a new state object.** Even if you don’t use a library like [Immutable](https://facebook.github.io/immutable-js/), you need to completely avoid mutation.
+
+Immutability is what lets [react-redux](https://github.com/gaearon/react-redux) efficiently subscribe to fine-grained updates of your state. It also enables great developer experience features such as time travel with [redux-devtools](http://github.com/gaearon/redux-devtools).
+
+To avoid mutating arrays you can use things like `.slice()`, `.concat()`, and spread (`...state`) to extract the elements you want to retain and combine them with any new elements you may want to add, while omitting elements you don’t want.
```
gaearon: things -> Array methods. Please make them links to the appropriate MDN pages.

docs/Troubleshooting.md
```diff
@@ -37,7 +33,31 @@ function todos(state = [], action) {
...
+
+To avoid mutating objects you’ll generally want to use a utility like `extend()` from Underscore, `assign()` from Lodash, or [`Object.assign()`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) or a polyfill for it.
+
+`Object.assign()` mutates its first argument, so make sure that you use it correctly. For example, instead of returning something like `Object.assign(state, { completed: true })` from your reducers, return `Object.assign({}, state, { completed: true })`. This way you don’t mutate the previous `state`.
+
+If you’re using Babel 6 you can alternatively use the [proposed object spread syntax](https://github.com/sebmarkbage/ecmascript-rest-spread) by enabling [`babel-plugin-transform-object-rest-spread`](http://babeljs.io/docs/plugins/transform-object-rest-spread/) or a preset that includes it. (At the time of this writing the proposal is stage 2 and therefore the plugin is included directly in [`babel-preset-stage-2`](http://babeljs.io/docs/plugins/preset-stage-2/) and indirectly in lower numbered stage presets.) Example:
```
gaearon: I’d say let’s wait for #1358 to get merged and just link to it rather than explain it in detail here.

gaearon:
I’m closing because it is a little outdated now (doesn’t merge cleanly). If you’d like to pick up the work on this, rebase it to be on top of the existing docs, and update it in response to my comments above, I’ll gladly reopen! (Fine to do as a separate PR too, if you’d like.) Cheers!

PR #1294, opened 2016-01-28
Add first-class support for store enhancers to createStore() API

gaearon:
(Updated to @timdorr’s suggestion in https://github.com/rackt/redux/pull/1294#issuecomment-175938621)

### TL;DR

```diff
- const createStoreWithMiddleware = applyMiddleware(thunk, logger)(createStore)
- const store = createStoreWithMiddleware(rootReducer, initialState)
+ const store = createStore(
+   rootReducer,
+   initialState,
+   applyMiddleware(thunk, logger)
+ )
```

### Fixing the Store Enhancer API

It’s often said that creating a store with enhancers like `applyMiddleware()` or `DevTools.instrument()` is weird because it just looks outlandish in JavaScript to compose functions in such a savage way:

```js
// bad
const createStoreWithMiddleware = applyMiddleware(thunk, logger)(createStore)
const store = createStoreWithMiddleware(reducer, initialState)

// and even worse
  applyMiddleware(thunk, logger),
  DevTools.instrument()
)(createStore)
const store = finalCreateStore(reducer, initialState)
```

Here, we propose to integrate the idea of store enhancers into the `createStore()` API itself. This would keep the concept of enhancers and their API exactly the same but will simplify applying them in real projects. This is also not a breaking change because we didn’t support any additional arguments to `createStore()` previously.

```js
// not bad
const store = createStore(
  reducer,
  initialState,
  applyMiddleware(thunk, logger)
)

// not too bad either
const store = createStore(
  reducer,
  initialState,
  compose(
    applyMiddleware(thunk, logger),
    DevTools.instrument()
  )
)
```

It has often been suggested that we do something about this uneasiness but unfortunately I don’t think there was ever any specific proposal that dealt with store enhancers correctly. Most proposals I saw were focused on the middleware which is just one kind of a store enhancer. I think that this proposal should satisfy the needs of the community in terms of a simple API without compromising on the power of store enhancers.

What do you think?

### DIFF
--- a/docs/advanced/AsyncActions.md
+++ b/docs/advanced/AsyncActions.md
@@ -138,7 +138,7 @@ Here’s what the state shape for our “Reddit headlines” app might look like
         {
           id: 42,
           title: 'Confusion about Flux and Relay'
-        }, 
+        },
         {
           id: 500,
           title: 'Creating a Simple Application Using React JS and Flux Architecture'
@@ -384,7 +384,7 @@ export function fetchPosts(subreddit) {
 >import 'babel-core/polyfill'
 >```
 
-How do we include the Redux Thunk middleware in the dispatch mechanism? We use the [`applyMiddleware()`](../api/applyMiddleware.md) method from Redux, as shown below:
+How do we include the Redux Thunk middleware in the dispatch mechanism? We use the [`applyMiddleware()`](../api/applyMiddleware.md) store enhancer from Redux, as shown below:
 
 #### `index.js`
 
@@ -397,12 +397,13 @@ import rootReducer from './reducers'
 
 const loggerMiddleware = createLogger()
 
-const createStoreWithMiddleware = applyMiddleware(
-  thunkMiddleware, // lets us dispatch() functions
-  loggerMiddleware // neat middleware that logs actions
-)(createStore)
-
-const store = createStoreWithMiddleware(rootReducer)
+const store = createStore(
+  rootReducer,
+  applyMiddleware(
+    thunkMiddleware, // lets us dispatch() functions
+    loggerMiddleware // neat middleware that logs actions
+  )
+)
 
 store.dispatch(selectSubreddit('reactjs'))
 store.dispatch(fetchPosts('reactjs')).then(() =>
--- a/docs/advanced/ExampleRedditAPI.md
+++ b/docs/advanced/ExampleRedditAPI.md
@@ -170,13 +170,15 @@ import rootReducer from './reducers'
 
 const loggerMiddleware = createLogger()
 
-const createStoreWithMiddleware = applyMiddleware(
-  thunkMiddleware,
-  loggerMiddleware
-)(createStore)
-
 export default function configureStore(initialState) {
-  return createStoreWithMiddleware(rootReducer, initialState)
+  return createStore(
+    rootReducer,
+    initialState,
+    applyMiddleware(
+      thunkMiddleware,
+      loggerMiddleware
+    )
+  )
 }
 ```
 
--- a/docs/advanced/Middleware.md
+++ b/docs/advanced/Middleware.md
@@ -272,6 +272,8 @@ The implementation of [`applyMiddleware()`](../api/applyMiddleware.md) that ship
 
 * To ensure that you may only apply middleware once, it operates on `createStore()` rather than on `store` itself. Instead of `(store, middlewares) => store`, its signature is `(...middlewares) => (createStore) => createStore`.
 
+Because it is cumbersome to apply functions to `createStore()` before using it, `createStore()` accepts an optional last argument to specify such functions.
+
 ### The Final Approach
 
 Given this middleware we just wrote:
@@ -305,13 +307,12 @@ Here’s how to apply it to a Redux store:
 ```js
 import { createStore, combineReducers, applyMiddleware } from 'redux'
 
-// applyMiddleware takes createStore() and returns
-// a function with a compatible API.
-let createStoreWithMiddleware = applyMiddleware(logger, crashReporter)(createStore)
-
-// Use it like you would use createStore()
 let todoApp = combineReducers(reducers)
-let store = createStoreWithMiddleware(todoApp)
+let store = createStore(
+  todoApp,
+  // applyMiddleware() tells createStore() how to handle middleware
+  applyMiddleware(logger, crashReporter)
+)
 ```
 
 That’s it! Now any actions dispatched to the store instance will flow through `logger` and `crashReporter`:
@@ -378,8 +379,8 @@ const timeoutScheduler = store => next => action => {
 }
 
 /**
- * Schedules actions with { meta: { raf: true } } to be dispatched inside a rAF loop 
- * frame.  Makes `dispatch` return a function to remove the action from the queue in 
+ * Schedules actions with { meta: { raf: true } } to be dispatched inside a rAF loop
+ * frame.  Makes `dispatch` return a function to remove the action from the queue in
  * this case.
  */
 const rafScheduler = store => next => {
@@ -472,15 +473,17 @@ const thunk = store => next => action =>
 
 
 // You can use all of them! (It doesn’t mean you should.)
-let createStoreWithMiddleware = applyMiddleware(
-  rafScheduler,
-  timeoutScheduler,
-  thunk,
-  vanillaPromise,
-  readyStatePromise,
-  logger,
-  crashReporter
-)(createStore)
 let todoApp = combineReducers(reducers)
-let store = createStoreWithMiddleware(todoApp)
+let store = createStore(
+  todoApp,
+  applyMiddleware(
+    rafScheduler,
+    timeoutScheduler,
+    thunk,
+    vanillaPromise,
+    readyStatePromise,
+    logger,
+    crashReporter
+  )
+)
 ```
--- a/docs/api/applyMiddleware.md
+++ b/docs/api/applyMiddleware.md
@@ -14,7 +14,7 @@ Middleware is not baked into [`createStore`](createStore.md) and is not a fundam
 
 #### Returns
 
-(*Function*) A store enhancer that applies the given middleware. The store enhancer is a function that needs to be applied to `createStore`. It will return a different `createStore` which has the middleware enabled.
+(*Function*) A store enhancer that applies the given middleware. The store enhancer signature is `createStore => createStore'` but the easiest way to apply it is to pass it to [`createStore()`](./createStore.md) as the last `enhancer` argument.
 
 #### Example: Custom Logger Middleware
 
@@ -37,8 +37,11 @@ function logger({ getState }) {
   }
 }
 
-let createStoreWithMiddleware = applyMiddleware(logger)(createStore)
-let store = createStoreWithMiddleware(todos, [ 'Use Redux' ])
+let store = createStore(
+  todos,
+  [ 'Use Redux' ],
+  applyMiddleware(logger)
+)
 
 store.dispatch({
   type: 'ADD_TODO',
@@ -56,12 +59,9 @@ import { createStore, combineReducers, applyMiddleware } from 'redux'
 import thunk from 'redux-thunk'
 import * as reducers from './reducers'
 
-// applyMiddleware supercharges createStore with middleware:
-let createStoreWithMiddleware = applyMiddleware(thunk)(createStore)
-
-// We can use it exactly like “vanilla” createStore.
 let reducer = combineReducers(reducers)
-let store = createStoreWithMiddleware(reducer)
+// applyMiddleware supercharges createStore with middleware:
+let store = createStore(reducer, applyMiddleware(thunk))
 
 function fetchSecretSauce() {
   return fetch('https://www.google.com/search?q=secret+sauce')
@@ -229,7 +229,12 @@ export default connect(
     let d = require('another-debug-middleware');
     middleware = [ ...middleware, c, d ];
   }
-  const createStoreWithMiddleware = applyMiddleware(...middleware)(createStore);
+
+  const store = createStore(
+    reducer,
+    initialState,
+    applyMiddleware(middleware)
+  )
   ```
 
   This makes it easier for bundling tools to cut out unneeded modules and reduces the size of your builds.
@@ -237,3 +242,5 @@ export default connect(
 * Ever wondered what `applyMiddleware` itself is? It ought to be an extension mechanism more powerful than the middleware itself. Indeed, `applyMiddleware` is an example of the most powerful Redux extension mechanism called [store enhancers](../Glossary.md#store-enhancer). It is highly unlikely you’ll ever want to write a store enhancer yourself. Another example of a store enhancer is [redux-devtools](https://github.com/gaearon/redux-devtools). Middleware is less powerful than a store enhancer, but it is easier to write.
 
 * Middleware sounds much more complicated than it really is. The only way to really understand middleware is to see how the existing middleware works, and try to write your own. The function nesting can be intimidating, but most of the middleware you’ll find are, in fact, 10-liners, and the nesting and composability is what makes the middleware system powerful.
+
+* To apply multiple store enhancers, you may use [`compose()`](./compose.md).
--- a/docs/api/compose.md
+++ b/docs/api/compose.md
@@ -20,40 +20,16 @@ This example demonstrates how to use `compose` to enhance a [store](Store.md) wi
 ```js
 import { createStore, combineReducers, applyMiddleware, compose } from 'redux'
 import thunk from 'redux-thunk'
-import * as reducers from '../reducers/index'
+import DevTools from './containers/DevTools'
+import reducer from '../reducers/index'
 
-let reducer = combineReducers(reducers)
-let middleware = [ thunk ]
-
-let finalCreateStore
-
-// In production, we want to use just the middleware.
-// In development, we want to use some store enhancers from redux-devtools.
-// UglifyJS will eliminate the dead code depending on the build environment.
-
-if (process.env.NODE_ENV === 'production') {
-  finalCreateStore = applyMiddleware(...middleware)(createStore)
-} else {
-  finalCreateStore = compose(
-    applyMiddleware(...middleware),
-    require('redux-devtools').devTools(),
-    require('redux-devtools').persistState(
-      window.location.href.match(/[?&]debug_session=([^&]+)\b/)
-    )
-  )(createStore)
-
-  // Same code without the `compose` helper:
-  //
-  // finalCreateStore = applyMiddleware(middleware)(
-  //   require('redux-devtools').devTools()(
-  //     require('redux-devtools').persistState(
-  //       window.location.href.match(/[?&]debug_session=([^&]+)\b/)
-  //     )(createStore)
-  //   )
-  // )
-}
-
-let store = finalCreateStore(reducer)
+const store = createStore(
+  reducer,
+  compose(
+    applyMiddleware(thunk),
+    DevTools.instrument()
+  )
+)
 ```
 
 #### Tips
--- a/docs/api/createStore.md
+++ b/docs/api/createStore.md
@@ -9,6 +9,8 @@ There should only be a single store in your app.
 
 2. [`initialState`] *(any)*: The initial state. You may optionally specify it to hydrate the state from the server in universal apps, or to restore a previously serialized user session. If you produced `reducer` with [`combineReducers`](combineReducers.md), this must be a plain object with the same shape as the keys passed to it. Otherwise, you are free to pass anything that your `reducer` can understand.
 
+3. [`enhancer`] *(Function)*: The store enhancer. You may optionally specify it to enhance the store with third-party capabilities such as middleware, time travel, persistence, etc. The only store enhancer that ships with Redux is [`applyMiddleware()`](./applyMiddleware.md).
+
 #### Returns
 
 ([*`Store`*](Store.md)): An object that holds the complete state of your app. The only way to change its state is by [dispatching actions](Store.md#dispatch). You may also [subscribe](Store.md#subscribe) to the changes to its state to update the UI.
@@ -49,3 +51,5 @@ console.log(store.getState())
 * For universal apps that run on the server, create a store instance with every request so that they are isolated. Dispatch a few data fetching actions to a store instance and wait for them to complete before rendering the app on the server.
 
 * When a store is created, Redux dispatches a dummy action to your reducer to populate the store with the initial state. You are not meant to handle the dummy action directly. Just remember that your reducer should return some kind of initial state if the state given to it as the first argument is `undefined`, and you’re all set.
+
+* To apply multiple store enhancers, you may use [`compose()`](./compose.md).
--- a/examples/async/store/configureStore.js
+++ b/examples/async/store/configureStore.js
@@ -3,13 +3,12 @@ import thunkMiddleware from 'redux-thunk'
 import createLogger from 'redux-logger'
 import rootReducer from '../reducers'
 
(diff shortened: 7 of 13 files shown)


bradwestfall:
+1

tedpennings:
This might be a stupid question, but how would I skip providing initial state. would this work?
```
createStore(
  reducer,
  undefined, // no initial state
  applyMiddleware(thunk, batchedUpdates, NewRelicErrorLoggingMiddleware)
)
```

holic:
@tedpennings According to the source, you'd just leave out the `initialState`, e.g.

```js
createStore(
  reducer,
  applyMiddleware(thunk, batchedUpdates, NewRelicErrorLoggingMiddleware)
)
```

anru:
Nice, I like it.

tedpennings:
awesome, thanks @holic -- I like this a lot :bowtie:

gaearon:
Yes, we can detect you didn't pass the initial state because valid state can never be a function.

tj:
:+1: I still have to track down some weird mis-config that is making middleware no-op, 99% it's something in my store stuff. Just my opinion but I find too many positional args looks confusing, they all end up looking like middleware or something.

gaearon:
>Just my opinion but I find too many positional args looks confusing, they all end up looking like middleware or something.

Can you throw a few alternatives off the top of your head just to get the ball rolling? I tried a few options with arrays but neither felt satisfactory.

timdorr:
:-1: from me.

What's wrong with encouraging good FP concepts?  Also, the way that reads is like the reducer and initial state are part of the "stack" of composed store enhancers. It's hard to tell when the configuration ends and the store enhancers begin.

What about a singular enhancer (much like a singular reducer), which can use `compose` to stack them:

```js
const enhancers = compose(
  applyMiddleware(thunk, logger),
  DevTools.instrument()
)

const store = createStore(
  reducer,
  initialState,
  enhancers
)
```

gaearon:
I like that this proposal allows us to drop `compose()` from the public API is it is no longer necessary for using the library. Those who want it elsewhere in the code can always use the lodash version.

timdorr:
> I like that this proposal allows us to drop compose() from the public API

Yeah, I agree with that. It's weird to provide that when someone might want to use Ramda or a homegrown solution. Redux isn't a FP library, it just leverages FP concepts.

tj:
> Can you throw a few alternatives off the top of your head just to get the ball rolling? I tried a few options with arrays but neither felt satisfactory.

No clue :D

gaearon:
>What's wrong with encouraging good FP concepts?

It's a parenthesis mess. `finalCreateStore` and `createStoreWithMiddleware` are incredible awkward to write and explain. In fact this was one of the reasons I still haven't recorded videos on the middleware—I can't find a way to teach them without writing very weird syntax. See also https://hughfdjackson.com/javascript/does-curry-help/ for a similar perspective.

Those who dig FPisms can always look inside or compose things manually but I've found that usually people turn off their hands at the sight of this blurb and copy-paste it across projects without understanding how it works. Providing a nicer top-level API is the first step towards helping them understand what's happening. 

>Also, the way that reads is like the reducer and initial state are part of the "stack" of composed store enhancers. It's hard to tell when the configuration ends and the store enhancers begin.

I agree with this criticism. Another problem is it makes the flow unobvious. Actions would hit the enhancers first top-down and then proceed to the reducer which would be weird. If you can offer an alternative way of stacking them that would be backwards compatible, wouldn't be awkward, would make it easy to just use `applyMiddleware()`, as well as combine it with DevTools, would allow to omit initial state, and wouldn't freak out at arrays as state, I'm keen to hear it!

tj:
> I've found that usually people turn off their hands at the sight of this blurb and copy-paste it across projects without understanding how it works.

This is totally me hahaha. My first impression when I see lots of distinct calls like that is "oh god, this is complicated", even if they're all doing fundamentally similar things.

AntJanus:
I'm not a big fan of the variadic function solution. The original solution is good but it does feel clunky. It feels like you should be able to apply middleware at the same time as creating the store. How about an `applyMiddleware` chain function? Like so:

```
const store = createStore(reducer, initialState).applyMiddleware(thunk, logger);
```

Or ExpressJS style (doesn't feel FP enough though):

```
const store = createStore(reducers, initialState);

store.use(thunk, logger);

//or
store.use(thunk);
store.use(logger);
```

Just throwing some thoughts out.

timdorr:
> It's a parenthesis mess.

Yeah, that's definitely true. This isn't a lisp! :smile: 

> If you can offer an alternative...

I think my suggestion satisfies this. Single reducer function, single enhancer function. Since we commonly export a root reducer in our examples and in the the real world, we can also export our enhancers the same way. You'd end up with this:

```js
import rootEnhancer from './enhancers'

const store = createStore(
  rootReducer,
  initialState,
  rootEnhancer
)
```
```js
export default const enhancers = compose(
  applyMiddleware(thunk, logger),
  DevTools.instrument()
)
```
Keeps parity with how you lay out reducers, so it's familiar. And it's backwards compatible, since it's just a new arg on `createStore`

gaearon:
No, there are very specific reasons why the underlying types have to be exactly as they are now. Store enhancers must be `createStore => createStore'`. When you use a store enhancer like DevTools it actually creates another store internally and gives you a projection of that store. We can only do that if we wrap `createStore` itself. We can't let you create a store because by this point it's too late. This is why those fluent APIs and `.use()` calls don't work here.

gaearon:
@timdorr I kinda like your suggestion. I really wanted to part ways with `compose()` but maybe this is our fate.

acdlite:
I think this is a good idea; however, I 100% agree with @tj here:

> Just my opinion but I find too many positional args looks confusing, they all end up looking like middleware or something. I think variadic is fine if it's all of the same type like applyMiddleware().

I would suggest making third argument an (optional) array of enhancers:

```js
const store = createStore(
   rootReducer,
   initialState,
   [applyMiddleware(thunk, logger), DevTools.instrument(), ...etc]
 )
```

gaearon:
So, per @timdorr:

```js
// not bad
const store = createStore(
  reducer,
  initialState,
  applyMiddleware(thunk, logger)
)

// not bad either
const store = createStore(
  reducer,
  initialState,
  compose(
    applyMiddleware(thunk, logger),
    DevTools.instrument()
  )
)
```

No variadic stuff, probably more static analysis friendly too. Thoughts?

tj:
@gaearon that looks pretty good to me

gaearon:
>I would suggest making third argument an (optional) array of enhancers:

I wanted to do this but then watch out:

```js
// I want to allow to omit the initial state
const store = createStore(
   rootReducer,
   [applyMiddleware(thunk, logger), DevTools.instrument(), ...etc]
 )

// Haha good luck telling array initial state from array of enhancers
const store = createStore(
   rootReducer,
   [{ lol: true }, { sad: true }],
   [applyMiddleware(thunk, logger), DevTools.instrument(), ...etc]
 )
```

acdlite:
@gaearon Hmm didn't think about omitting `initialState`. So... we would type check the second param to see if it's a function? Eh, I'm not a huge fan of that but idk.

BerkeleyTrue:
How about a single argument function?

```js
const store = createStore({
  reducer,
  initialState,
  enhancer: applyMiddleware(thunk, logger)
})
```
And/Or

```js
const store = createStore({
  reducer,
  initialState,
  enhancer: [thunk, logger]
})
```

tj:
why not just make the initial state mandatory? looks gross but at least it's obvious

acdlite:
@tj If we could go back in time I'd be 100% for making the initial state mandatory.

tj:
@acdlite what about just in this scenario though?

tj:
@BerkeleyTrue second one is kinda nice :+1:  I guess it'd be missing the functional feel though BUT I must say it's awesome to `import` less stuff for composing and whatnot

gaearon:
>why not just make the initial state mandatory? looks gross but at least it's obvious

That’s definitely an option. Still might be easy to forget it during refactoring (especially when you add middleware to a project that didn't use it), and it's hard for us to provide a good error message if you mess it up. Also yeah, `undefined` holes in function calls feel gross.

bradwestfall:
@gaearon I don't think it's gross, I think it makes it more clear honestly

BerkeleyTrue:
Both could be done at once, and since reducer is currently always a function it could be done in a backwards compatible way. If first arg is type function, then old api and warn of deprecation. if type object, new api.

tj:
@BerkeleyTrue's object suggestion might be nice that way, enough to probe to treat it as a new api

gaearon:
To be fair composing is not a very common scenario anyway. Most people are content with `applyMiddleware` and haven't bothered with setting up DevTools so the common case is just one line.

gaearon:
>@gaearon I don't think it's gross, I think it makes it more clear honestly

It makes it clear for people who know that but I guarantee you people will file issues saying something doesn't work because they forgot to put it in there when they added middleware to a non-server-rendered project that never used `initialState`.

acdlite:
IMO the grossness of passing `undefined` as the initial state is really just exposing the hidden grossness of relying on default parameters + an init action to specify initial state. Alas, probably too late to fix that :D

bradwestfall:
Btw, for me, passing `undefined` is gross, I thought we were talking about passing `{}` as a means of no initial state

timdorr:
> Alas, probably too late to fix that :D

This is on loop in my head now: https://www.youtube.com/watch?v=BsKbwR7WXN4

timdorr:
I'm a **huge** fan of keyword/non-positional args. Love, love, love them! But while it's quick and easy for us in ES2015-land, it's going to be more verbose for our ES5 friends. So, there's a small tradeoff there to be aware of.

acdlite:
@timdorr Haha yeah I know I'm fighting a losing battle but...

Imagine if this were a typed language. You would *think* the type signature for a Redux reducer is

```
State -> Action -> State
```

but if you follow idiomatic Redux it's actually

```
Maybe State -> Action -> State
```

acdlite:
@timdorr Huh is it really that much more verbose? I don't see what you mean.

BerkeleyTrue:
> it's going to be more verbose for our ES5 friends 

I'm not sure what you mean by this. Can you elaborate?

timdorr:
It's only a slight increase:
```js
const store = createStore({
  reducer: reducer,
  initialState: initialState,
  enhancer: applyMiddleware(thunk, logger)
})
```
Not earth shattering, but not really fun to program if you're not using ES2015 \*shrug*

gaearon:
@BerkeleyTrue @tj

I'd rather avoid the “configuration object” API. We’d have to add checks for typos, support both versions to be backwards compatible, tell people to use the new version for “advanced” behaviors, figure out whether to allow `enhancer`, `enhancers`, or both, have to put up with lingo code like `enhancers: applyMiddleware(thunk)` which seems to introduce cryptic terminology in app code for no gain, and worst of all, we’ll get more feature requests to put stuff into that configuration object :smile: . I think configuration object means we can't figure out what the API does. Better for us to choose the right focus.

(I understand it's very subjective though)

tj:
I actually prefer that hahaha, `({foo,bar})` looks meaningless to me, I just see args half the time

timdorr:
And for those in IE8 purgatory (I think. It's been a while since I've done ES3):
```js
var store = createStore({
  "reducer": reducer,
  "initialState": initialState,
  "enhancer": applyMiddleware(thunk, logger)
});

**Comment by acdlite** [comment]
> And for those in IE8 purgatory

Pfft, as far as I'm concerned they can go to hell

**Comment by acdlite** [comment]
^ That was a (lame) joke, btw, I'm actually a nice guy

**Comment by gaearon** [comment]
Nah, you only need to quote reserved words like `default` for IE8.

**Comment by gaearon** [comment]
I’m pretty confident I want to get @timdorr’s version in.

It’s a simple change, a purely cosmetic one. Still has you using `compose()` if you want multiple enhancers, but they read much nicer. Leaves us a room to do something more drastic in the future with variadic arguments if we choose to. (Compose result is also enhancer so adding support for multiple enhancers without explicit `compose()` will be backwards compatible.)

If anybody feels like it, please contribute some tests, docs, and tweak the examples, and make a PR against this PR. Thanks for a great discussion!

**Comment by acdlite** [comment]
@gaearon Wait did we figure out the API for omitting initial state?

**Comment by gaearon** [comment]
Current API in this PR lets you do that because it only allows a single enhancer (which has to be a function). We assume top-level state object is not a function (sounds like a pretty safe assumption).

**Comment by acdlite** [comment]
Hmm that's what I thought. Not crazy about it, but don't hate it either.

**Comment by mikekidder** [comment]
Definately like @timdorr version, thanks for opening topic up for discussion Dan

**Comment by hunterc** [comment]
I think it makes the most sense to compose the middleware functions outside of `createStore` and pass the composed functions in. That way `createStore` handle middleware very much the same way it handles reducers.

**Comment by gaearon** [comment]
>Not crazy about it, but don't hate it either.

This is a sign of a good tradeoff :wink: 

With our usage we really don’t want to add churn. Removing small impediments to learning Redux or understanding Redux apps with no breakage cost goes a long way.

**Comment by gaearon** [comment]
@hunterc Definitely. The suggested API accommodates that.

**Comment by hunterc** [comment]
@gaearon Yep. I'm liking that solution. :) By the way, thanks for the great lib!

**Comment by michaeljonathan** [comment]
>  wanted to do this but then watch out:
> ```
> // Haha good luck telling array initial state from array of enhancers
> const store = createStore(
>    rootReducer,
>    [{ lol: true }, { sad: true }],
>    [applyMiddleware(thunk, logger), DevTools.instrument(), ...etc]
>  )
> ```

@gaearon But if Redux doesn't accept an array of enhancers, does it mean that we have to use `compose()` anyway for multiple store enhancers? Which prevents dropping `compose()` off the public API.

(aside)
Crazy idea:
What happens if it starts to look like this instead?
```
newStore()
.withReducer(rootReducer)
.withInitialState(initialState)
.withEnhancers([applyMidldleware(thunk, logger), DevTools.instrument()])
.create()
```

**Comment by rafaelrinaldi** [comment]
I personally dislike the idea of messing with object keys (specially if they're optional) and chaining. :+1: for @timdorr's solution. Looks classy :tophat:

**Comment by AdamMaras** [comment]
I'm sure I'm outnumbered here, but I really like the configuration object proposal, with support for both enhancer and enhancers. It may add terminology, but to me it seems more clear and verbose than positional arguments.

**Comment by tedpennings** [comment]
re making initial state mandatory: I'm :-1:  on anything that makes initial state mandatory because that seems antagonistic to `combineReducers`. 

I worked on a medium-sized Redux application [that shipped in late 2015](https://blog.newrelic.com/2015/11/12/apm-error-analytics/) that uses the `(Maybe State, Action) -> State` signature to encapsulate a lot of smaller initial states in reducers zipped together with `combineReducers`. We have 11 reducers that are combined into the root reducer. Each reducer defines its own initial state in a slightly different shape. Generally, the initial state provides defaults and empty data placeholders that render as a loading state. But the individual reducer state trees vary a bit, so I want to keep that DRY.

If initial state were mandatory, I guess I would create my store by first calling my root reducer with `undefined` and an empty action (`like { type: '@@redux/INIT' }`) to make the reducers produce their initial state. That works, but it's ugly boilerplate that smells like a design tradeoff.  If we make initial state mandatory, I think we need to add a big caveat to `combineReducers` that it produces a leaky abstraction because there isn't a useful initial state encapsulation.

**Comment by AprilArcus** [comment]
:thumbsup: to named argument objects, but if the `reducer` arg is non-optional it should probably be a positional parameter? this is the convention in python, anyway - positional arguments are mandatory, named arguments are optionals.

```
var store = createStore(reducer, {
  initialState: initialState,
  enhancer: applyMiddleware(thunk, logger)
})
```

I guess the alternative would be to throw a `TypeError` if the `reducer` key isn't found on the params object.

**Comment by timdorr** [comment]
@AprilArcus Unfortunately, that would be impossible to maintain any sort of backwards compatibility. The 2nd arg has always been initialState, which may be an object. So, we're unable to distinguish what that object represents (config or initial state). We would have to make a a very breaking change that would likely cause a lot of upgrade hassles. 

I really wish we had kwargs in JS, but that's not the world we live in :smile:

**Comment by gaearon** [comment]
>But if Redux doesn't accept an array of enhancers, does it mean that we have to use compose() anyway for multiple store enhancers? Which prevents dropping compose() off the public API.

Right. But the result is still better than what we have now. Removing `compose()` wasn’t a goal by itself, the goal was to simplify common usage which I believe the current proposal does. 

As for the fluent API proposal, I don’t think turning one function into a factory with different intermediate methods is something we want to do. Overall I am not a fan of fluent APIs.

>re making initial state mandatory: I'm :-1: on anything that makes initial state mandatory because that seems antagonistic to combineReducers.

No, this is not something we even considered. When we talked about making initial state mandatory we only meant about *having to* specify `undefined` as the second argument before getting to the new third argument. I don’t want to do this either and the current proposal works around that. This is not related to how `combineReducers()` works and we are definitely not touching that part ;-)

**Comment by yelouafi** [comment]
Maybe I'm speaking non-sens; But I think Redux doesn't need store enhancers neither compose. 

I can understand applyMiddleware composition has clear semantics: action travels a chain of middleware that can intercept/modify or let the action passthrough

But for store enhancers, this is not obvious to me: I mean the idea of a store enhancer is pretty obvious, we wrap a store in an enhanced store which adds some feature. But *composing enhancers* what's the meaning of that ? Why a store enhancer should receive an enhanced createStore from a previous store enhancer ? put another way, does *composing enhancements* has a any meaning ?

It feels like store enhancers are different from middlewares. Each store enhancer is (or seems) independent from others so all it needs is the original store and not an already enhanced one.

May be I miss something obvious here, but I think if Redux would provide the actual action to its subscribers as argument (actually subscribers are called without arg), one could achieve what he wants without store enhancers; and even better, he wouldn't have to monkey-patch the store as enhancers actually do (each enhancer has its own dispatch version which has some awkward side-effects as it creates a barrier between multiples dispatch chains of different enhancers)

**Comment by johanneslumpe** [comment]
@yelouafi If an enhancer would only receive the original store, then the last enhancer would "win" and its result would be passed to the middleware. But if you have multiple enhancers they must stack.

A middleware cannot receive the final state after the chain has been run. A store enhancer on the other hand can get a copy of the previous state and then a copy of the next state and for example diff the two (like devtools). Or it can log the dispatch duration. Afaik these cannot be solved with normal middleware because you need to be able to do something before and after the actual dispatch.

**Comment by gaearon** [comment]
>Each store enhancer is (or seems) independent from others so all it needs is the original store 

Please check out how Redux DevTools is implemented. Consider how to make it work with async middleware. I think you'll see why we came up with the current design. If store was created beforehand, I don't see how either would work.

**Comment by yelouafi** [comment]
@johanneslumpe I understand, and I think this a consequence of how store enhancers works actually (monkey-patching). But what about the other solution, if we provide subscribers with the dispatched action as argument; The we could achieve the same thing enhancers can do

```javascript
myEnhancer(store, options) {
 
 let lastState = store.getState()

  function process(oldState, newState, action) {
   ...
  }  

  store.subscribe(action => {
     const newState = store.getState()
     process(lastState, newState, action)
     lastState = newState
  })

}
```

The subscriber/enhancer above has all the necessary info to build its logic: old/new state and current action. And this hasn't the drawbacks of the actual enhancers (multiple dispatch chains) Because there is only one store version, so multiple enhancers can also inter-communicate via store.dispatch

gaearon:
Can you make a PR to DevTools repo showing how this would work? I still don't see it because DevTools enhancer needs to wrap the reducer and "lift" both getState() and dispatch().

yelouafi:
@gaearon Could make it (when I'll have the time of couse :) ). But couldn't something like this work (i mean the idea, impl. details may vary)?

```javascript
function instrument(store, options) {

 function liftedAction(action, oldState, newState) { ... }
 function liftedReducer(liftedState, liftedAction) { ... }
  
 const devToolsStore = createStore(liftedReducer)

  function process(oldState, newState, action) {
    devToolsStore.dispatch(liftedAction(action, oldState, newState))
  }  

  store.subscribe(action => {
     const newState = store.getState()
     process(lastState, newState, action)
     lastState = newState
  })

}
```

The idea of providing actions to subscirbers could also serve other meanings: can build an action rx stream and implement side effect with rx observables; ie it connect Redux to the outside world in a more natural way.

gaearon:
How would you make an existing `store` magically change its state based on what `devToolsStore` tells it?

yelouafi:
>How would you make an existing store magically change its state based on what devToolsStore tells it?

With the above solution. The simplest way I see is to have it builtin into the store api (some resetState method), I don't think this is a pb since it'll only serve for developer tools.

gaearon:
I like the current system because it doesn't require internal methods like setState() which no doubt will be abused. I don't see a big issue with how store enhancers work but I'm happy to consider specific alternatives. We spent some hair pulling figuring out the current API and while I understand things are not perfect, I don't have time to reinvent it from the ground up again. If you have specific proposals I'm happy to consider a joint RFC showing new (specific and verified to run correctly) API for composing "new" store enhancers and how that would solve real problems justifying the churn in the ecosystem that would come out of that.

yelouafi:
@gaearon I understand. All what I said was a bit of quick thinking :). Perhaps keeping the actual API and providing the action to subscribers as an alternative way (unless there are issues with that), wiil gives a chance to compare how the 2 models works in real-life

gaearon:
We can implement your "process" proposal as middleware, just like sagas today. Can't we?

yelouafi:
The issue in my case is other: I used an enhancer like method to implement runSaga which can run sagas outside the middleware env (server side rendering, code splitting ...). I didn't use compose but the result would be similar : [patch dispatch function to get the dispatched action](https://github.com/yelouafi/redux-saga/blob/master/src/runSaga.js#L18-L23). But it has some drawbacks; as in https://github.com/yelouafi/redux-saga/issues/48

>saga called with runSaga can dispatch actions to store and all sagas from middleware will see it, but actions dispatched from middleware sagas can't be taken by saga called with runSaga

Since dispatch of *inside* (passed to the middleware) is not the same as dispatch *outside* (the final that the enhancer get); the runSaga can only patch the final dispatch returned by applyMiddleware enhancer; So the intermediate dispatch (passed to redux-saga middleware) is not patched and then runSaga can't intercept those actions

I thought if store.subscribe can guarantee the delivery of each action then it wont be necessary to enhance/patch the store. Maybe there other cases where the store enhancer needs just the actual action, so it won't be necessary to complicate the store setup if they could just use subscribe

gaearon:
I don't yet understand but I feel like I'm close. Can you create a 10 line repo showing the problem?

yelouafi:
Here is a [jsbin demo](http://jsbin.com/gezizu/9/edit?js,console).

There are 2 middlewares sagas and 1 saga outside (with runSaga). the outside saga dispatch a RUNSAGA_ACTION and both middleware sagas take it;

Then the 1st middleware saga dispatch a MIDDLEWARE_SAGA action; but only the 2nd middleware saga can take it; the saga outside don't see it;

gaearon:
Demi is empty. Wrong link?

yelouafi:
>Demi is empty. Wrong link?

didn't understand, pb with the link or with or running the demo; re-tried it and it works for me

http://jsbin.com/gezizu/9/edit?js,console

and gist

 https://gist.github.com/yelouafi/baf20b6e7e2685d6f08d

yelouafi:
perhaps a lifted reducer like used in devtools will work; Is this the only way a store enhancer can get an action from the store

gaearon:
Maybe JSBin is broken on mobile. I'll check when I'm at desktop. Can you please file a separate issue? I'll be happy to discuss there as seems mostly unrelated to this thread.

yelouafi:
Ok, will create an issue. Thanks!

gaearon:
Out in 3.1.0.

PR #1299, opened 2016-01-28
Deprecate redux-immutablejs in favour of redux-immutable

gajus:
Deprecate for reasons outlined in https://github.com/gajus/redux-immutable/issues/12#issuecomment-176189454

### DIFF
--- a/docs/introduction/Ecosystem.md
+++ b/docs/introduction/Ecosystem.md
@@ -86,7 +86,7 @@ On this page we will only feature a few of them that the Redux maintainers have 
 * [redux-actions](https://github.com/acdlite/redux-actions) — Reduces the boilerplate in writing reducers and action creators
 * [redux-act](https://github.com/pauldijou/redux-act) — An opinionated library for making reducers and action creators
 * [redux-transducers](https://github.com/acdlite/redux-transducers) — Transducer utilities for Redux
-* [redux-immutablejs](https://github.com/indexiatech/redux-immutablejs) — Integration tools between Redux and [Immutable](https://github.com/facebook/immutable-js/)
+* [redux-immutable](https://github.com/gajus/redux-immutable) — Used to create an equivalent function of Redux `combineReducers` that works with [Immutable.js](https://facebook.github.io/immutable-js/) state.
 * [redux-tcomb](https://github.com/gcanti/redux-tcomb) — Immutable and type-checked state and actions for Redux
 * [redux-mock-store](https://github.com/arnaudbenard/redux-mock-store) — Mock redux store for testing your app
 


gaearon:
Any reason why you don't want to export `combineReducers` as a drop-in replacement? I understand the desire for flexibility but it would be nice to have `createCombineReducers({ debug: process.env.NODE_ENV !== 'production' })` as the default.

gajus:
@gaearon Redux implementation of `combineReducers` uses [`process.env.NODE_ENV` to enable/ disable debug mode](https://github.com/rackt/redux/blob/v3.0.6/src/combineReducers.js#L112-L117). It is only my opinion, but third-party libraries should not rely on environment variables to control their behaviour, e.g. in our deployment process we are using `production-{name of the server}` `NODE_ENV` value to make different builds. This unintentionally included "development" build of Redux in the production bundle. To avoid this, library/function behaviour must be controlled only through the public API.

I did not want to change Redux `combineReducers` API (I could have had `options` as the second parameter of the function). Therefore, a sound compromise was to have a function that would build `combineReducers` with preset "debug" behaviour.

I would advise `redux` to either take a similar approach or update `combineReducers` ARI to avoid earlier described pitfall.

gajus:
I can of course export `combineReducers` in addition to `createCombineReducers` with preset debug value (as you suggested), but that would be of little value to people since they would still need to import `createCombineReducers` to differentiate between development and production builds.

gajus:
> I would advise redux to either take a similar approach or update combineReducers API to avoid earlier described pitfall.

If you see this as a worthy discussion, I will open a GitHub issue.

gaearon:
>@gaearon Redux implementation of combineReducers uses process.env.NODE_ENV to enable/ disable debug mode. It is only my opinion, but third-party libraries should not rely on environment variables to control their behaviour, e.g. in our deployment process we are using production-{name of the server} NODE_ENV value to make different builds. 

I agree in principle but this is an established convention in the React ecosystem. Redux just tries to be a good citizen here. If you're not envifying React, you'll have quite slow builds, so all we do is follow the same convention as React, and if you want to skip that, you can just use pre-envified UMD builds. I think libraries complementary to Redux (at least those we officially recommend) should stick to the same conventions as Redux.

gaearon:
>in our deployment process we are using production-{name of the server} NODE_ENV value to make different builds.

Since React and Redux and React-Redux and a ton of other libraries in React ecosystem already use NODE_ENV for that, maybe you could just use a different env variable name?

gajus:
Published [`redux-immutable@3.0.0`](https://github.com/gajus/redux-immutable/tree/v3.0.0) implementing the discussed changes.

gaearon:
Thanks for cooperation :+1: 
I know that existing conventions often aren't the best but it's better to avoid fragmenting community over this.

gajus:
Just working for the better tomorrow. : )

PR #1301, opened 2016-01-28
Add browserify config

e-jigsaw:
Replace `process.env.NODE_ENV` by adding this config when using browserify.

React also written this config: https://github.com/facebook/react/blob/0ce28c6e88448af6eda69df66088b88e7248bbb7/packages/react/package.json#L29-L33

Thanks

[Review on Reviewable](https://reviewable.io/reviews/rackt/redux/1301)

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,5 +81,10 @@
         "*.js"
       ]
     }
-  ]
+  ],
+  "browserify": {
+    "transform": [
+      "loose-envify"
+    ]
+  }
 }


gaearon:
Don't we also need to make it a dependency then?

gaearon:
Also if we go with that, we need to add this to React Redux as well.

e-jigsaw:
@gaearon 

> Don't we also need to make it a dependency then?

No, Just `"browserify": "..."`

> Also if we go with that, we need to add this to React Redux as well.

Yes, I will send same PR to react-redux if this PR merged

timdorr:
LGTM and shouldn't hurt anything to add it. :+1:

gaearon:
Thanks

e-jigsaw:
@gaearon Thanks, I sent same PR to react-redux.

gaearon:
Out in 3.1.0.

gaearon:
FYI #1304

PR #1302, opened 2016-01-28
Update applyMiddleware enhancer to use new createStore signature

acdlite:
3.1 changed the signature of `createStore()`. This isn't considered a breaking change because existing enhancers will continue to work; however, going forward, it should be expected that a store enhancer returns a `createStore` function with the same signature as its input.

See https://twitter.com/acdlite/status/692772885186764802 for context

Still needs tests :)

### DIFF
--- a/src/applyMiddleware.js
+++ b/src/applyMiddleware.js
@@ -17,8 +17,8 @@ import compose from './compose'
  * @returns {Function} A store enhancer applying the middleware.
  */
 export default function applyMiddleware(...middlewares) {
-  return (createStore) => (reducer, initialState) => {
-    var store = createStore(reducer, initialState)
+  return (createStore) => (reducer, initialState, enhancer) => {
+    var store = createStore(reducer, initialState, enhancer)
     var dispatch = store.dispatch
     var chain = []
 


gaearon:
Maybe just `...rest`? Now that enhancers ever have to deal with other enhancers.

acdlite:
Yeah true. Although let's ask the question: should the enhancer passed to `createStore` be applied *before* or *after* the enhancer in question (`applyMiddleware` in this case)?

I think this is right (before) but... hmm.

gaearon:
We don't ever want to encourage mixing both patterns. We'll transition docs to only use the new pattern, and the fact that the old one works will become an implementation detail.

acdlite:
Even if we don't encourage it, I still think it's important as an implementation detail. At the least, it will be important for enhancer authors.

Problem in a nutshell:

```js
const finalCreateStore = DevTools.instrument()(createStore)
const store = finalCreateStore(
  reducer,
  initialState,
  applyMiddleware(thunk, logger)
)
```

Which gets executed first: the DevTools, or the middleware?

gaearon:
In this case, middleware will wrap the previously wrapped-with-DevTools function.
Specifying `enhancer` means specifying "the last wrapper".
Seems right to me.

(But really, nobody should do it)

gaearon:
I trust that this works but please feel free to add tests. I'll release it as a patch bump now.

gaearon:
Out in 3.1.1.

acdlite:
@gaearon I don't think that's right. `applyMiddleware()` gets applied first because it's passed all the way through to the base `createStore()`.

gaearon:
Oh, right. Lol this is counter-intuitive. But then again, it's not a breaking change since it's a new way of doing things. I don't think people will mess this up in their apps but we'll see. Even if they do mess up, we'll just offer to use the new way.

acdlite:
Alright as long as we're aware of it :)

PR #1305, opened 2016-01-28
Add loose-envify dep

dereklieu:
Here's the missing dependency in `package.json`.


timhudson:
This addresses issue https://github.com/rackt/redux/issues/1304

gaearon:
Is it enough to make it a dev dependency? They are only installed when you run npm install in the project folder. I feel like this should be a regular dep.

timdorr:
Yeah, it needs to be a regular dep. Check #1306.

PR #1306, opened 2016-01-28
Add loose-envify as a dep for Browserify users.

### LINKED ISSUE #1304: Missing loose-envify dependency
Hey, thanks for the awesome work on redux.

Just installed 3.1.1. `package.json` calls for a [`loose-envify`](https://github.com/rackt/redux/blob/master/package.json#L87) transform, but it's not listed as a dependency anywhere and so is breaking my build.

timdorr:
Fixes #1304

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,9 @@
     "url": "https://github.com/rackt/redux/issues"
   },
   "homepage": "http://rackt.github.io/redux",
+  "dependencies": {
+    "loose-envify": "^1.1.0"
+  },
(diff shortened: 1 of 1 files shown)


gaearon:
Can we verify this works?

timdorr:
If someone can pull off of master, they can confirm that way. But I don't see any reason why it wouldn't. Invariant does the same thing: https://github.com/zertosh/invariant/blob/master/package.json

gaearon:
:+1:

PR #1307, opened 2016-01-28
Fix unknown action type handling

msafi:
As discussed in #1054, and if I understand correctly, the private action handling test is not really possible if the user has access `ActionTypes.INIT`. 

So, I changed the way `combineReducers` inits reducers by making it do a random probe and `undefined` state checks simultaneously.

I modified the tests to make sure `combineReducers` implementation throws if the given reducer could potentially return `undefined` either for an `undefined` state or an unknown action type.

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -57,21 +57,9 @@ function assertReducerSanity(reducers) {
     if (typeof initialState === 'undefined') {
       throw new Error(
         `Reducer "${key}" returned undefined during initialization. ` +
-        `If the state passed to the reducer is undefined, you must ` +
-        `explicitly return the initial state. The initial state may ` +
-        `not be undefined.`
-      )
-    }
-
-    var type = '@@redux/PROBE_UNKNOWN_ACTION_' + Math.random().toString(36).substring(7).split('').join('.')
-    if (typeof reducer(undefined, { type }) === 'undefined') {
-      throw new Error(
-        `Reducer "${key}" returned undefined when probed with a random type. ` +
-        `Don't try to handle ${ActionTypes.INIT} or other actions in "redux/*" ` +
-        `namespace. They are considered private. Instead, you must return the ` +
-        `current state for any unknown actions, unless it is undefined, ` +
-        `in which case you must return the initial state, regardless of the ` +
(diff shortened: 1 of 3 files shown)


gaearon:
Hey, sorry for not getting back to this sooner. This change still kinda eludes me sometimes. Can you clarify:

1. What would be the meaning of this? What are the scenarios it protects against, and why is it better than what we have right now?
2. Is this a breaking change? Can it possibly break *any* projects?
3. Does it have any impact on Redux DevTools or related tooling? Can you try searching `@@redux/INIT` in GitHub projects to check that?

Thanks.

msafi:
Hey Dan,

1. So, in the current code, Redux does two checks separately. First, it checks that if [`undefined` state](https://github.com/msafi/redux/blob/f3af5458600a70f487406000676a49adc309a737/src/combineReducers.js) is passed, the reducer knows to return a value. Second, it checks that if [unknown action type](https://github.com/reactjs/redux/blob/63fe8503c1a04487f78284d1abddfd87f537f1e7/src/combineReducers.js#L64) is passed, the reducer knows to return a value. My change combines these checks, makes the error message clearer, and simplifies the code. It doesn't protect against anything.
2. I don't see how it could be a breaking change. It's the same functionality, only simplified.
3. Regarding Redux DevTools, yes, [it would break it](https://github.com/gaearon/redux-devtools/blob/b030896048fd2c236fe4ea8ae50286905ca8983c/src/instrument.js#L264)! And possibly other projects? :cry: To solve this, I could restore [`PROBE_UNKNOWN_ACTION`](https://github.com/reactjs/redux/blob/63fe8503c1a04487f78284d1abddfd87f537f1e7/src/combineReducers.js#L63), and have `combineReducers` use it to assert reducer sanity.

All in all, this isn't an important change. I just made it as a quick first contribution and it is turning out to be more confusing than it's worth, so feel free to close it! :smile:

timdorr:
> To solve this, I could restore PROBE_UNKNOWN_ACTION, and have combineReducers use it to assert reducer sanity.

This sounds like the more sane solution to me. Having a separate probe action from the core init action avoids problems when trying to overload its purpose like this.

PR #1313, opened 2016-01-29
Minor :: use cross-env; npm script support

mikekidder:
This PR just adds support to set ENV variables when using npm scripts, that supports Windows


gaearon:
This seems great. Can we use the same solution for Webpack builds? Then we can unify two configs into one.

mikekidder:
Yes, was thinking the same.  Happened across a tweet that mentioned it 

https://twitter.com/_ericelliott/status/692057061073534976

PR #1319, opened 2016-01-30
throw exception if nextReducer is not a function

neebz:
Fix for: https://github.com/rackt/redux/issues/1318

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -159,6 +159,11 @@ export default function createStore(reducer, initialState, enhancer) {
    * @returns {void}
    */
   function replaceReducer(nextReducer) {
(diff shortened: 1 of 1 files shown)


neebz:
@svenanders thanks  - I was about to do that but there is just one `describe` block of `createStore` - should the `replaceReducer` test go inside the same describe or should we create separate ?

gaearon:
It should be fine in the same block.

gaearon:
Thank you for the effort. I went ahead and merged #1321 — it had a test, followed the existing style, and didn’t have conflicts. Keep an eye on the repo, we may have more low-hanging fruit later :-)

PR #1320, opened 2016-01-30
replaceReducer() should throw if argument is not a function fix #1318

### LINKED ISSUE #1318: replaceReducer() should throw if argument is not a function
This is an easy one, please feel free to send a PR.
We want to add a `typeof` check in `replaceReducer` just like we have when the store is created.

Andersos:
Fix for: #1318

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -159,8 +159,12 @@ export default function createStore(reducer, initialState, enhancer) {
    * @returns {void}
    */
   function replaceReducer(nextReducer) {
-    currentReducer = nextReducer
-    dispatch({ type: ActionTypes.INIT })
+    if (typeof nextReducer === 'function') {
+      currentReducer = nextReducer
+      dispatch({ type: ActionTypes.INIT })
+    } else {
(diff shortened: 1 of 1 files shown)


src/createStore.js
```diff
@@ -159,8 +159,12 @@ export default function createStore(reducer, initialState, enhancer) {
    * @returns {void}
    */
   function replaceReducer(nextReducer) {
-    currentReducer = nextReducer
-    dispatch({ type: ActionTypes.INIT })
+    if (typeof nextReducer === 'function') {
```
gaearon: Nitpicking: let's make it an early exit, like in the beginning of `createStore`.

```js
if (...) {
  throw ...
}

...
```

gaearon:
Thank you for the effort. I went ahead and merged #1321 — it had a test, followed the existing style, and didn’t have conflicts. Keep an eye on the repo, we may have more low-hanging fruit later :-)

Andersos:
Maybe next time =)

PR #1325, opened 2016-01-30
subscribe to throw when listener is not a function

JamieDixon:
Fixes: https://github.com/rackt/redux/issues/1324

### DIFF
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -548,4 +548,12 @@ describe('createStore', () => {
       store.replaceReducer(() => {})
     ).toNotThrow()
   })
+
+  it('throws if listener is not a function', () => {
+    const store = createStore(reducers.todos)
+
+    expect(() =>
+      store.subscribe()
+    ).toThrow();
+  })
 })
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -567,6 +567,5 @@ describe('createStore', () => {
     expect(() =>
       store.subscribe(undefined)
     ).toThrow()
-
   })
 })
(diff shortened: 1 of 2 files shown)


test/createStore.spec.js
```diff
@@ -548,4 +548,12 @@ describe('createStore', () => {
       store.replaceReducer(() => {})
     ).toNotThrow()
   })
+
+  it('throws if listener is not a function', () => {
+    const store = createStore(reducers.todos)
+
+    expect(() =>
+      store.subscribe()
+    ).toThrow()
```
gaearon: Let's check a few more data types: null, undefined, and string should throw.

JamieDixon: Added a few extra checks for the types mentioned. I've gone with the assertion style I've seen elsewhere rather than something like:

```
const cases = [undefined, null, ''];
    cases.map(arg =>
      expect(() =>
        store.subscribe(arg))
      .toThrow())
```

test/createStore.spec.js
```diff
@@ -548,4 +548,25 @@ describe('createStore', () => {
...
+      store.subscribe('')
+    ).toThrow()
+
+    expect(() =>
+      store.subscribe(null)
+    ).toThrow()
+
+    expect(() =>
+      store.subscribe(undefined)
+    ).toThrow()
```
gaearon: Nit: please remove the extra newline

JamieDixon: +1 removed in latest commit.

gaearon:
Out in 3.1.6, thanks!

Issue #1042, opened 2015-11-16
jsnext:main should point to a distributable ES6 bundle


dchambers:
It's not currently spelt out in the wiki page, but [jsnext:main](https://github.com/rollup/rollup/wiki/jsnext:main) is apparently meant to point to a distributable ES6 bundle rather than to the source code. For example, see:

> Unfortunately, Redux's jsnext:main points to its source code, not a distributable ES6 bundle. Because of that, the build process for this repo needs to transpile Redux's source code. Redux uses Babel 5, this repo uses Babel 6, causing the build to fail. These kinds of incompatibilities are why jsnext:main should not be used to point to source code.

taken from [this comment](https://github.com/BladeRunnerJS/bundle-all-spike/issues/1#issuecomment-156475479).

gaearon:
Hmm. What is a “distributable ES6 bundle”? Something that doesn’t have `import` and `export` statements?

danmartinez101:
@dchambers So basically what we need is to run babel on src to most of the transforms, but exclude the module transforms so that import and export remain in the transpiled `src` directory?

dchambers:
> Hmm. What is a “distributable ES6 bundle”? Something that doesn’t have import and export statements?

Good question! I did wonder myself, but assumed it was only me that didn't quite get it :-). @Rich-Harris, can you clarify please?

danmartinez101:
Other than that possibility, it could either mean that it needs to be bundled in to a single file (which I think defeats the purpose of Rollup, so I imagine it is not that case) or that the es7 features used in the codebase are the blocker. There isn't extensive usage of es7 but the object spread operation would probably be enough to make it "not es6 distributable" 

I'm sure Rich-Harris can provide a solid answer, though.

Rich-Harris:
Sure, I'll try and clarify – but with the caveat that this is *my* interpretation! Unfortunately as a community we haven't quite figured out a consensus on how to transition to ES6 modules – there's lots of room for different views on how exactly to go about it. There's a thread over here though if you'd like to participate --> https://github.com/jsforum/jsforum/issues/5

[...]

PR #1327, opened 2016-01-30
Drop broken jsnext:main

taion:
Closes https://github.com/rackt/redux/issues/1042

Per https://github.com/rackt/redux/issues/1042#issuecomment-164312922, it looks like there's not much benefit in adding a proper ES2015 module build, and the current `jsnext:main` pointing at untranspiled code doesn't accomplish anything other than breaking rollup users.

As I recall, there are remain other reasons for continuing to distribute `src/` (Flow-related?), so I didn't drop that from `files`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,6 @@
   "version": "3.1.5",
   "description": "Predictable state container for JavaScript apps",
   "main": "lib/index.js",
-  "jsnext:main": "src/index.js",
   "files": [
     "dist",
     "lib",


gaearon:
Shall we do something like https://github.com/rackt/react-redux/pull/256/files? I'm open to distributing a proper ES6 module build.

taion:
Ah – I didn't see that, and thought #1042 was lingering. Yes, that's a better way to go if you want to distribute an ES2015 build. Easier with Babel 5, too. I'll close this out for now, then.

taion:
BTW, this is all we have to do with Babel 5: https://github.com/rackt/react-router/pull/2530.

PR #1331, opened 2016-01-31
[RFC] Cache bound action creators with WeakMap

gaearon:
This is an experimental optimization. I do see some speedups on Chrome and Firefox when rendering many components (which happen to call `bindActionCreators`) but I'm not sure whether this optimization is worth it, and whether it has downsides on other browsers/engines. **I’m not very knowledgeable in JS performance so I hope somebody else can share their perspective on this.**

In my testing this improves performance when you create many connected components of the same type. For example, when a list view adds many rows. It is wasteful to bind the same action creator to the same `dispatch` function over and over again.

Here, we use `WeakMap` if available to cache the bound action creators. Using `WeakMap` allows us to prevent leaks on the server where you'd have a separate store instance on every request.

Does this make sense?

### DIFF
--- a/src/bindActionCreators.js
+++ b/src/bindActionCreators.js
@@ -2,6 +2,40 @@ function bindActionCreator(actionCreator, dispatch) {
   return (...args) => dispatch(actionCreator(...args))
 }
 
+var caches
+var resolveBoundActionCreator
+
+if (
+  typeof WeakMap === 'function' &&
+  Object.prototype.toString.call(new WeakMap()) === '[object WeakMap]'
+) {
+  caches = new WeakMap()
+  resolveBoundActionCreator = bindActionCreatorWithCache
+} else {
+  resolveBoundActionCreator = bindActionCreator
+}
+
+function getCache(dispatch) {
+  if (!caches) {
+    return
+  }
+  var cache = caches.get(dispatch)
+  if (!cache) {
+    cache = new WeakMap()
+    caches.set(dispatch, cache)
+  }
+  return cache
+}
+
+function bindActionCreatorWithCache(actionCreator, dispatch, cache) {
+  var boundActionCreator = cache.get(actionCreator)
+  if (!boundActionCreator) {
+    boundActionCreator = bindActionCreator(actionCreator, dispatch)
+    cache.set(actionCreator, boundActionCreator)
+  }
+  return boundActionCreator
+}
+
 /**
  * Turns an object whose values are action creators, into an object with the
  * same keys, but with every function wrapped into a `dispatch` call so they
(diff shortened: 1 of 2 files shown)


sompylasar:
:+1: makes sense!

timdorr:
Shouldn't this be a perf improvement on react-redux instead? If components repeatedly binding action creators, it should maintain an internal cache. 

It seems like a lot of complexity to add here, which makes the code less comprehensible. It's also for a specific use of `bindActionCreators`, so it seems weird to try attack this lower level. It would be like trying to get a Redux-specific perf improvement into Lodash.

gaearon:
>It's also for a specific use of bindActionCreators

What other use case is there?

gaearon:
Unfortunately the only case we can handle well in React Redux is when the user specifies AC as an object. But I'm inclined to agree with you that it is too low level to fix it here.

Additionally, while this improves perf on Chrome and Firefox, I see equivalent slowdown in Safari.

timdorr:
More so, I really think the complexity increase is an important issue. Redux should remain at its core very simple, easy to grok, and generic in implementation. 

Did you do that testing over a few thousand iterations and with a warmup period? I'd honestly expect v8, Spidermonkey, et al. to identify our code as a hot area and swap it out for native, considering it gets executed so often.

sompylasar:
@timdorr This looks more like a memory optimization, not CPU.

timdorr:
@sompylasar The `WeakMap` is useful to keep the cache from growing unbounded and causing a memory leak. But it's mainly CPU because we keep the bound action creator cache amongst all calls of the function, so we shouldn't have to reallocate new anonymous functions for each binding. True, there is some memory benefit as well, but the process of allocating those function instances is going to be more of the issue.

gaearon:
Closing as too prescriptive and not a good tradeoff.

>More so, I really think the complexity increase is an important issue. Redux should remain at its core very simple, easy to grok, and generic in implementation.

I’d say if we can make Redux more performant without changing public API by writing code in an optimized way I think it’s worth the increase in implementation complexity. For example I see React pooling objects where possible and avoiding allocations as great internal optimizations that are worth the less obvious code. Not that we have many chances to do that though.

sompylasar:
@timdorr The allocations happen once per connect, right? So, there is no repetitive CPU-intensive work, while memory for all these equal functions is retained. What am I missing? Thanks.

gajus:
In terms of optimising `mapDispatchToProps`, I am using [`mapActionsToProps`](https://github.com/gajus/redux-composer/blob/7c54f7fb05609d2a14a2511efa2fb7d821c61ead/src/app/mapActionsToProps.js#L1-L16) function to construct all action creators and bind them to my store. Then I can simply use [`mapActionsToProps` in place of `mapDispatchToProps`](https://github.com/gajus/redux-composer/blob/7c54f7fb05609d2a14a2511efa2fb7d821c61ead/src/app/containers/SandboxContainer.js#L18-L19). This allows me to [call whatever action](https://github.com/gajus/redux-composer/blob/7c54f7fb05609d2a14a2511efa2fb7d821c61ead/src/app/containers/SandboxContainer.js) within the smart components.

Does anyone see a drawback to this approach?

PR #1334, opened 2016-01-31
Add todos example described in Usage with React

urbanvikingr:


gaearon:
This is now in via d7b54ee34a109bdc62ebe43d19095b63f52bd168.

PR #1339, opened 2016-02-01
Use lodash v4.

jdalton:
Attempt 2 of https://github.com/rackt/redux/pull/611.


webpack.config.js
```diff
@@ -14,10 +14,18 @@ var config = {
     libraryTarget: 'umd'
   },
   plugins : [
+    {
+      apply: function apply(compiler) {
+        compiler.parser.plugin('expression global', function expressionGlobalPlugin() {
+          this.state.module.addVariable('global', "(function() { return this; }()) || Function('return this')()")
```
gaearon: I'm curious how much going to great lengths like Lodash does is buying us. Should we settle for this, or should we do a more sophisticated check like https://github.com/rackt/redux/pull/611#issuecomment-177679332?

Also, are we breaking anyone with this? Are there environments where `Object.getPrototypeOf` (as we had before) would “just work” but `Function('return this')()"` would fail? I'm thinking stuff like Firefox extensions etc.

jdalton: You stated that double compiling is an edge case. One that webpack clearly doesn't support. This is going above and beyond to support them in the primary case. If they are double compiling and using their code in extensions then yes it could be an issue.

Lodash's UMD is erm robust because folks do some crazy stuff and we have to support it. I think your case is fine though.

gaearon:
Thanks for following through with this!

jdalton:
Np. Future versions of the lodash modules may remove the global reference too.

jdalton:
The next release will remove the `global` references. See https://github.com/lodash/lodash/blob/npm/isPlainObject.js.

phated:
@jdalton when will that version be released, can it be a 4.1.1 that is released soon?

jdalton:
Moved discussion to #1346.

PR #1340, opened 2016-02-01
When not to use Redux

wordofchristian:
[This article ](http://www.code-experience.com/getting-out-of-javascript-fatigue/) raises some good points.

>After the initial Flux pattern (which was so simple, it was boilerplate and a Dispatcher module) an enormous amount of Flux libraries emerged. This seemed to stress out people. Unnecessarily, in my opinion: If a huge number of libraries appear for a problem, it means that the problem is not hard, so why worry about it?

> Besides, people forgot to mention that there is a good chance you don't need Flux at all. Too bad it's (afaik) not mentioned in the otherwise well-curated docs of Redux on which the community has settled for now.

This is my first crack at some guidelines for newcomers who are struggling to decide whether or not they need Redux.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -3,6 +3,7 @@
 * [Read Me](/README.md)
 * [Introduction](/docs/introduction/README.md)
   * [Motivation](/docs/introduction/Motivation.md)
+  * [When Not To Use](/docs/introduction/WhenNotToUse.md)
   * [Three Principles](/docs/introduction/ThreePrinciples.md)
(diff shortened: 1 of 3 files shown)


gaearon:
This is a great start. Can you rewrite it to be in the style of the rest of the documentation rather than linking / quoting tweets?

gaearon:
Closing for inactivity. Please let us know if you’d like to pick it up again and address the above comment, and we can reopen this. Cheers!

PR #1341, opened 2016-02-01
Lowercase private actions

askmatey:
A popular convention in JavaScript is to only capitalize constructor functions or ES6 classes.

The `actionTypes` variable is a **bare object** — and should therefore be lowercased.

In some cases this might be acceptable

```js
import * as ActionTypes from '../actions'
```

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -1,4 +1,4 @@
-import { ActionTypes } from './createStore'
+import { actionTypes } from './createStore'
 import isPlainObject from 'lodash/isPlainObject'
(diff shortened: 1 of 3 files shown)


gaearon:
I think it’s fairly common to also use pascal case for constant enumerations.

In any case, `ActionTypes` is not a part of the public API though. It is only there for our internal tests. We might have made it part of public API accidentally when we told people `require('redux/lib/*')` is now supported usage, but we'll probably remove it from `createStore` exports in the next major for this reason.

PR #1345, opened 2016-02-01
Warn When dispatching During Middleware Setup

### LINKED ISSUE #1240: Dispatching in a middleware before applyMiddleware completes
Preface: This may or may not be a bug in Redux.

Say you have a middleware that intercepts actions and emits async events to dispatch later on. We do something like this in `redux-simple-router`:

``` js
function middleware(store) {
  history.listen(location => { store.dispatch(updateLocation(location)) })

  return next => action => {
      if (action.type !== TRANSITION) {
        return next(action)
      }
      const { method, arg } = action
      history[method](arg)
    }
}
```

The key bit is the history listener at the top. The middleware intercepts actions of type `TRANSITION` and runs a history command (`push`, for example). That comes back to us via the listener and we dispatch an `updateLocation` action to update the store with the current location. 

Now, let's say you have a little bit of analytics middleware that watches for updateLocation events and tells your analytics service there's a new page view. This works A-OK because `store.dispatch` above has had the middleware applied to it and will include the analytics middleware in the dispatch.

Here's the problem: That listener fires when you're applying the middleware, so calling `store.dispatch` is [calling the un-middlewared `dispatch`](https://github.com/reactjs/redux/blob/c5d9a8a11897ea558c348e02eaf76038563d57b9/src/applyMiddleware.js#L27) and our analytics middleware won't see any event.

Should Redux handle this? It might be better handled by the history middleware, or by the analytics middleware, or by a non-middleware approach. But it certainly does end up creating some confusion for users when code operates differently the first time you run it.

timdorr:
Closes #1240

Any thoughts on the wording that was chosen? I assume we want to just warn and not throw a hard error here.

### DIFF
--- a/docs/advanced/Middleware.md
+++ b/docs/advanced/Middleware.md
@@ -268,11 +268,15 @@ The implementation of [`applyMiddleware()`](../api/applyMiddleware.md) that ship
 
 * It only exposes a subset of the [store API](../api/Store.md) to the middleware: [`dispatch(action)`](../api/Store.md#dispatch) and [`getState()`](../api/Store.md#getState).
 
-* It does a bit of trickery to make sure that if you call `store.dispatch(action)` from your middleware instead of `next(action)`, the action will actually travel the whole middleware chain again, including the current middleware. This is useful for asynchronous middleware, as we have seen [previously](AsyncActions.md).
+* It does a bit of trickery to make sure that if you call `store.dispatch(action)` from your middleware instead of `next(action)`, the action will actually travel the whole middleware chain again, including the current middleware. This is useful for asynchronous middleware, as we have seen [previously](AsyncActions.md). There is one caveat when calling `dispatch` during setup, described below.
 
 * To ensure that you may only apply middleware once, it operates on `createStore()` rather than on `store` itself. Instead of `(store, middlewares) => store`, its signature is `(...middlewares) => (createStore) => createStore`.
 
 Because it is cumbersome to apply functions to `createStore()` before using it, `createStore()` accepts an optional last argument to specify such functions.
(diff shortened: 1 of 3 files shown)


gaearon:
Can we absolutely not make it work? If we apply middleware from the innermost to the outermost, it seems like we could gradually update the outer `dispatch` reference as we fold them, can we not?

timdorr:
@yelouafi has a potential option here: https://github.com/rackt/redux/issues/1240#issuecomment-178140708

I'm not married to this PR, so I'd be more than happy to close it if we can find a solution. This just exists to make our job easier if we don't find one.

gaearon:
I guess what’s happening here is I don’t understand the current semantics without a test ;-)

gaearon:
To clarify, I see that you added a warning, and you test for it. But I want to understand *what actually happens* if you have several middlewares that attempt to dispatch actions in definition phase.

timdorr:
OK, I can work up something more elaborate.

gaearon:
I think we should just throw and call it a day. Warning will be frustrating because user can't control it, and middleware author will have to remove it anyway due to complaints.

gaearon:
I’ll close because I’d like to see a hard error here, not a warning, and the PR has been inactive for a while. If you update this please reopen; otherwise somebody else might give it a shot if they find the time. Thanks!

timdorr:
Sorry, I was meaning to get back to this but keep forgetting. I'll try to get it clean up in the next couple days.

gaearon:
Thanks, no trouble. I’m just cleaning up non-actionable PRs so new contributors aren’t confused.

PR #1347, opened 2016-02-02
Use Babel 6

gaearon:
This builds on top of #1123. It also adds a `prepublish.js` script that calls `es3ify` because Babel 6 currently has a bug preventing ES3ification plugins from working. In addition, instead of relying on a preset, we embrace the config to have precise control over the plugins we are using, and their options.


SimenB:
@gaearon ES3 transform is fixed now: babel/babel#3368

PR #1349, opened 2016-02-02
Update to lodash v4.2.0.

### LINKED ISSUE #1346: 3.2.0 Global not defined with in isPlainObject
screen shot 2016-02-01 at 4 55 12 pm

I'm building our app files with browserify, babel 6.4, envify, react-redux, react 0.14 and the bundles are throwing this error that global is not defined. I take it envify or loose envify is not factoring out the globals?

For now I have reverted to 3.1.7 of redux and 4.1.2 of react-redux but of course it would be nice to see a real fix.

jdalton:
This closes #1346.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
   },
   "homepage": "http://rackt.github.io/redux",
   "dependencies": {
-    "lodash": "^4.1.0",
+    "lodash": "^4.2.0",
     "loose-envify": "^1.1.0"
   },
   "devDependencies": {
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -14,14 +14,6 @@ var config = {
     libraryTarget: 'umd'
   },
   plugins: [
-    {
-      apply: function apply(compiler) {
-        compiler.parser.plugin('expression global', function expressionGlobalPlugin() {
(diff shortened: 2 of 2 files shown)


stoeffel:
How about including [`lodash.isPlainObject`](https://www.npmjs.com/package/lodash.isplainobject) instead of the complete lodash?

gaearon:
@stoeffel It doesn't really matter because we import just that module.

stoeffel:
Agreed.
We're not using redux directly, but we use react-dnd. 
We depend on react-dnd with a fixed dependency, but react-dnd depends on dnd-core (not fixed) and dnd-core depends on redux (not fixed).
Our webpack was configured to alias lodash to one global lodash, I guess the intention behind that was to reduce client-size. This lead to a failing build because we are still using lodash v3, which wouldn't have happened if redux would use `lodash.isPlainObject` or the dependencies in dnd-core and react-dnd was fixed.
But IMO we need to change that anyway, because other libs will probably update to lodash v4 soon.

gaearon:
Yes, that's a big assumption to make.
Wanna send a PR to update Lodash in DnD?

stoeffel:
> Wanna send a PR to update Lodash in DnD?

Will try to find some time to create a PR, but maybe someone will be faster.

PR #1356, opened 2016-02-03
Update counter examples' "Increment if odd" to increment on negative odd numbers too.

nelsonjchen:


gaearon:
Fair enough. Can you also fix the regular `counter` example as well as all occurrences, if any, of this phrase in the docs (`md` files)?

nelsonjchen:
Ah, I'm a newbie whose just perusing the documentation to understand Redux. But I'll give a quick look through the repo.

nelsonjchen:
Updated PR. I also added a test for that in the fix for counter. As far as the rest of the repo, I can't find anything else with `/%.*2.*===.*1/`.

gaearon:
Appreciated!

PR #1357, opened 2016-02-03
use isEmpty variable in asnyc example

Aweary:
use a variable `isEmpty` instead of accessing `posts.length` twice. Makes it more readable, and short enough for the whole expression to exist on one line.


gaearon:
:+1:

PR #1362, opened 2016-02-04
RFC: syncHistoryWithStore()

gaearon:
As discussed in rackt/react-router-redux#257, some features of `react-router-redux` such as dedicated action creators do not seem necessary with the API changes in `react-router@2.0`. In addition, it seems to me that middleware is suboptiomal API for the DevTools replay functionality.

This pull request is an attempt at starting a discussion around what `react-router-redux` could look like if instead of using middleware, we wrapped the `history` object itself. To me, this seems a more natural approach to this problem, and seems to fix some issues with the current DevTools support such as rackt/react-router-redux#252.

In this PR we:

* Remove dependency on `react-router-redux` and use `react-router` as is
* Demonstrate how we can preserve record/replay/time travel features while using vanilla `history` API
* Implement a function that is similar in spirit to the original `redux-simple-router` in terms of its simple API

If there is interest on `react-router-redux` side and there are no apparent problems with this approach, we may want to use that as its evolution. Otherwise we may publish something like this independently. I have not considered server rendering here, so please let me know if I am way off, and this approach is broken in some way.

Finally, let me acknowledge that I took 80% of the logic from `react-router-redux`. My only contribution here is slightly tweaking the API to emphasize vanilla `react-router` usage and fixing some edge cases in DevTools support.

I will link to this PR on `react-router-redux` as well. Big thanks to @jlongster, @taion, @timdorr, and @naw for their work and help with this.

Related issues:

* https://github.com/rackt/react-router-redux/issues/138
* https://github.com/rackt/react-router-redux/issues/248
* https://github.com/rackt/react-router-redux/issues/252
* https://github.com/rackt/react-router-redux/pull/255
* https://github.com/rackt/react-router-redux/issues/257

### DIFF
--- a/examples/real-world/index.js
+++ b/examples/real-world/index.js
@@ -1,12 +1,17 @@
 import 'babel-polyfill'
 import React from 'react'
 import { render } from 'react-dom'
+import { browserHistory } from 'react-router'
 import Root from './containers/Root'
 import configureStore from './store/configureStore'
+import { syncHistoryWithStore } from './syncHistoryWithStore'
 
 const store = configureStore()
+const history = syncHistoryWithStore(browserHistory, store, {
+  adjustUrlOnReplay: true
+})
 
 render(
-  <Root store={store} />,
+  <Root store={store} history={history} />,
   document.getElementById('root')
 )
--- a/examples/real-world/reducers/index.js
+++ b/examples/real-world/reducers/index.js
@@ -1,7 +1,7 @@
 import * as ActionTypes from '../actions'
 import merge from 'lodash/merge'
 import paginate from './paginate'
-import { routeReducer } from 'react-router-redux'
+import { reducer as routing } from '../syncHistoryWithStore'
 import { combineReducers } from 'redux'
 
 // Updates an entity cache in response to any action with response.entities.
@@ -50,8 +50,7 @@ const rootReducer = combineReducers({
   entities,
   pagination,
   errorMessage,
-  routing: routeReducer
+  routing
 })
-
 
 export default rootReducer
--- a/examples/real-world/syncHistoryWithStore.js
+++ b/examples/real-world/syncHistoryWithStore.js
@@ -87,9 +87,23 @@ export function syncHistoryWithStore(history, store, {
   return Object.assign({}, history, {
     // The listeners are subscribed to the store instead of history
     listen(listener) {
-      return store.subscribe(() =>
-        listener(getLocationInStore(true))
-      )
+      // History listeners expect a synchronous call
+      listener(getLocationInStore(true))
+
+      // Keep track of whether we unsubscribed, as Redux store
+      // only applies changes in subscriptions on next dispatch
+      let unsubscribed = false
+      const unsubscribeFromStore = store.subscribe(() => {
+        if (!unsubscribed) {
+          listener(getLocationInStore(true))
+        }
+      })
+
+      // Let user unsubscribe later
+      return () => {
+        unsubscribed = true
+        unsubscribeFromStore()
+      }
     },
 
     // It also provides a way to destroy internal listeners
--- a/examples/real-world/syncHistoryWithStore.js
+++ b/examples/real-world/syncHistoryWithStore.js
@@ -1,0 +1,103 @@
+export const WILL_NAVIGATE = '@@react-router/WILL_NAVIGATE'
+
+const initialState = {
+  locationBeforeTransitions: null
+}
+
+// Mount this reducer to handle location changes
+export const reducer = (state = initialState, action) => {
+  switch (action.type) {
+    case WILL_NAVIGATE:
+      // Use a descriptive name to make it less tempting to reach into it
+      return {
+        locationBeforeTransitions: action.locationBeforeTransitions
+      }
+    default:
+      return state
+  }
+}
+
+export function syncHistoryWithStore(history, store, {
+  // Specify where you mounted the reducer
+  selectLocationState = state => state.routing,
+  adjustUrlOnReplay = false
+} = {}) {
+  let initialLocation
+  let currentLocation
+  let isTimeTraveling
+  let unsubscribeFromStore
+  let unsubscribeFromHistory
+
+  // What does the store say about current location?
+  const getLocationInStore = (useInitialIfEmpty) => {
+    const locationState = selectLocationState(store.getState())
+    return locationState.locationBeforeTransitions ||
+      (useInitialIfEmpty ? initialLocation : undefined)
+  }
+
+  // Whenever store changes due to time travel, keep address bar in sync
+  const handleStoreChange = () => {
+    const locationInStore = getLocationInStore(true)
+    if (currentLocation === locationInStore) {
+      return
+    }
+
+    // Update address bar to reflect store state
+    isTimeTraveling = true
+    currentLocation = locationInStore
+    history.replace(locationInStore)
+    isTimeTraveling = false
+  }
+
+  if (adjustUrlOnReplay) {
+    unsubscribeFromStore = store.subscribe(handleStoreChange)
+    handleStoreChange()
+  }
+
+  // Whenever location changes, dispatch an action to get it in the store
+  const handleLocationChange = (location) => {
+    // ... unless we just caused that location change
+    if (isTimeTraveling) {
+      return
+    }
+
+    // Remember where we are
+    currentLocation = location
+
+    // Are we being called for the first time?
+    if (!initialLocation) {
+      // Remember as a fallback in case state is reset
+      initialLocation = location
+
+      // Respect persisted location, if any
+      if (getLocationInStore()) {
+        return
+      }
+    }
+
+    // Tell the store to update by dispatching an action
+    store.dispatch({
+      type: WILL_NAVIGATE,
+      locationBeforeTransitions: location
+    })
+  }
+  unsubscribeFromHistory = history.listen(handleLocationChange)
+
+  // The enhanced history uses store as source of truth
+  return Object.assign({}, history, {
+    // The listeners are subscribed to the store instead of history
+    listen(listener) {
+      return store.subscribe(() =>
+        listener(getLocationInStore(true))
+      )
+    },
+
+    // It also provides a way to destroy internal listeners
+    dispose() {
+      if (adjustUrlOnReplay) {
+        unsubscribeFromStore()
+      }
+      unsubscribeFromHistory()
+    }
+  })
+}
--- a/examples/real-world/syncHistoryWithStore.js
+++ b/examples/real-world/syncHistoryWithStore.js
@@ -22,6 +22,17 @@ export function syncHistoryWithStore(history, store, {
   selectLocationState = state => state.routing,
   adjustUrlOnReplay = false
 } = {}) {
+  // Fail early if the reducer is not mounted
+  if (typeof selectLocationState(store.getState()) === 'undefined') {
+    throw new Error(
+      'Expected the routing state to be available either as `state.routing` ' +
+      'or as the custom expression you can specify as `selectLocationState` ' +
+      'named argument in the `syncHistoryWithStore()` options. Did you ' +
+      'forget to put the `reducer` exported from `syncHistoryWithStore` into ' +
+      'your `combineReducers()` call?'
+    )
+  }
+
   let initialLocation
   let currentLocation
   let isTimeTraveling
--- a/examples/real-world/syncHistoryWithStore.js
+++ b/examples/real-world/syncHistoryWithStore.js
@@ -56,7 +56,10 @@ export function syncHistoryWithStore(history, store, {
     // Update address bar to reflect store state
     isTimeTraveling = true
     currentLocation = locationInStore
-    history.replace(locationInStore)
+    history.transitionTo(Object.assign({},
+      locationInStore,
+      { action: 'PUSH' }
+    ))
     isTimeTraveling = false
   }
 
--- a/examples/real-world/containers/App.js
+++ b/examples/real-world/containers/App.js
@@ -1,6 +1,6 @@
 import React, { Component, PropTypes } from 'react'
 import { connect } from 'react-redux'
-import { push } from 'react-router-redux'
+import { browserHistory } from 'react-router'
 import Explore from '../components/Explore'
 import { resetErrorMessage } from '../actions'
 
@@ -17,7 +17,7 @@ class App extends Component {
   }
 
   handleChange(nextValue) {
-    this.props.push(`/${nextValue}`)
+    browserHistory.push(`/${nextValue}`)
   }
 
   renderErrorMessage() {
@@ -56,20 +56,18 @@ App.propTypes = {
   // Injected by React Redux
   errorMessage: PropTypes.string,
   resetErrorMessage: PropTypes.func.isRequired,
-  push: PropTypes.func.isRequired,
   inputValue: PropTypes.string.isRequired,
   // Injected by React Router
   children: PropTypes.node
 }
 
-function mapStateToProps(state) {
+function mapStateToProps(state, ownProps) {
   return {
     errorMessage: state.errorMessage,
-    inputValue: state.routing.location.pathname.substring(1)
+    inputValue: ownProps.location.pathname.substring(1)
   }
 }
 
 export default connect(mapStateToProps, {
-  resetErrorMessage,
-  push
+  resetErrorMessage
 })(App)
--- a/examples/real-world/containers/RepoPage.js
+++ b/examples/real-world/containers/RepoPage.js
@@ -72,8 +72,8 @@ RepoPage.propTypes = {
   loadStargazers: PropTypes.func.isRequired
 }
 
-function mapStateToProps(state, props) {
-  const { login, name } = props.params
+function mapStateToProps(state, ownProps) {
+  const { login, name } = ownProps.params
   const {
     pagination: { stargazersByRepo },
     entities: { users, repos }
--- a/examples/real-world/containers/Root.dev.js
+++ b/examples/real-world/containers/Root.dev.js
@@ -2,15 +2,15 @@ import React, { Component, PropTypes } from 'react'
 import { Provider } from 'react-redux'
 import routes from '../routes'
 import DevTools from './DevTools'
-import { Router, browserHistory } from 'react-router'
+import { Router } from 'react-router'
 
 export default class Root extends Component {
   render() {
-    const { store } = this.props
+    const { store, history } = this.props
     return (
       <Provider store={store}>
         <div>
-          <Router history={browserHistory} routes={routes} />
+          <Router history={history} routes={routes} />
           <DevTools />
         </div>
       </Provider>
--- a/examples/real-world/containers/Root.prod.js
+++ b/examples/real-world/containers/Root.prod.js
@@ -1,14 +1,14 @@
 import React, { Component, PropTypes } from 'react'
 import { Provider } from 'react-redux'
 import routes from '../routes'
-import { Router, browserHistory } from 'react-router'
+import { Router } from 'react-router'
 
 export default class Root extends Component {
   render() {
-    const { store } = this.props
+    const { store, history } = this.props
     return (
       <Provider store={store}>
-        <Router history={browserHistory} routes={routes} />
+        <Router history={history} routes={routes} />
       </Provider>
     )
   }
--- a/examples/real-world/containers/UserPage.js
+++ b/examples/real-world/containers/UserPage.js
@@ -72,8 +72,8 @@ UserPage.propTypes = {
   loadStarred: PropTypes.func.isRequired
 }
 
-function mapStateToProps(state, props) {
-  const { login } = props.params
+function mapStateToProps(state, ownProps) {
+  const { login } = ownProps.params
   const {
     pagination: { starredByUser },
     entities: { users, repos }
--- a/examples/real-world/package.json
+++ b/examples/real-world/package.json
@@ -24,7 +24,6 @@
     "react-dom": "^0.14.7",
     "react-redux": "^4.2.1",
     "react-router": "2.0.0-rc5",
-    "react-router-redux": "^2.1.0",
     "redux": "^3.2.1",
     "redux-logger": "^2.4.0",
     "redux-thunk": "^1.0.3"
--- a/examples/real-world/store/configureStore.dev.js
+++ b/examples/real-world/store/configureStore.dev.js
@@ -1,26 +1,19 @@
 import { createStore, applyMiddleware, compose } from 'redux'
-import { syncHistory } from 'react-router-redux'
-import { browserHistory } from 'react-router'
-import DevTools from '../containers/DevTools'
 import thunk from 'redux-thunk'
-import api from '../middleware/api'
 import createLogger from 'redux-logger'
+import api from '../middleware/api'
 import rootReducer from '../reducers'
-
-const reduxRouterMiddleware = syncHistory(browserHistory)
+import DevTools from '../containers/DevTools'
 
 export default function configureStore(initialState) {
   const store = createStore(
     rootReducer,
     initialState,
     compose(
-      applyMiddleware(thunk, api, reduxRouterMiddleware, createLogger()),
+      applyMiddleware(thunk, api, createLogger()),
       DevTools.instrument()
     )
   )
-
-  // Required for replaying actions from devtools to work
-  reduxRouterMiddleware.listenForReplays(store)
 
   if (module.hot) {
     // Enable Webpack hot module replacement for reducers
--- a/examples/real-world/store/configureStore.prod.js
+++ b/examples/real-world/store/configureStore.prod.js
@@ -1,6 +1,4 @@
 import { createStore, applyMiddleware } from 'redux'
-import { syncHistory } from 'react-router-redux'
-import { browserHistory } from 'react-router'
 import thunk from 'redux-thunk'
 import api from '../middleware/api'
 import rootReducer from '../reducers'
@@ -9,6 +7,6 @@ export default function configureStore(initialState) {
   return createStore(
     rootReducer,
     initialState,
-    applyMiddleware(thunk, api, syncHistory(browserHistory))
+    applyMiddleware(thunk, api)
   )
 }


timdorr:
https://github.com/rackt/redux/blob/routing-history/examples/real-world/syncHistoryWithStore.js

(Just posting so I have quick access to take a look at the important file :smile:)

gaearon:
>(Just posting so I have quick access to take a look at the important file :smile:)

Also notice how the usage changed. There is no middleware or store enhancer, it works as a single function call:

```js
const store = configureStore()
const history = syncHistoryWithStore(browserHistory, store, {
  adjustUrlOnReplay: true
})
```

You still *do* need to mount the reducer but you can use the regular history rather than the special action creators.

examples/real-world/syncHistoryWithStore.js
```diff
@@ -0,0 +1,103 @@
...
+      }
+    default:
+      return state
+  }
+}
+
+export function syncHistoryWithStore(history, store, {
+  // Specify where you mounted the reducer
+  selectLocationState = state => state.routing,
+  adjustUrlOnReplay = false
```
timdorr: Since there are no action creators exposed and the state key is named to discourage access, it would seem the primary purpose is to sync up the URL bar after a replay. In that case, should this default to true?

timdorr:
> You still do need to mount the reducer but you can use the regular history rather than the special action creators.

Why not make this a store enhancer so the user doesn't have to do that?

examples/real-world/syncHistoryWithStore.js
gaearon: Defaulting to true means that if you hydrate state from e.g. `localStorage` URL will adjust to that initial state. It can be desirable in development and in production "replay" scenarios but it might be surprising to people who just store some Redux state in `localStorage`. This is why I'd rather see it opt-in.

timdorr:
Ah wait, hit enter too soon. We're not returning the store, so we can't enhance it. Hmm...

gaearon:
>Why not make this a store enhancer so the user doesn't have to do that?

How would we inject a reducer into an existing reducer combination? Also some people want to use Immutable so it's not very good to assume a specific way of putting it there. And wrapping all app's state just because we want to "own" `routing` key is not very good either. I think this is the lesser evil here.

examples/real-world/syncHistoryWithStore.js
```diff
@@ -0,0 +1,103 @@
...
+    })
+  }
+  unsubscribeFromHistory = history.listen(handleLocationChange)
+
+  // The enhanced history uses store as source of truth
+  return Object.assign({}, history, {
+    // The listeners are subscribed to the store instead of history
+    listen(listener) {
+      return store.subscribe(() =>
+        listener(getLocationInStore(true))
```
gaearon: TODO (optimization): check if location has changed and don't call listener if it hasn't.
Should be added after tests are in place.

examples/real-world/syncHistoryWithStore.js
```diff
@@ -0,0 +1,103 @@
...
+      }
+    }
+
+    // Tell the store to update by dispatching an action
+    store.dispatch({
+      type: WILL_NAVIGATE,
+      locationBeforeTransitions: location
+    })
+  }
+  unsubscribeFromHistory = history.listen(handleLocationChange)
```
taion: Normally history enhancers don't call `listen` until after someone calls `listen` on the wrapped history, but I think this is okay in this case.

examples/real-world/syncHistoryWithStore.js
```diff
@@ -0,0 +1,103 @@
...
+  return Object.assign({}, history, {
+    // The listeners are subscribed to the store instead of history
+    listen(listener) {
+      return store.subscribe(() =>
+        listener(getLocationInStore(true))
+      )
+    },
+
+    // It also provides a way to destroy internal listeners
+    dispose() {
```
gaearon: TODO (design): can we agree on a standard way of disposing resources required by histories, if any? Then we could call underlying `history.dispose()` in case it exists.

examples/real-world/syncHistoryWithStore.js
```diff
@@ -0,0 +1,103 @@
...
+      // Use a descriptive name to make it less tempting to reach into it
+      return {
+        locationBeforeTransitions: action.locationBeforeTransitions
+      }
+    default:
+      return state
+  }
+}
+
+export function syncHistoryWithStore(history, store, {
```
gaearon: This is when I get to say “thank goodness histories are not classes, I told ya!”

examples/real-world/syncHistoryWithStore.js
```diff
@@ -0,0 +1,103 @@
...
+  const handleStoreChange = () => {
+    const locationInStore = getLocationInStore(true)
+    if (currentLocation === locationInStore) {
+      return
+    }
+
+    // Update address bar to reflect store state
+    isTimeTraveling = true
+    currentLocation = locationInStore
+    history.replace(locationInStore)
```
taion: Comparable code in react-router-redux is https://github.com/rackt/react-router-redux/blob/b6a9c1bb51521bf6f5f416de394658d586683d9f/src/index.js#L80-L100

This is a little bit subtle – if you call `history.replace` there, `history` will actually emit a new `location` that then won't match what the reducer saw. One option is to use `location.key` and `transitionTo`, but then you end up with https://github.com/rackt/react-router-redux/issues/164 if you're not careful (and I don't understand the fix there).

examples/real-world/index.js
```diff
@@ -1,12 +1,17 @@
...
 import React from 'react'
 import { render } from 'react-dom'
+import { browserHistory } from 'react-router'
 import Root from './containers/Root'
 import configureStore from './store/configureStore'
+import { syncHistoryWithStore } from './syncHistoryWithStore'
 
 const store = configureStore()
+const history = syncHistoryWithStore(browserHistory, store, {
+  adjustUrlOnReplay: true
```
gaearon: Note: if you pass `false` **DevTools history time travel will still work in UI**, it just won't update the address bar.

examples/real-world/syncHistoryWithStore.js
```diff
@@ -0,0 +1,103 @@
...
+      locationBeforeTransitions: location
+    })
+  }
+  unsubscribeFromHistory = history.listen(handleLocationChange)
+
+  // The enhanced history uses store as source of truth
+  return Object.assign({}, history, {
+    // The listeners are subscribed to the store instead of history
+    listen(listener) {
+      return store.subscribe(() =>
```
taion: I'm not sure what the guarantees with `store.subscribe` are, but `history.listen` is expected to synchronously invoke the listener with the current `location`.

examples/real-world/syncHistoryWithStore.js
timdorr: You can detect that, because the initial location on load is a "POP" action.

examples/real-world/syncHistoryWithStore.js
gaearon: How do I reproduce the issue? I tried pressing Back and then toggling that action in DevTools, and it worked for me.

examples/real-world/syncHistoryWithStore.js
gaearon: Ah, spot on!

timdorr:
> How would we inject a reducer into an existing reducer combination?

The only way I can think of is the same way `applyMiddleware` does it, by wrapping dispatch. Not that we should actually do that, of course. 

I'm just looking out for the person that forgets one part of the install process and misses a key reducer. Then when they click on stuff nothing happens. Doesn't really fail gracefully.

examples/real-world/syncHistoryWithStore.js
taion: On react-router-redux? It looks like that was fixed with https://github.com/rackt/react-router-redux/issues/164 having been closed. More than that I don't know.

What you will see, though, is that e.g. if you use hash history, the location key in the URL then won't match the location key in the Redux store.

examples/real-world/syncHistoryWithStore.js
taion: The standard way is that we have this cruft in like a half dozen different places by now: https://github.com/rackt/history/blob/v2.0.0-rc3/modules/createBrowserHistory.js#L104-L132.

This would be easier if histories were classes, but they're not :stuck_out_tongue:

examples/real-world/syncHistoryWithStore.js
gaearon: Should be fixed in e271f55a471939ff0bea341e22c837914ca8b995.

taion:
Could you make this a store enhancer that just overwrites `history.listen` in-place? Mutating `history` isn't the cleanest thing in the world, but it might work.

examples/real-world/syncHistoryWithStore.js
timdorr: The initial POP location can't be applied back to the browser location, so reseting the state in Dev Tools doesn't reset the URL bar. Replacing was a quick solution to achieve that.

Perhaps a `go(-history.length)` would be better?

gaearon:
>I'm just looking out for the person that forgets one part of the install process and misses a key reducer. Then when they click on stuff nothing happens. Doesn't really fail gracefully.

Fixed in 0db894c91617329d965b444314cf9c4ce916a7ed.

gaearon:
>Could you make this a store enhancer that just overwrites history.listen in-place? Mutating history isn't the cleanest thing in the world, but it might work.

What would be the point of making it a store enhancer? What problem would it solve?

timdorr:
Less wiring up by the user is the main purpose. We can also better hide the location state.

examples/real-world/syncHistoryWithStore.js
taion: Why does that not break for toggling browser-level back actions? Those are `POP`s too.

taion:
@gaearon My https://github.com/rackt/redux/pull/1362#issuecomment-179589961 was in response to @timdorr's https://github.com/rackt/redux/pull/1362#issuecomment-179584321.

examples/real-world/syncHistoryWithStore.js
timdorr: Because we short circuit: https://github.com/rackt/history/blob/master/modules/createBrowserHistory.js#L70

taion:
Better than that though would be a

```js
const { history, storeEnhancer } = wrapHistory(history);
```

or something.

gaearon:
>>How would we inject a reducer into an existing reducer combination?
>The only way I can think of is the same way applyMiddleware does it, by wrapping dispatch. 

`applyMiddleware()` does not inject reducers into existing reducer combination so I’m confused as to what you mean exactly. Once you have a root reducer, you can’t “crack open it”. It’s a function, and even store enhancers can’t just put their stuff there.

DevTools *wraps* the app root reducer and provides a slice of the state to the app, if that is what you mean. I tried this approach today too. My first attempt *was* a store enhancer. However this meant that DevTools themselves would see the app state as `{ app, routing }` where `app` is your actual root state. This is unfortunate because in the code you would still refer to things as `state.something` whereas in the DevTools you would see them as `state.app.something`. This is why I dropped this approach. Now I don’t see anything a store enhancer could buy us that a history enhancer doesn’t already buy.

As much as I like store enhancers, experimenting with this problem today made me fairly confident they are an inferior solution to a history enhancer in this particular case.

gaearon:
>Less wiring up by the user is the main purpose. We can also better hide the location state.

Can you send a PR against this PR? I don’t understand how we can hide it.

examples/real-world/syncHistoryWithStore.js
taion: Right, but that short-circuiting should also come into play when you toggle on any `POP` action, such as those created by using the browser's back button, no?

timdorr:
> I’m confused as to what you mean exactly.

I was only postulating on an idea. No idea how it would actually work. Feel free to file this line of reasoning under the "I dunno. It sounds like a good idea" category :smile:

gaearon:
Ah, okay :smile: . I tried this today but I realized my mistake as I described in https://github.com/rackt/redux/pull/1362#issuecomment-179591549.

examples/real-world/syncHistoryWithStore.js
gaearon: How would I tell between the first POP action and a persisted session that happens to end on a POP?

examples/real-world/syncHistoryWithStore.js
timdorr: We listen to `onpopstate` to make it work, I believe: https://github.com/rackt/history/blob/master/modules/createBrowserHistory.js#L60

examples/real-world/syncHistoryWithStore.js
gaearon: Yeah, we're purposefully eavesdropping here. Can make this lazy if necessary although then we have a risk of missing initial actions (I think).

examples/real-world/syncHistoryWithStore.js
timdorr: So do we need to mutate the initialState's location in some way so going back to it programmatically isn't an issue?

taion: Okay, so looking at that short-circuiting logic – this suggests that when toggling a `POP` action, we won't update the browser URL, with the current react-router-redux implementation.

It's probably not a huge issue here, to be honest.

gaearon: Haha I’m totally lost on what you folks are talking about. If any of those problems are relevant for my implementation here, can you please provide steps to reproduce them inside this PR? I’d be very grateful!

taion: - If you toggle actions, Redux and the browser (the upstream history, really) will disagree about `location.key`; if you e.g. refresh the page, then you may not see the same `location.state`, if the state has been mutated for that key
  - If you use the deprecated `history.setState` API, something weird will happen
  - Also, for hash history, the `?_k=` query key in the URL won't match `location.key`

timdorr: @taion Yes, that was the case until I switched to a replace. 

@gaearon The downside of the replace is it doesn't rewind the browser history to match. Go click on some links, hit reset in dev tools, and then notice you're able to hit back in your browser still. Don't know if that's a real issue, or just a quirk to live with.

taion: Let's discuss the current react-router-redux issues separately; it's just noise here.

gaearon: @timdorr I think once you're toggling browser actions in DevTools with `adjustUrlOnReplay = true` you should not expect Back button to function correctly. I don't really see a big issue with this.

valscion:
This RFC really seems to be *just* what I wanted `react-router-redux` / `redux-simple-router` be. We are actually just using that library for tying up to `history` — not interacting with react router anymore as it was no longer necessary. We still love to have the ability to use the same library for that, too.

We have a piece of state living in the URL of the app currently, and our reducer uses the `UPDATE_LOCATION` action and the `action.location.query` given by `react-router-redux` to update domain specific filter ID array (in reducer):

```js
case UPDATE_LOCATION:
  const { location: { query } } = action;
  newActiveVenueIds = Array.from((query && query.activeVenueIds) || []).map(Number);
  break;

// and later

if (newActiveVenueIds.length > 0) {
  newFilteredEventRequests = newEventRequests.filter(
    (eventRequest) => newActiveVenueIds.indexOf(eventRequest.venue_id) >= 0
  );
} else {
  newFilteredEventRequests = newEventRequests;
}

// later

return {
  eventRequests: newEventRequests,
  eventRequestsByVenue: newEventRequestsByVenue,
  activeVenueIds: newActiveVenueIds,
  filteredEventRequests: newFilteredEventRequests
};
```

We trigger history changes from relevant action creators using thunk middleware while only modifying the relevant query parameter and keeping the others intact. We don't want to care about any other query parameter besides the `activeVenueIds` there is.

```js
// Action creators
import { routeActions } from 'redux-simple-router';

function queryForRemoveFilterByVenue({venueId, query}) {
  const { activeVenueIds } = query;
  const index = activeVenueIds.indexOf(venueId);
  const newActiveVenueIds = [
    ...activeVenueIds.slice(0, index),
    ...activeVenueIds.slice(index + 1)
  ];
  const newQuery = {...query, activeVenueIds: newActiveVenueIds};
  if (newActiveVenueIds.length === 0) {
    // Have to delete the key completely to remove from URL
    delete newQuery.activeVenueIds;
  }
  return newQuery;
}

function queryForAddFilterByVenue({venueId, query}) {
  const newActiveVenueIds = [
    ...(query.activeVenueIds || []), venueId
  ];
  return {...query, activeVenueIds: newActiveVenueIds};
}

export function toggleFilterByVenue(venueId) {
  venueId = String(venueId); // eslint-disable-line no-param-reassign
  return (dispatch, getState) => {
    const { routing: { location } } = getState();
    const query = location.query || {};
    const activeVenueIds = query.activeVenueIds || [];
    const newQuery =
      activeVenueIds.indexOf(venueId) !== -1 ?
        queryForRemoveFilterByVenue({venueId, query}) :
        queryForAddFilterByVenue({venueId, query});
    dispatch(routeActions.push({pathname: location.pathname, query: newQuery}));
  };
}
```

I figured that we really don't need to dispatch a `routeActions.push` in there if we could just use the singleton history object. A `history.push` would be equivalent for us.

I'd really like to store the `activeVenueIds` state _only_ in the URL. That way we have only one place where that data resides.

> If there is interest on react-router-redux side and there are no apparent problems with this approach, we may want to use that as its evolution. Otherwise we may publish something like this independently

Either of these approaches would be great for me. I don't fancy building a history reducer and action dispatcher myself that much, given how difficult e.g. devtools integration will be.

With this RFC, all I need to do is configure the history to be a singleton available on client side and import that on the action creator side:

```js
import { createHistory, useQueries } from 'history';
import { parse as parseQueryString, stringify as stringifyQuery } from 'qs';

let history;
if (typeof document !== 'undefined') {
  history = useQueries(createHistory)({
    parseQueryString,
    stringifyQuery
  });
}

export default history;
```

We have separate builds of the same modules for server rendering via `react-rails`, so that's why I had to wrap the `useQueries` call inside a browser availability check.

Testing against that `history` singleton will need module level stubbing with something like [`babel-plugin-rewire`](https://github.com/speedskater/babel-plugin-rewire), though, as singletons are quite nasty to stub...

examples/real-world/syncHistoryWithStore.js
taion: @gaearon Just to finish up this discussion, can you try doing

```
history.transitionTo({ ...locationInStore, action: 'PUSH' })
```

for this line? I believe it will be more correct and will avoid the odd edge cases I mentioned above.

(alternatively, `action: 'REPLACE'` to replace the current history stack frame instead of pushing a new one)

gaearon:
I agree action creators can be useful. However I don’t think they need to be coupled or necessary for the replay functionality. I see `middleware + history action` and `reducer + syncing history` as two separate concerns which, no doubt, can be included in a single library, but don’t have to be coupled together.

taion:
I think I'm pretty happy with leaving action creators out of scope now that we have singleton histories.

I've discussed a _few_ edge cases where singleton histories aren't ideal with @ryanflorence, but I really don't care that much about them – in practice just using the singleton should be fine.

examples/real-world/syncHistoryWithStore.js
gaearon: This seems to preserve the key. I don’t see whether it impacts the behavior in any important way, but thanks for the tip. Fixed by 2e42dfa96c2fb0397a51bd68323d9c37783fe804.

taion: It just avoids the edge cases I mentioned in https://github.com/rackt/redux/pull/1362#discussion_r51826397. There will still be a mismatch in `action`, but that's much less risky. Thanks!

gaearon:
cc @ezekielchentnik as this is conceptually similar to his [redux-history](https://github.com/ezekielchentnik/redux-history) project but slightly more ambitious (DevTools work + store is treated as primary source of truth rather than a copy). Maybe we can publish this as evolution of `redux-history`? There doesn’t seem to be anything specific to React Router here, and maybe it doesn’t make sense to call this `react-router-redux`. I know nobody is happy with the churn, so maybe we can see which approach wins in the long term, and avoid breaking `react-router-redux` users.

timdorr:
On the naming part: React Router 2.0 hides the History project a little better. You're still talking to history objects, but they're provided via the Router's APIs. And there's no peerDep on history. So, it's visibility is going away and I think most people, from a "brand" perspective, are going to associate this with React Router vs History.

I would be interested in taking react-router-redux in this direction. We can split it into two exports: The action creator/middleware part, and the reducer/history sync part. Import whatever part you need. If we can get some code contributions from @ezekielchentnik or @taion or you, that would be fantastic. I'm just a steward of the project, so it's whatever the community makes of it.

taion:
redux-simple-router has never actually dealt with React Router, though.

react-router-redux doesn't _strictly_ make sense as a name, but it's probably the least confusing name for most users.

timdorr:
No, but the brand association is there. It's purely a naming thing, now that history is fading more into the background.

examples/real-world/syncHistoryWithStore.js
timdorr: I'm confused. Why would it be surprising that restoring your serialized state, which syncs to history, would restore your URL bar? Wouldn't that be an expected behavior?

gaearon:
I’m happy to contribute to react-router-redux but I heard concerns from @jlongster about API churn in a (not very good) initial PR I sent a couple of days ago. I don’t want to impose my opinions on this project. However I truly believe its current architecture doesn’t fit Redux as well as this proposal, so this is how I started thinking about maybe starting another library. If react-router-redux maintainers are willing to change its API again, I’m happy to find some time to contribute this back within the next couple of weeks.

examples/real-world/reducers/index.js
```diff
@@ -1,7 +1,7 @@
 import * as ActionTypes from '../actions'
 import merge from 'lodash/merge'
 import paginate from './paginate'
-import { routeReducer } from 'react-router-redux'
+import { reducer as routing } from '../syncHistoryWithStore'
```
benmonro: why not just name the thing 'routing' in syncHistoryWithStore?  I do this exact same thing in my code and my hunch is that other people do as well.  :smile:

timdorr:
I don't consider the API churn as big of an issue as @jlongster might. \*shrug\* It would be different if we were just moving things around for the sake of moving them around, but this is a big philosophical shift for the library. And I think it's a good one that educates users about the APIs they should be using, rather than hopping onto a cargo cult library that it's going to turn into with Router 2.0. I would like to make it a history syncing library first and "oh yeah, hey, it also does this action creator stuff, but don't worry too much about that because Router has some great new APIs!" It's really Router and History 2.0 that is driving a lot of this. We should be keeping up with them, rather than just applying patches to translate old patterns to new APIs.

I'm going to get started on something like that tonight to get the ball rolling on discussion and improvements. This has already been a great move forward for the library and I really appreciate you taking the time to look into this with such detail, @gaearon!

jlongster:
I'm definitely ok with another API change, I'd rather that than starting yet another project. My initial implementation was super, super simple and only aimed to solve the serialization problem. In the side project that I used it in, I only used react-router's APIs. As the project gained popularity, a lot of bugs were exposed *mainly* because of people trying to use APIs *outside* of react-router (the react-router-redux APIs to read and control location). I've been supportive of refactorings that supported those use-cases more, but I've never loved the slow movement towards "make redux drive everything".

This is much more in line with the roots of the project. I'm happy to see it. I think we'll get some pushback from users, but that shouldn't stop us from doing what we think is the best. Careful thought towards how we communicate this would be good.

I'm in the middle of a bunch of stuff at work, and my wife is about to have a baby any moment now, so I can't look too closely at this. But from what I see I like it, and I defer many of the details to you all. I'll try to take a closer look over the next few days if I can.

gaearon:
@jlongster 

Thank you for taking your time to respond! We’ll definitely think about communicating this correctly and treading lightly. Moving action creators and middleware to a separate export from history enhancer might be the best idea in terms of casing less churn overall while advancing the project.

jlongster:
@gaearon "separate export"? I hope you don't mean separate project. There are already a lot of small moving pieces in the react/redux ecosystem. If you mean still provide them in the library as something that can be used, that sounds good, and we can deprecate over a release or two and remove them eventually.

timdorr:
@jlongster I'm thinking something like `import { syncHistoryToStore, routerMiddleware } from 'react-router-redux'`

gaearon:
>"separate export"? I hope you don't mean separate project.

I was thinking something like

```js
import { push, replace, middleware } from 'react-router-redux/actions'
```

and

```js
import { reducer, keepStateInStore } from 'react-router-redux/sync'
```

but maybe it doesn’t really make sense in ES6 land.

What I’m trying to communicate is the “action creators + middleware to interpret them” and the “history enhancer + reducer” are two separate parts of the project that can be used independently. You don’t have to use them both.

jlongster:
I see, great. That sounds much better than cutting users off from the way they currently do things in one release.

taion:
Do we really need the actions when we can do `browserHistory.push`? I can come up with examples where they might be necessary, but I'm not convinced we need them in most cases.

gaearon:
>Do we really need the actions when we can do browserHistory.push? I can come up with examples where they might be necessary, but I'm not convinced we need them in most cases.

I agree they are not necessary. However, as I gathered from https://github.com/rackt/react-router-redux/issues/257, people do rely on them for several things:

* performing history actions on non-singleton history object when data fetching without passing it from the context
* asserting that correct actions have been dispatched—easy to test return values of action creators
* logging them as they go through other middleware

So I wouldn't want to completely remove them—only to de-emphasize them and explain that they are not necessary. As we figure out pain points and address them in React Router eventually we may deprecate them, but we need to make sure we’re not leaving the users out in the cold for scenarios they care about and that aren’t easy to codemod.

jlongster:
Hmmmm that's a good point. Is there nothing indication the internal routing state changes in your new approach? Does that mean given some state and some actions, we can no longer reliably replay those actions and generate the same state all the time?

timdorr:
@jlongster There's [a dispatched action](https://github.com/rackt/redux/blob/routing-history/examples/real-world/syncHistoryWithStore.js#L93-L96) and the type is exported as well, so you can look for it in your middleware.

gaearon:
The way it works now, **if you use `syncHistoryWithStore(history, store)` there is both state and actions, the user just doesn’t trigger them:

```js
function syncHistoryWithStore(history, store) {
  // ...

  history.listen(location => {
    // ...
    store.dispatch({ type: NAVIGATE, location })
  })

  // ...
}
```

This is how this PR allows users to interact with `history` object and still get full record/replay.

----------

In this case, providing `{ push, pop }` would be just a convenience but they wouldn’t be actual actions. They would just be instructions to the (optional!) `middleware` that would interpret them and call `history` methods for the user. At this point, *if the user is using `syncHistoryWithStore()`*, they will trigger real actions and cause store updates.

jlongster:
@gaearon Nice, thanks for explaining. I guess I just need to look at this closer which I'll do soon.

gaearon:
This is why I’m stressing that these are two complementary but separate things.

(Not in this PR:) `react-redux-router/actions` exports `push`, `pop`, etc, and `middleware(history)`. All that middleware would do is to convert `dispatch(push(stuff))` into `history.push(stuff)`. This solves the problem of dependency injecting `history` and making action creators testable without any connection to record/replay functionality. In fact those “actions” would never even reach the reducers.

`react-redux-router/sync` is when you want Redux to store your routing state. It exports `reducer` and `syncHistoryWithStore(history, store)`. It does not know anything about the any of the stuff in the first paragraph. `syncHistoryWithStore` will return you an object that behaves like `history` (has same methods) **but is backed by Redux `store`**. For example, `enhancedHistory.listen()` actually becomes `store.subscribe()`. Under the hood, it subscribes to the passed `history`, but rather than call the subscribers, it dispatches an action. **Essentially it “routes” history updates through Redux so that the store becomes a source of truth for anything interacting with such `enhancedHistory` really interacts with the store.** DevTools compatibility and record/replay naturally fall out of this, with some manual mode to correct the URL bar. Of course in this case you would need to mount the provided reducer into your tree.

These two things can work together but I want to highlight there is no shared code or dependencies between them. You can use one, other, or both.

timdorr:
Opened up rackt/react-router-redux#259 to build this out on `react-router-redux`

ezekielchentnik:
@gaearon @timdorr I'm down to contribute and collaborate any way I can.  From my perspective, I've always seen history functionality completely independent of react-router (although, likely to be used together).  From experience, I've worked on projects that don't need full featured react-router, but could still take advantage of stuffing history/location into the store.

> What I’m trying to communicate is the “action creators + middleware to interpret them” and the “history enhancer + reducer” are two separate parts of the project that can be used independently. You don’t have to use them both.

I agree completely with this sentiment.  Allowing a user to piece things together in a simple way can go along way for adoption and account for multiple uses (what if react-router is no longer defacto? what if you only need a query param for a fetch?  what if you're not using react?).  

I've got some code laying around where I did split subscribing to store, and listening to history so they could be used independently without it being awkward, but had trouble deciding on the api naming ...  I'll try to get a PR/gist to continue discussion.  Cheers.

gaearon:
This is now superseded by https://github.com/rackt/react-router-redux/pull/259 so let’s continue the discussion and contributions there.

Thank you to everyone, and looking forward to react-router-redux@4.0.

PR #1363, opened 2016-02-04
RFC related: Alternative Routing History Sync With history.go

mjrussell:
Regarding https://github.com/rackt/redux/pull/1362#discussion_r51826397

To my understanding, by doing a `history.replace` you are essentially causing another history transition which will create a new key (used by `rackt/history`). This key ties your state and could be modified and not returned correctly on a refresh. This it definitely an edge case but I was curious if it was possible to return the user to the exact "state" that they were at the time of the action. So I used `history.go` while keeping a history index of where the browser was at the time of the location.

One thing that does **not work** in my attempt is going forward, this is due to an extra action being dispatched on the `go` (despite the `isTimeTraveling = true`). You'll notice that the key is identical to the state that it tried to reset to.

### DIFF
--- a/examples/real-world/syncHistoryWithStore.js
+++ b/examples/real-world/syncHistoryWithStore.js
@@ -1,7 +1,8 @@
 export const WILL_NAVIGATE = '@@react-router/WILL_NAVIGATE'
 
 const initialState = {
-  locationBeforeTransitions: null
+  locationBeforeTransitions: null,
+  historyIndex: 0
 }
 
 // Mount this reducer to handle location changes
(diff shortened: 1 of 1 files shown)


mjrussell:
Pinging @taion @gaearon @timdorr for their thoughts. Might be a little out of left field, but I thought it was worth looking into

timdorr:
I was thinking something like this would work. Does it function well in testing?

mjrussell:
The main issue right now is that it dispatches an extra redux action on the history.go(). I gave up for the night but will look into it again tomorrow.

gaearon:
I'm curious: do we care about actually moving user in history? I think that if the user enables "adjust URL" option, the current "always replace" behavior so fine. The only constraint we care about is that actions, when replayed, should change URL in the correct way.

Is this about preserving `state`? Does it get lost if you don't `go` to the correct entry?

taion:
I don't think this isn't quite right – can't you arbitrarily manipulate the sequence of played actions in DevTools? I think we'd be better off not preserving history this way.

mjrussell:
I think https://github.com/rackt/redux/commit/2e42dfa96c2fb0397a51bd68323d9c37783fe804 achieves where I was trying to go, closing this!

gaearon:
Thanks to @taion for the tip!

PR #1372, opened 2016-02-06
Use lodash-es in ES Modules build

gaearon:
Since it’s not hard to do, I decided to give it a go.
This rewrites `lodash` imports in the ES Modules build (#1369) to use `lodash-es` instead.
Not that it makes a huge difference, but it’s kinda nice.


jdalton:
So cool I can't even.

PR #1374, opened 2016-02-09
typo

michaelghinrichs:
was reading the docs, noticed an extra `'` hanging around :)

### DIFF
--- a/docs/api/applyMiddleware.md
+++ b/docs/api/applyMiddleware.md
@@ -14,7 +14,7 @@ Middleware is not baked into [`createStore`](createStore.md) and is not a fundam
 
 #### Returns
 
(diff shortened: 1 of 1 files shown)


gaearon:
This is on purpose although I agree it’s not very clear. What I meant to say is it takes an “original” `createStore` and returns a “different” `createStore`. Hence `'` at the end, like `A` and `A'` math notation. I’m not a math person though :smile: .  Just not sure how to signal this better.

michaelghinrichs:
ahhh the good ole A prime notation :)

Sounds good! I will see if I can come up with something that signals it better.

Thanks for all that you do!

gaearon:
Thank you for helping out :wink:

Issue #1375, opened 2016-02-09
Error message while running a reducer -- could be more informative?


sawan:
Hi, I have the following code structure defined for a reducer:

```javascript
const SET_NAME = 'SET_NAME'

function setName(value){
    return {
        type: SET_NAME ,
        name: value
    }
}

function nameReducer(state = '', action){
    switch(action.type){
        case SET_NAME:
           return action.name
        default:
            return state
    }
}

const personalDataReducer = combineReducers({
    nameReducer,
    // some other reducers ...
   })

// ... all hooked up via:
// store --> container component  -->   presentation component --> drop down list 
// as per ToDoList example in Redux docs.
```
This works fine. 

The problem arises when I mistakenly change the *name* variable in the action function:

```javascript
function setName(value){
    return {
        type: SET_NAME ,
        changed: value
    }
}
```
After which I get a an error that reads:

> Uncaught Error: Reducer "nameReducer" returned undefined handling "SET_NAME". To ignore an action, you must explicitly return the previous state.

This is a rather confusing message -- what does __"undefined handling"__ mean? 

1. I am definitely dispatching the right action type.
2. The correct reducer is getting invoked and the correct action type (SET_NAME) is being matched.
3. The problem is trying to access a non-existent property the action. End result being, the reducer fails to run properly.

__"undefined handling"__  is specially confusing given (2).

Could the error message be improved? Better stacktrace?

 I apologize if this sounds silly, JavaScript and it's idioms are relatively new to me. 

Code responsible is in [combineReducers.js L129-L133](https://github.com/rackt/redux/blob/master/src/combineReducers.js#L129-L133)

gaearon:
Does “returned undefined *while* handling” sound better? The error message describes what happened: you returned `undefined` from the reducer when you were handling SET_NAME action.

johnsoftek:
I suggest:

`Uncaught Error: Given action "SET_NAME", reducer "nameReducer" returned undefined. To ignore an action, you must explicitly return the previous state.`

gaearon:
Would you like to send a pull request?

johnsoftek:
Done.

gaearon:
Fixed by #1378.

sawan:
Wonderful! Many thanks and sorry for not testing the return result of the reducer -- that would have clarified the message right away.

PR #1378, opened 2016-02-09
Reword error message for 'reducer returned undefined'

johnsoftek:
Minor rewording of error message to address issue #1375.

Old message:
` Uncaught Error: Reducer "nameReducer" returned undefined handling "SET_NAME". To ignore an action, you must explicitly return the previous state.`

New message:
`Uncaught Error: Given action "SET_NAME", reducer "nameReducer" returned undefined. To ignore an action, you must explicitly return the previous state.`

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -7,7 +7,7 @@ function getUndefinedStateErrorMessage(key, action) {
   var actionName = actionType && `"${actionType.toString()}"` || 'an action'
 
   return (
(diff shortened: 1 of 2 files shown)


sompylasar:
Maybe surround "undefined" with backticks, like a code fragment?

gaearon:
>Maybe surround "undefined" with backticks, like a code fragment?

Maybe, but then this needs to be done consistently in all other error messages. I don’t mind this as a separate PR to discuss.

PR #1382, opened 2016-02-10
Update README.md

jbovenschen:
After merging #1339 this project does have a dependency.

I think it is better to remove the "and has no dependencies" from the README.md.
Maybe it would be better to explain why lodash is added as a dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ Redux is a predictable state container for JavaScript apps.
 It helps you write applications that behave consistently, run in different environments (client, server, and native), and are easy to test. On top of that, it provides a great developer experience, such as [live code editing combined with a time traveling debugger](https://github.com/gaearon/redux-devtools).
 
 You can use Redux together with [React](https://facebook.github.io/react/), or with any other view library.  
-It is tiny (2kB) and has no dependencies.
+It is tiny (2kB).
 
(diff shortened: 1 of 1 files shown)


gaearon:
Yeah, this came up before. Technically it has a dependency but we are talking about two ten-line files, not the *whole* Lodash. We use just those two modules, and similar code was previously bundled into Redux, so there is no sensible size difference.

Can we reword it in a way that suggests that *it’s 2kB including the dependencies*? This way we make it clear Lodash doesn’t make the build large.

jbovenschen:
I guess it would be nice to have a sentence something like the following; 
It is tiny (2kB) including lodash utilities. ?

Because it makes it more clear that it has just one dependency that is lodash and it is using lodash for providing utility functions from lodash.

gaearon:
I don’t want to mention Lodash specifically. That we are using it is an implementation detail nobody should worry or care about.

Maybe like this?

>It is tiny, even when bundled with the dependencies (2kB).

acdlite:
> It is tiny (2kB, including dependencies).

?

gaearon:
@acdlite :+1:

jbovenschen:
I think that one is great i will change it tonight!

gaearon:
:+1:

PR #1388, opened 2016-02-11
Update ExampleTodoList.md

benny856694:
incorrect use of spread operator on object

### DIFF
--- a/docs/basics/ExampleTodoList.md
+++ b/docs/basics/ExampleTodoList.md
@@ -71,10 +71,10 @@ const todo = (state, action) => {
         return state
       }
 
-      return {
-        ...state,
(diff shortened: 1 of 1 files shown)


gaearon:
Would you mind also looking through other doc pages to make sure we don’t use object spread there unless specifically mentioning that it is a proposal?

benny856694:
@gaearon , ok

benny856694:
@gaearon , one more finding: [components/TodoList.js](http://redux.js.org/docs/basics/UsageWithReact.html).
I don't know why this code compiles but above code doesn't compile? should this code be corrected with expanded property?

gaearon:
It works inside JSX because JSX already includes this syntax. So it is fine there.

PR #1399, opened 2016-02-14
Clarify comment in compose.js

krawaller:
This PR updates the documentation comment in `compose.js` to clarify that the rightmost function can have multiple arguments, as it provides the signature for the resulting composite function.


gaearon:
Can you please update the `.md` docs accordingly? (In case there is a similar sentence there.)

krawaller:
Doh, good catch, should've though of that. Added now!

gaearon:
:+1:

PR #1403, opened 2016-02-15
Replace universal example with async-with-routing and async-universal

xulien:
hi,

this is my proposal related to https://github.com/reactjs/react-router-redux/issues/232#issuecomment-175197622


examples/universal-with-router/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "dependencies": {
+    "babel-polyfill": "^6.3.14",
+    "babel-register": "^6.4.3",
+    "express": "^4.13.3",
+    "qs": "^4.0.0",
+    "react": "^0.14.7",
+    "react-dom": "^0.14.7",
+    "react-redux": "^4.2.1",
+    "react-router": "^2.0.0",
+    "react-router-redux": "^3.0.0",
```
gaearon: Can we please use 4.x instead? I think there’s a beta out. It has a slightly different API but it’s better.

gaearon:
Hi, thank you for contributing to examples.
I thought about it a bit more, and I’m not sure this is the way we want to go.
(Sorry for the churn!)

I  don’t think we want any more counter examples. In fact it would be better to change the existing `universal` example to do something else. I would say that `async` example is the best candidate for “universalification”.

I would prefer if we:

* Add `async-with-routing` example that was based on async but also used React Router. It would change URLs when the subreddit changes. No need for react-router-redux there.

* Rename `universal` to `async-universal`. Again, no need for react-router-redux there. It should build on top of `async-with-routing` but show asynchronous data fetching in universal app. The counter example there doesn’t really make sense so we can get rid of it.

* Keep `real-world` example as the only example using `react-router-redux` because it is the only example using DevTools which is what that package is primarily for. 

Does this make sense?

examples/universal-with-router/package.json
xulien: sure, it's done

xulien:
No problem
I'm newcomers into react world, and I'm steal learning redux; I haven't yet read async chapter, but I'll ;)
But I have some questions in mind, and I don't know if it's already discuss somewhere, or if those are stackoverflow questions.

how to deal with initialState on huge app? all the data store should be loaded on each "url entry"? or should we populate the initialState accordingly to url location on server side and duplicate the router.js? is there already an elegant way to do that?

and finally should I stick on simple declarative router, or using a dynamic router?

gaearon:
On the server side, you could

* match the route using `match()` from React Router
* use some convention like a static `fetchData()` on route handlers that returns an async action
* wait for all async actions to finish dispatching
* render app with the store that has prefilled data.

You don’t need to worry about `initialState` argument on `createStore()`. To prefill it on the server, dispatch actions.

xulien:
thank you, I will try to produce something at the earliest

FarhadG:
I was looking for some examples using react router with a universal app and I'm glad I stumbled upon this. Thank you, @xulien, and I'm really looking forward to seeing @gaearon's suggestions implemented.

I did have one question: where I run this app, it seems a bit slow when changing routes. Are you having a similar experience? I would imagine for such a simple app, with minimal data and thin components to be instant.

xulien:
I'm not sure if I took the right way to track URL changes, but it's a first proposal to async-router

xulien:
@FarhadG my pleasure :)
about your question, I tried both and I haven't seen differences !?

examples/async-with-routing/containers/Reddit.js
```diff
@@ -0,0 +1,100 @@
...
+
+  static contextTypes = {
+    router: PropTypes.object
+  }
+
+  constructor(props) {
+    super(props)
+    this.handleChange = this.handleChange.bind(this)
+    this.handleRefreshClick = this.handleRefreshClick.bind(this)
+    this.rootPath = location.pathname.substring(0, location.pathname.lastIndexOf('/') + 1)
```
gaearon: Is this necessary? Can’t we just navigate to `/stuff`?

examples/async-with-routing/containers/Reddit.js
```diff
@@ -0,0 +1,100 @@
...
+  componentWillReceiveProps(nextProps) {
+    const { dispatch, params } = this.props
+    if ( nextProps.params.id !== params.id ) {
+      dispatch(selectReddit(nextProps.params.id))
+      dispatch(fetchPostsIfNeeded(nextProps.params.id))
+    }
+  }
+
+  handleChange(nextReddit) {
+    this.context.router.push(this.rootPath + nextReddit)
```
gaearon: seems like 

```js
this.context.router.push(/`${nextReddit}`)
```

would have worked here

examples/async-with-routing/routes.js
```diff
@@ -0,0 +1,18 @@
...
+import IndexRoute from 'react-router/lib/IndexRoute'
+import Redirect from 'react-router/lib/Redirect'
+
+import Layout from './components/Layout'
+import Home from './components/Home'
+import Sample from './components/Sample'
+import Reddit from './containers/Reddit'
+
+export default (
+      <Route path="/" component={Layout}>
```
gaearon: Let’s align route definitions by two spaces like everything else.

examples/universal-with-router/server/server.js
```diff
@@ -0,0 +1,105 @@
...
+  noInfo: true,
+  publicPath: webpackConfig.output.publicPath
+}))
+app.use(webpackHotMiddleware(compiler))
+
+// This is fired every time the server side receives a request
+app.use(handleRender)
+
+function handleRender(req, res) {
+
```
gaearon: Nit: empty line

examples/universal-with-router/server/server.js
```diff
@@ -0,0 +1,105 @@
...
+  })
+
+}
+
+function renderFullPage(html, initialState) {
+  return `
+    <!doctype html>
+    <html>
+      <head>
+        <title>Redux Universal Example</title>
```
gaearon: This “async with routing” example

gaearon: Oh wait. It’s “universal with routing” :smile:

examples/universal-with-router/client/index.js
```diff
@@ -0,0 +1,25 @@
+import 'babel-polyfill'
+import React from 'react'
+import { render } from 'react-dom'
+import { Provider } from 'react-redux'
+import { Router, browserHistory } from 'react-router'
+import { syncHistoryWithStore } from 'react-router-redux'
```
gaearon: From my earlier comment:

>Rename universal to async-universal. Again, **no need for react-router-redux there.**

You can remove it :wink: . I already added it to the only example where it makes sense: `real-world`.

examples/universal-with-router/common/actions/index.js
```diff
@@ -0,0 +1,42 @@
+export const SET_COUNTER = 'SET_COUNTER'
+export const INCREMENT_COUNTER = 'INCREMENT_COUNTER'
+export const DECREMENT_COUNTER = 'DECREMENT_COUNTER'
+
+export function set(value) {
```
gaearon: Let’s kill all of this with fire. I don’t think the “counter” example makes any sense for the universal example. Instead we should copy the components and logic form `async` and fetch *that* in the new (supposedly renamed) `async-universal`.

gaearon:
Thank you for your work on this! I left a few comments inline.

Please skim through my earlier comment too: https://github.com/reactjs/redux/pull/1403#issuecomment-184218441.

Rather than create `universal-with-router`, I would like to kill the existing `universal` example and create an `async-universal` example that combines the two. The current “counter” example in `universal` doesn’t make sense so I wanted to get rid of it.

Basically `async-universal` should be = `async-with-routing` you just did + universal rendering and data fetching on the server.

examples/universal-with-router/client/index.js
xulien: Sure, I will! but I got my first star from @FarhadG for this example... it's hard for me to delete it right now ;)

xulien:
I hadn't miss it, but I'm waited review on async-router. async-universal should coming soon

Thanks for review

gaearon:
Great, thanks! Can you please remove the irrelevant parts for now so it’s easier to review, and get lint to pass? Then I can merge this PR and you can work on the next one.

gaearon:
@xulien Don’t forget to ping me when you’re done and ready for review :wink:

xulien:
@gaearon I'm done :)
async-router and async-universal

gaearon:
Can you please make it so when `async-with-routing` starts, it redirects you to `/reactjs`? Otherwise we have inconsistent history where you start at `/` but after changing it there and back, end up on `/reactjs`. I’m not sure what the best place for this code is but I trust you to figure it out. (It should be generic though and shouldn’t rely on `/reactjs` specifically.)

xulien:
> The devil is in the detail
@gaearon I'm ready for review

examples/async-universal/common/api/reddit.js
```diff
@@ -0,0 +1,14 @@
+import { fetchPostsIfNeeded, selectReddit } from '../actions'
+
+export function fetchData(store, subreddit, callback) {
+
+  if (!subreddit) return callback()
```
gaearon: Let's just make this function return a Promise instead?

gaearon: I don't think `api` is a good folder name for this by the way. It's just a leftover from earlier example.

My intention with this example is for it to *also* have routing. 

I would like to:

* Add routing here 
* Move this function to the container component
* Add code to gather all `fetchData()` from matching routes before rendering instead of hard coding a particular call

Thanks!

examples/async-universal/common/components/Layout.js
```diff
@@ -0,0 +1,11 @@
+import React, { Component } from 'react'
+
+export default class Layout extends Component {
```
gaearon: Is this component necessary? It doesn't seem to do anything.

xulien:
@gaearon something like that?

examples/async-universal/.babelrc
```diff
@@ -0,0 +1,9 @@
+{
+  "presets": ["es2015", "react"],
+  "plugins": ["transform-class-properties"], // needed to get local context.router
```
gaearon: We don’t want to have anything later than ES2015 in the examples. Just change the code to assign static properties instead:

```js
class MyComponent extends Component {}
MyComponent.contextTypes = ...
```

examples/async-universal/common/containers/Reddit.js
```diff
@@ -0,0 +1,115 @@
...
+    router: PropTypes.object
+  }
+
+  constructor(props) {
+    super(props)
+    this.handleChange = this.handleChange.bind(this)
+    this.handleRefreshClick = this.handleRefreshClick.bind(this)
+  }
+
+  static fetchData(dispatch, params) {
```
gaearon: Since we shouldn’t use `static` keyword, let’s put this method below, i.e.

```js
Reddit.contextTypes = ...
Reddit.fetchData = ...
```

examples/async-universal/common/routes.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import Route from 'react-router/lib/Route'
+
+import Reddit from './containers/Reddit'
+
+export default(
```
gaearon: Nit: missing space

examples/async-universal/common/routes.js
```diff
@@ -0,0 +1,10 @@
...
+  <Route path="/" component={Reddit}>
+    <Route path=":id" component={Reddit}/>
```
gaearon: This config is a bit weird. Routes describe component nesting structure. `Reddit` being *inside* `Reddit` doesn’t quite make sense to me. I’m not very good at router configs these days, but I think there should be a way to express this more correctly.

examples/async-universal/server/server.js
```diff
@@ -0,0 +1,90 @@
...
+  match({
+    routes,
+    location: req.url
+  }, (error, redirectLocation, renderProps) => {
+    if (error) {
+      res.status(500).send(error.message)
+    } else if (redirectLocation) {
+      res.redirect(302, redirectLocation.pathname + redirectLocation.search)
+    } else if (renderProps) {
+
```
gaearon: Nit: unnecessary newline. Also the indentation is off (it jumps to 4 spaces while it’s 2 spaces everywhere else).
In general I think there is no need to have each two lines delimited by newlines here.

examples/async-universal/webpack.config.js
```diff
@@ -0,0 +1,49 @@
...
+        loaders: ['babel'],
+        exclude: /node_modules/,
+        include: __dirname
+      }
+    ]
+  }
+}
+
+
+// When inside Redux repo, prefer src to compiled version.
```
gaearon: I recently removed this block from all examples because it was causing some issues. Let’s remove it here as well.

examples/async-with-routing/components/Layout.js
```diff
@@ -0,0 +1,11 @@
+import React, { Component } from 'react'
+
+export default class Layout extends Component {
```
gaearon: `Layout` is confusing. I don’t mind the component per se but right now it looks pointless. If you can use it to make the page look nicer in some way, go for it, but an empty component that just renders its children is not very useful.

gaearon: I suggest you to keep it but rename it to `App` and add some sort of header there, for example, the link to the source code like in the `real-world` example.

examples/async-universal/common/routes.js
gaearon: OK, I get it now. You used to have `Layout` but I asked to remove it, and this is why it looks strange now. As I also write below, please add `Layout` back but make it more useful. You can call it `App` instead and add a header to it that links to the source, like in `real-world` example. I believe this should be enough to use what seems to be a more correct route config (with `IndexRoute` and a `Route` inside a generic app-wide route).

examples/async-with-routing/webpack.config.js
```diff
@@ -0,0 +1,49 @@
...
+        exclude: /node_modules/,
+        include: __dirname
+      }
+    ]
+  }
+}
+
+
+// When inside Redux repo, prefer src to compiled version.
+// You can safely delete these lines in your project.
```
gaearon: Let’s remove this block, as I wrote above.

examples/async-universal/server/server.js
```diff
@@ -0,0 +1,90 @@
+/* eslint-disable no-console, no-use-before-define */
```
gaearon: It is confusing that this single example uses CommonJS in server, and all other examples use ES6. This inconsistency is surprising. I understand why you did this but I’d say we should write this file in CommonJS for consistency’s sake. Later we can change all examples to use ES6 on the server as a single chunk of work.

gaearon: Let’s not add per-file ESLint exceptions. These rules exist in the rule config, so just disabling them in a file begs the question why we use that config at all. I would suggest complying with `no-use-before-define` (we do so elsewhere) and moving `no-console` `disable` + `enable` pair specifically to the line where it is being used. (I’m not sure if was an issue in the existing examples as well, but we definitely shouldn’t introduce more per-file rule overrides.)

gaearon:
This is shaping up really nicely! I’m sorry I’m slow at reviewing. This is already very good but I’d love to make a few more changes. I left the new comments inline.

buzinas:
@xulien are you still gonna work on this? I would really appreciate if you could :)

xulien:
@buzinas I was really (really) busy this last weeks and probably this one too, but I want to finish it soon as possible

buzinas:
@xulien Thanks for the fast response!

gaearon:
@xulien 

If you’re too busy let us know, and somebody else can pick it up from here. I really want to get this in asap because in a way it’s blocking some other work on examples, as we can’t make changes to `async` example without running into inconsistencies here.

xulien:
@gaearon

hell time ended last week
> somebody else can pick it up

it's miiine, my preeecious! and it seems I'm close to finish it.

xulien:
> I’d say we should write this file in CommonJS for consistency’s sake

if I write server.js in CommonJS, it's the whole async-universal example I must rewrite in CommonJS, and it will not be more consistent !?

gaearon:
>if I write server.js in CommonJS, it's the whole async-universal example I must rewrite in CommonJS, and it will not be more consistent !?

No, just `server.js`. Alternatively you can replace `server.js` of all other examples with the ES6 version. It shouldn’t be much work since they’re pretty much identical.

xulien:
I suddenly doubt
When i read "write this file in CommonJS", I understood "don't use babel on server.js".
but i having slept on it, and it seems it's more like "just replace const by var and import by require". 

Second purpose is the right one?

gaearon:
>Second purpose is the right one?

Yea, that’s what I meant (just for the sake of consistency).

xulien:
argl, sorry for noise

xulien:
@gaearon ready for (probably) the last review

mxstbr:
Awesomeee, good work @xullen! Can't wait for this to be merged :fire:

xulien:
@mxstbr great thx, but I'll waiting review before to shout victory! ( I'm shaking on each push :fearful: )

xulien:
this works is in safe hands, then i close this pull request.

gaearon:
👍 Thanks for your work! Hopefully I’ll get to it soon.

gaearon:
Reopening because I got stuck with #1800. There are still some things that need more work here, I’ll try to post a review when I get more time from other projects.

timdorr:
Reclosing in favor of #1883.

Issue #1401, opened 2016-02-15
Redux typings


ulfryk:
Current typings on *DefinitelyTyped* seem to be outdated and not so strict. 
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ffceea9dd124d277c4597c7bd12930666ec074c5/redux/redux.d.ts

```
// Type definitions for Redux v1.0.0
// Project: https://github.com/rackt/redux
```

I have prepared up-to-date version with more strict approach to type safety -> [see it here](https://gist.github.com/ulfryk/69981ccfb488647b6ee3).

Please update it or add typings to this repo. Or I can update it myself. I can help updating those typings also on future releases :)

gaearon:
Please send a PR to DefinitelyTyped. We don’t maintain them ourselves but your help is very welcome.

aikoven:
@ulfryk As you could see, it's been a while since https://github.com/DefinitelyTyped/DefinitelyTyped/pull/7908 was submitted, and its predecessor https://github.com/DefinitelyTyped/DefinitelyTyped/pull/6629 was stuck for even longer time without approval.

It seems that DefinitelyTyped model outgrown itself and we should move to [Typings](https://github.com/typings/typings).

ulfryk:
@aikoven - yes I see - my proposition is also on it's way to stuck ;) ( [8093](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/8093)

wbuchwalter:
Another option would be to maintain them directly in the repo, like what is done for [ng-redux](https://github.com/wbuchwalter/ng-redux), with an `index.d.ts` at the root.
This seems to be the best solution since TypeScript 1.6, see [Typings for npm packages](https://github.com/Microsoft/TypeScript/wiki/Typings-for-npm-packages), and doesn't require any action from the user to get the typings working.
Your call @gaearon, I can understand your wish of keeping redux free of specific languages considerations as well.

gaearon:
I’m happy to keep it in our repo if somebody is willing to maintain them.

wbuchwalter:
:+1: 
We have been ~10 recurrent people debating on this issue in the various PR at definitlTtyped, so I think there will be plenty of people willing to help on that, myself included.
Plus, if you make a change to Redux's API, outdated typings will only affect TypeScript user, so it will force the community to act, and have no impact on non-TS users.

[...]

PR #1413, opened 2016-02-16
Add TypeScript definitions

aikoven:
Original issue: https://github.com/reactjs/redux/issues/1401


gaearon:
Do we want somebody else to review these too, or is this good to merge?

ulfryk:
1. I'd suggest adding tests for typings (like in DT)
2. Please take a look at my proposition -> https://gist.github.com/ulfryk/69981ccfb488647b6ee3 . 

for example

```
export function createStore<S>(reducer: Reducer<S>, initialState?: S, enhancer?: StoreEnhancer): Store<S>;
```

does not give strict rules of use, look at this instead:

```
interface IStoreCreator extends Function {
    <S>(reducer: IReducer<S>): Store<S>;
    <S>(reducer: IReducer<S>, enhancer: IStoreEnhancer<S>): Store<S>;
    <S>(reducer: IReducer<S>, initialState: S): Store<S>;
    <S>(reducer: IReducer<S>, initialState: S, enhancer: IStoreEnhancer<S>): Store<S>;
}
…
interface IStoreEnhancer<S> extends Function {
    (createStore: IStoreCreator): IStoreCreator;
}
…
const createStore: IStoreCreator;
```

etc. etc.

Of course I'm not talking about naming convention (ignore `I` prefix if you want). What is the most useful in my proposition - it's the strict set of rules how  `createStore()` can be used.

gaearon:
If we can add tests for these, we should add tests.

index.d.ts
```diff
@@ -0,0 +1,51 @@
+export interface ActionCreator {
+  (...args: any[]): any;
+}
+
+export type Reducer<S> = (state: S, action: any) => S;
+
+export type Dispatch = (action: any) => any;
```
ulfryk: It would be great if try to not use too much `any` - it's just like working without support of typscript… :(

Consider more strict approach like:

```
interface Action {
    type: any;
}

interface Reducer<S> extends Function {
    <S>(state: S, action: Action): S;
}

interface ReducersMapObject<S> {
    [key: string]: Reducer<S>
}

interface Dispatch extends Function {
    <A extends Action>(action: A): A;
}
```

Or even better

```
interface Action<T> {
    type: T;
}

interface Reducer<S, T> extends Function {
    <S>(state: S, action: Action<T>): S;
}
…
interface Dispatch<T> extends Function {
    <A extends Action<T>>(action: A): A;
}
```

so anyone can implement his own strict actions:

```
enum ActionType {A, B, C}
interface MyAction implements Redux.Action<ActionType> {
    type: ActionType;
    payload: MyPayloadType;

```

or dynamic ones:

```
interface MyAction implements Redux.Action<any> {}
```

index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+export interface MiddlewareAPI<S> {
+  dispatch: Dispatch;
+  getState: () => S;
+}
+
+export interface Middleware {
+  <S>(api: MiddlewareAPI<S>): (next: Dispatch) => Dispatch;
+}
+
+export class Store<S> {
```
ulfryk: It's not a class because you can not import it from Redux (or 'redux') lib…

Trying this:

```
import { Store } from "redux";
const myStore = new Store({myStorKeys:andValues});
```

or just

```
const myStore = new Redux.Store({myStorKeys:andValues});
```

will compile but throw an error in runtime - this is makes static typing a bit useless :(

aikoven:
@ulfryk As far as I can see from `createStore` source, enhanced store creator has only one signature:
https://github.com/reactjs/redux/blob/master/src%2FcreateStore.js#L49
```js
return enhancer(createStore)(reducer, initialState)
```
But you are right about original `createStore` signatures. I'll make a fix.

You are also right about type constraint on `action` argument for `Reducer`.
But `Dispatch` can accept argument of any type, not only `Action`, if user uses some middleware.

@gaearon DefinitelyTyped uses TypeScript code to see if it compiles correctly against definitions. I've added tests for original PR: https://github.com/aikoven/DefinitelyTyped/blob/redux/redux/redux-tests.ts. But I'll need to figure out how to actually run them here.

Will do it tomorrow.

ulfryk:
@aikoven 

## Enhanced creator

You're right - `enhancer` has only one signature. But its signature (of course without `<S>`) is reusable:

```
interface IStoreEnhancer extends Function {
    (createStore: IStoreCreator): IStoreCreator;
}
```

here:

```
function applyMiddleware<S>(...middlewares: IMiddleware<S>[]): IStoreEnhancer;
```

and here:

```
interface IStoreCreator extends Function {
    <S, A>(reducer: Reducer<S, A>): Store<S>;
    <S, A>(reducer: Reducer<S, A>, enhancer: IStoreEnhancer): Store<S>;
    <S, A>(reducer: Reducer<S, A>, initialState: S): Store<S>;
    <S, A>(reducer: Reducer<S, A>, initialState: S, enhancer: IStoreEnhancer): Store<S>;
}
```

## Reducer

It can be still a generic type:

```
interface Reducer<S, A> extends Function {
    <S, A>(state: S, action: A): S;
}

interface IStoreCreator extends Function {
    <S, A>(reducer: Reducer<S, A>): Store<S>;
   …
```

so anyone who uses some middleware and need many types of _action-like-things_ can use 

```
const rootReducer = (state: MyState, a: any): MyState => {…};
const myStore = createStore(rootReducer);
```

mikekidder:
Has this PR been compared to what has already been submitted to [typings/registry](https://github.com/typings/registry/blob/master/npm/redux.json)?

Which one takes precedence if someone does "typings search redux" not realizing a definition file exists in Redux repo?  I can see it bringing confusion, and in turn, issues opened on Redux repo.

index.d.ts
aikoven: Correct. Not sure how it became a class here.

index.d.ts
aikoven: Why isn't this sufficient?
```ts
interface Action {
  type: string;
}
```

If you want to have strict actions you can just do:
```ts
enum ActionType {A, B, C}
interface MyAction<P> {
    type: ActionType;
    payload: P;
}
```
And it will be assignable to `Action`.

Also, reducer should not be constrained to accept only some selected actions. It should accept any possible action and bypass ones it doesn't need to handle.

index.d.ts
```diff
@@ -0,0 +1,60 @@
+export interface Action {
+  type: string;
```
aikoven: Is this correct? Do we expect action type to always be `string`, or should it be `any`?

index.d.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export type Reducer<S> = (state: S, action: Action) => S;
+
+export type Dispatch = (action: any) => any;
```
aikoven: We can't avoid `any` here. Middleware potentially allows `dispatch` to accept anything: functions (thunks), promises, etc. Its return type could also be anything.

index.d.ts
```diff
@@ -0,0 +1,60 @@
...
+export type StoreEnhancer = <S>(next: StoreCreator<S>) => StoreCreator<S>;
+
+export const createStore: StoreCreator;
+
+export function combineReducers<S>(reducers: {[key: string]: Reducer<any>}): Reducer<S>;
+export function applyMiddleware<S>(...middlewares: Middleware[]): StoreEnhancer;
+
+
+export interface ActionCreator {
+  (...args: any[]): any;
```
aikoven: Once more, action creator here is anything that can be bound to `Dispatch`, so its return type is anything that `Dispatch` can accept.

aikoven:
@mikekidder I guess once this PR gets merged it will become official typings for Redux and Typings registry should point to it.

aikoven:
Added some tests using https://github.com/adamcarr/typescript-definition-tester.
For now it requires to add `typescript` and `chai` to `devDependencies`, until this is resolved: https://github.com/adamcarr/typescript-definition-tester/issues/1

index.d.ts
ulfryk: Maybe 

```
export interface Action<T> {
    type: T;
}
```

index.d.ts
ulfryk: We can. Just make it generic:

```
interface Dispatch extends Function {
    <A>(action: A): A;
}
```

or if dispatch returns output changed by middleware:
```
interface Dispatch extends Function {
    <A, O>(action: A): O;
}
```

This is one of typings that is necessary for strict typing of any redux based app. If dispatch can return anything and can't be parametrised then static typing is of no value here…

aikoven: How do we use it? Considering first option we'd have:
```ts
export interface Store<S> {
  dispatch: Dispatch<any>;
  // ...
}
```

So the only place where type parameters would be useful is the app code, e.g.
```ts
@connect(
  state => /* ... */,
  (dispatch: Dispatch<Action>) => {
    return {
      someActionCreator: () => dispatch(someActionCreator)
    }
  }
)
```

But nothing prevents you to set up custom `Dispatch` type with constraints that fit you needs:
```ts
type MyDispatch = (action: MyAction) => MyAction;
```
Then you can use your strict `MyDispatch` in place of `Dispatch` anywhere and it will type-check correctly.

IMO if we add type parameters here, we'd have to write `Dispatch<any>` most of the time, because concrete type constraints depend on the middleware(s) you use.

aikoven: I see how your proposal can be useful: if we put
```ts
type Dispatch<A, O> = (action: A) => O;
```
Then original `Store` definition (no middleware) would be:
```ts
interface Store<S> {
  dispatch: Dispatch<Action, Action>;
}
```
Then we can strengthen `Middleware` type something like:
```ts
interface Middleware {
  <S, A, O>(api: MiddlewareAPI<S>): 
    <D extends Dispatch<any, any>>(next: D) => D & Dispatch<A, O>;
}
``` 
Thus we could always have strong typings for `dispatch` after all store enhancers were applied.
I need to experiment with this a bit.

index.d.ts
aikoven: The only case I see for this is your example with `enum`:
```ts
enum ActionType {
   A, B, C
}
```
But where would it be useful? We can use it in reducer:
```ts
function reducer(state, action: Action<ActionType>) {
  // ...
}
```
But that would be incorrect: `action` argument here is not only your user-defined actions, it can also be `{type: '@@redux/INIT'}` or any other action used by third-party redux library.

index.d.ts
aikoven: Here's what I've come up with:
```ts
type StoreEnhancer<NewDispatch extends Dispatch<any, any>> =
  <D extends Dispatch<any, any>>(next: StoreCreator<D>) => 
    StoreCreator<D | NewDispatch>;
```
It takes store creator with original dispatch and returns one where dispatch is either original or new.

But I couldn't get it to work with enhancers that don't add new dispatch signature, e.g. logger middleware:
```ts
const logger: Middleware</* ? */> = ({getState}) => {
  return <D extends Dispatch<any, any>>(next: D) =>
    (action: /* ? */) => {
      console.log('will dispatch', action)
      let returnValue = next(action)
      console.log('state after dispatch', getState())
      return returnValue
    }
}
```

ulfryk: Nice way but it is becoming quite complicated, isn't it? 

This is not the best way in my opinion:

```
type Dispatch<A, O> = (action: A) => O;
```

…but instead you can try this:

```
type Dispatch = <A, O>(action: A) => O;
```

Also I've checked source code ( see: https://github.com/reactjs/redux/blob/master/src/createStore.js#L143 ) - `.dispatch()` returns reference to action passed in so it actually is:

```
type Dispatch = <A>(action: A) => A;
```

And then :

```
interface Store<S> {
    dispatch: Dispatch;
}
```

So when anyone uses his store:

```
myStore.dispatch({anything})  // ok

// and if we have variable 'x' of type 'XType'
const done = myStore.dispatch(x); // ok
// TS compiler will infer 'done'  type as 'XType'


const newAction: MyAction = …;
myStore.dispatch<MyAction>(newAction);  // ok
const doneAction = myStore.dispatch(newAction); // ok and compiler inferred 'const doneAction: MyAction;'
const doneAction: MyAction = myStore.dispatch(newAction); // ok
const doneAction: MyAction = myStore.dispatch<MyAction>(newAction); // ok

const otherAction: OtherAction = …;
const otherDoneAction: MyAction = myStore.dispatch(otherAction); // compile error
const otherDoneAction: OtherAction = myStore.dispatch<MyAction>(otherAction); // compile error
```

Above one can be strict if parametrised by final user or almost as elastic as `(a: any) => any` without type parameters passed. So this is a win-win :)

index.d.ts
```diff
@@ -0,0 +1,60 @@
...
+export type Dispatch = (action: any) => any;
+
+export interface MiddlewareAPI<S> {
+  dispatch: Dispatch;
+  getState: () => S;
+}
+
+export interface Middleware {
+  <S>(api: MiddlewareAPI<S>): (next: Dispatch) => Dispatch;
+}
```
ulfryk: Middleware definition below seems to be enough, but…

```typescript
interface Middleware<S> extends Function {
    (store: Store<S>): (next: Dispatch) => Dispatch;
}
```

It actually does not return a dispatch but function mapping input of type A to output of type B:

```typescript
interface Middleware<S, A, B> extends Function {
    (store: Store<S>): (next: Dispatch) => (action: A) => B;
}
```

But in this case `Middleware` will always have to be parametrised with both type parameters. We can avoid this ( *but I'm NOT sure if we should* ) in the same manner as with Dispatch:

```typescript
interface Middleware extends Function {
    <S, A, B>(store: Store<S>): (next: Dispatch) => (action: A) => B;
}
```

It's not always so easy to add static type definitions to code written in dynamically typed language… ;)

index.d.ts
```diff
@@ -0,0 +1,60 @@
...
+                                    f2: (a: B) => C,
+                                    f3: (a: A) => B): (a: A) => D;
+export function compose<A, B, C, D, E>(f1: (b: D) => E,
+                                       f2: (a: C) => D,
+                                       f3: (a: B) => C,
+                                       f4: (a: A) => B): (a: A) => E;
+
+// Minimal typing for more than 4 parameters
+export function compose<Result>(f1: (a: any) => Result,
+                                ...functions: Function[]): (a: any) => Result;
```
ulfryk: Maybe it can be even better if we decide what should be the input type of composed function:

```typescript
export function compose<R, I>(fn1: (arg: any) => R, 
    ...functions: Function[]): (arg: I) => R;
// I for input, R for Result
```

index.d.ts
aikoven: That's fine for default dispatch (no middleware):
```ts
type DefaultDispatch = <A extends Action>(action: A) => A;
```
The problem here is that dispatch doesn't always return its input. Consider `redux-thunk`:
```ts
store.dispatch(dispatch => 42)  // returns 42
```
After applying thunk middleware `dispatch` gets type:
```ts
type ThunkDispatch = <O>(thunk: (dispatch, getState) => O) => O;
typeof store.dispatch == DefaultDispatch | ThunkDispatch;
```
Things get worse when we try to define types for `thunk` arguments. In case of single `thunk` middleware it is obviously `DefaultDispatch`:
```ts
type ThunkDispatch =
  <O, S>(thunk: (dispatch: DefaultDispatch, getState: () => S) => O) => O;
```
But what if we had e.g. Promise middleware in front of `thunk` middleware? We should be able to dispatch promises from `thunk`:
```ts
type PromiseDispatch = <P extends Promise<any>>(promise: P): P;
type ThunkDispatch =
  <O, S>(thunk: (dispatch: DefaultDispatch | PromiseDispatch, getState: () => S) => O) => O;
```
This means that `thunk` middleware should know about previously applied middlewares.

index.d.ts
aikoven: Let's try to implement `thunk` middleware in TypeScript as an example. Start without any types:
```ts
const thunkMiddleware = ({dispatch, getState}) =>
  (next) => (action) => {
    return typeof action === function ? action(dispatch, getState) : next(action)
  }
```
Now what types can we add here? Keep in mind that there may me other middlewares applied before `thunk`, so `dispatch` here can potentially accept anything, e.g. promises. Same for `next`, same for `action`.
Middleware is standalone and doesn't know anything about `dispatch` type prior to when it was applied.

gaearon:
@aikoven @ulfryk 

I added you both to collaborators with write access and leave this one to you :wink: . Please work it out between yourselves how/whether you’d like to collaborate, and merge this when you think it’s good enough. I also trust you to maintain these and will ping you if we introduce any changes to the API.

Thanks!

index.d.ts
ulfryk:
```typescript
import { MyState } from './wherever-my-store-is-declared'

interface Middleware<S, A, B> extends Function {
    (store: MiddlewareAPI<S>): (next: Dispatch) => (action: A) => B;
}

type ThunkAction = ((d: Dispatch, gs: () => MyState) => ThunkAction) | Object;

const thunkMiddleware:  Middleware<MyStore, ThunkAction, ThunkAction> = 
  ({dispatch, getState}) =>
    (next) => (action) => {
      return typeof action === function ? action(dispatch, getState) : next(action)
    }
```

Does it do the job ?

ulfryk: It shows that elasticity of `thunkMiddleware` stays in oposition to type safety. Anyway it can be always done in that way:

```typescript
import { MyState } from './wherever-my-state-is-declared';

const thunkMiddleware:  Middleware<MyStore, any, any> = 
  ({dispatch, getState}) =>
    (next) => (action) => {
      return typeof action === function ? action(dispatch, getState) : next(action)
    }
```

;)

ulfryk:
@gaearon - Thank you very much for trust :) We'll try to do our best!

index.d.ts
ulfryk: Tricky situation - JS (and other dynamically typed langs) encourages to build structures that accept anything and handle type inside (runtime reflection) while static typing is about splitting it to handle different inputs/outputs in separate functions/methods.

Thunk (and probably many other) middleware is so elastic, that best solution is to use just:

```typescript
interface DefaultDispatch {
    <A>(action: A) => A;
    <A, B>(action: A) => B;
}
```

And use it with thunk:

```typescript
const dispatch: DefaultDispatch;

…
…  dispatch<any>() …
```

ulfryk: Also consider this:

```typescript
type ThunkDispatch =
  <O, S>(thunk: (dispatch: DefaultDispatch | PromiseDispatch, getState: () => S) => O;
```
is exactly same as:

```typescript
type Thunk<S, O> = (dispatch: Dispatch, getState: () => S) => O;
type ThunkDispatch = <S, O>(thunk: Thunk<S, O>) => O;
```

Which could be replaced with:

```typescript
type ThunkDispatch  = <Th, O>(thunk: Th) => O;
// Just set type parameter `Th` as `Thunk<SomeState, TheOutput>` and `O` as `Output`
```

Which is exactly same as second overload here (:
```typescript
interface Dispatch extends Function {
    <A>(action: A) => A;
    <A, B>(action: A) => B;
}
```

Proposed promise dispatch is actually dispatch:
```typescript
type PromiseDispatch = <P extends Promise<any>>(promise: P): P;
// what is
const dispatch: Dispatch;
… dispatch<Promise<any>>(promise);
```

aikoven: It makes more sense for me now. Thanks.
So basically we don't constraint types of dispatch argument and return value, but add ability for user to specify them if they want?

index.d.ts
aikoven: The problem is — `thunkMiddleware` is standalone and it can't have knowledge of what `MyStore` is.

aikoven:
@ulfryk Updated typings. Please also see comments for outdated diffs.

I have one concern for `Action` type yet. It's clear that user will have to extend `Action` to describe shape of actions in their app. So if we have
```ts
interface Action {
  type: string;
}
```
then user can have e.g. FSA like this:
```ts
interface MyAction<P> extends Action {
  payload: P;
}
```
or free-shape action:
```ts
interface MyAction extends Action {
  [key: string]: any;
}
```
If you want to constraint `type` property:
```ts
type ActionType = 'A' | 'B' | 'C';

interface MyAction extends Action {
  type: ActionType;
  // ...
}
```
Although if `ActionType` was enum, this wouldn't compile, because [String Enums](https://github.com/Microsoft/TypeScript/issues/3192) aren't supported yet. We could loosen constraint for `type` to be `any`, but I'm not sure if we should: values of enum are integers, which would make it hard to analyze action log with DevTools.

Still [Redux docs](http://redux.js.org/docs/basics/Actions.html) don't enforce `type` to be string:

> Types should typically be defined as string constants.

cc @gaearon

index.d.ts
ulfryk: Right. I think it's the best we can do - will satisfy both strict typing freaks (like me) and elasticity lovers. I think the overloaded one does the job in a best possible way:

```typescript
interface Dispatch extends Function {
    <A>(action: A): A;
    <T, O>(action: T): O;
}
```

ulfryk:
@aikoven - about Action interface - we have 2 possibilities:

## Strict

```typescript
interface Action<T> {
    type: T;
}
// So user will always have to specify type of `type` property
```

then user can have e.g. FSA like this:

```typescript
interface MyAction<P> extends Action<string> {
  payload: P;
}

// or

interface MyAction<P, T> extends Action<T> {
  payload: P;
}
```

or free-shape action:

```typescript
interface MyAction extends Action<string> {
  [key: string]: any;
}
```

If you want to constraint type property:

```typescript
type ActionType = 'A' | 'B' | 'C';

interface MyAction extends Action<ActionType> {
  // ...
}
```

and TS enums will also work:

```typescript
type ActionType = { A, B, C }

interface MyAction extends Action<ActionType> {
  // ...
}
```

or

## Elastic

```typescript
interface Action {
    type: any;
}
```

It works in all your examples plus TS enums, but User always have to define his own interface/type for actions if he want's stricter mode.

## Conclusion

IMO stricter version is much better:

```typescript
interface Action<T> {
    type: T;
}
```

index.d.ts
```diff
@@ -0,0 +1,81 @@
+export interface Action {
+  type: string;
+}
+
+
+/* reducers */
+
+export type Reducer<S> = <A extends Action>(state: S, action: A) => S;
+
+export function combineReducers<S>(reducers: {[key: string]: Reducer<any>}): Reducer<S>;
```
ulfryk: I think that reducers map should be defined separately:

```typescript
export interface ReducersMapObject {
    [key: string]: Reducer<any>;
}

export function combineReducers<S, M extends ReducersMapObject>(reducers: M): Reducer<S>;
```

So anyone can easily make his own type for this corresponding with his `State` type if one was defined:

```typescript
interface MyReducers extends ReducersMapObject {
    'posts': Reducer<Post[]>;
    'user': Reducer<MyUser>;
    … 
}

const reducers: MyReducers = …;

const myRootReducer = combineReducers<Reducer<State>, MyReducers>(reducers);
```

index.d.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export interface Dispatch {
+  <A>(action: A): A;
+  <A, B>(action: A): B;
+}
+
+export interface Store<S> {
+  dispatch: Dispatch;
+  getState(): S;
+  subscribe(listener: () => void): () => void;
```
ulfryk: One more lil suggestion here - lets create type/interface for function returned from subscribe so user can declare variable which will hold reference to it in future:

```typescript
// index.d.ts
export interface Unsubscribe extends Function {
    (): void;
}

export interface Store<S> {
    …
    subscribe(listener: () => void): Unsubscribe;
    …
}

// real example from my code:
@Component({selector: 'my-app', templateUrl: '/src/app.html'})
export class App implements OnInit, OnDestroy {
    constructor(@Inject('AppStore') private appStore: Store<IAppState>) {}
    …
    
    // I can declare it with proper type;
    private unsubscribe: Unsubscribe;
    
    // and assign value later
    public ngOnInit(): void {
        this.unsubscribe = this.appStore.subscribe(() => …doSomethingWith(this.appStore.getState())… );
    }

    public ngOnDestroy(): void {
        this.unsubscribe();
    }

}
```

aikoven: That makes sense. One question: what's the reasoning behind `extends Function`? Isn't it redundant when interface already has call signature?

index.d.ts
aikoven: I see one drawback here: until TypeScript supports [default type parameters](https://github.com/Microsoft/TypeScript/issues/2175) we'd have to always specify second parameter which adds verbosity.

How about overloads:
```ts
export function combineReducers<S>(reducers: ReducersMapObject): Reducer<S>;
export function combineReducers<S, M extends ReducersMapObject>(reducers: M): Reducer<S>;
```

index.d.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export interface ActionCreator {
+  <O>(...args: any[]): O;
+}
+
+export function bindActionCreators<
+  T extends ActionCreator|{[key: string]: ActionCreator}
+>(actionCreators: T, dispatch: Dispatch): T;
+
+
```
ulfryk: ## Action creators:

I think that overloading is much better than intersection:

```typescript
// or maybe "<A  extends Action>" but due to middleware issues better without "extends Action"
interface ActionCreator<A> extends Function {
    (...args: any[]): A;
}

interface ActionCreatorsMapObject {
    [key: string]: <A>(...args: any[]) => A;
}

interface ActionCreatorsBinder {
    <A>(actionCreator: ActionCreator<A>, dispatch: Dispatch): ActionCreator<A>;
    <M extends ActionCreatorsMapObject>(actionCreators: M, dispatch: Dispatch): M;
    (actionCreators: ActionCreatorsMapObject, dispatch: Dispatch): ActionCreatorsMapObject;
}
```

I've tested it with such piece of useless code:

```typescript

const oneCreatorA = bindActionCreators<Action>(a => a, store.dispatch);
const oneCreatorB = bindActionCreators(oneCreatorA, store.dispatch);
const oneCreatorC = bindActionCreators(oneCreatorB, store.dispatch);

interface Ott extends ActionCreatorsMapObject {
    one: IActionCreator<Action>;
    two: IActionCreator<Action>;
    three: IActionCreator<Action>;
    four: IActionCreator<Action>;
}

const xCreatorA = bindActionCreators({
    one: a => a,
    two(b: Action) {return b;},
    three: oneCreatorA
}, store.dispatch);

const cMap: Ott = {
    one: a => a,
    two(b: Action) {return b;},
    three: oneCreatorA,
    four: oneCreatorC
};

const xCreatorB: ActionCreatorsMapObject = bindActionCreators(xCreatorA, store.dispatch);
const xCreatorC: Ott = bindActionCreators(cMap, store.dispatch);
```

index.d.ts
ulfryk: I did not test it but i think that `Function` has few things like `bind` that are not declared with just call signature… But I may be wrong. It may be redundant anyway.

index.d.ts
ulfryk: Overloading makes sense :+1:

But have to be well tested ;)

index.d.ts
aikoven: I've checked, call signature is enough.

index.d.ts
aikoven: > ```ts
> export interface ActionCreatorsBinder {
>     <A>(actionCreator: ActionCreator<A>, dispatch: Dispatch): ActionCreator<A>;
>     <M extends ActionCreatorsMapObject>(actionCreators: M, dispatch: Dispatch): M;
>     (actionCreators: ActionCreatorsMapObject, dispatch: Dispatch): ActionCreatorsMapObject;
> }
> ```

Third overload seems redundant here, we can omit type parameter in second overload and it will be inferred as `ActionCreatorsMapObject`.

aikoven: Also consider the difference between
```ts
<A>(actionCreator: ActionCreator<A>, dispatch: Dispatch): ActionCreator<A>;
```
and
```ts
<A extends ActionCreator<any>>(actionCreator: A, dispatch: Dispatch): A;
```
I guess the latter is stronger because let's you constraint not only return type of ActionCreator, but its full signature including argument types.

aikoven:
@ulfryk Pushed the changes and added tests, please review.

Also we have to take care of typings for other Redux-related libs to make them consistent with these.

index.d.ts
ulfryk: Second one here:

```typescript
<A extends ActionCreator<any>>(actionCreator: A, dispatch: Dispatch): A;
```

is not typesafe as it enforces `any`.

I think we should find better way…

index.d.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export interface ActionCreatorsMapObject {
+  [key: string]: ActionCreator<any>;
+}
+
+
+export function bindActionCreators<A extends ActionCreator<any>>(
+  actionCreator: A, dispatch: Dispatch
+): A;
```
ulfryk: hmm.. still not sure if it's the best solution.. As I mentioned in the conversation before: https://github.com/reactjs/redux/pull/1413#discussion_r53292869

index.d.ts
ulfryk: @aikoven also you said

> Third overload seems redundant here, we can omit type parameter in second overload and it will be inferred as `ActionCreatorsMapObject`.

And that is great :) 
It gives possibility to ignore type parameter if one chooses straight `ActionCreatorsMapObject`.

gaearon:
Note that Travis build fails on Node 5.

index.d.ts
aikoven: > ```ts
> <A extends ActionCreator<any>>(actionCreator: A, dispatch: Dispatch): A;
> ```
> is not typesafe as it enforces any.

I'm not sure I understand. This signature allows doing
```ts
bindActionCreators<ActionCreator<MyAction>>(...)
```
and even
```ts
bindActionCreators<(text: string) => MyAction>(...)
```
So type parameter is not enforced.

In contrast, if we used this signature:
```ts
<A>(actionCreator: ActionCreator<A>, dispatch: Dispatch): ActionCreator<A>;
```
to bind action creator of type `(text: string) => MyAction`, then `bindActionCreators` would return value of type `(...args: any[]) => MyAction`, effectively erasing constraints on its arguments.

Also, I found that we don't cover cases when `bindActionCreators` argument and return value have different types, e.g. if action creator returns thunk, then bound action creator will return what thunk returned.

ulfryk: Yeah, you're right. My mistake:

```typescript
<A extends ActionCreator<any>>(actionCreator: A, dispatch: Dispatch): A;
```
Is totally ok.

> Also, I found that we don't cover cases when `bindActionCreators` argument and return value have different types, e.g. if action creator returns thunk, then bound action creator will return what thunk returned.

So maybe third overload:

```typescript
<A extends ActionCreator<any>, B extends ActionCreator<any>>(actionCreator: A, dispatch: Dispatch): B;
```

luisrudge:
IMO, the action interface shouldn't be generic. type should always be a string.

This is the best alternative to me
```js
interface Action {
  type: string;
}
```

mohsen1:
This is great! Thank you for doing this. I'm on my phone but i can try it on my project to see if everything is working.

aikoven:
@luisrudge I think so too. However, it isn't stated anywhere that action type *must* be string.

mohsen1:
I use `enum`s for action types which translate to numbers:

https://github.com/contacts-mvc/redux-react/blob/master/constants/ActionTypes.ts#L1

ulfryk:
@luisrudge - I myself use typescript enums for action types so such definition would make all these typings useless for me and many other ( like @mohsen1 ). Some may use  e.g. `Symbol` for that…

Type definitions that @aikoven builds here step-by-step with care and love should reflect real situation. And real situation is highly complicated because JS is dynamically typed and it will accept anything here. So generic type (overloaded to meet real life expectations) is the one reflecting real situation.

index.d.ts
ulfryk: :+1:

index.d.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export function bindActionCreators<M extends ActionCreatorsMapObject>(
+  actionCreators: M, dispatch: Dispatch
+): M;
```
ulfryk: After whole discussion I'd suggest adding another overloads here for thunk action creators:

```typescript
export function bindActionCreators<A extends ActionCreator<any>, B extends ActionCreator<any>>(
    actionCreator: A, dispatch: Dispatch
): B;
export function bindActionCreators<M extends ActionCreatorsMapObject, N extends ActionCreatorsMapObject >(
    actionCreators: M, dispatch: Dispatch
): N;
```

this should be complete then…

aikoven:
@luisrudge @mohsen1 @ulfryk How about we add
```ts
interface ActionBase<T> {
  type: T;
}
```
for advanced usage and
```ts
interface Action extends ActionBase<string> {
}
```
for standard string types?

ulfryk:
In my opinion it's overkill. Anyone can jest define his own:

```typescript
type MyAction = Action<string>;
```

ulfryk:
@aikoven - I feel that we are almost there. I'll read it through once more tomorrow and probably thumbs up will go :)

aikoven:
It would be nice if other folks reviewed this as well. @wbuchwalter @xogeny

gaearon:
I will post the link on Twitter.

aikoven:
We may also want to wait until https://github.com/adamcarr/typescript-definition-tester/pull/2 gets merged to keep `devDependencies` clean.

wbuchwalter:
I'm very happy with that PR. Big improvements over the current definitions at DefinitlyTyped, so :+1:

mikekidder:
This current iteration looks good.  Ship it!  Thanks to @aikoven and @ulfryk for all the effort on this.

ichpuchtli:
If you add JSDoc style comments to the types you get awesome autocomplete/intellisense in VS Code and others editors. Great for people diving into redux.

joshuacc:
These type definitions look :sparkles: to me.

luisrudge:
can we use union types here? 

```
interface Action {
  type: string | Number;
}
```

xogeny:
FYI, I'm digging into this.  I'm converting one of my open source projects to use this version.

Main sticking point was `Action`.  I swear I read somewhere that `type` had to be a `string` but it seems I was wrong.

As I see it, there are two important uses for `Action`.  The first is to say that the reducer takes an `Action` as an argument.  In this case, it is clear that the argument type needs to be `Action<any>` since the reducer needs to be able to accept all actions in the application and we cannot know *a priori* what type `type` will have for those.

The other important use is for implying a type constraint on *our actions* (*i.e.,* the `Action`s we create ourselves).

I think it would be clearer to delineate these two as different types.  Honestly, `redux` really only uses `Action<any>`.  It neither uses nor cares about the type of `type`.  So specifying a type constraint on `type` is only helpful to people writing their own actions.  Said another way, having a type parameter on `Action` doesn't add any real type safety because `redux` doesn't specify the type of `type`.

So, I would propose that we leave `Action` as:

```
interface Action {
  type: any;
}
```

Since that accurately expresses `redux`s definition of what an action is (no more, no less).  But also include:

```
interface TypedAction<T> extends Action {
  type: T;
}
```

That allows people to impose type constraints on their actions when defining them but avoids lots of `Action<any>` all over their reducer definitions.  Of course, users can simply do stuff like this (without needing `TypedAction`):

```
interface MyAction<T,P> extends Action {
  type: T;
  payload: P;
}
```

Now they get the type constraints on `type` and any *others* they wish to specify as well.  So `TypedAction` might be good for being more explicit, but I don't think it is super important.

To add one final point, in my migration most changes were to `Action<any>`.  Only in a few cases, did I actually need to use `Action<string>`.  My case might be unusual in that I'm extending from `Action` (in one place) to [define an `Operation` type](https://medium.com/@mtiller/observations-on-redux-routing-and-application-architecture-d7dff6609485#.mbo4tbgj8).

gaearon:
>I swear I read somewhere that type had to be a string but it seems I was wrong.

We only require that it isn’t `undefined`. Although we definitely recommend strings.

xogeny:
As a completely anecdotal point of reference, I checked and I changed `Action` to `Action<any>` 15 times in the code I just converted (to use this PR) and to `Action<string>` only 2 times.  Both times I used `Action<string>`, I was defining a type so I could have used either a `TypedAction` definition or simply inlined the specification of `type` into my type.  So I'm inclined to drop the type parameter from `Action`.

index.d.ts
jaysoo: Having `any[]` as ActionCreator arguments makes it unsafe. For example, take the following example:

```
const square: ActionCreator<number> = (x: number) => x * x;
const x: number = square(1);
const y: number = square('bad'); // unsafe but works
```

This problem is compounded when you use `ActionCreatorsMapObject` or `bindActionCreators` since even the output is unsafe.

I don't know what the proper solution is, or if it is worth fixing this right now. I'm gonna take a look later tonight to see if there are any good solutions.

jaysoo:
Added a small note on `ActionCreator` and type safety.

https://github.com/reactjs/redux/pull/1413#discussion_r53500086

I don't want this to be a blocker, but would appreciate any ideas to help with safety.

ulfryk:
@luisrudge - Current proposition covers this union type. Just do:

```typescript
type SomeAction = Action<string | number>;
```

ulfryk:
@xogeny - I think that:

```typescript
interface Action {
  type: any;
}
```

is elastic but does not encourage type safety… JS is flexible. Static typing adds lot of safety. Flexibility is the cost. I think that parametrised `Action<T>` type is best solution for most cases:

You can define one interface/type `type AnyAction = Action<any>` and use it in all places where you had to change `Action` to `Action<any>`

index.d.ts
ulfryk: Actually there are no restriction on shape of action creator. Our proposition is just reflection of real world. But you of course can lock in safer pattern inside your project, just:

```typescript
interface MyActionCreator<A> extends ActionCreator<A> {
  (x: A): A;
}

const square: MyActionCreator<number> = (x: number) => x * x;
const x: number = square(1);
const y: number = square('bad'); // unsafe but works
```

but of course in this situation it would be much better to let typescript infer types properly:

```typescript
const square = (x: number) => x * x; // inferred as    (x: number) => number
const x = square(1); // safe and compiles, x is inferred as number
const y = square('bad'); // unsafe and breaks compilation
```

xogeny:
@ulfryk First, I totally agree that static typing adds a lot.  You will not get any argument out of me on that. 

My point, just to reiterate, is that `redux` *itself* imposes no type constraints.  The type of action recognized by a reducer is not included in the reducer type (nor should it, because of the way reducers  that take potentially different types of actions are composed).  As such, knowing the type of `type` gives no benefit because the `Reducer` type is *generic* (with respect to action type).

I actually thought about the `type AnyAction = Action<any>` approach.  But then every instance of `Action` in `index.d.ts` would be changed to `AnyAction` and it would be really confusing because `Action` would then be effectively gone (as far as the `redux` API is concerned).  That's why I think it is better to make `Action` include `type: any` and have an alternate interface (*e.g.,* `TypedAction`) for cases where it makes sense to impose type constraints.  Then `Action` is used consistently in the description of the API but more strongly typed actions can still be defined.

As I mentioned before, in migrating to these type definition, almost all my work was in changing `Action` to `Action<any>`.  That is a lot of work *with no static typing benefit*.  In the cases where I could benefit from static typing, I used it.  But I don't think it is good "marketing" to force people to do all the work of specifying a type parameter when it doesn't given any benefit (especially when what I'm proposing doesn't preclude providing typing information when useful).

index.d.ts
```diff
@@ -0,0 +1,111 @@
...
+  subscribe(listener: () => void): Unsubscribe;
+  replaceReducer(reducer: Reducer<S>): void;
+}
+
+export interface StoreCreator {
+  <S>(reducer: Reducer<S>): Store<S>;
+  <S>(reducer: Reducer<S>, initialState: S): Store<S>;
+
+  <S>(reducer: Reducer<S>, enhancer: StoreEnhancer): Store<S>;
+  <S>(reducer: Reducer<S>, initialState: S, enhancer: StoreEnhancer): Store<S>;
```
Pajn: These can be simplified to
```typescript
export interface StoreCreator {
  <S>(reducer: Reducer<S>, enhancer?: StoreEnhancer): Store<S>;
  <S>(reducer: Reducer<S>, initialState: S, enhancer?: StoreEnhancer): Store<S>;
}
```
while still providing the same type information and it better describes the actual API.

index.d.ts
```diff
@@ -0,0 +1,111 @@
+export interface Action<T> {
+  type: T;
+}
+
+
+/* reducers */
+
+export type Reducer<S> = <A extends Action<any>>(state: S, action: A) => S;
```
Pajn: As this is the only place the Action type is used, why is it generic?
Better make the type property any directly in the interface.

Pajn:
I totally agree with @xogeny, why should Redux care for what types you use in your application?
As far as Redux is concerned, the type is not generic (Redux doesn't know what that is) and the type of the type property is any as there are no way to disallow undefined in the Typescript type system.

If you define an Action type in your program with a specific type for the type property it would also fulfill the Redux Action type, you don't need to extend it or anything like that. Typescript isn't a classical language, it doesn't need an inheritance chain to fulfill an interface.

index.d.ts
```diff
@@ -0,0 +1,111 @@
...
+                                    f2: (a: B) => C,
+                                    f3: (a: A) => B): (a: A) => D;
+export function compose<A, B, C, D, E>(f1: (b: D) => E,
+                                       f2: (a: C) => D,
+                                       f3: (a: B) => C,
+                                       f4: (a: A) => B): (a: A) => E;
+
+// Minimal typing for more than 4 parameters
+export function compose<I, R>(f1: (a: any) => R,
+                              ...functions: Function[]): (a: I) => R;
```
Pajn: The most inner and the returned function does not necessarily take just a single argument, see https://github.com/reactjs/redux/pull/1399/files

So for all these the returned function must have an untyped parameterlist of `...args` to not create false negatives that can only be solved by casting to `Function`

index.d.ts
ulfryk: :+1:

ulfryk:
@xogeny , @Pajn  - I think I understand your point of view now. And I tend to agree.

@aikoven - what do you say?

index.d.ts
ulfryk: Like this maybe:

```
…
    f4: (a: A, ...args: any[]) => B): (a: A, ...args: any[]) => E;
```

and
```
…
    ...functions: Function[]): (a: I, ...args: any[]) => R;
```

It is really hard to type it strictly… Maybe it should be done just like tuples are in Scala - with limit to ~24 composed functions ?

Pajn: But then you assume that the rightmost function takes at least one argument, which might be false.

Pajn:
Personally I'mt not a fan of `<A, B>(action: A) => B`, it gives exactly the same type safety as `(action) => any`. And if you want to type it afterwards all you have done is moving the type casts into the generic type parameters.

mikemorton:
Please don't force me to use `string` to differentiate Action types.  Ideally, I can differentiate my types using typescript classes and check which type using `instanceof`.

ichpuchtli:
Hey guys please consider adding short summaries in JSDoc format. i.e.
```ts
export interface Store<S> {
   /**
    * Adds a change listener. It will be called any time an action is dispatched, and some part of the
    * state tree may potentially have changed. You may then call getState() to read the current state tree
    * inside the callback.
    * 
    * To unsubscribe the change listener, invoke the function returned by subscribe.
    */        
    subscribe(listener: () => void) : Unsubscribe;
}
```

ulfryk:
@mikemorton - Can you elaborate?

index.d.ts
ulfryk: Right.

ulfryk:
@Pajn - of course we can extend type to:

```typescript
export interface Dispatch {
    (action: any): any;
    <A>(action: A): A;
    <A, B>(action: A): B;
}
```

So it's useful for you and useful for me ;)

xogeny:
@ulfryk - But how is it useful.  If you have `<A,B>(action: A): B`, how would you use it.  As far as I can tell, you'd have to do something like `ret = <MyAction,MyReturn>dispatch(action)`.  I think @Pajn's point is then you might as well do `ret = dispatch(action) as MyReturn`.  It is shorter and doesn't complicate the definition of `Dispatch`.  The issue is that `<A,B>(action: A): B` doesn't *enforce* any type safety because you just specify `A` and `B` yourself...nothing gets inferred so there are no real type constraints here.  At least as far as I can see.

gaearon:
>Ideally, I can differentiate my types using typescript classes and check which type using instanceof.

This is something we discourage although I know TypeScript community feels different. We want actions to be plain objects because then you can serialize them, record and replay them. If you rely on `instanceof` both serialization and hot reloading would break. This is why we only allow plain objects as actions.

>Please don't force me to use string to differentiate Action types. 

This is right, types don’t have to be strings. For example some people like to have `Symbol` types, although this is also problematic for serialization.

ulfryk:
@xogeny 

> The issue is that <A,B>(action: A): B doesn't enforce any type safety because you just specify A and B yourself...nothing gets inferred so there are no real type constraints here.

Actually seems that you are right. 

At first we dropped `(action: any): any;` and replaced it with much stricter `<A>(action: A): A;` (which is true for basic dispatch in redux implementation). But then it appeared that it may be false after applying middleware so things complicated and we needed `<A,B>(action: A): B`. @Pajn pointed out that that the second one actually is equal to `(action: any): any;`.

You are also right - it is safe enough to use `ret = dispatch(action) as MyReturn` or classic `ret = <MyReturn>dispatch(action)`. And it for sure makes both parametrised overloads redundant here.

What do yo say @aikoven ? Should we go back to:

```typescript
type Dispatch = (action: any) => any;
```

?

xogeny:
w.r.t. `Dispatch`, my opinion is that a `Store` should have a `dispatch` method with the signature `(action: Action) => Action`.

Yes, I know that application of middleware can wrap the `Store` in such a way that the `dispatch` method can become something else (*e.g.,* `(action: Action) => Promise<T>`).  But it is my contention that the result is **no longer a `Store`**, but something else.

The reason I think this is important it that a `Store` introduces a type constraint (that `dispatch` returns something with the same type as its argument).  Other "types" of stores might support other type constraints.  It is *useful* to enforce that type constraint (for static type checking reasons).  Making it `(action: any) => any` doesn't provide any useful type constraint.  For that matter, neither does `<A,B>(action: A) => B` since the type constraint being expressed isn't coming from the *store*, but is just being arbitrarily imposed.

The only way to retain a meaningful type constraint (that I can see) is to actually have the type constraint reflected *in the store itself*.  That means that applying middleware should fundamentally change the type of the store (because, after all, *it does*).

aikoven:
I tend to agree with @xogeny [comment](https://github.com/reactjs/redux/pull/1413#issuecomment-186399421): typings should reflect constraints that are imposed by Redux API, nothing more.
E. g. reducer `action` argument must have `type` property; `state` argument and return value must have the same type. 

Considering `dispatch` overloads: I first thought that by *type-safety* @ulfryk meant using `dispatch` like this:
```ts
store.dispatch<MyStrictAction>(action)
```
which would keep us from accidentally passing wrong argument. Although I would not use this approach myself and would let compiler do all the inference.

@xogeny I've experimented with `Store` having default `dispatch` signature `(action: Action) => Action` and letting `StoreEnhancer` provide extra `dispatch` signatures, see https://github.com/reactjs/redux/pull/1413#discussion_r53141654

Although it gets really complicated when extra `dispatch` signature have to reference final `dispatch`, e.g.:
```ts
type Thunk<S, O> = (dispatch: FinalDispatch, getState?: () => S) => O;
type ThunkDispatch = <S, O>(thunk: Thunk<S, O>) => O;

type FinalDispatch = DefaultDispatch | ThunkDispatch;
```
We get recursive types that aren't easy to build.

Pajn:
While I would like to keep the types should be kept the way Redux uses them, it would be just impossible to use Redux with Typescript without casting the store to different things all the time.

Why?
Because the ecosystem is just too diverse an built around composing and replacing which is great concept but doesn't play very well with types. For example nearly all applications will include some kind of middleware to do async actions, may it be using thunks or promises (or something else). By doing that basic thing which most users will do, they will get false negatives if the dispatch is typed as `(action: Action) => Action`.
With both `(action: Action) => Action` and `(action) => any` most applications would need to modify that type both to get type saftey and to get away from false type errors. However with `(action) => any` all applications would just work and the users who need or want better type safety can still specify a stricter type in their application (which they would need to do anyway).

Getting trickery with `StoreEnchancer` doesn't help very much as a middleware can change that type as well so that trickery would need to allow types to flow from the middlewares, getting unioned and flow into the store enhancers and finally flow into the dispatch method. And even if that would be possible to type with some insanely complex type definition all middleware and store enhancer would need to be typed as well, otherwise we would end up with `(action) => any` anyway.

Pajn:
As Typescript automatically discovers the typings file, having to strict typings would be a breaking change requiring a new major version as otherwise everyone would start to get compilation failures.

joshuacc:
It's probably worth pointing out that if you want more type safety than the official Redux definitions give you, it's relatively easy to write a set of wrapping functions with stricter type definitions. For example:

```typescript
interface MyAction {
    type: "INCREMENT" | "DECREMENT";
}

function myDispatch(action:MyAction) {
    dispatch(action);
}
```

aikoven:
So I did more experimenting with `StoreEnhancer` being able to *extend* `Store` type. 

My latest approach was based on adding type parameter to `StoreCreator` that specifies type of `Store` extension:
```ts
export interface StoreCreator<E extends {}> {
  <S>(reducer: Reducer<S>): Store<S> & E;
  // ...
}

export const createStore: StoreCreator<{}>;  // no extension for default store creator
```

In turn, `StoreEnhancer` can extend `StoreCreator` extension:
```ts
export type StoreEnhancer<E extends {}> =
  <F extends {}>(next: StoreCreator<F>) => StoreCreator<E & F>;
```

Middleware is able to add new dispatch signature:
```ts
export function applyMiddleware<D extends Dispatch>(
  middleware: Middleware<D>
): StoreEnhancer<{dispatch: D}>;
```

However, subsequent experimenting failed, because enhanced `Store` gives:
```ts
const store = createStore<S>(reducer, applyMiddleware(middleware));

typeof store ==  Store<S> & {dispatch: D};
typeof store.dispatch == ((action: Action) => Action) & D;
```
but correct dispatch type should be
```ts
typeof store.dispatch == ((action: Action) => Action) | D;
```

# Conclusion
We should probably stick with simplest `dispatch` type:
```ts
type Dispatch = (action: any) => any;
```

ulfryk:
It seems to reflect reality. :+1:

aikoven:
What about `combineReducers` and `bindActionCreators`? Should we drop extra type parameters there as well?

aikoven:
# combineReducers
We currently have two overloads for `combineReducers`. 
## First overload
```ts
export function combineReducers<S>(reducers: ReducersMapObject): Reducer<S>;
```
But without type parameter this
```ts
const reducer = combineReducers<MyState>(reducers)
```
becomes this:
```ts
const reducer: Reducer<MyState> = combineReducers(reducers)
```
which is more verbose. So I'd leave type parameter here.

## Second overload
```ts
export function combineReducers<S, M extends ReducersMapObject>(
  reducers: M
): Reducer<S>;
```
With second type parameter:
```ts
const reducer = combineReducers<MyState, MyReducers>(reducers)
```
Without:
```ts
const reducer = combineReducers<MyState>(<MyReducers>reducers)
```
which has almost same verbosity. So I'd drop this overload.

# bindActionCreators
```ts
export function bindActionCreators<A extends ActionCreator<any>>(
  actionCreator: A, dispatch: Dispatch
): A;

export function bindActionCreators<
  A extends ActionCreator<any>,
  B extends ActionCreator<any>
>(actionCreator: A, dispatch: Dispatch): B;
```
Not sure if we can drop one of these, because they describe different cases: first is when `dispatch` returns its argument, and second is when `dispatch` returns something else, e.g. Thunk case.

ulfryk:
:+1:  for dropping **2nd** overload in `combineReducers`

ulfryk:
# bindActionCreators

have 4 overloads actually:

```typescript
export function bindActionCreators<A extends ActionCreator<any>>(
  actionCreator: A, dispatch: Dispatch
): A;

export function bindActionCreators<
  A extends ActionCreator<any>,
  B extends ActionCreator<any>
>(actionCreator: A, dispatch: Dispatch): B;

export function bindActionCreators<M extends ActionCreatorsMapObject>(
  actionCreators: M, dispatch: Dispatch
): M;

export function bindActionCreators<
  M extends ActionCreatorsMapObject,
  N extends ActionCreatorsMapObject
>(actionCreators: M, dispatch: Dispatch): N;
```

And I think that 1st and 3rd are really useful for all situations where dispatch returns what it gets.

2nd and 4th are a little verbose, but dropping them in favour of `bindActionCreators(ac: any, d: Dispatch): any` will lead to situation in which 1st and 3rd become useless. So I'm for keeping all for overloads here.

aikoven:
Updated PR. This update removes redundant type parameters, fixes `compose` definitions and adds JSDoc.

xogeny:
I don't use `combineReducers` or `bindActionCreators`, so I don't have a strong opinion there.  But I just grabbed the latest `index.d.ts` and dropped it in with some of my code, and it worked perfectly.  So :+1: for the current definitions of `Action`, `Reducer` and `Store`.

@aikoven I applaud your effort to try and formulate a clever type system based approach for gradually widening the type for stores.  Too bad it didn't work, but thanks for trying.

Finally, big thanks to @gaearon for agreeing to bundle these bindings!  I'm super excited about having these bindings in the NPM module. :clap: 

From my perspective, this is good to merge.  Everything I need/use looks good!

ulfryk:
I also tested new typings in my app and all I can say is :+1: and kudos to @aikoven

aikoven:
@Pajn @wbuchwalter Could you guys please review it one last time?

wbuchwalter:
:+1: 
I also tested the new defs with my different projects, everything looks good!

Pajn:
Sorry for not being active recently. I think the current one looks great, very nice to get the documentation too!

mohsen1:
Nice job! Please release this in a major version.

xogeny:
:clap: :clap:

Very excited to see this merged. Thanks everyone.

ulfryk:
:cake:

gaearon:
Will we need to bump a major for this? This can potentially break people, right?

aikoven:
I also don't think that bumping major is necessary. Things can break because some of DefinitelyTyped typings dependent on Redux may become incompatible. But bumping major won't help here because DefinitelyTyped isn't good at versioning anyway.

I will run DT tests against these new typings in a couple of days to see which ones should be updated.
Also I think that maybe we should make official typings for them as well instead of updating DT.

gaearon:
I’m confused :sob: 

>But bumping major won't help here because DefinitelyTyped isn't good at versioning anyway.

If we release this as a minor version and `npm install` can break people’s builds because of that, then we must bump major. If they have problems after `npm install redux@latest` it’s fine because we can say it’s a breaking change. The most important part for me is that people have have `"redux": "^3.0.0"` in their `package.json` are not going to have broken builds on `npm install` after we release this.

So do we need to bump major or not? (There are a few more things I will get into Redux 4 anyway so it’s not a big deal.)

joshuacc:
@gaeron Yes, this can break people's builds if they are using a community type definition or a custom definition of their own. I vote for bumping major.

gaearon:
That’s what I wanted to know. We’ll get this into Redux 4 then.

Pajn:
@joshuacc Really?
In my tests TS takes the community one before the one in the package so should only break if the user removes the community one (which would break anyway as there are no typings in that case)

joshuacc:
@Pajn I believe so, but am away from my computer for the weekend, so can't verify with 100% certainty. It may be different depending on whether using ambient or external typings.

wbuchwalter:
As far as I can tell, ambient modules should take precedence over external ones. 
See section 11.3.1 of TypeScript specifications: 
>* If the import declaration specifies a top-level module name and the program contains an
AmbientModuleDeclaration (section 12.2) with a string literal that specifies that exact name, then
the import declaration references that ambient module.
* If the import declaration specifies a top-level module name and the program contains no
AmbientModuleDeclaration (section 12.2) with a string literal that specifies that exact name, the
name is resolved in a host dependent manner (for example by considering the name relative to a
module name space root). If a matching module cannot be found an error occurs.

So this should not break anyone.

mohsen1:
I didn't know ambient module take precedences. If that's the case it shouldn't be breaking changes for my case.

Ciantic:
Umm, what are these: https://github.com/andrew-w-ross/typings-redux the "typings" (the official replacement of tsd which is now deprecated) installed these. I don't know what is the tsdm @Pajn is referring to, it probably is some fork and not preferred?

Pajn:
@Ciantic tsdm is a tool for using type definitions from npm https://github.com/shyiko/tsdm
I much prefer using npm than other solutions that are not as refined, however it's not the "official typescript way".

Ciantic:
@Pajn but if tsd says its deprecated and redirects all of it's users to typings, that be becomes standard on which all typings must be available. I gather the redux typing is available from typings once they become released?

Ciantic:
@Pajn also, typings is partly to blame for the confusing user experience. It does not say whose typings it installs:

```
typings install redux --save
? Found redux typings for NPM. Continue? (Y/n) Y
Installing redux@3.0.6 (NPM)...
└── (No dependencies)
```

It says its coming from some npm, but not which, when one looks at the typings.json however it says:

```
{
...
"redux": "github:andrew-w-ross/typings-redux#2e0b57e18c6240d17fde045573d74e649762437b"
...
}
```

which clearly points to someone elses NPM.

Pajn:
@Ciantic This is about the Redux typings beeing avalible in the actual redux module so that you can use it without installing any typings at all.

Ciantic:
@Pajn I know, but once they become available, people still run "typings install package" blindly  because they are accustomed to do so, it should give or somehow install the same official typings and not some unofficial ones.

Ciantic:
I opened an issue to typings: https://github.com/typings/typings/issues/320 I think it should be more clear whose typings it installs, I could have thought from it's output it installed some sort of official typings from NPM.

aikoven:
Please note #1526

Igorbek:
This too #1537

aikoven:
For everyone concerned with stronger typings for `dispatch`, please see @Igorbek's proposal: #1537.

It's based on module/interface augmentation introduced in TS 1.8. So to add `dispatch` signature for e.g. `redux-thunk`, you would do:

```ts
// redux-thunk.d.ts

declare module "redux" {
    export interface Dispatch<S> {
        <R>(asyncAction: (dispatch: Dispatch<S>, getState: () => S) => R): R;
    }
}
```

This looks like a decent solution to me, but I'd like to hear from others before merging.

gaearon:
@aikoven 

Sorry I haven’t been maintaining Redux closely for the past couple of months.
What’s up with these? Are we good to cut a minor release with this, or shall we wait for more changes?

aikoven:
@gaearon IMO we should wait at least for #1537.
Also there's a discussion #1526 considering `StoreEnhancer`/`StoreCreator` types.

Ciantic:
Would it be impossible to throw the contents of these: https://github.com/reactjs/redux/blob/master/index.d.ts to inside the declare namespace in here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/redux/redux.d.ts

I had to do it manually couple of times already.

Currently most people are getting their types from there, at least it would match the official types in upcoming release this way.

gaearon:
@aikoven OK. Ping me here or, if I miss it, on Twitter, when you’re ready to cut it.

aikoven:
@Ciantic We'll probably have to copy these typings to DT, because it seems there's no other way to treat other typings in DT that are dependent on Redux.

aikoven:
@gaearon We're good to go

gaearon:
@aikoven What’s the latest consensus, is doing this in a minor bump okay?

aikoven:
I've tested for my build: DT typings take precedence over these, so they don't break anything. New Redux typings seem to be ignored unless user takes some steps i.e. manually unlink other dependent libraries' typings from old DT Redux typings so that TS compiler could resolve imports from `redux` to new ones.

So yes, it is safe to release this as a minor.

gaearon:
What if the user had no typings at all? Is this possible in TypeScript?

Igorbek:
@gaearon if user hadn't DT typings, the typing shipped with the module would be resolved. That said, a user that just installed 'redux' module can start using TypeScript without need to install DT or other external typings.

gaearon:
Then this is a breaking change.

1. User has no typings
2. We add typings in a minor version
3. Their build process fails

Is this correct?

Igorbek:
I don't think it will fail. If user had no typings it wouldn't be able to use any explicit redux types.

Igorbek:
Anyway, I might misunderstand and if it's possible to break a build somehow, it's a breaking change that needs major.

aikoven:
Can we then bump major without waiting until other issues from 4.0 milestone are resolved? There's increasing number of people who want this released asap. How about RC version?

Igorbek:
Personally, I'm pretty sure that adding typing bindings is a non-breaking change that theoretically might break build in very synthetic case. Anyway, I'm not insisting.

Ciantic:
> 1. User has no typings
> 2. We add typings in a minor version
> 3. Their build process fails

Then the user is not using TypeScript, because it's not possible to go past point number one on this list.

Because you can't use TypeScript without typings, they get big honking error message if they try to use redux without typings. It just doesn't compile until there is some error message and they have to either write any declaration themselves or install them from DT.

Though, most of the people using TS has to deal with wildly changing typings anyway, so if there is some hypothethical case where someone has used redux without typings, and if typings entry appears it suddenly uses TS, it's a weird one.

ulfryk:
I 100% agree with @Ciantic

aikoven:
@gaearon I guess the consensus so far is that bumping major is not necessary here.

gaearon:
Out in 3.4.0.

PR #1420, opened 2016-02-18
Update example for conditional applyMiddleware

dcefram:
The example of the tip for apply a middleware threw some errors to me. The reason is because an array was passed to applyMiddleware.

We should pass each elem of the array as an argument first to get it to work.


gaearon:
Oh, thanks.

Issue #1429, opened 2016-02-21
Config issue of UglifyJsPlugin in webpack.config.js


arturparkhisenko:
When i saw [that(webpack.config.js#L27-L33)](https://github.com/reactjs/redux/blob/master/webpack.config.js#L27-L33) config, I wonder why i didn't know about some options, so i looked up them  [here(uglifyjs2)](https://github.com/mishoo/UglifyJS2#usage) and [here(lisperator/compress)](http://lisperator.net/uglifyjs/compress), after that my configs looks like:
```js
new webpack.optimize.UglifyJsPlugin({
  compress: { //compressor? didnt found that in docs
    pure_getters: true,
    unsafe: true,
    warnings: false
    // unsafe_comps: true, //not found
    // screw_ie8: true //not found
  }
})
```

gaearon:
Usually it’s best to check in the source code. Even when docs for these tools are not the best, source code never lies. :wink: 

* [`options.compressor`](https://github.com/webpack/webpack/blob/366ee00fc42731c6acd0b445f4eeafa1e9424c96/lib/optimize/UglifyJsPlugin.js#L14-L16)
* [`unsafe_comps`](https://github.com/mishoo/UglifyJS2/blob/bdd8e34f635db60765f1fd5c8b5355e71ee16095/lib/compress.js#L1066)
* [`screw_ie8`](https://github.com/mishoo/UglifyJS2/blob/bdd8e34f635db60765f1fd5c8b5355e71ee16095/lib/compress.js#L1986)

arturparkhisenko:
@gaearon Thank you Dan! Srry, my bad, i'll look in source code next time! And also now i'm going to use this options too ;)

gaearon:
No problem :smile: . I didn’t know how these options worked either so I took them from other projects. The only useful course of action is to compare the build size before and after the changes.

PR #1430, opened 2016-02-21
Updated UglifyJsPlugin configuration

arturparkhisenko:
for https://github.com/reactjs/redux/issues/1429


gaearon:
Please see my reply in https://github.com/reactjs/redux/issues/1429#issuecomment-186879117.

Rather than blindly tweak options according to the docs, I encourage you to check source code for the tools. Additionally, you can try running `npm run build:umd:min` before and after your changes, and compare the output size. This way it’s easier to tell whether the options were necessary or not.

PR #1433, opened 2016-02-22
+ Feedback.md

arasmussen:
Add Feedback.md with a link to Product Pains and an <iframe> that embeds a Product Pains widget inline for feedback specifically relevant to Redux.

### DIFF
--- a/docs/Feedback.md
+++ b/docs/Feedback.md
@@ -1,0 +1,6 @@
+# Feedback
+
+We appreciate feedback from the community. You can post feature requests and bug reports on [Product Pains](https://productpains.com/product/redux).
+
+<iframe width="100%" height="400px" scrolling="no" frameBorder="0" src="https://productpains.com/widget.html?token=07268479-03f5-d5b8-2626-0320223ff1ee"></iframe>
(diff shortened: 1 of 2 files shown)


arasmussen:
Note that the iframe doesn't render in GitHub but should render fine on the actually docs website.

arasmussen:
And you can see a demo of what the widget looks like at https://productpains.com/widget.html?token=07268479-03f5-d5b8-2626-0320223ff1ee :)

docs/Troubleshooting.md
```diff
@@ -1,6 +1,6 @@
 # Troubleshooting
 
-This is a place to share common problems and solutions to them.
```
gaearon: Spaces are significant in Markdown.

arasmussen:
`git checkout HEAD^ docs/Troubleshooting.md`

arasmussen:
(Removed whitespace changes, ready for another look)

mxstbr:
Thanks very much @arasmussen!

gaearon:
@mxstbr Have you tried publishing the docs btw? You should be able to, I think.

gaearon:
I think `npm run docs:publish` should just work. If it doesn’t work for you let me know and we’ll figure it out!

PR #1435, opened 2016-02-23

rajaraodv:
In the current todos Example, AddTodo component doesn't follow the Container component pattern properly like others in the example. This was initially very confusing to me. So I've updated it by creating a new component **AddTodoForm** presentation component and making AddTodo a proper container component.

### DIFF
--- a/examples/todos/components/AddTodoForm.js
+++ b/examples/todos/components/AddTodoForm.js
@@ -1,0 +1,28 @@
+import React from 'react'
+
+
+let AddTodoForm = ({ onSubmit }) => {
+  let input
+
+  return (
+    <div>
+      <form onSubmit={e => {
+        e.preventDefault()
+        if (!input.value.trim()) {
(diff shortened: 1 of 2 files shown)


gaearon:
To be honest I’d leave it as is. As explained in [Usage with React](http://redux.js.org/docs/basics/UsageWithReact.html):

>Other Components

>Sometimes it’s hard to tell if some component should be a presentational component or a container. For example, sometimes form and function are really coupled together, such as in case of this tiny component:

> [ ... ]

>* AddTodo is an input field with an “Add” button

>Technically we could split it into two components but it might be too early at this stage. It’s fine to mix presentation and logic in a component that is very small. As it grows, it will be more obvious how to split it, so we’ll leave it mixed.

This actually happens in real apps pretty often, and I wanted to highlight the fact that it’s OK and you don’t have to prematurely overmodularize.

PR #1439, opened 2016-02-24
fixed real world example links

ipeters90:
Updated all the broken real world example links.


gaearon:
Thanks man, very much appreciated! 😅

ipeters90:
:+1:

PR #1445, opened 2016-02-25
Solved recursion issue, now "dispatch" directly calls reducer

mananvpanchal:
Solved recursion issue, now "dispatch" directly calls reducer from middleware. This can be used to bypass middleware chain for some actions. We just need to call dispatch instead of next in that case

### DIFF
--- a/src/applyMiddleware.js
+++ b/src/applyMiddleware.js
@@ -24,7 +24,7 @@ export default function applyMiddleware(...middlewares) {
 
     var middlewareAPI = {
       getState: store.getState,
-      dispatch: (action) => dispatch(action)
+      dispatch: (action) => store.dispatch(action)
     }
     chain = middlewares.map(middleware => middleware(middlewareAPI))
     dispatch = compose(...chain)(store.dispatch)


src/applyMiddleware.js
```diff
@@ -24,7 +24,7 @@ export default function applyMiddleware(...middlewares) {
 
     var middlewareAPI = {
       getState: store.getState,
-      dispatch: (action) => dispatch(action)
+      dispatch: (action) => store.dispatch(action)
```
sompylasar: With this change, the assignment above is not needed:
```
var dispatch = store.dispatch
```
-- only the variable declaration for the following line:
```
```

mananvpanchal: Yes you are right. There still room for little improvements.  I started discussion at #1446 you can share your view there

gaearon:
Thank you for sending a PR!
However I don’t think we want a change like this for the reasons described here: https://github.com/reactjs/redux/issues/1446#issuecomment-188811044

Issue #1315, opened 2016-01-30
RFC: waitFor() equivalent in combineReducers() using generators


gaearon:
I know I’ve said many times that `waitFor()` is an anti-pattern and I still think that for most apps that is true. However some apps with complex data calculations have to either:

* (1) Create complex memoized selector hierarchies from normalized data
* (2) Pass data around from parent to child reducers in specific order and add extra arguments to reducers

Both are legitimate ways to solve the problem. However it feels like there is some middle ground for a Flux-like shortcut that would help you write (2) without burdening the parent reducer with the knowledge of the child reducer data dependencies.

In other words, the neat feature of `waitFor()` is how it allows one store (in our case, reducer) to say “I want to use your calculation in my calculation” declaratively, without writing the plumbing code to make this happen. This is especially important in large teams where plumbing code is where the merge conflicts inevitably happen.

I don’t propose to add any form of `waitFor()` to Redux itself. (At least, not yet ;-). There even have been [attempts](https://github.com/dtinth/redux-waitfor) at implementing it in userland. I just had this idea today: what if reducers inside a special version of `combineReducers` (it can be in userland too) could be declared as generators, **and yielding another reducer was the equivalent of `waitFor()`?**.

```js
function a(state = 0, action) {
  switch (action.type) {
  case 'increment':
    return state + 1
  case 'decrement':
    return state - 1
  default:
    return state
  }
}

function b(state = 0, action) {
  switch (action.type) {
  case 'increment':
    return state + 10
  case 'decrement':
    return state - 10
  default:
    return state
  }
}

function* c(state = 0, action) {
  const aState = yield a
  const bState = yield b
  return state + Math.abs(aState - bState)
}

const counters = combineReducers({ a, b, c })
```

Note how `c` has its own `state` but is able to ask `a` and `b` for their next state without calling them. The special `combineReducers()` implementation treats generators differently, asking which reducers (declared on the same level and assumed to be unique, of course) those reducers depends upon, and makes sure that dependencies are resolved, and the relevant state is returned back by `yield`. Just like before, all reducers are called before the combined next state is returned—the only difference is that `yield` inside one child reducer causes another child reducer to execute first, and its return state may be used for a dependent computation.

This provides an easy way to accumulate dependent state without buying into the whole selector business and with zero API other than custom `combineReducers`. And who knows, maybe the default `combineReducers` could support it one day if this is not a dumb idea and can be done performantly.

[...]

PR #1447, opened 2016-02-25
provide the updated parent state to each reducer

chriddyp:
hi folks!

In my app I've found that some reducers occasionally need access to other parts of the state in order to compute their own state.

My fork just passes in the parent state for each reducer to (optionally) use.

This probably isn't the best way to do this as `combineReducers` aims to keep each slice of the store logically independent. I'm curious to hear how other folks have solved this.

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -126,7 +126,7 @@ export default function combineReducers(reducers) {
       var key = finalReducerKeys[i]
       var reducer = finalReducers[key]
       var previousStateForKey = state[key]
-      var nextStateForKey = reducer(previousStateForKey, action)
+      var nextStateForKey = reducer(previousStateForKey, action, nextState)
       if (typeof nextStateForKey === 'undefined') {
         var errorMessage = getUndefinedStateErrorMessage(key, action)
         throw new Error(errorMessage)


d6u:
I found it most of the time I can solve this by making derived values computed when connecting with components by using https://github.com/reactjs/reselect. For two reason:

- Store structure can be minimal this way, no dependencies inside store
- Storing dependencies inside store can be confusing. Imagine this scenario: you make two API calls to the server, say tags, and posts, posts has tag_id on it. So in order to display the tag name on post you need to link posts with tags. If you calculate that in reducer, you could ended up having to handle the case when one arrived but the other didn't twice, one inside posts reducer, one in tags reducer. This duplication in large app is hard to manage.

chriddyp:
Good suggestion @d6u !

I've solved this for my particular use case by constructing an additional "`fullState`" reducer outside of combine reducers:


```
slicedReducers = combineReducers({stateSlice1, stateSlice2, stateSlice3});

module.exports = function(fullState, action) {
    let nextState = slicedReducers(fullState, action);
    return fullStateReducer(nextState, action);
}
```

gaearon:
Yep, both approaches work and I’m not sure whether we want to be opinionated here.
See also discussion in #1315.

Issue #1064, opened 2015-11-23
Namespacing action creators


Zacqary:
I've been able to easily create namespaced bits of my store (e.g. to track the state of a single UI element) using `combineReducers`:

```
let topLevelReducers = {
    aTopLevelReducer: function() { /* Reducer Function */ },
    namespacedReducers: Redux.combineReducers(objectOfReducersToNamespace)
}
```

It's not very easy to do this with action creators using the standard API, since `bindActionCreators` is expecting a flat object containing only functions.

It doesn't look like there's a way to get `bindActionCreators` to recursively bind any child objects it finds inside the object it's passed. Is there a way of doing this in the standard Redux API already?

gaearon:
There's no standard way to do this. It might look like this:

```js
import { bindActionCreators } from 'redux';

export default function bindActionCreatorsDeep(actionCreators, dispatch) {
  return Object.keys(actionCreators).reduce((result, key) => {
    if (!actionCreators[key]) {
      return result;
    }
    switch (typeof actionCreators[key]) {
    case 'object':
      result[key] = bindActionCreatorsDeep(actionCreators[key], dispatch);
      break;
    case 'function':
      result[key] = bindActionCreators(actionCreators[key], dispatch);
      break;
    default:
      break;
    }
    return result;
  }, {});
}
```

We won't include this in core, but if you decide to publish it as a module, let us know!

Issue #716, opened 2015-09-13
Allow grouped actions in bindActionCreator


sbycrosz:
I know that I could write my own and I'm doing so at the moment.
Binding every actions as a props result in lots of props being passed down into the nested property chain. So instead of

```javascript
bindActionCreators({
    createCard,
    displayCardForm,
    updateCardForm,
    hideCardForm
    createList,
    displayListForm,
    updateListForm,
    hideListForm
  }, dispatch);
```

Should we allow
```javascript
bindActionCreators({
    cardAction: {
      createCard,
      displayCardForm,
      updateCardForm,
      hideCardForm
    },
    listAction: {
      createList,
      displayListForm,
      updateListForm,
      hideListForm
    }
  }, dispatch);
```

the same syntax with es6 import

```javascript
import * as CardAction from '../actions/cards';
import * as ListAction from '../actions/lists';

bindActionCreators({
    CardAction,
    ListAction
}, dispatch);

```

And then we just need to pass actions group instead of each bounded action separately.
Will do a pull request later if this is a good idea.

gaearon:
I think we should keep core helpers simple. (Same goes for `combineReducers` that also doesn't support nesting out of the box.)

If you use a library like Lodash, you can write `mapValues({ cardActions, listActions }, actions => bindActionCreators(actions, dispatch))`. And indeed you can write a custom helper that does this, and publish it for others to use.

I don't think we want this in core, as we're entering a very opinionated territory with this.

PR #1451, opened 2016-02-27
Allowing nested action creators in bindActionCreators

lmuntaner:
- New function mapNestedValues using traverse
- bindActionCreators uses mapNestedValues
- new specs for new feature


gaearon:
Thank you for taking time to contribute!

Alas, I would rather *not* take this because deeply nested action creators will be slower to bind (unless you only do it once), and we don’t want to encourage slow patterns. I am also concerned about `action.js` exporting some object by mistake which would get interpreted by `bindActionCreators` as something that needs deep traversal which, with unlucky circumstances, can be very expensive but easy to overlook.

This has been discussed before and it’s fairly easy to implement in userland, so you are welcome to publish a separate package that implements this.

Previous discussions:

* https://github.com/reactjs/redux/issues/716
* https://github.com/reactjs/redux/issues/1064
* https://github.com/reactjs/react-redux/issues/132#issuecomment-144780772

Sorry!

lmuntaner:
Thanks for taking time to go over the PR and responding.

I understand your concerns.

timrichd:
Wheres that separate package?

markerikson:
Not sure the PR author actually published one, but you can see the code changes in the diff.  Also, https://github.com/stutrek/redux-action-wrapper looks like an equivalent approach.

lmuntaner:
Yep, we never publish any package regarding this, sorry.

stegano:
> Thank you for taking time to contribute!
> 
> Alas, I would rather _not_ take this because deeply nested action creators will be slower to bind (unless you only do it once), and we don’t want to encourage slow patterns. I am also concerned about somebody’s `actions.js` exporting some object by mistake which would get interpreted by `bindActionCreators` as something that needs deep traversal which, with unlucky circumstances, can be very expensive but easy to overlook.
> 
> This has been discussed before and it’s fairly easy to implement in userland, so you are welcome to publish a separate package that implements this.
> 
> Previous discussions:
> 
> * #716
> * #1064
> * [reduxjs/react-redux#132 (comment)](https://github.com/reduxjs/react-redux/issues/132#issuecomment-144780772)
> 
> Sorry!

Hi @gaearon 

Could you reconsider allowing nested objects?
(i'm not good at english, please understanding 😅)

Unlike in the past, we can detect and prevent wrong action objects from being bound in advance by using type system such as Typescript or flow.

Libraries like [_typesafe-actions_](https://github.com/piotrwitek/typesafe-actions) make utilizing of nested object actions(e.g. createAsyncAction) but, use bindActionCreators you must manually list all actions.

I respect your opinion, but this issue too old, so i'll try request.

Thanks and Happy New Year 😀

markerikson:
@stegano Dan hasn't worked on Redux in years, and waking up a years-old issue thread just to ping someone is poor etiquette. Please don't do that.

`bindActionCreators` was really built to serve the needs of `connect`.  Now that we have the React-Redux hooks API, it's much less important.

We have no plans to change the core API at this point.

PR #1455, opened 2016-02-28
RFC: remove React Transform from examples

gaearon:
Inspired by @thejameskyle’s rant on Twitter (no offense taken :smile: ), I propose to remove `babel-preset-react-hmre` from examples for the following reasons:

1. Both plugin transform and proxy wrapping are rather invasive and have edge cases
2. It’s **experimental tech** and beginners shouldn’t have to deal with its warts
3. It doesn’t currently reload functional components or container functions like `mapStateToProps`
4. In Redux we rarely care about preserving component local state when hot reloading! Preserving store state is usually enough.

This is a proof of concept of “vanilla” hot reloading (for now, only for a single example).

It has the following traits.

## Pros

1. It is absolutely non-invasive and doesn’t change semantics of your code
2. It hot reloads any components, whether classes or functions
3. It preserves the store state on hot reload
4. It still handles errors on mount and hot reloading “fixes” them
5. It works for functions like `mapStateToProps`

## Cons

1. It does *not* preserve the local state of components
2. It does *not* preserve the DOM so the components are remounted

What do you think? Should we proceed with this and do this for every example?

### DIFF
--- a/examples/async/index.js
+++ b/examples/async/index.js
@@ -1,15 +1,43 @@
 import 'babel-polyfill'
 import React from 'react'
-import { render } from 'react-dom'
+import ReactDOM from 'react-dom'
 import { Provider } from 'react-redux'
-import App from './containers/App'
 import configureStore from './store/configureStore'
 
 const store = configureStore()
+const rootEl = document.getElementById('root')
 
-render(
-  <Provider store={store}>
-    <App />
-  </Provider>,
-  document.getElementById('root')
-)
+let render = () => {
+  const App = require('./containers/App').default
+  ReactDOM.render(
+    <Provider store={store}>
+      <App />
+    </Provider>,
+    rootEl
+  )
+}
+
+if (module.hot) {
+  // Support hot reloading of components
+  // and display an overlay for runtime errors
+  const renderApp = render
+  const renderError = (error) => {
+    const RedBox = require('redbox-react')
+    ReactDOM.render(
+      <RedBox error={error} />,
+      rootEl
+    )
+  }
+  render = () => {
+    try {
+      renderApp()
+    } catch (error) {
+      renderError(error)
+    }
+  }
+  module.hot.accept('./containers/App', () => {
+    setTimeout(render)
+  })
+}
+
+render()
--- a/examples/async/webpack.config.js
+++ b/examples/async/webpack.config.js
@@ -15,7 +15,6 @@ module.exports = {
   plugins: [
     new webpack.optimize.OccurenceOrderPlugin(),
     new webpack.HotModuleReplacementPlugin(),
-    new webpack.NoErrorsPlugin()
   ],
   module: {
     loaders: [
--- a/examples/async/.babelrc
+++ b/examples/async/.babelrc
@@ -1,8 +1,3 @@
 {
-  "presets": ["es2015", "react"],
-  "env": {
-    "development": {
-      "presets": ["react-hmre"]
-    }
-  }
+  "presets": ["es2015", "react"]
 }
--- a/examples/async/package.json
+++ b/examples/async/package.json
@@ -40,10 +40,10 @@
     "babel-loader": "^6.2.0",
     "babel-preset-es2015": "^6.3.13",
     "babel-preset-react": "^6.3.13",
-    "babel-preset-react-hmre": "^1.0.1",
     "expect": "^1.6.0",
     "express": "^4.13.3",
     "node-libs-browser": "^0.5.2",
+    "redbox-react": "^1.2.2",
     "webpack": "^1.9.11",
     "webpack-dev-middleware": "^1.2.0",
     "webpack-hot-middleware": "^2.2.0"


17cupsofcoffee:
I think this is a good idea - having the transform plugin in the examples sorta gives the impression that it's something that you _need_ to have in order to get hot reloading working with Redux. In fact, I was kinda under that impression myself until reading this!

gaearon:
Yeah a lot of people tend to get that impression.

jnleonard3:
Given all of this talk about "JavaScript fatigue", the less dependencies you need, the better.

That said, I definitely want to see react-transform advertised heavily as a DX improvement/plugin for redux - once you get comfortable using vanilla redux.

examples/async/index.js
```diff
@@ -1,15 +1,34 @@
...
+const rootEl = document.getElementById('root')
+
+function render(rootReactEl) {
+  ReactDOM.render(
+    <Provider store={store}>
+      {rootReactEl}
+    </Provider>,
+    rootEl
+  )
+}
```
thejameskyle: Why not put the try-catch here and call `render()` below so even if you reload and get a redbox it will still update with new changes?

ericclemmons:
:+1: from me.  I agree that this doesn't belong in these examples, since it's not directly-related to Redux.

examples/async/index.js
gaearon: Ah, good call. My goal was to make sure RedBox doesn't get into the production build but I can put some more conditionals to ensure that.

danmartinez101:
Sounds like a good move to me

dlmanning:
This is what we've drifted towards as our project has gotten larger. Component HMR is just too finicky to maintain, and all the state we care about preserving is in the store

frontsideair:
I like this, it demonstrates there's nothing magic about hmre.

Phoenixmatrix:
This is 100x better, and handles all the common scenarios for 1/10th of the complexity. I'm in love with it.

examples/async/index.js
```diff
@@ -1,15 +1,43 @@
...
+const rootEl = document.getElementById('root')
 
-render(
-  <Provider store={store}>
-    <App />
-  </Provider>,
-  document.getElementById('root')
-)
+let render = () => {
+  const App = require('./containers/App').default
```
briandipalma: What's wrong with leaving this as an ES2015 import? Is there a technical reason why this must be changed to an inline require?

tanepiper:
:+1: from me, the hmre plugin caused me a lot of headaches a couple of weeks back (mostly documented here: https://gist.github.com/tanepiper/ea41aca1dda6473af738). It was a weird issue I never did resolve in the end, except to switch back to vanilla `react-transform-hmr` which worked out the box as expected.

dayletter:
@gaearon I just tried your modified example, RedBox shows correctly when an error is thrown, but after removing the error, the following occurs: `Uncaught TypeError: Cannot read property '_currentElement' of null`, and hot reloading no longer works. Is it a bug or is it not supported?

gaearon:
@dayletter 

Good catch, this indeed happens with stable React. I was testing with the master version of React. I checked `0.15.0-alpha.1` and it doesn’t seem to have this problem. I would imagine that this was fixed with the initial pass on adding error boundaries in https://github.com/facebook/react/pull/5602/files#diff-c807d0f5ebbc12c44d9479dea693bef6R379.

examples/async/index.js
justingreenberg: dynamic modules are not officially supported since the loader spec is still a work in progress.
webpack v2 supports the ES2015 draft proposal, ie:

```js
let render = System.import('./containers/App').then(App => {
  ReactDOM.render(
    <Provider store={store}>
      <App />
    </Provider>,
    rootEl
  )
})
```

i believe this is also supported in babel, using a polyfill and system module formatter

gaearon: The reason it’s an inline require is because we want to import *the latest version* after hot update.
Babel transpiles `import` to `require()` so we have no opportunity to `require()` again.

In Webpack 2, if we disable Babel modules transform, we can use a normal `import` because Webpack will keep it a live binding and update the imported value.

imranismail:
Tried with alpha react, the same error happens

gaearon:
Weird. Are you sure it was 0.15 alpha, you installed it in example’s own folder and that you restarted the server?

examples/async/index.js
ericclemmons: Also, `import` statements have to be top-level, not nested.

briandipalma: Thanks @gaearon makes sense.

justingreenberg: @gaearon @ericclemmons is correct—although imports are not hoisted by webpack v2 they still have to be at the top level. the system loader maintains bindings though—hmr works like a charm with v2 using the snippet above, just had to remove the deprecated occurrence order plugin :)

imranismail:
Sorry my bad, it was some other error that looks the same.

gaearon:
Please let me know if you can reproduce it!

c089:
@gaearon Do you have an idea how this would work with apps using `reactjs/react-router`, e.g. the `real-world` example? It seems router lost it's `replaceRoute` function at some point and documents that the routes cannot be replaced with a warning. However that means also all react components inside aren't re-rendered. They do when I add a random `key` to the `Router`, but that seems like quite a horrible hack:

```
const renderApp = () => {
  const routes = require('../app/routes').default;
  const app = (
    <Provider store={store}>
      <Router routes={routes} history={history} key={Math.random()}/>
    </Provider>
  );
  render(app, rootNode);
};

renderApp();

if (module.hot) {
  module.hot.accept('../app/reducer', () => {
    store.replaceReducer(require('../app/reducer').default);
  });
  module.hot.accept('../app/routes', () => {
    setTimeout(() => {
      renderApp();
    });
  });
}

```

gaearon:
If you have a `<Root>` component in another file that renders `<Provider><Router /></Provider>` you should be able to use the approach in this PR without adding a key.

c089:
Indeed, thanks :+1:

chibicode:
@gaearon I support this change but have a question:

> It does not preserve the DOM so the components are remounted

I am fetching data on `componentDidMount()` on the top-level component. And when I modify the child component, the top-level component's `componentDidMount()` gets called. I guess this can't be avoided?

gaearon:
With this approach, it can’t. You can, however, add some logic in the action creator to prevent fetches if the data is already available locally. IMO it is a good thing to do anyway (e.g. for instant Back button when you add routing).

chibicode:
@gaearon thanks so much for a quick reply! Got it. I've added caching in the action creator before but invalidating the cache when data got updated from some other action became hairly, so I just defaulted to fetching every time. Thanks!

0x80:
I'm failing to get this working with react-router. Below is the source of my Root component. The Dummy is just a function returning a span with some text. If I remove the Router+Route and place Dummy directly under Provider, it updates with HMR. When the Router and Route components are there I get the `The following modules couldn't be hot updated...`. 

Then I inline the Dummy function in the Root file like `function Dummy() {
  return <span>"Update me x"</span>
}`, editing it will update the display too.

I feel I've tried everything by now :/ What could possibly to wrong?

```javascript
import { PropTypes } from 'react'
import { Provider } from 'react-redux'
import { Router, Route } from 'react-router'
import Dummy from './Dummy'

Root.propTypes = {
  history: PropTypes.object.isRequired,
  routes: PropTypes.element.isRequired,
  store: PropTypes.object.isRequired
}

export default function Root ({store, history}) {
  return (
    <Provider store={store}>
      <Router history={history}>
        <Route path='/' component={Dummy}/>
      </Router>
    </Provider>
  )
}
```

gaearon:
It’s impossible to say what goes wrong if you provide a project reproducing the issue :wink: .

0x80:
I've managed to work it out by putting the routes in a separate file and importing it in the Root module.

```javascript
import { PropTypes } from 'react'
import { Provider } from 'react-redux'
import { Router, Route } from 'react-router'
import routes from '../routes'

Root.propTypes = {
  history: PropTypes.object.isRequired,
  store: PropTypes.object.isRequired
}

export default function Root ({store, history}) {
  return (
    <Provider store={store}>
      <Router history={history}>
        {routes}
      </Router>
    </Provider>
  )
}
```

Before, I passed the routes as a prop into Root at the dom render function like @c089 did in his example, but that doesn't seem to work for me. And passing the routes to module.hot.accept displays a warning on update that routes can't be hot updated as discussed here https://github.com/reactjs/react-router/issues/2704.

So although I can work around it for now, I still don't quite understand how to handle routes properly in relation to HMR.

examples/async/index.js
```diff
@@ -1,15 +1,43 @@
...
+  }
+  render = () => {
+    try {
+      renderApp()
+    } catch (error) {
+      renderError(error)
+    }
+  }
+  module.hot.accept('./containers/App', () => {
+    setTimeout(render)
```
just-boris: Why do you need `setTimeout` there? I have never seen it in other hot reload examples?

gaearon:
>Before, I passed the routes as a prop into Root at the dom render function like @c089 did in his example, but that doesn't seem to work for me.

Yes, that wouldn’t work. Hot updates follow the import chain. If you import routes from `Root`, and `index.js` calls `module.hot.accept('./Root', ...)`, it will work because this update is handled. If you import routes from `index.js` and don’t explicitly handle them with something like `module.hot.accept('./routes', ...)`, it will not work.

examples/async/index.js
gaearon: I think the original motivation was to avoid errors thrown in `render` blowing up HMR updates. Now we handle those errors in `try/catch` so maybe it’s not necessary anymore. I’ll look into it again after React 15 comes out and maybe remove it.

just-boris: Thanks for the information! I am working on my own hotloader for something, so I wonder how it is done in other implementations.

0x80:
Thanks @gaearon for explaining! I did try `module.hot.accept('./routes', ...)` from index.js, but then received the error from react-router saying it couldn't hot update the routes. Maybe I missed something. I will give it another try later.

gaearon:
I think we’re talking past each other :smile: . I did not mean for you to actually do `module.hot.accept('./routes'` because indeed, React Router doesn’t know what to do with updated routes anyway. This is exactly why instead I suggested you to import routes in `Root` component and import a `Root` component in `index.js` so it is always different on every hot reload, and the router doesn’t try to reconcile new props—it is just replaced because `Root` type changed.

0x80:
Haha ok I will NOT try that again then! :+1:

sunyang713:
Hi, so with this new update, do you still need a separate server.js to run an express server with webpack-dev-middleware and webpack-hot-middleware? Or can you revert back to webpack-dev-server?

examples/async/index.js
krigar: How would this work with `react-router`, I assume you have to require each component this way opposed to importing?

0x80:
@sunyang713 It only relies on standard webpack hmr. You can use both the hot middleware and the standard dev server with hmr enabled.

Dema:
Guys,so what is the most current way to use react hot reload with redux? Do you have any boilerplace project? https://github.com/gaearon/react-transform-boilerplate seems to be kind of deprecated.

gaearon:
What you see in this PR is what I recommend for Redux apps after React 15 comes out. I hope to work on a more complete solution that would replace RT later.

examples/async/index.js
mgcrea: Would love an example with `react-router`, tried but failed to make it work (encountering `Warning: [react-router] You cannot change <Router routes>; it will be ignored` like errors).

examples/async/webpack.config.js
```diff
@@ -15,7 +15,6 @@ module.exports = {
   plugins: [
     new webpack.optimize.OccurenceOrderPlugin(),
     new webpack.HotModuleReplacementPlugin(),
-    new webpack.NoErrorsPlugin()
```
burdiyan: Why do you remove NoErrorsPlugin?

gaearon:
@sunyang713

Dev server vs express middleware is completely irrelevant to hot reloading. The only reason I prefer middleware in new examples is because it's easier to show how to combine it with server rendering than when you have two separate servers. You can use WebpackDevServer *absolutely fine* with either this setup or React Transform. This choice is completely unrelated.

examples/async/index.js
tomduncalf: Yeah I'm also struggling to get this working in a real life application, ideally I'd just have to put the `module.hot.accept` somewhere "high up" (e.g. my entry point) and then changes would bubble up there and it would reload the whole thing, but I can't seem to get this working

examples/async/webpack.config.js
wdhorton: He's displaying them with react-redbox instead (see line 24). If he continued using the NoErrorsPlugin, Webpack would suppress the errors, and Redbox would just never be rendered anyway.

gaearon: No, this is not the reason. Redbox only catches runtime errors, it doesn’t help with syntax errors.
NoErrorsPlugin has upsides and downsides. I’m likely to add it back. Please see https://github.com/gaearon/react-transform-boilerplate/issues/122 for a list of its upsides and downsides.

wdhorton: That makes more sense, sorry for the incorrect explanation

examples/async/index.js
tomduncalf: Also the reloading doesn't seem to bubble up quite as I expect - that is, if I have a component `Parent` that requires a component `Child`, and I have `module.hot.accept('components/Parent', ...`, it won't handle changes on `Child` (due to the `components/Parent` argument I guess?). Not really clear how all this is meant to work so a real world example in an app with multiple components, routing, redux etc. would be amazing :)

silvenon:
I really benefit from React Transform even though I'm Redux, because lots of times I'm using state for things that are awkward to keep in the Redux store, like toggling a modal. A modal is an example where React Transform shines, because sometimes you need a few clicks to open a modal again, so without it you would have to do that every time you save a file.

examples/async/index.js
gaearon: @tomduncalf 

Please see the discussion: [here's how to make it work with React Router](https://github.com/reactjs/redux/pull/1455#issuecomment-192857829).

stoikerty: I've been following the discussion about vanilla HMR and decided to use that approach when upgrading to webpack in my [universal-dev-toolkit](https://github.com/stoikerty/universal-dev-toolkit).

I'm using `react-router` with a separate `routes`-file and a `RouterContainer` that serves as the root-component which is reloaded [in the `app.js`-file](https://github.com/stoikerty/universal-dev-toolkit/blob/master/src/client/app.js) using the approach in this PR. It all seems to work well for me so far. When I add a new view to the router, everything refreshes accordingly. Perhaps this could serve as a starting point for someone wishing to implement it themselves?

If you want to add Redux into the toolkit, you can wrap the `<Provider store={store}>` around the `RouterContainer` like @0x80 did (see previous comment).

tomduncalf: Thanks @gaearon for the pointer and @stoikerty for the information. I'm still struggling to get this working - when updating a component, I either get the `component doesn't know how to update itself` message if I say `module.hot.accept('containers/Root'...)`, or `Warning: [react-router] You cannot change <Router routes>; it will be ignored` if I say `module.hot.accept('routes'...)` (which is the component containing the routes).

I suspect that what is happening is that there is no clear "include path" from my component up to the root component due to the way my app is structured, hence the `doesn't know how to update` message. There is a clear path from the updated component to the `Routes` component, but whatever reason routes doesn't want to be reloaded.

I need to take a look at my setup in more detail and work out why this might be happening... not a big deal for now anyway, I'm sticking with class components and `react-transform` for this project :)

stoikerty: @tomduncalf If you provide a more comprehensive example where it doesn't work, say in a repo or a gist, I'm willing to have a look for you. It sounds like the component you're trying to update might be outside of the tree that is on the hot-reload path.

tomduncalf: Thanks @stoikerty, that is my suspicion too - my structure is a little more complex than the examples, because I have split the application into separate modules (which are in essence individual single page applications, with their own entry point) but with common code for some parts.

I will put together a minimal reproducible example when I have a moment, and it would be great to get your feedback if I am unable to get it working! Thanks

examples/async/index.js
sunyang713: hey, how are you able to use the old `render` function without changing the contents of the render? doesn't `render` use the un-updated `App` component?

Following the HMR code for reducers in your other examples, wouldn't the callback of `module.hot.accept()` need to look like this:

```javascript

// render takes in a component to render
let render = (App) => ReactDOM.render(App, rootEl)

 /* . . . */

module.hot.accept('./containers/App', () => {
  const NextApp = require('./containers/App').default
  render(NextApp)
})
```


here is the working code I used for my app: https://github.com/ADI-Labs/calendar-web/blob/master/src/index.js

stoikerty: Line 11 `const App = require('./containers/App').default` always gets the updated component. Inline require calls do that by their very nature of being a function instead of a top-level `import`. The fact that the module isn't passed into `render` doesn't matter in this case because as soon as the render-function is run, the variable definition inside `render` in turn runs the `require`-function that triggers a re-require of the module.

Even though the code inside `render` is the same, that doesn't mean that the functions that are called within render don't produce a different result. Perhaps you're confusing class-constructors and regular functions.

zhangmq:
I wan't to build isomorphic app with react-router async routes, failed with function component too. I found this hmr solution, but it's hart to implement Root for async routes. I write code below, it works fine, but ugly.  need some advise for better solutions.

hot wrapper
```javascript
const hot = (r, a) => {
  if (module.hot) {
    let Inner = r();

    const wrapped = class Wrapped extends Component {
      constructor(props) {
        super(props);

        a(() => {
          Inner = r();
          this.forceUpdate();
        });
      }

      render() {
        return <Inner {...this.props} />;
      }
    };

    return wrapped;
  }

  return r();
};
```
routes
```javascript
{
    path: '/',
    component: hot(
      () => require('./containers/Application/Application'),
      (update) => module.hot.accept('./containers/Application/Application', update)
    ),
    indexRoute: {
      getComponent(nextState, cb) {
        require.ensure([], require => {
          cb(null, hot(
            () => require('./containers/Home/Home'),
            (update) => module.hot.accept('./containers/Home/Home', update)
          ));
        });
      }
    },
    childRoutes: [
      {
        path: 'about',
        getComponent(nextState, cb) {
          require.ensure([], require => {
            cb(null, hot(
              () => require('./containers/About/About'),
              (update) => module.hot.accept('./containers/About/About', update)
            ));
          });
        },
        onEnter: requireLogin
      }
    ]
  }
```

crysislinux:
A note:

I think it's better to also wrap the reducer  with try catch block to avoid broken HMR.

```js
if (module.hot) {
  // Enable Webpack hot module replacement for reducers
  module.hot.accept('../reducers', () => {
    try {
      const nextRootReducer = require('../reducers').default;
      store.replaceReducer(nextRootReducer);
    } catch (error) {
      /* eslint-disable no-console */
      console.error('error: ', error);
      /* eslint-enable no-console */
    }
  });
}
```

gaearon:
Hey everyone!

While I’m still probably going to keep vanilla HMR in Redux examples for simplicity, I just released an alpha of React Hot Loader 3 which handles functional components, needs very little configuration (especially with regards to server rendering or production builds), works great with higher order components (so you can hot reload `mapStateToProps`) and contains other improvements. It’s not 100% ready yet but I encourage you to [check it out](https://github.com/gaearon/react-hot-boilerplate/pull/61)!

examples/async/index.js
```diff
@@ -1,15 +1,43 @@
 import 'babel-polyfill'
 import React from 'react'
-import { render } from 'react-dom'
+import ReactDOM from 'react-dom'
```
davidterry01: Why use ReactDOM.render() instead of plain render()?

justingreenberg: render is being assigned on line 10...

timdorr:
@gaearon Can we close this out then?

timdorr:
Superceded by #1883

Sawtaytoes:
I saw a comment here about react-router being supported only with synchronous routes:
https://github.com/gaearon/react-hot-boilerplate/blob/next/src/App.js

Is it possible it works with async routes now? I moved my code into a separate `root.jsx` file to match the examples, and it works so long as I swap my `<Router />` component for a `<div />`. This works with async routes in `1.3.0`.

Sadly, on-save with the code as-is, it refreshes the page :/.

## client.jsx (entry)
```jsx
import React from 'react'
import { render } from 'react-dom'
import { match } from 'react-router'
import { AppContainer } from 'react-hot-loader'

// Root Component
import Root from 'root'

// Polyfills
import 'react-fastclick'
import 'utilities/polyfills'

// Store and Routes
import { history } from 'utilities/store'
import routes from './routes'

// Router
match({ history, routes }, (error, redirectLocation, renderProps) => {
	if (redirectLocation) {
		window.location = redirectLocation.pathname
	}

	render(
		<AppContainer>
			<Root renderProps={renderProps} />
		</AppContainer>
	, document.getElementById('root'))

	if (module.hot) {
		module.hot.accept('./root', () => {
			const RootContainer = require('./root').default
			render(
				<AppContainer>
					<RootContainer renderProps={renderProps} />
				</AppContainer>
			, document.getElementById('root'))
		})
	}
})
```

## root.jsx
```jsx
import React, { Component } from 'react'
import { Provider } from 'react-redux'
import { browserHistory, Router } from 'react-router'

// Store and Routes
import { store } from 'utilities/store'

export default class Root extends Component {
	render() { return (
		<Provider store={store}>
			<Router history={browserHistory} {...this.props.renderProps} />
		</Provider>
	)}
}
```

Repo w/ Update Branch:
https://github.com/Sawtaytoes/Ghadyani-Framework-Webpack-React-Redux/tree/feature/update-react-hot-loader

Update Commit:
https://github.com/Sawtaytoes/Ghadyani-Framework-Webpack-React-Redux/commit/c675c42301e20f83abd8f1cdb1ba8b7978f3fe04

PR #1460, opened 2016-02-29
initialState is necessary in reducer

Frezc:


gaearon:
Good catch, thank you.

Issue #1458, opened 2016-02-28
Rewrote todos app(examples/todos) to maximize unit testability.


weicheng113:
Hi Guys,

I rewrote todos app. I think testability is quite important. You can find it [here](https://github.com/weicheng113/redux). The idea behind it is as follows. What do you think?

# Purpose

Rewrote todos app([`examples/todos`](https://github.com/weicheng113/redux/tree/master/examples/todos)) to maximize *unit testability*.

### Theory and Reasoning
Components creation and wiring should happen right in the factories and application entry point, which is one of the keys to make components easy to be unit tested. 

This is a practice in OOP world, which was mentioned Nat Pryce(author of [Growing Object Oriented Software, Guided by Tests](http://www.amazon.com/Growing-Object-Oriented-Software-Guided-Tests/dp/0321503627)), [Misko Hevery](http://misko.hevery.com/)(Creator of AngularJS) and other places. 

The process of components creation is like an inverted tree, using small components to assemble large ones, started from leaf components. There are two kinds of components, finite instance components and infinite instance components. And infinite ones will be created by factories.

### Todos as an Example([`examples/todos`](https://github.com/weicheng113/redux/tree/master/examples/todos))
[`src/components/App.js`](https://github.com/weicheng113/redux/blob/master/examples/todos/src/components/App.js) is the entry point of app. We create and wire up components here. 

The components, `AddTodo`(1 instance), `TodoList`(1 instance), `Footer`(1 instance) as well as `Link`(3 instances), are examples of finite instance components. Whereas, the component, Todo, is an example of infinite instance component, which will be created by component factory, `TodoFactory` in this case.

As components only ask for minimal direct dependencies at construction, their logic can be easily unit tested, regardless of whether it is a leaf component(e.g., `Link`) or a branch component(e.g., `TodoList`). 

Unit tests(or specs) for components are under [`test/components`](https://github.com/weicheng113/redux/tree/master/examples/todos/test/components).

Cheers, Cheng

gaearon:
Isn’t solving that the point of [shallow rendering](https://facebook.github.io/react/docs/test-utils.html#shallow-rendering)? You can test that parent components render correctly without relying on the details about their children.

weicheng113:
@gaearon , No, this is a general idea for writing a testable app, regardless tools and languages. I find some related links below.
[...]

PR #1463, opened 2016-02-29
Rewrote todos app(examples/todos) to maximize unit testability.

weicheng113:
This is PR for discussion in #1458. Please have a look the changes under src/components and test/components, ignoring other parts for now.


gaearon:
Hi!

I appreciate the work you put into this but ultimately I don’t think this is something we want in the official examples. Our position is that testing functions is easier than testing classes. Factories also add quite a bit of unnecessary, in my opinion, indirection, whereas shallow rendering lets you abstract away implementation details of children in what I find to be a more satisfactory way without diverging from React conventions.

That said, I’m glad you found something that works better for you, and I’m happy to discuss specific problems you have with our current suggested approaches in #1458.

Thanks!

cwei-bgl:
@gaearon , I have made a new PR #1488, in which I have removed all unrelated parts. The topic I focused on here was to show a potential better way to make an app more **unit testable**, which is a known way in OOP world. The approach, which was already explained in the beginning, is a general idea, which can be applied regardless of the use of classes or functions.

Now that we have a comparable PR(#1471), let me try to point out the key different between them below:
a. Component creation is top down tree structure. Child components are created by parents where required in original approach. 

Whereas, in the approach I propose the component creation is aninverted tree. There are only two places where component creation happens, which are application entry and factories. Application entry consists of three top level containers(they have to stay at very top) and their role is composing or assembling the app. And a function, displayTodo, plays factory role for infinite instance components, regardless its form, a function or a class.

b. #1471 has to depend on shallow rendering to accomplish kind of unit testing. The other approach does not.


Let me try to point out key consequences and observations with the two approach:
a. Unit test, or a better name Focused test, is an isolated test against a component or object under test. It is to test whether a component behaves correctly assuming all its collaborators are correct. 

In the original top-down approach, as components are created directly in place, there is no way to mock out. In other words, it mixes component creation logic and business/ui logic.

b.The need of a special tool shallow rendering for testing may be a hint that components are not written in a test-friendly way.Take TodoList.spec.js for example, it will become an integration test without shallow rendering, as it contains children or collaborator code. 
Depending on how well shallow rendering is, such tests may have more reasons, than it should be, to fail. 

Having said that, I feel this may be better to post to react, which I will do.

Cheers, Cheng

PR #1464, opened 2016-03-01
setState + shallow rendering are now working fine

igorsantos07:
Updating docs regarding React updates on Shallow Rendering: https://github.com/facebook/react/issues/4019#issuecomment-172470697

### DIFF
--- a/docs/recipes/WritingTests.md
+++ b/docs/recipes/WritingTests.md
@@ -337,9 +337,9 @@ describe('components', () => {
 })
 ```
 
-#### Fixing Broken `setState()`
+#### Fixing Broken `setState()` in older React versions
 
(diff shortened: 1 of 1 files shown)


gaearon:
Shall we also remove jsdom from all example tests?

fshowalter:
:+1: for removing jsdom. We only use it when testing components that interact with external (non-React) libraries and thus can't be fully covered with shallow rendering.

Dropping jsdom removes the potential point of confusion regarding node compatiblity. If the examples lock jsdom to `7.x.x` and a newcomer is using node `0.10.x` or `0.12.x`, they may think it's Redux that's incompatible.

mxstbr:
Opened a new issue to track removing jsdom: #1473 

I think we can merge this for now and remove the notice when jsdom is removed!

PR #1472, opened 2016-03-03
misprint

alexeybondarenko:


mxstbr:
Actually, `naïve` is the correct word there. This is the definition according to the [reference.com dictionary](http://dictionary.reference.com/browse/naive):

> having or showing a lack of experience, judgment, or information; credulous:

which is exactly what we want to convey here. That solution is actually not a very good one, but is probably the one most people think of first when they've never done this before. Thus, they show a *lack of experience*.

Thanks for submitting this PR anyway, we appreciate you taking the time to do so!

alexeybondarenko:
Thank you for explanation.

PR #1476, opened 2016-03-04
convert to stateless function component

kvzivn:


mxstbr:
The other components in this example are all built with the `class` syntax, and I'm not sure it's worth convoluting what we actually want to show (redux) by having these concepts mixed in. What we could do is convert all of them to stateless, which would make more sense.

What do you think @gaearon?

mxstbr:
Talked to Dan, we want to keep the examples consistent so converting one component to the stateless version doesn't make much sense. 

Thanks for the PR anyway, we appreciate it!

gaearon:
Yep, if we do that, we should really do that everywhere possible.

PR #1477, opened 2016-03-04
docs: replace slice calls to map

just-boris:
As it was discussed in https://github.com/reactjs/redux/pull/1468, ES5 array methods like `map` and `filter` are recommended approach rather than array slicing.

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -53,14 +53,15 @@ function todos(state = [], action) {
       ]
     case 'COMPLETE_TODO':
       // Return a new array
-      return [
-        ...state.slice(0, action.index),
-        // Copy the object before mutating
-        Object.assign({}, state[action.index], {
-          completed: true
-        }),
-        ...state.slice(action.index + 1)
-      ]
+      return state.map((todo, index) => {
+        if(index === action.index) {
+          // Copy the object before mutating
+          return Object.assign({}, todo, {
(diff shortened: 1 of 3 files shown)


mxstbr:
IMO this looks much cleaner and easier to understand, thanks very much @just-boris!

I'm in a hurry right now, so I'll do a functional test soon to make sure everything still works as expected and then merge!

just-boris:
I made changes only in Markdown files. So all code there wasn't actually executed during build.

@mxstbr How are gooing to test this?

mxstbr:
Oh right, my bad, I didn't look at the code changes!

Looked through it now, LGTM, thanks very much @just-boris!

docs/Troubleshooting.md
```diff
@@ -53,14 +53,15 @@ function todos(state = [], action) {
...
-      return [
-        ...state.slice(0, action.index),
-        // Copy the object before mutating
-        Object.assign({}, state[action.index], {
-          completed: true
-        }),
-        ...state.slice(action.index + 1)
-      ]
+      return state.map((todo, index) => {
+        if(index === action.index) {
```
gaearon: Nit: we leave space after `if` pretty much everywhere else.
Would you mind fixing this as a separate PR?

just-boris: No problem, attached it to #1478

PR #1478, opened 2016-03-04
docs: fix surrounding text after changes in #1477

just-boris:
Just found out that some text mentions that reducer uses `slice` to create a new array. Once we changed it in #1477, we should change surrounding text as well.

Does new text look good for you?

I really should more carefully read the text before sending that PR. Sorry about that.


gaearon:
That looks great, thank you!

PR #1484, opened 2016-03-06
Warn When dispatching During Middleware Setup

### LINKED ISSUE #1240: Dispatching in a middleware before applyMiddleware completes

timdorr:
Recreates #1345
Closes #1240

Switched this to throw an error now.

### DIFF
--- a/docs/advanced/Middleware.md
+++ b/docs/advanced/Middleware.md
@@ -268,11 +268,15 @@ The implementation of [`applyMiddleware()`](../api/applyMiddleware.md) that ship
 
 * It only exposes a subset of the [store API](../api/Store.md) to the middleware: [`dispatch(action)`](../api/Store.md#dispatch) and [`getState()`](../api/Store.md#getState).
 
-* It does a bit of trickery to make sure that if you call `store.dispatch(action)` from your middleware instead of `next(action)`, the action will actually travel the whole middleware chain again, including the current middleware. This is useful for asynchronous middleware, as we have seen [previously](AsyncActions.md).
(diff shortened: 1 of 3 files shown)


timdorr:
Ugh, forgot the branch :)

PR #1487, opened 2016-03-06
Fix Object.assign on callAPIMiddleware

renanvalentin:
The payload property wasn't being passed to Object.assign.


gaearon:
I think there was no intention to make it an actual property of the `action` object. We don’t use FSA in the examples. Thanks for PR though!

Issue #1481, opened 2016-03-06
Use enzyme in tests


gaearon:
We should start using [enzyme](https://medium.com/airbnb-engineering/enzyme-javascript-testing-utilities-for-react-a417e5e5090f) in component tests. It is great and the community seems to have converged on it. We should migrate all our existing tests from `ReactTestUtils` to `enzyme`, and use it for any new component tests.

I don’t plan to work on this, but **if you’d like, please leave a comment stating your intention, and then send a PR when you have something to show so we keep a conversation.** If you feel you don’t have the time, don’t worry—just let us know so another person can pick up the task.

Help wanted :wink:

fshowalter:
I'll take it.

mjw56:
I've been working on updating the ```todomvc``` example. That example had some existing integration tests. I haven't had the opportunity to try ```enzyme``` yet. I can work on converting those existing tests over to ```enzyme```.

gaearon:
@fshowalter 

Thanks, please do! It seems that `counter` example is the best candidate right now. We plan to completely replace `todomvc` so it might not be worth trying there. Please send a PR after you’ve done with `counter` and we can figure out next steps

gaearon:
@fshowalter 

Thank you for the counter update! If you’d like, feel free to add enzyme tests to either of these examples:

* https://github.com/reactjs/redux/tree/master/examples/shopping-cart
* https://github.com/reactjs/redux/tree/master/examples/tree-view
* https://github.com/reactjs/redux/tree/master/examples/real-world

They currently have no component tests at all.

fshowalter:
Will do.

borisyankov:
Hey I am interested in adding enzyme tests too to the projects with none. May be I'll start with the real world one?

gaearon:
@fshowalter @borisyankov Please work it out between yourselves :wink:

fshowalter:
@borisyankov real world is all yours :wink:

fshowalter:
@gaearon did you want to keep the tests under /test or colocate them with the source in a __tests__ folder ala React?

gaearon:
I would keep them under `/test` so we don’t need to change all other projects. We can rearrange later if we desire so.

fshowalter:
Sounds good.

tomchentw:
Quick question here. Is there any codemod for transitioning from `ReactTestUtils` to `enzyme`? I thought it would be nice to have one to automate this procedure and benefits the community.

[...]

PR #1490, opened 2016-03-07
add shopping cart example tests (for #1481)

fshowalter:
Adds component and reducer tests for the shopping cart example. Component tests use enzyme.

@gaearon please let me know if you'd like a different format for the reducer tests. I took a best-guess shot based on the counter example, but I'm happy to change them.


examples/shopping-cart/package.json
```diff
@@ -28,9 +30,14 @@
     "babel-preset-es2015": "^6.3.13",
     "babel-preset-react": "^6.3.13",
     "babel-preset-react-hmre": "^1.1.1",
+    "cross-env": "^1.0.7",
+    "enzyme": "^2.0.0",
     "express": "^4.13.3",
     "json-loader": "^0.5.3",
+    "react-addons-test-utils": "^0.14.7",
```
gaearon: Do we need these? Does `enzyme` peerDep on them or is it here by accident?

gaearon:
This looks great, thank you.

examples/shopping-cart/test/reducers/index.spec.js
```diff
@@ -0,0 +1,85 @@
+import expect from 'expect'
+import { getTotal, getCartProducts } from '../../reducers'
+
+describe('reducers', () => {
```
gaearon: Maybe worth changing to `selectors`? They are grouped with reducers in the code for convenience but we should probably label them differently.

examples/shopping-cart/package.json
fshowalter: Needed if you're using React 0.14. Enzyme works with 0.13 or 0.14 but only requires `react-addons-test-utils` if you're using 0.14. Thus, they can't peerDep it, but they do document it. 

`react-dom` is required in 0.14 too, but we're using shallow rendering so we don't need it.

examples/shopping-cart/test/reducers/index.spec.js
fshowalter: Good call. Updated.

timdorr:
This is great! Very readable and comprehendible tests!

PR #1491, opened 2016-03-07
Fixes a documentation typo in basics/Reducers

himynameisdave:
Found an error (at least I think it is) when working through the todo tutorial.

```javascript
return state.map(todo, index) => {
  if (index === action.index) {
    return Object.assign({}, todo, {
      completed: true
    })
  }
    return todo
  }
```

...has been changed to...

```javascript
return state.map((todo, index) => {
  if (index === action.index) {
    return Object.assign({}, todo, {
      completed: true
    })
  }
    return todo
  })
```

Hope this helps! :smile:

### DIFF
--- a/docs/basics/Reducers.md
+++ b/docs/basics/Reducers.md
@@ -212,14 +212,14 @@ function todos(state = [], action) {
         }
       ]
     case COMPLETE_TODO:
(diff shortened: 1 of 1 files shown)


gaearon:
Oh, thanks.

PR #1492, opened 2016-03-08
Unnecessary re-evaluation of subscribers & leaking in dispatch

hnry:
Specifically this line, https://github.com/reactjs/redux/blob/master/src/createStore.js#L175

Just looking through the src this line stuck out to me because if there was no "mutex" in place this is would be a race cond. (Though the mutex doesn't cover the subscriber part, so if it were somehow in a multi-threaded environment, the subscribers could possibly not be in sync here, though even in a single- threaded environment subscribers are still able to be added, explanation further down.)

Right now there is no benefit to re-evaluate the length on every loop and goes against https://github.com/reactjs/redux/blob/master/src/createStore.js#L85 
"The subscriptions are snapshotted just __before__ every `dispatch()` call." (emphasis mine)

So we should cache as things are suppose to be "frozen" anyway.

===
Related to this issue, specifically "The subscriptions are snapshotted just before every `dispatch()` call."

If this is correct spec, this does not happen by the way. New subscribers can be leaked into a dispatch __during__ dispatch (during the mutex lock). As reducers can add subscribers and they _will_ be called when listeners are called. 

This pull does not address that. As I am unsure if this is truly unintended. But I can add to this pull. I am opening it up with minimal changes to get the conversation going on these issues.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -172,7 +172,7 @@ export default function createStore(reducer, initialState, enhancer) {
     }
 
     var listeners = currentListeners = nextListeners
-    for (var i = 0; i < listeners.length; i++) {
+    for (var i = 0, l = listeners.length; i < l; i++) {
       listeners[i]()
     }
 


timdorr:
I would feel a lot better about this if there was a test case. I'm not sure if it's practically able to be triggered.

> As reducers can add subscribers

Nope, they cannot. Otherwise they aren't pure functions and would violate the predictable nature of Redux.

hnry:
@timdorr Test case for the length caching? Well caching the length is not fixing a bug, it's just removing unnecessary part of re-evaluating the length of subscribers every time it loops. Since the length should never change / subscribers can't be added during dispatch, why does the length need to be recalculated?

As for 
> Nope, they cannot. Otherwise they aren't pure functions and would violate the predictable nature of Redux.
Right. I was just bringing up that it can be done, whether a user does it or not is one thing, the option is there. It is the same thing as this https://github.com/reactjs/redux/blob/master/src/createStore.js#L164 with this line they don't have the option to at all.

timdorr:
Length is a property, not a function call. Nothing is getting reevaluated on each loop. There is no performance difference here, as it will just deref a pointer when this gets compiled down to IR/machine code by any of the major JS engines.

Your original post made it seem like there is a subtle spec violation here. Is there one? That's what there should be a test case for.

hnry:
Yes with most modern JITs the performance doesn't matter. But that's beside the point. I guess I should rephrase, performance being equal, I am referring to the fact it will re-evaluate because the length property _could_ change versus it can never change if it were cached. The potential (theoretically at least) for mutations on the length can exist in the future.

So if performance doesn't matter, then we are talking about programmer's _intent_. And that is not our intention to ever have the length of the array to change, so why does the possibility exist? 

1. Fine, you can say under no _current_ circumstances will this occur now. Fair enough, I was just raising the point that it could lead to something unexpected, and if both codes are the same performance wise, _why not play it safe_?

2. Like I mentioned this is just a pull request to get discussion going on possible changes. But as per https://github.com/reactjs/redux/blob/master/src/createStore.js#L85 if this is still the intention, then the current code does _not_ do that. As per the actual source it does not do this. It actually snapshots _after_ dispatch. Which is _why_ you can subscribe during a dispatch in a reducer.

There are many different ways this can be addressed but I am not sure what route the maintainers would like to take or any.

I've pointed out the simplest and easiest solution is caching the length of the subscribers. If you cache the length before dispatch as per L85 comments says it does, then even if the length changed (more subscribers get added as explained in 2. this _is_ possible), you still _only_ call the same listeners. And whether you care about 2., it does more closely follow the comment in L85 as far as programmer's intent.

hnry:
I am actually going to close this, I don't feel like this pull is heading anywhere. As I don't understand why I need to explain myself so much or feel like I need to defend myself. I don't feel like this is a partnership in a problem, but the opposite.

This was meant to be a trivial change... and not a proposition for a major change. It seemed obvious to me there is mismatch in the code. But I am not the maintainer, I understand that.

Length should never change as per comments in the code, so why should the possibility for the loop to evaluated a different length exist?
For the subscribers being snapshotted as the comments say before, but the code clearly shows after. Which is correct?  Like I said in the original post, I don't know which is correct so I am opening for discussion before I make the right changes. But the code or the comment of what the code does, one of them is obviously wrong.

As for reducers ability to add subscribers, this can be done as of now as I mentioned. If docs are enough fine, but there's already existing code in other places to put limits on things on top of the docs.

gaearon:
Hey, thanks for PR. I think that we rely pretty heavily on listeners *not* changing during dispatch. We have safeguards in place to ensure that. (See how we copy array lazily if subscribers change during dispatch.)

Therefore we don't want to add a safeguard *here* because it might prevent a more important bug from surfacing early in the future. If subscribers *do* get mutated we want to fail early rather than make such a bug dormant and unnoticed.

hnry:
The lazy array copying happens after dispatch, not during.

As mentioned several times already, you can add new subscribers during a dispatch through a reducer:
https://jsbin.com/zomunu/1/edit?js,console

There is a safeguard to stop dispatching from a dispatch, but not subscribing. Why safeguard one but not the other?
This introduces side effects from leaking into redux. With this I can actually dispatch from a dispatch.
https://jsbin.com/zomunu/edit?js,console
___warning__ this is an infinite loop since dispatching from a dispatch is recursive. But you can open it in a new tab to see I'm not lying.


And the safeguard I proposed was just to start discussion, I was open to whatever method after discussion as mentioned in the original post.

gaearon:
There is no need to make this emotional :wink: 

>There is a safeguard to stop dispatching from a dispatch, but not subscribing. Why safeguard one but not the other?

You’re right—it should have been safeguarded. There is no particular reason we don’t safeguard subscriptions while reducer is running other than we forgot to do it. I’m happy to accept a pull request that prevents subscriptions and unsubscriptions while the reducer is running, analogous to how we prevent dispatches. Since it’s technically a breaking change, it would need to be submitted against the `next` branch.

Thanks for pointing out the issue! I apologize if my tone was dismissing because this was not my intention. I was just proposing that we should fail hard for cases we don’t want to support rather than have half-working safeguards against them.

I would also suggest that in the future, to avoid bikeshedding and confusion, you include the failing test case in the pull request. This would make it much easier to see what problem you are trying to *solve* rather than debate whether the problem exist.

Thank you again!

hnry:
@gaearon I admit after a bit I am emotional, but it's not about whether I can make this pull happen. It's that I feel I'm repeating myself because people do not believe me. If it was a clear "I understand, but it's not an issue" then ok.

I closed the issue because I did not feel it was going to go anywhere, and I did not want it to force a resolution if maintainers did not feel there was a problem.

Everything I've been saying is just an expansion or repeating my original post. I say "hey reducers can subscribe", I'm told they cannot. I've been saying subscriptions happen _during_ and not _before_ but I'm told this is not true either. 

I mean if someone took even 1 minute to look at the code itself they would realize I am not making this stuff up. But point noted that I will have a clear example from the get go next time.

And I see your point about bike shedding, but there is a __bigger__ issue that I keep trying to point out.
Again if people would assume I'm not making things up or assume I and others have valid points to contribute they can see this.

I will make a new issue we can talk further about it there and we can leave the misunderstandings here.

timdorr:
@hnry Sorry if I contributed to any feelings of excess resistance. We definitely want contributions on this project and I'm sorry if I discouraged that for you. 

Much like Dan said, a test was suggested because it's the clearest way to get across the issue you're trying to fix. It also helps prevent a regression later on.

Interestingly, assuming you could change the spec, you could do this more simply as a reversed loop:
```js
for (var i = listeners.length; i >= 0; i--) {
```
I think my main style issue is excess terseness. That would be better represented as a `var length = listeners.length` on the line beforehand. That would be more comprehensible to more people. Single letter variables are often a source of confusion, especially for beginners.

PR #1471, opened 2016-03-03
Add todos example tests

jontewks:
Per issue #1397

PR #1493, opened 2016-03-08
add todos enzyme tests

mjw56:
This PR adds `enzyme` tests for the `todos` example `components`

This is part of #1481

### DIFF
--- a/examples/todos/package.json
+++ b/examples/todos/package.json
@@ -4,7 +4,7 @@
   "description": "Redux Todos example",
   "scripts": {
     "start": "node server.js",
-    "test": "cross-env NODE_ENV=test mocha --recursive --compilers js:babel-register --require ./test/setup.js",
+    "test": "cross-env NODE_ENV=test mocha --recursive --compilers js:babel-register",
     "test:watch": "npm test -- --watch"
   },
   "repository": {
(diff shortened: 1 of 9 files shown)


gaearon:
I specifically didn’t mention `todos` in #1481 because @jontewks was already working on it in #1471. I should have communicated this better :disappointed_relieved: .

Ultimately we want to use enzyme in all our tests so this probably supersedes @jontewks’s work. I apologize for not handling it better.

Can you please make sure that this includes whatever tests existed in #1471?

mjw56:
Yeah this was my fault, I should have checked the open PR's first. I reviewed #1471 and it looks like this includes all the tests from there.

whoisandie:
Aren't the component tests for todo-with-undo the same as todos? Trying to figure out if there's any other component test that has to be added. The action and reducer tests are missing though. Since redux-undo is already tested, minor tweaks in action and reducer tests might be sufficient.

examples/todos/package.json
```diff
@@ -31,9 +31,10 @@
     "babel-preset-react-hmre": "^1.1.1",
     "babel-register": "^6.3.13",
     "cross-env": "^1.0.7",
+    "enzyme": "^2.0.0",
     "expect": "^1.8.0",
     "express": "^4.13.3",
-    "jsdom": "^5.6.1",
+    "jsdom": "^8.1.0",
```
gaearon: Shouldn’t we be able to get rid of `jsdom` now?

mjw56: Yeah, we can safely remove ```jsdom``` now. Good catch, thanks. I've updated the PR with 0baa696 to reflect this.

timdorr:
Can this be reworked to also remove/replace [the existing React Test Utils tests](https://github.com/reactjs/redux/tree/master/examples/todomvc/test/components)?

mjw56:
@timdorr sure, we can do that. If anyone else is interested, please feel free to chime in. I wouldn't be able to start these until tomorrow.

mjw56:
@timdorr I started by converting one file, does this look ok to you?

timdorr:
Thanks for the efforts @mjw56. We're going to go down the route of using react-test-renderer instead.

PR #1494, opened 2016-03-08
add tree view example tests (for #1481)

fshowalter:
Adds a component spec to the tree view example.


gaearon:
Thanks, great work!

PR #1495, opened 2016-03-08
Add babel 6 instructions in WritingTests.md

p16:


timdorr:
I don't think this is relevant to the scope of this file. If you're starting to write tests and already have Babel set up, you should have already set up your .babelrc at this point.

p16:
@timdorr Fair enough :) 

I'm not an expert in redux nor babel, I faced the issue while starting to test my redux app and it took me a bit to figure out why it wasn't working :) That's why I thought it could be useful to others as well.

PR #1500, opened 2016-03-09
Added new dev dependencies to package.json

jennabelle:
Added 'webpack-dev-middleware' and 'webpack-hot-middleware' to package.json because lines 2-3 of server.js requires them both


timdorr:
You're looking at the wrong package.json. They're in [the local package.json file](https://github.com/reactjs/redux/blob/master/examples/shopping-cart/package.json).

PR #1503, opened 2016-03-09
add todos-with-undo enzyme tests

whoisandie:
Add missing action and reducer tests along with enzyme component tests.


timdorr:
Thanks for the efforts, @whoisandie. We're going to go down the route of using react-test-renderer instead!

PR #1508, opened 2016-03-10
Fix docs, response instead of req in ExampleRedditAPI.md

frooeyzanny:
Just `req` -> `response`.


mxstbr:
Good catch, that is indeed more in line with the [other](https://github.com/reactjs/redux/blob/0613a94c95eda272c7fca2292875f46b63994b68/docs/recipes/ReducingBoilerplate.md#L210) [examples](https://github.com/reactjs/redux/blob/7edf4bd9227bcc7da64709b519b04a6e78ee946f/examples/real-world/middleware/api.js#L29)!

mxstbr:
Thanks very much for taking the time!

PR #1516, opened 2016-03-11
Update ComputingDerivedData.md

ellbee:
Here's an update for the Computing Derived Data doc that I have been tinkering with for a couple of weeks. To be honest I found it pretty hard to write, any criticisms or suggested improvements are very welcome!

### DIFF
--- a/docs/recipes/ComputingDerivedData.md
+++ b/docs/recipes/ComputingDerivedData.md
@@ -1,6 +1,6 @@
 # Computing Derived Data
 
-[Reselect](https://github.com/faassen/reselect.git) is a simple library for creating memoized, composable **selector** functions. Reselect selectors can be used to efficiently compute derived data from the Redux store. 
+[Reselect](https://github.com/reactjs/reselect) is a simple library for creating memoized, composable **selector** functions. Reselect selectors can be used to efficiently compute derived data from the Redux store.
 
 ### Motivation for Memoized Selectors
 
@@ -54,7 +54,7 @@ We would like to replace `getVisibleTodos` with a memoized selector that recalcu
 
 Reselect provides a function `createSelector` for creating memoized selectors. `createSelector` takes an array of input-selectors and a transform function as its arguments. If the Redux state tree is mutated in a way that causes the value of an input-selector to change, the selector will call its transform function with the values of the input-selectors as arguments and return the result. If the values of the input-selectors are the same as the previous call to the selector, it will return the previously computed value instead of calling the transform function.
 
-Let's define a memoized selector named `getVisibleTodos` to replace the non-memoized version above:
+Let’s define a memoized selector named `getVisibleTodos` to replace the non-memoized version above:
 
 #### `selectors/index.js`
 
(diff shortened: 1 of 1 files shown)


docs/recipes/ComputingDerivedData.md
```diff
@@ -130,6 +130,194 @@ const VisibleTodoList = connect(
...
+```
+
+We also need a way to give each instance of a container access to its own private selector. The `mapStateToProps` argument of `connect` can help with this.
+
+**If the `mapStateToProps` argument supplied to `connect` returns a function instead of an object, it will be used as a factory function to create an individual `mapStateToProps` function for each instance of the container.**
+
+In the example below `mapStateToPropsFactory` creates a new `getVisibleTodos` selector, and returns a `mapStateToProps` function that has exclusive access to the new selector:
+
+```js
+const mapStateToPropsFactory = () => {
```
gaearon: I’m a bit concerned we are calling the same concept “factory” and “creator” on the same page.
Maybe we could change that to `makeGetVisibleProps()`, `makeMapStateToProps()`?

ellbee: Yes, fair point.

docs/recipes/ComputingDerivedData.md
```diff
@@ -130,6 +130,194 @@ const VisibleTodoList = connect(
...
+
+const VisibleTodoList = connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(TodoList)
+
+export default VisibleTodoList
+```
+
+A selector created with `createSelector` only returns the cached value when its set of arguments is the same as its previous set of arguments. If we alternate between rendering `<VisibleTodoList listId="1" />` and `<VisibleTodoList listId="2" />`, the shared selector will alternate between receiving `{listId: 1}` and `{listId: 2}` as its `props` argument. This will cause the arguments to be different on each call, so the selector will always recompute instead of returning the cached value. We'll see how to overcome this limitation in the next section.
```
gaearon: Nit: can you please use `’` instead `'`

docs/recipes/ComputingDerivedData.md
```diff
@@ -130,6 +130,194 @@ const VisibleTodoList = connect(
...
+We also need a way to give each instance of a container access to its own private selector. The `mapStateToProps` argument of `connect` can help with this.
+
+**If the `mapStateToProps` argument supplied to `connect` returns a function instead of an object, it will be used to create an individual `mapStateToProps` function for each instance of the container.**
+
+In the example below `makeMapStateToProps` creates a new `getVisibleTodos` selector, and returns a `mapStateToProps` function that has exclusive access to the new selector:
+
+```js
+const makeMapStateToProps = () => {
+  const getVisibleTodos = makeGetVisibleTodos();
+  return (state, props) => {
```
gaearon: Maybe declare it explicitly?

```js
const mapStateToProps = ...
return mapStateToProps
```

docs/recipes/ComputingDerivedData.md
```diff
@@ -130,6 +130,194 @@ const VisibleTodoList = connect(
...
+In order to share a selector across multiple `VisibleTodoList` components **and** retain memoization, each instance of the component needs its own private copy of the selector.
+
+Let's create a function named `makeGetVisibleTodos` that returns a new copy of the `getVisibleTodos` selector each time it is called:
+
+#### `selectors/todoSelectors.js`
+
+```js
+import { createSelector } from 'reselect'
+
+const getVisibilityFilter = (state, props) => state.todoLists[props.listId].visibilityFilter
```
gaearon: Let’s put some line breaks here, it’s getting too long.
```js
const getVisibilityFilter = (state, props) =>
  state.todoLists[props.listId].visibilityFilter
```

same for `getTodos`

docs/recipes/ComputingDerivedData.md
```diff
@@ -130,6 +130,201 @@ const VisibleTodoList = connect(
...
+We also need a way to give each instance of a container access to its own private selector. The `mapStateToProps` argument of `connect` can help with this.
+
+**If the `mapStateToProps` argument supplied to `connect` returns a function instead of an object, it will be used to create an individual `mapStateToProps` function for each instance of the container.**
+
+In the example below `makeMapStateToProps` creates a new `getVisibleTodos` selector, and returns a `mapStateToProps` function that has exclusive access to the new selector:
+
+```js
+const makeMapStateToProps = () => {
+  const getVisibleTodos = makeGetVisibleTodos()
+  return (state, props) => {
```
gaearon: Can you change it here, too?

```js
const mapStateToProps = ...
return mapStateToProps
```

gaearon:
Great work. Will merge after this last nit. Thank you very much.

docs/recipes/ComputingDerivedData.md
```diff
@@ -1,10 +1,10 @@
 # Computing Derived Data
 
-[Reselect](https://github.com/faassen/reselect.git) is a simple library for creating memoized, composable **selector** functions. Reselect selectors can be used to efficiently compute derived data from the Redux store. 
+[Reselect](https://github.com/reactjs/reselect.git) is a simple library for creating memoized, composable **selector** functions. Reselect selectors can be used to efficiently compute derived data from the Redux store.
 
 ### Motivation for Memoized Selectors
 
-Let's revisit the [Todos List example](../basics/UsageWithReact.md):
+> The examples in this section are based on the [Redux Todos List example](http://redux.js.org/docs/basics/UsageWithReact.html).
```
gaearon: What is the reason for this link change?

docs/recipes/ComputingDerivedData.md
```diff
@@ -130,6 +130,202 @@ const VisibleTodoList = connect(
...
+  makeMapStateToProps,
+  mapDispatchToProps
+)(TodoList)
+
+export default VisibleTodoList
+```
+
 ## Next Steps
 
 Check out the [official documentation](https://github.com/rackt/reselect) of Reselect as well as its [FAQ](https://github.com/rackt/reselect#faq). Most Redux projects start using Reselect when they have performance problems because of too many derived computations and wasted re-renders, so make sure you are familiar with it before you build something big. It can also be useful to study [its source code](https://github.com/rackt/reselect/blob/master/src/index.js) so you don’t think it’s magic.
```
gaearon: `rackt/reselect` URL needs updating

docs/recipes/ComputingDerivedData.md
```diff
@@ -1,10 +1,10 @@
 # Computing Derived Data
 
-[Reselect](https://github.com/faassen/reselect.git) is a simple library for creating memoized, composable **selector** functions. Reselect selectors can be used to efficiently compute derived data from the Redux store. 
+[Reselect](https://github.com/reactjs/reselect.git) is a simple library for creating memoized, composable **selector** functions. Reselect selectors can be used to efficiently compute derived data from the Redux store.
```
gaearon: `.git` at the end is not necessary in this link

docs/recipes/ComputingDerivedData.md
ellbee: Oh, I didn't do it on purpose, it's because of a copy paste from the reselect docs. I'll switch it back.

mxstbr:
Awesome, I reckon this is ready – really useful resource!

PR #1519, opened 2016-03-14
fix: SET_VISIBILITY_FILTER should return state

axelmm:
I'm not sure ... but found this inconsistency in docs:
returning a few "Object.assign(..." followed by just "return action.filter", rather not "new state"-type - am I right?

### DIFF
--- a/docs/basics/Reducers.md
+++ b/docs/basics/Reducers.md
@@ -250,7 +250,9 @@ Let’s explore reducer composition more. Can we also extract a reducer managing
 function visibilityFilter(state = SHOW_ALL, action) {
   switch (action.type) {
   case SET_VISIBILITY_FILTER:
-    return action.filter
+    return Object.assign({}, state, {
+      visibilityFilter: action.filter
+    })
   default:
     return state
   }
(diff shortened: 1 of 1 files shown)


gaearon:
This is incorrect. In this case, the string *is* the state, so we return it. We want `action.filter` to become the next version of the state. Its initial state is also not `{ filter: 'SHOW_ALL' }`, it is `'SHOW_ALL'`.

Reducers don’t have to manage an *object* state. Sure, it’s the most common case, but state can also be a number, a string, an array, or something else.

axelmm:
OK, I should read text (already bolded) more carefully.  Correct me if I'm wrong.

Theese decomposed/combined reducers are its domain (slice of state object) type aware, don't care/manage whole state (object in this case) anymore. 

They are still fired with all kind (whole app set) of actions but in majority returning (its slice) state not affected (default) as not matching (for mutation).
There can be an action affecting a few slices of object state.

gaearon:
>Theese decomposed/combined reducers are its domain (slice of state object) type aware, don't care/manage whole state (object in this case) anymore.

Yep.

>They are still fired with all kind (whole app set) of actions but in majority returning (its slice) state not affected (default) as not matching (for mutation).

Yep.

>There can be an action affecting a few slices of object state.

Yep.

:+1:

PR #1521, opened 2016-03-16
Fix Mocha example to work with failed tests

chrisblossom:
Using the existing example, Mocha will never call done() if the test fails resulting in  `Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.`

Relevant PR: https://github.com/arnaudbenard/redux-mock-store/pull/32

### DIFF
--- a/docs/recipes/WritingTests.md
+++ b/docs/recipes/WritingTests.md
@@ -122,7 +122,7 @@ describe('async actions', () => {
     nock.cleanAll()
   })
 
-  it('creates FETCH_TODOS_SUCCESS when fetching todos has been done', (done) => {
+  it('creates FETCH_TODOS_SUCCESS when fetching todos has been done', () => {
     nock('http://example.com/')
       .get('/todos')
       .reply(200, { body: { todos: ['do something'] }})
@@ -133,13 +133,11 @@ describe('async actions', () => {
     ]
     const store = mockStore({ todos: [] })
 
-    store.dispatch(actions.fetchTodos())
+    return store.dispatch(actions.fetchTodos())
       .then(() => {
         const actions = store.getActions()
 
         expect(actions[0].type).toEqual(types.FETCH_TODOS_REQUEST)
-
-        done()
       })
   })
 })


gaearon:
Let’s use `.then(done, done)` instead?

chrisblossom:
I thought .then(done).catch(done) was more self explanatory for documentation purposes. Fixed as requested.

gaearon:
Yeah, I’m not really sure what people prefer. I try to avoid `.then(cont).catch(cont2)` because usually I don’t want `cont2` to catch exceptions occurring in `cont`. In this case it doesn’t matter. Also `Promise.prototype.catch()` is not part of Promises A+. In this case it doesn’t matter either as they are part of the language now.

I don’t know. What do others think? If `then(done).catch(done)` is more explanatory then perhaps we should use that instead.

chrisblossom:
I've have avoided the .then(success, fail) because of https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns (The .then(success, fail) anti-pattern
 section). Feel it is a bit more succinct as well.

gaearon:
I still don’t understand why that is claimed to be an “anti-pattern” to be honest.
I write code like

```
callAPI().then(
  (result) => dispatch({ type: 'SUCCESS', result }),
  (error) => dispatch({ type: 'ERROR', error })
)
```

all the time and it exactly the behavior I want.

In this context,

```
callAPI().then(
  (result) => dispatch({ type: 'SUCCESS', result })
).catch(
  (error) => dispatch({ type: 'ERROR', error })
)
```

would be wrong because I would **not** want runtime errors caused by side effects from dispatch of `SUCCESS`, such as an error thrown by a component or a reducer, to be reported to the user as an app error. I would want it to be unhandled, and to be reported to the crash analytics as an unhandled rejection.

(cc @ForbesLindesay who I believe had thoughts on this)

fshowalter:
Mocha understands promises, just return the promise and drop `done()` entirely.

```
// Promise test example with mocha and expect
it('should execute promise', () => {
    function success() {
      return {
        type: 'FETCH_DATA_SUCCESS'
      };
    }

    function fetchData() {
      return dispatch => {
        return fetch('/users.json') // Some async action with promise
          .then(() => dispatch(success()));
      };
    }

    const store = mockStore({});

    return store.dispatch(fetchData())
      .then(() => {
        expect(store.getActions()[0]).toEqual(success())
      })
})
```

chrisblossom:
@fshowalter thanks for the clarification. Updated PR.

jsmapr1:
I was just coming here because of this issue. I noticed the expected actions is not used anymore. It should work if you run this. That will test the FETCH_TODOS_SUCCESS is called and the response is passed. Instead of just testing FETCH_TODOS_REQUEST.

```
 it('creates FETCH_TODOS_SUCCESS when fetching todos has been done', () => {
    nock('http://example.com/')
      .get('/todos')
      .reply(200, { body: { todos: ['do something'] }})

    const expectedActions = [
      { type: types.FETCH_TODOS_REQUEST },
      { type: types.FETCH_TODOS_SUCCESS, body: { todos: ['do something']  } }
    ]
    const store = mockStore({ todos: [] })

    return store.dispatch(actions.fetchTodos())
      .then(() => {
        const actions = store.getActions()
        expect(actions).toEqual(expectedActions)
      })
  })
```

ModuloM:
Did you test what you've proposed @jsmapr1?
Because it didn't work for me :(

But this does (with Mocha exemple this time):
```
return store.dispatch(actions.fetchTodos()).then(() => {
  expect(store.getActions()).to.deep.equal(expectedActions)
}).then(done, done)
```

jsmapr1:
I did and it worked for me and I just tried it again. Not sure why it didn't work for you, but I see that you are sill calling `done` in the next promise. You shouldn't have to do that for the reasons mentioned above. Maybe that's the problem? If it still doesn't work post the error and I'll see if I can recreate.

ModuloM:
Oh yes, I think this is because you are using Jasmine (toEqual is in Jasmine if I'm not wrong). And the problem comes from the way Mocha treats promises. In the `then` statement you should pass success callback, failure callback. So `.then(done, done)` seems to be the way to go.

jsmapr1:
Nope. I'm using mocha. toEqual is part of the [expect library](https://github.com/mjackson/expect#toequal) which is being used on these tests.

chrisblossom:
https://github.com/reactjs/redux/pull/1549 merged solving same issue.

PR #1524, opened 2016-03-17
Add shouldComponentUpdate note to UsageWithReact

joshkel:
I'm learning React + Redux and was having trouble figuring out whether I needed to implement `shouldComponentUpdate`, since performance discussions of React frequently mention it but Redux doesn't mention it at all. I suggest adding an explicit mention here in case anyone else has the same question I did.


joshwcomeau:
I believe the reason it's undocumented is because _how_ it does its optimizations is an implementation detail and subject to change.

I could be wrong though, and it could still be worth documenting. It's a question I had as well :)

gaearon:
Most of the optimizations are actually achieved by memoizing the last rendered element rather than shouldComponentUpdate for reasons that take quite a bit of time to explain. So I wouldn’t go into details there. Can we amend the text to just mention React Redux prevents unnecessary re-renders without explaining how?

joshkel:
Maybe I should explain my rationale a bit more...

The React [Advanced Performance](https://facebook.github.io/react/docs/advanced-performance.html) guide talks about implementing shouldComponentUpdate or using PureRenderMixin to improve performance, as long as state/props are immutable.

Since Redux uses immutable state, trying to figure out how to apply the Advanced Performance guide to Redux made me wonder if I still needed to worry about implementing shouldComponentUpdate or PureRenderMixin for my Redux components. Based on my reading of the Advanced Performance guide, it seemed like a trivial optimization to make (and so the kind of thing that Redux should do itself), but the Redux docs didn't give any details on whether or not it did. I didn't want to not make trivial optimizations, but I didn't want to waste time reimplementing what Redux did, so I ended up going source-diving to try and satisfy myself.

If you don't want to go into details, then is this revision any better? (Or would there be a better section of the docs for addressing this? Or am I worrying too much, and it doesn't need to be addressed at all?)

> You could write a container component by hand, but we suggest instead generating container components with the React Redux library's [`connect()`](https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options) function, which provides many useful optimizations to prevent unnecessary re-renders.  (One result of this is that you shouldn't have to worry about the [React performance suggestion](https://facebook.github.io/react/docs/advanced-performance.html) advice of implementing `shouldComponentUpdate` yourself.)

gaearon:
Yeah this looks fine. :+1: 
Please make sure you don’t have a double space after editing (it exists in the original commit).
Thanks!

joshkel:
I updated the wording. Thank you.

docs/basics/UsageWithReact.md
```diff
@@ -245,7 +245,7 @@ export default App
...
+Now it’s time to hook up those presentational components to Redux by creating some containers. Technically, a container component is just a React component that uses [`store.subscribe()`](../api/Store.md#subscribe) to read a part of the Redux state tree and supply props to a presentational component it renders. You could write a container component by hand, but we suggest instead generating container components with the React Redux library’s [`connect()`](https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options) function, which provides many useful optimizations to prevent unnecessary re-renders. (One result of this is that you shouldn't have to worry about the [React performance suggestion](https://facebook.github.io/react/docs/advanced-performance.html) of implementing `shouldComponentUpdate` yourself.)
```
gaearon: Nit: `shouldn't` -> `shouldn’t`

joshkel: Fixed. Thanks for catching that.

gaearon:
Thanks!

PR #1525, opened 2016-03-17
add async example tests (for #1481)

fshowalter:
This adds specs for the async example. It required a few changes:

* actions
  * `isomorphic-fetch` changed to be called via it's global in order to mock it with `fetch-mock`
* `App.js` 
   * changed to export the un-connected component for test.
   * changed to pass it's action creators via props in order to mock them in the specs.
   * changed to fetch in  `componentWillMount` instead of `componentDidMount` to test via shallow-rendering. This keeps with React's advice about when to fetch data, so win-win.


timdorr:
This will need to be rewritten because of #1883. Please resubmit!

Issue #1531, opened 2016-03-18
installation issue


missinglink:
Is node >12 required? I couldn't find a reference to it in the readme or an 'engines' section in package.json

```
$ npm i

...
> cross-env BABEL_ENV=es babel src --out-dir es

/redux/build/use-lodash-es.js:4
      ImportDeclaration(path) {
                       ^
SyntaxError: Unexpected token (
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at /redux/node_modules/babel-core/lib/transformation/file/options/option-manager.js:177:20
    at Array.map (native)
    at Function.normalisePlugins (/redux/node_modules/babel-core/lib/transformation/file/options/option-manager.js:155:20)
...
```

```
$ node --version
v0.12.12

$ npm --version
2.14.9
```

```
$ git log -1
commit 175f32a99368980e24a4277ad9a3b2a4d012331f
Merge: 80d11be 74b1188
Author: Tim Dorr <timdorr@timdorr.com>
Date:   Fri Mar 11 13:16:20 2016 -0500

    Merge pull request #1515 from arnaudbenard/update-redux-mock-store-docs
    
    update redux-mock-store documentation
```

timdorr:
[We only test Node 4.x.x and v5.x.x](https://github.com/reactjs/redux/blob/master/.travis.yml#L3-L4). You will need one of those versions to build Redux.

missinglink:
Adding an [engines section](https://docs.npmjs.com/files/package.json#engines) to your npm will emit a warning to old-schooler like me :)

timdorr:
Could you submit a PR to add that? Seems like low-hanging fruit if you want to make a contribution :smile: I would do it, but I'm mobile.

missinglink:
no problem, here you go https://github.com/reactjs/redux/pull/1532

on closer inspection it might be better handled with an entry to the readme since the way npm handles that flag is inconsistent.

gaearon:
It is required for contributing. We should add this information to https://github.com/reactjs/redux/blob/master/CONTRIBUTING.md.

It is, however, not required for using Redux. `npm install redux` will work fine.

PR #1532, opened 2016-03-18
add engines section to package.json

missinglink:
re: https://github.com/reactjs/redux/issues/1531

this setting *should* emit a warning to users who are still running node <4

...although according to [this](http://www.marcusoft.net/2015/03/packagejson-and-engines-and-enginestrict.html) that isn't exactly true

when running `git clone && npm i` it doesnt work *but* when installing via `npm i redux` it *does*.

¯\_(ツ)_/¯

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,6 +104,9 @@
     "typescript-definition-tester": "0.0.4",
     "webpack": "^1.9.6"
   },
+  "engines": {
+    "node" : ">=4"
+  },
   "npmName": "redux",
   "npmFileMap": [
     {


timdorr:
@gaearon Is this OK to add? It currently fails to build on older versions of Node.

gaearon:
I don’t think we want this. Yeah you can’t contribute or run tests before Node 4, but you can *use* it just fine which I think what `engines` is all about. You can install Redux on 0.10 and run it if you wish.

timdorr:
@missinglink You were attempting to build Redux, not use it in a project, right? I think that's what I was missing from the previous issue. Built versions work in 0.10 and up, but you'll need 4 and higher to develop.

missinglink:
agh ok thanks, that solves the issue for me, thanks!

PR #1535, opened 2016-03-20
Async example using redux-saga

StuartHarris:
This is a new example that is an exact port of the `async` example, except that it uses `redux-saga` instead of `redux-thunk`. This can improve testability (no mocking or dependency injection needed) and can allow cleaner separation between UI and long lived flows. Also action creators can now just create synchronous actions and need not concern themselves with doing any work.

All the asynchronous coordination is in `sagas/reddit` and is fully tested. No changes have been made to reducers or components.

- [x] port `async` example to use `redux-saga`
- [x] add tests
- [x] move to new example (`async-saga`)


StuartHarris:
Oops! Please don't pull this! It should have been raised against our fork's master.
It's just an example of using `redux-saga` :-)

StuartHarris:
Just realised this has already been done :-) https://github.com/yelouafi/redux-saga/tree/master/examples/async

PR #1536, opened 2016-03-21
Update examples to use functional components

nickuraltsev:


examples/real-world/containers/Root.dev.js
```diff
@@ -1,24 +1,20 @@
...
-      <Provider store={store}>
-        <div>
-          <Router history={history} routes={routes} />
-          <DevTools />
-        </div>
-      </Provider>
-    )
-  }
-}
+const Root = ({ store, history }) =>
```
Aweary: Is React able to parse the `displayName` from named arrow functions? If not it might be good to add a`displayName` property to the exported functions for stuff like `ReactPerf`.

Unless the new `ReactPerf` that @gaearon is working on handles that differently?

nickuraltsev:
@Aweary I've just checked that `ReactPerf` shows the correct names of the functional components in the updated examples. Please see below:

reactperf

timdorr:
This was a part of #1800, which was superceded by #1883. Can you refactor and resubmit now that the examples have changed quite a bit?

PR #1538, opened 2016-03-22
Corrected assertion

coryhouse:
toEqual doesn't exist in Chai. to.equal does though. Perhaps this assumed a different assertion library? 

Regardless, must use deep.equal in Chai to compare objects - otherwise Chai will do a reference type comparison, so the existing example would fail.

### DIFF
--- a/docs/recipes/WritingTests.md
+++ b/docs/recipes/WritingTests.md
@@ -61,7 +61,7 @@ describe('actions', () => {
       type: types.ADD_TODO,
       text
     }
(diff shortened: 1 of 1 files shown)


kjbekkelund:
It's not using Chai, it's using https://www.npmjs.com/package/expect, which has `toEqual`. You'll see this a bit above your fix:

```
import expect from 'expect'
```

coryhouse:
How'd I miss that?! Sigh. Thanks for the quick reply.

PR #1539, opened 2016-03-24
Update AsyncActions.md

ntdb:
The `isFetching` value should be checked first to avoid multiple requests.


ntdb:
Well... I realized the problem with this immediately after submitting it. Sorry for the noise!

PR #1545, opened 2016-03-25
Added index.d.ts to package.json

ahz:
Maybe I am missing something but it seems index.d.ts should be added to package.json to get into distributed npm package.


mxstbr:
Good catch, thanks very much @ahz!

gaearon:
Out in 3.4.0.

Issue #1469, opened 2016-03-02
Optimize TravisCI runs


zpao:
If you look at https://travis-ci.org/reactjs/redux/builds, it's taking > 40 minutes to run TravisCI tests. This is incredibly long and it's making me fatigued.

Can we do something to optimize these runs?

gaearon:
Maybe it’s worth removing `build:examples` and `test:examples` from the CI. Examples are taking most of the time, especially their independent `npm install`s.

naw:
I'm not very familiar with TravisCI, but if you're using their container-based infrastructure, you should be able to cache the node_modules folder for each separate example.

https://docs.travis-ci.com/user/workers/container-based-infrastructure/

https://docs.travis-ci.com/user/caching

zpao:
I agree that caching could help out here, but be careful of how cache invalidation is done (or not done, which is where it hurts). I've seen cases where CI continued fine but a patch version upgrade of a npm package broke things so we didn't know for a while. I think the top-level caching of node_modules is supposed to work and get invalidated when your deps change but that might be wrong.

naw:
If I'm understanding correctly, you could prevent such a scenario by having CI execute `npm update` instead of `npm install` on top of the (potentially stale) cached packages. 

This would result in the _exact_ package versions that would be installed by `npm install` when there is no cache, wouldn't it?

gaearon:
I haven’t looked into those links yet but I think we should, if possible,

* *not* cache the root `node_modules` for extra safety
* cache `node_modules` for examples, as they are not critical

timdorr:
> not cache the root node_modules for extra safety

It shouldn't be caching `node_modules` by default, just the `~/.npm`. That gives a boost to the download and native code build times for everything, which is about 164MB on my current system. As new modules are released within the semver restrictions set up within `package.json`, they will be downloaded fresh and added to the cache for the next build. 

Plus, we've gotten on with this in place all this time with no issues that I can remember :smile:

timdorr:
OR what if we move the examples out to another repo? There are fairly regular additions, so a monorepo format for a `redux-examples` repo would scale well without affecting the core project's builds.

gaearon:
I’d rather keep the examples here.

@timdorr Can you look into optimizing the speed of Travis? Maybe enabling caching for `node_modules` inside examples would be enough but I don’t know.

timdorr:
[...]

PR #1548, opened 2016-03-27
avoid npm install for each example to make build faster

just-boris:
This goes as addition to #1469

Now you are doing npm install twice for each example. I guess that single npm install at the start of a build is enough. All necessary modules for examples can be resolved from the single `node_modules` directory at project root. It allows to not duplicate some huge packages like `webpack` and `babel` with lots of dependencies.

**UPD** Build took 6 minutes. Doesn't work on Node 4 because of npm dedupe must be called before build that starts automatically after `npm install`


just-boris:
Properly enumerated all dependencies, I made it work with npm v2 as well.

Both builds on npm 2 & 3 are faster than current.

mxstbr:
I'm not sure what the reasoning was behind each example having it's own `package.json`, but I'm sure there was one. @gaearon?

gaearon:
The reasoning is so that people can copy paste examples independently as standalone projects to tweak, and get a better sense of how a single project is organised.

Also making something like react-router a dev dependency of Redux just seems wrong. I see where you're coming from but we used to have examples without their own package.json in the beginning, and this confused people. ("I copied the project and now it doesn't run!")

So I'm closing. Thanks for taking time to look into this.

PR #1515, opened 2016-03-11
update redux-mock-store documentation

arnaudbenard:
I have updated https://github.com/arnaudbenard/redux-mock-store to be framework agnostic and have less errors.

This PR updates the documentation with the new syntax.

PR #1549, opened 2016-03-28
Update docs - Writing tests - Async actions text example

ro-savage:
TLDR: **This PR fixes the test in Async Actions documentation to correctly pass/fail**

The docs (but not the [website](http://redux.js.org/docs/recipes/WritingTests.html)) were recently updated for v2 of [redux-mock-store](https://github.com/arnaudbenard/redux-mock-store) by the creator @arnaudbenard himself in #1515.

However best I can tell the changes do not appear to be correctly testing the async actions.

They appear to be:
- Not testing that the fetch was called.
- Not testing the entire action, only the type.
- Only testing 'FETCH_TODOS_REQUEST' instead of 'FETCH_TODOS_REQUEST' and 'FETCH_TODOS_SUCCESS'.

This PR fixes the test to correctly pass/fail.

Please let me know if I am mistaken.

I would also be great if the website was updated with master when you get the chance, so that people aren't trying to follow the docs with the old api.

### DIFF
--- a/docs/recipes/WritingTests.md
+++ b/docs/recipes/WritingTests.md
@@ -134,13 +134,11 @@ describe('async actions', () => {
     const store = mockStore({ todos: [] })
 
     store.dispatch(actions.fetchTodos())
-      .then(() => {
-        const actions = store.getActions()
-
-        expect(actions[0].type).toEqual(types.FETCH_TODOS_REQUEST)
-
-        done()
+      .then(() => { // return of async actions
+        expect(store.getActions()).toEqual(expectedActions)
       })
+      .then(done) // test passed
+      .catch(done) // test failed
   })
 })
 ```


arnaudbenard:
I agree with the `done` in the `catch` statement, good find. It was updated in the README after I made the PR.

```
Not testing the entire action, only the type.
Only testing 'FETCH_TODOS_REQUEST' instead of 'FETCH_TODOS_REQUEST' and 'FETCH_TODOS_SUCCESS'.
```

It's more of an implementation choice. In the previous version you could only test the whole action and it was a problem for many users. For the sake of clarity we could put the tests and all the actions.

ro-savage:
Thanks for the comments @arnaudbenard

Ot could be rewritten i as:

```js
.then(() => { // return of async actions
  const actions = store.getActions() // Get he actions that were created

  expect(actions[0].type).toEqual(types.FETCH_TODOS_REQUEST) // You can test directly against the action type

  expect(actions[1]).toEqual(expectedActions[1])  // Or against the entire mocked payload
})
```

If you think it would be useful for people to see examples of both?

mxstbr:
I like the `done` in the `catch`, it's clearer and more concise that way. :+1:

arnaudbenard:
@ro-savage You are welcome to add more examples in the `redux-mock-store` repo if you think some are missing :+1:

PR #1556, opened 2016-03-30
update compose function for more lambda style

jabez128:
update compose function for more lambda style

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -9,4 +9,4 @@
  * (...args) => f(g(h(...args))).
  */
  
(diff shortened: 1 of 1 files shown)


andreieftimie:
Much harder to read and maintain having everything in a single line.

gaearon:
Thanks for the PR! I think we want to be a little more explicit in the intent. The version you propose strike me as more elegant but harder to understand to people looking at this code for the first time so I decided not to take this. Nevertheless I appreciate the time you took to contribute!

PR #1563, opened 2016-03-31
no need to use 'store' namespace

jabez128:
no need to use 'store' namespace


gaearon:
Thanks for the PR! We tweaked those lines so this is actually needed again. I'm closing for that reason.

PR #1566, opened 2016-04-01
Simpler with mocha's built-in promise support

hoegrammer:
Updated test docs following http://tobyho.com/2015/12/16/mocha-with-promises/


mxstbr:
Thanks @hoegrammer, this is a lot nicer!

PR #1571, opened 2016-04-03
FAQ page updates

markerikson:
Added a Table of Contents, plus shorter anchor links than Github generates.  Some additional editing and article links.  Also added a new Gitbook plugin to mostly fix trying to go directly to a link with a hash anchor in it.

### DIFF
--- a/book.json
+++ b/book.json
@@ -3,7 +3,7 @@
   "structure": {
     "summary": "docs/README.md"
   },
-  "plugins": ["edit-link", "prism", "-highlight", "github"],
+  "plugins": ["edit-link", "prism", "-highlight", "github", "anker-enable"],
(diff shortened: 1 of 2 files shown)


markerikson:
Merged and published.  Appears to be working well - the TOC shows up in the docs page, and opening a new tab with a URL like http://redux.js.org/docs/FAQ.html#performance-clone-state goes directly to that question.  I think that's going to be a _very_ nice improvement to the FAQ's usefulness.  Hand someone a direct link to the question, with a shorter URL than "#here-is-the-entire-slugified-text-of-the-original-question", and they can go straight to it.

gaearon:
Looks great. Thank you!

PR #1572, opened 2016-04-03
Adding Symbol For Actions

ev1stensberg:
Hi! Been trying to add Symbol as a feature of Redux. As you know, symbols are immutable and this would be a great 2nd choice for people that will have static actions to use. While implementing this, I ran into two issues.

1." Move function declaration to program root " - This has to be adjusted to have symbols as a second action creator.
2. Invalid code -  I'm not sure if the code is 100%. What we could do, is to make functions out of the mapping of the symbols for `Object.key(mykey)` , for it to fully function.

I decided make two different instances of the actioncreator. The first try of embedding a for loop inside this snippet:

 ` for(var i = 0; i < keys.length; i++) {
    var key = keys[i]
    var actionCreator = `actionCreators[key]

    if(typeof actionCreator === 'function') {
      boundActionCreators[key] = bindActionCreator(actionCreator, dispatch)
    }
  }
  return boundActionCreators
}`

...  was ok ( a for loop within the for loop), but it sent two requests to the dispatcher/actioncreator head function, which resulted in an error in the debugger. The solution was to make two separate actioncreator functions, but also, this would be embedded as an conditional if the first actioncreator could not find an object or an function, like a fallback. This also failed since the debugger wanted it to be passed to the "program root". 

**So, why this PR?**

 1. Support Symbols
2. Addressing fallback actioncreators

Also, if someone would like to fix this, please, do help, it would be awesome!  :tiger: 

Let me add that you should watch the PR code, since I'm terrible to explain stuff.


markerikson:
(I'll leave managing this PR to someone who's been around longer than I have.)

While the contribution and effort is appreciated, I doubt that it will be accepted.  One of Redux's explicit goals is that actions should be serializable, and using Symbols in actions generally goes against that basic principle.  See the [FAQ answer discussing serializable actions](http://redux.js.org/docs/FAQ.html#why-should-type-be-a-string-or-at-least-serializable-why-should-my-action-types-be-constants) for more information.

ev1stensberg:
Yea! But symbols are immutable, it could open up the possibility to have nested actions as one single action. Also it avoids clashes and could spare you for the same action name conflicts in large apps. However, they aren't 100% pure, which is a bit shady. I also, doubt it, but it would be such a nice thing to have, maybe I'll convert it into a library.

Phoenixmatrix:
If immutability is your main requirement, I wouldn't be surprised if using an Immutable.js record would just work. If you don't care about dev tools and serialization, and just want the uniqueness aspect, you can use a symbol as the value of the type of the action.

Though I'll be honest, countless people make the leap that actions should be symbols, before realizing the tradeoffs just aren't worth it.

ev1stensberg:
To clear up, my intended use case for symbols would be to add nesting actions into an existing one instead of using Object.assign etc...

PR #1559, opened 2016-03-31
[WIP] refactor to eslint airbnb, remove babel-eslint

knowbody:
This is a work in progress on moving to `eslint-config-airbnb` and removing `babel-eslint`

#### Todo
- [x] - remove `babel-eslint`
- [x] - remove `eslint-config-rackt`
- [x] - add `eslint-config-airbnb`
- [x] - fix the linting according to airbnb style guide
- [ ] - merge awaitng PRs (as per: https://github.com/reactjs/redux/issues/1554#issuecomment-203576976), and apply new eslint config to them
- [ ] - apply new linting to the docs

@gaearon let me know if you have any feedback on this

PR #1574, opened 2016-04-03
Added ES6 scope and error log

ev1stensberg:
Added local scoping from ES6 syntax with const to static variables. Also added error log to getState function to notify if the user decides to do something tricky.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -73,7 +73,10 @@ export default function createStore(reducer, initialState, enhancer) {
   function getState() {
     return currentState
   }
-
+  
+  if (typeof getState !== 'function' || getState == 'undefined') {
+    throw new Error('Expected the getState to be a function.')
+  }
   /**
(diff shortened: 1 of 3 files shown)


src/createStore.js
```diff
@@ -73,7 +73,10 @@ export default function createStore(reducer, initialState, enhancer) {
   function getState() {
     return currentState
   }
-
+  
+  if (typeof getState !== 'function' || getState == 'undefined') {
```
gaearon: `getState` is a locally defined function. It is guaranteed to be a function by the code unless I’m missing something.

gaearon:
I’m up for replacing `var` with `const` or `let` but we should do it consistently across the codebase, not just in a few methods. This is already being done in #1559 so I’ll close this out. Thanks for the PR!

ev1stensberg:
Yeah, did that at first, but considering some variables were used across scopes, I left them as variables, but Mateusz covers this smoothly!

PR #1575, opened 2016-04-03
Added Error log

ev1stensberg:
Added error log if user decides to be a superhero.

### DIFF
--- a/src/bindActionCreators.js
+++ b/src/bindActionCreators.js
@@ -36,6 +36,15 @@ export default function bindActionCreators(actionCreators, dispatch) {
   }
 
   var keys = Object.keys(actionCreators)
+
+  if (typeof keys !== 'object' || !keys instanceof Function || keys === null) {
+
+    throw new Error(
+      `Properties of keys should be enumerable values. Have you modified 'keys'
+      in the source code with slice/push or any other method?`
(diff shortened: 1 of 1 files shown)


gaearon:
I’m a bit confused. `keys` is a local variable obtained from `Object.keys()`. How can the user modify it?
We could potentially warn if `keys` is an empty array but I’m not sure it’s worth the trouble either.

ev1stensberg:
This is just a dummy-check to warn/output something if the user is doing anything that he/she should not do. I added warning at first, but not sure if the user should pass the build if he/she has tweaked it like that. Anyhow, I can change it to warn if the array is empty or if he/she has done anything stupid, but this is a waste of code and it will take space. Anyways, thanks for the response! -Closing this PR.

gaearon:
I think if the user changes the source, they’re on their own. We can’t reliably protect against such cases, and I don’t think we should. I’ve never seen a library do that. I mean, if they want to change how it works, presumably they can also remove any checks we add to prevent this :wink: .

ev1stensberg:
Yeah, true. I think some might have it. In use cases where they want the user to tweak the library on their own while still having a functionality they can follow. But that's extremely rare, I assume :smile:

Issue #465, opened 2015-08-11
Middleware not dispatched through on init


phated:
I'm not sure if this has been discussed before (quick search got me no where), but I am noticing that no middleware is executed when redux does its `init` actions.  Is this on purpose or just something that has come out of the functional composition nature of redux?  I would like to use middleware to ensure payload objects on each of my actions, but the `init` actions cause things to explode due to not passing through middleware.

acdlite:
> Is this on purpose or just something that has come out of the functional composition nature of redux?

Both :) `init` action is a no-op action that exists solely to prime the store with an initial state, using default parameters with reducers:

```js
function reducer(state = initialState, action) {...}
```

Since `state` is undefined at initialization, and `init` is a no-op action, `initialState` is returned as... the initial state.

> I would like to use middleware to ensure payload objects on each of my actions, but the init actions cause things to explode due to not passing through middleware.

Not sure I understand what the problem is. Could you elaborate?

acdlite:
For context, here's where the init action is dispatched: https://github.com/gaearon/redux/blob/breaking-changes-1.0/src/createStore.js#L140

phated:
The thing that lead me down this path was just a code-style preference in my reducers.

This is a poorly thought out example of the style I was looking to achieve.
```js
function status(state = initial, { type, payload }){
    const { notification, isError } = payload;

    switch(type){
        case UPDATE_STATUS:
            return { notification, isError };
        default:
            return state;
    }  
}
```

As you can probably tell, moving the destructuring from payload to the top of the reducer keeps the switch statements much simpler/cleaner, especially as things start to grow, or assignment is being done, etc.

However, this pattern fails because on Redux's `init`, there is no payload.  I __definitely__ don't think Redux should add a payload property to the actions it dispatches for `init`, but I would like to put a middleware in front of my reducers that guarantees a payload property that is an object.  This still falls apart because the `init` action isn't passed through middleware.

I understand that the `init` action is a no-op, but middleware is supposed to act upon actions (whether they are no-ops or not), so it seems strange that middleware is skipped for initialization.

Another thing that I would probably want in the future is to log the `init` actions to know everything is booting up correctly throughout the system.  This could be logged by logging middleware if `init` was passed through the middleware stack.

I understand the example I gave that got me to this place could be solved by using a default for the `payload` property of my destructure, but hopefully we can look at this from a high-level perspective and not get bogged down in my simple example.

acdlite:
[...]

PR #1576, opened 2016-04-04
Proof of concept: fully create store before dispatching init action.

tappleby:
Not ready to be merged in this state, looking for feedback. Proof of concept for my [comment](https://github.com/reactjs/redux/issues/465#issuecomment-205086273) in #465.

The idea is to split up creating + initializing the store into separate phases, ensuring the store enhancer is applied before the first `@@INIT` action is dispatched.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -35,20 +35,7 @@ export var ActionTypes = {
  * @returns {Store} A Redux store that lets you read the state, dispatch actions
  * and subscribe to changes.
  */
-export default function createStore(reducer, initialState, enhancer) {
-  if (typeof initialState === 'function' && typeof enhancer === 'undefined') {
-    enhancer = initialState
-    initialState = undefined
-  }
-
-  if (typeof enhancer !== 'undefined') {
-    if (typeof enhancer !== 'function') {
-      throw new Error('Expected the enhancer to be a function.')
-    }
-
-    return enhancer(createStore)(reducer, initialState)
-  }
-
+export function createStore(reducer, initialState) {
   if (typeof reducer !== 'function') {
     throw new Error('Expected the reducer to be a function.')
   }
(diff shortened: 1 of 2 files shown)


gaearon:
Wow, that’s a really neat approach. Is there any way we can forbid applying “old-style” enhancers now?

tappleby:
Ive been trying to think of a way we could throw/warn, but not sure if its possible to detect.

tomkis1:
That's exactly what I did locally, this approach has one significiant drawback, it will not ensure that `@@INIT` action is dispatched as first action (see my [comment](https://github.com/reactjs/redux/issues/465#issuecomment-205047706) in original issue)

eg. `redux-devtools` will dispatch its own init function before `redux/@@INIT` and populates appstate beforehand... this will unfortunatelly not solve our issue in `redux-elm` where we would need the first action (be that either `redux-devtools` init or redux init) dispatched with all the middlewares applied.

src/createStore.js
```diff
@@ -210,3 +192,42 @@ export default function createStore(reducer, initialState, enhancer) {
...
+    initialState = undefined
+  }
+
+  if (typeof enhancer !== 'undefined') {
+    if (typeof enhancer !== 'function') {
+      throw new Error('Expected the enhancer to be a function.')
+    }
+  }
+
+  var finalCreateStore = enhancer ? enhancer(createStore) : createStore
```
tomkis1: Enhancer can potentially dispatch an action which will be called before `ActionTypes.INIT` therefore this does not ensure that first action has all the middlewares applied.

gaearon:
>eg. redux-devtools will dispatch its own init function before redux/@@INIT and populates appstate beforehand

Can you point to where this happens? I don’t see it dispatching any initial actions in the source.

>Enhancer can potentially dispatch an action which will be called before ActionTypes.INIT therefore this does not ensure that first action has all the middlewares applied.

If we go ahead with this, we’ll forbid dispatching while enhancers/middleware are being created, kinda like #1485 but more generic.

tomkis1:
> Can you point to where this happens? I don’t see it dispatching any initial actions in the source.

https://github.com/gaearon/redux-devtools/blob/master/src/instrument.js#L151 `recomputeStates` will ensure calling that mutation beforehand, strictly speaking it does not `dispatch` the action but pre-populates app state which as i've explained for example won't work for us in `redux-elm`

> If we go ahead with this, we’ll forbid dispatching while enhancers/middleware are being created, kinda like #1485 but more generic.

If that's acceptable then this may be a way to go.

gaearon:
>https://github.com/gaearon/redux-devtools/blob/master/src/instrument.js#L151 recomputeStates will ensure calling that mutation beforehand, strictly speaking it does not dispatch the action but pre-populates app state which as i've explained for example won't work for us in redux-elm

Can we change DevTools to fix this? In other words, if the behavior in PR was accepted in Redux, can we change DevTools code to be compatible with Redux Elm?

tomkis1:
> Can we change DevTools to fix this? In other words, if the behavior in PR was accepted in Redux, can we change DevTools code to be compatible with Redux Elm?

It's not only about Redux Elm, there are workarounds but I am thinking more in general.

Think about this: It's not good idea to check whether type of Action is `@@INIT`, for user it should be transparent they should never be aware of the action, user should only know that there is **some** action which ensures proper initialization of app state.

In my opinion this is the only reason why `@@INIT` function should be executed with middlewares otherwise people should never rely on its presence. You said that we could forbid dispatching while store is being created, but this will solve only part of the problem, because the middleware can potentially provide non-`undefined` app state beforehand. 

Conclusion: Would this PR even make sense then? Because if people really want to bootstrap their app, they should probably provide their own Bootstrapping action because it's **the only reliable way** given the fact  that some Enhancer could break it.

tomkis1:
Anyway, re-visited original problem and it looks like https://github.com/reactjs/redux/issues/465#issuecomment-130418992 justifies it.

gaearon:
All about tradeoffs. Being slightly hacky in two or three projects is better in my view than forcing everyone to add a custom `dispatch()` call and update all tutorials and docs. 

As far as I remember the whole reason why we put initial action there in the DevTools is *because* of the problem this PR fixes. Unless I miss something I think it won’t be necessary if we go ahead with this PR. If this is correct it would solve both your and my problem :-)

tappleby:
@gaearon what was the original reason for the INIT action?

gaearon:
@tappleby Not sure what you mean. What was its purpose in DevTools, or generally?

tappleby:
Meant generally. Was asking because I thought we might be able to do something like `initialState = reducer(initialState, { type: ActionTypes.INIT })` but then realized thats pretty much the same thing as dispatching.

gaearon:
Closing in favor of #1702 but big thanks for providing the inspiration.

PR #1578, opened 2016-04-05
Improve typing of StoreEnhancer

Igorbek:
Following up https://github.com/reactjs/redux/pull/1526#issuecomment-205403474 here is a PR that enhance typyings for `StoreEnhancer`.

- make `StoreEnhancer` to be parameterized by state type, because it's always being called with known type of the state, so that it opens a way to create state-specific enhancers
- when a generic version of `StoreEnhancer` is needed, newly introduced `GenericStoreEnhancer` can be used, which a function with same signature but has a state type parameter; obviously it's assignable to any typed store enhancer
- `StoreEnhancer` exposes a `StoreEnhancerStoreCreator<S>` as a result, that is a stricter version of `StoreCreator` which strictly expects to get a reducer and initial state (not optionals)
- `applyMiddlewares` now returns `GenericStoreEnhancer` as the middlewares are generic
- tests that show/prove usage of the new signature have also been added

/cc @aikoven

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -194,9 +194,8 @@ export interface Store<S> {
  * @template S State object type.
  */
 export interface StoreCreator {
-  <S>(reducer: Reducer<S>, enhancer?: StoreEnhancer): Store<S>;
-  <S>(reducer: Reducer<S>, initialState: S,
-      enhancer?: StoreEnhancer): Store<S>;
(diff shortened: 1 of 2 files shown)


aikoven:
Looks great to me. 
Cc @ulfryk @Pajn

ulfryk:
Absolutely :+1:

PR #1579, opened 2016-04-05
Load data in right lifecycle event

jungejason:
It's better to load the data in `componentWillMount` instead of `componentDidMount`. When we do it in the former we are setting the status of data to be `fetching` so before `renders()` runs, so that it will actually render the loading screen.

### DIFF
--- a/docs/recipes/ReducingBoilerplate.md
+++ b/docs/recipes/ReducingBoilerplate.md
@@ -213,7 +213,7 @@ class Posts extends Component {
     )
   }
 
-  componentDidMount() {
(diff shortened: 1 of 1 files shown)


Aweary:
[Per the React docs](https://facebook.github.io/react/docs/component-specs.html#mounting-componentdidmount):

>If you want to integrate with other JavaScript frameworks, set timers using setTimeout or setInterval, **or send AJAX requests**, perform those operations in this method.

In this example the `loadData` method uses `fetch` to make an AJAX request, making `componentDidMount` the correct lifecycle method.

PR #1581, opened 2016-04-06
This could be a bug

emguy:
Let me know if I am wrong. I am still learning it.


Aweary:
Hey @emguy, this example is using [deconstructing assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment). 

`const { SHOW_ALL } = VisibilityFilters` is transpiled to `const SHOW_ALL = VisibilityFilters.SHOW_ALL` via Babel.

Your change here would be the equivalent of `const SHOW_ALL = VisibilityFilters.SHOW_ALL.SHOW_ALL` which would be incorrect.

emguy:
You are right. Thanks for pointing out.

PR #1582, opened 2016-04-06
Push reducer key to finalReducerKeys

mjw56:
This saves a call to ```Object.keys``` by explicitly pushing the reducer key to ```finalReducerKeys``` when ```typeof reducers[key] === 'function'```.

I guess this might only matter if there were a lot of reducers. Even then, I'm not sure how much of a performance boost this would provide. Either way, just wanted to note something I observed!


gaearon:
I'd say there's less chance of a mistake if we let Object.keys() do the work and this code only runs once for each reducer group so I would not do it. Thanks for the effort though!

PR #1589, opened 2016-04-07
Remove side-effects from action ADD_TODO

stenver:
When prepearing for class where we teach Redux, I came upon this acton
ADD_TODO that had side-effects. This contrasts a lot with everything I
say prior to that example where I talk a lot about how you should never
have side-effects in actions, reducers, etc.

This PR removes the side-effect so new people who are beginners would
not get the wrong idea.

### DIFF
--- a/docs/basics/ExampleTodoList.md
+++ b/docs/basics/ExampleTodoList.md
@@ -29,11 +29,10 @@ render(
 #### `actions/index.js`
 
 ```js
-let nextTodoId = 0
-export const addTodo = (text) => {
+export const addTodo = (text, id) => {
   return {
     type: 'ADD_TODO',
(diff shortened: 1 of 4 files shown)


sacho:
If you remove a `todo` from the middle and then add a new `todo`, you will have two `todos` with the same `id`.

stenver:
Good catch, thanks. I fixed that problem

gaearon:
While I see where you're coming from, this is intentional. In Redux only reducers have to be pure. One of the reasons action creators are a useful pattern is *because* they can have side effects and you don’t *have to* write code like this. So I appreciate your effort but I think we would like to keep it this way to highlight action creators don’t have to be pure.

PR #1590, opened 2016-04-07
Add a note that action creators do not have to be pure

stenver:
As a result of this PR:
https://github.com/reactjs/redux/pull/1589#issuecomment-206778709

I understood that some of my understanding about Action Creators were
false. This PR aims to make the fact that action creators are the places
where to make API calls more explicit and clearer for future
readers.

### DIFF
--- a/docs/basics/Actions.md
+++ b/docs/basics/Actions.md
@@ -99,7 +99,7 @@ boundCompleteTodo(index)
 
 The `dispatch()` function can be accessed directly from the store as [`store.dispatch()`](../api/Store.md#dispatch), but more likely you'll access it using a helper like [react-redux](http://github.com/gaearon/react-redux)'s `connect()`. You can use [`bindActionCreators()`](../api/bindActionCreators.md) to automatically bind many action creators to a `dispatch()` function.
 
-Action creators can also be asynchronous and have side-effects. You can read about [async actions](../advanced/AsyncActions.md) in the [advanced tutorial](../advanced/README.md) to learn how to handle AJAX responses and compose action creators into async control flow.
+Action creators can also be asynchronous and have side-effects. You can read about [async actions](../advanced/AsyncActions.md) in the [advanced tutorial](../advanced/README.md) to learn how to handle AJAX responses and compose action creators into async control flow. Don’t skip ahead to async actions until you’ve completed the basics tutorial, as it covers other important concepts that are prerequisite for the advanced tutorial and async actions.
 
 ## Source Code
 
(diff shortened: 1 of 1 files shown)


docs/basics/Actions.md
```diff
@@ -51,7 +51,7 @@ Finally, we’ll add one more action type for changing the currently visible tod
 
 ## Action Creators
 
-**Action creators** are exactly that—functions that create actions. It's easy to conflate the terms “action” and “action creator,” so do your best to use the proper term.
+**Action creators** are exactly that—functions that create actions. It's easy to conflate the terms “action” and “action creator,” so do your best to use the proper term. Note that action creators *do not* have to be pure and *can* have side-effects. Don't worry about that before you get into [advanced async action creators](../advanced/AsyncActions.md).
```
gaearon: Minor nit: please use ’ instead of '.

gaearon: Why do we need to link to the advanced section here at all?

stenver: Ok, I will remove it. Especially since it's in the bottom of the page

stenver:
Updated

gaearon:
One potentially confusing thing is we don’t talk much about what “pure” or “side effects” mean until the next section about reducers where we introduce these concepts. If you were reading for the first time, this passage would probably be confusing. Is it worth calling that out in the reducer section instead? What do you think?

stenver:
You are very right. I thought about how I could word this better or if this was simply my own fault that I missed this information initially. After pondering on how I could fit this information better on somewhere. I then came upon this solution. What do you think of this?

It felt to me that the advanced footnote was in kind of awkward place where it was easy to dismiss it. It also did not convey this critical piece of information that you can do advanced things with actions and action creators.

docs/basics/Actions.md
```diff
@@ -99,6 +99,8 @@ boundCompleteTodo(index)
 
 The `dispatch()` function can be accessed directly from the store as [`store.dispatch()`](../api/Store.md#dispatch), but more likely you'll access it using a helper like [react-redux](http://github.com/gaearon/react-redux)'s `connect()`. You can use [`bindActionCreators()`](../api/bindActionCreators.md) to automatically bind many action creators to a `dispatch()` function.
 
+Action creators can also be asynchronous and have side-effects. You can read about [async actions](../advanced/AsyncActions.md) in the [advanced tutorial](../advanced/README.md) to learn how to handle AJAX responses and compose action creators into async control flow.
+
```
gaearon: I would still add something like “Don’t skip ahead to async actions until you’ve completed this tutorial, as it covers other important concepts that are prerequisite for the advanced tutorial and async actions.”

stenver:
Updated as separate commit which can be squashed

gaearon:
Thanks!

stenver:
Redux has the best documentation I have ever seen. Thank you for that :)

PR #1592, opened 2016-04-08
remove initial assignment of `dispatch` in applyMiddleware to make it more expressive

mz026:
In the original version, the `dispatch` local variable in `applyMiddleware` stands for
different things before and after the `compose` call.
It'd be more expressive to separate "the original one on store" and "the one with middlewares applied"

### DIFF
--- a/src/applyMiddleware.js
+++ b/src/applyMiddleware.js
@@ -19,7 +19,8 @@ import compose from './compose'
 export default function applyMiddleware(...middlewares) {
   return (createStore) => (reducer, initialState, enhancer) => {
     var store = createStore(reducer, initialState, enhancer)
-    var dispatch = store.dispatch
+    var storeDispatch = store.dispatch
+    var dispatch
     var chain = []
 
     var middlewareAPI = {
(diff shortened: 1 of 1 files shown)


Aweary:
I think that accessing `dispatch` as a method of `store` via `store.dispatch` is pretty clear, providing the same essential information that `storeDispatch` does. Is there another advantage of localizing it to `storeDispatch`?

mz026:
When first reading the code, I was pretty confused by `middlewareAPI.dispatch`: why wrapping `dispatch` into a function that simply returns directly without doing nothing else? That is, why not pass `store.dispatch` to it like this:

```javascript
var middlewareAPI = {
  getState: store.getState,
  dispatch: store.dispatch
}
```

or like this:

```javascript
var middlewareAPI = {
  getState: store.getState,
  dispatch
}
```

Then I realized the purpose of wrapping `dispatch` in the `middlewareAPI` was that the local variable `dispatch` changed after this line

```javascript
```

so we were actually using **the `dispatch` with middleware** in `middlewareAPI.dispatch` via closure.
However, the assigning statement `var dispatch = store.dispatch` shadowed this purpose, and by separating them can make this purpose clearer.

gaearon:
It used to be this way some long time ago. Was changed later to appease Flow but we never merged Flow support anyway so I suppose there's no point to keeping it this way. Let's get this merged.

gaearon:
FWIW this broke middleware that relied on `dispatch` during initialization.
We will be deprecating this pattern but for now, I reverted this in https://github.com/reactjs/redux/pull/1647.

PR #1593, opened 2016-04-09
Warn if user uses previous state

ev1stensberg:
## Description 

So.. This PR is 50/50. It needs some work in my eyes. I'm not 100% sure how to check the previous state vs the current one, so please help out on this if I misunderstood this completely in the PR or haven't got the code to do what I wanted it to.

### Why this PR?

We want the user to know that he is setting the previous state. Maybe he/she doesn't even need to do that. This helps users to maybe refactor the code they want in a more effective matter.

### Functionality 

It warns possible "empty" states/previousStates that are the same.

### What has been changed?

The only thing added is a new warning function along with a conditional to check if this is true. Note that I'm not sure if I set it up correctly. It ran through tests fine, but would be convenient to double-check this.

## Note

A lot of small patches, as I had a mixed thought about checking if the string is empty or if we just could do a normal conditional between two variables. Naturally last option was more reasonable.

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -18,7 +18,7 @@ function printWarningOnEmptyState(action) {
 
   return (
     `Given action ${actionName} returned your previous state` +
-    `Are you sure you want this action to be empty?`
+    `Are you sure you need this action to be empty?`
   )
 }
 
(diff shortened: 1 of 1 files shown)


gaearon:
I think it is perfectly fine for reducers to return their current state when they don’t care about a particular action. In fact this is what makes Redux reasonably fast: most of the state tree keeps referential identity and only small part changes so we can efficiently rerender what needs to be renderered. So I think we won’t be doing this. If you run this on examples I think you’ll get a lot of warnings where things are perfectly fine. Thanks for the PR though!

PR #1595, opened 2016-04-09
Avoid every execution and allocation

aimwhy:
```javascript

  if (!funcs.length) {
    return (...args) => args[0]
  } else {
    const last = funcs[funcs.length - 1]
    const rest = funcs.slice(0, -1)
    return (...args) => rest.reduceRight((composed, f) => f(composed), last(...args))
  }
}
```

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -10,15 +10,12 @@
  */
  
 export default function compose(...funcs) {
-  const isEmpty = funcs.length === 0
-  const last = isEmpty ? null : funcs[funcs.length - 1]
-  const rest = isEmpty ? [] : funcs.slice(0, -1)
-    
-  return (...args) => {
-    if (isEmpty) {
-      return args[0]
-    }else {
-      return rest.reduceRight((composed, f) => f(composed), last(...args))
-    }
+  if (funcs.length) {
+  	const last = funcs[funcs.length - 1];
+  	const rest = funcs.slice(0, -1);
+  	
+  	return (...args) => rest.reduceRight((composed, f) => f(composed), last(...args));
+  } else {
+    return (...args) => args[0];
   }
 }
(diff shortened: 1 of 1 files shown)


src/compose.js
```diff
@@ -10,14 +10,15 @@
  */
  
 export default function compose(...funcs) {
+  const isEmpty = funcs.length === 0
```
gaearon: Let’s do it the other way: return another function early if there are no `funs`.

```js
  if (!funcs.length) {
    return (...args) => args[0];
  }

  // ...

}
```

Then we can remove those ternaries and conditions below.

aimwhy:
>Expected indentation of 4 space characters but found 2  ?
> Mixed spaces and tabs  ? 


Error: The Travis CI build could not complete due to an error  


Too young for me  and too simple for my think  ( ▼-▼ )

src/compose.js
```diff
@@ -10,14 +10,12 @@
...
-    const last = funcs[funcs.length - 1]
-    const rest = funcs.slice(0, -1)
-
-    return rest.reduceRight((composed, f) => f(composed), last(...args))
+  if(funcs.length){
+    const last = funcs[funcs.length - 1];
+    const rest = funcs.slice(0, -1);
+    
+    return (...args) => rest.reduceRight((composed, f) => f(composed), last(...args));
+  }else{
```
gaearon: Let’s flip the condition and make the `0` case an early return.

```js
if (!...) {
  return ...
}

...
```

aimwhy: Just started using github
why The Travis CI build Report an error always ?
i dont use local editor ,only use explore
ok, put commented

gaearon:
I don’t understand why linter crashes either :disappointed:

gaearon:
Thanks!

dd1994:
too young, too simple, 这也能膜 🌝

PR #1601, opened 2016-04-12
Add redux-actions-assertions

dmitry-zaets:
redux-actions-assertions has assertions for multiple testing frameworks and libraries, like:
- [chai](#chai)
- [expect](#expect)
- [expect.js](#expectjs)
- [should](#should)


gaearon:
Thanks!

dmitry-zaets:
Thank you too!

PR #1604, opened 2016-04-12
Very minor updates to the comments in index.d.ts.

adamdonahue:
## CHANGE SUMMARY

* Made a handful of very small changes to the comments to `index.d.ts`, fixing a grammatical error (data is plural), and removing some slightly misleading text.  See inline comments in the PR for details.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -37,6 +37,9 @@ export interface Action {
  * side-effects. This is what enables exciting features like hot reloading and
  * time travel.
  *
+ * *Do not put calls into reducers that have themselves are impure or may
+ * may have side-effects.*
+ *
  * @template S State object type.
  */
 export type Reducer<S> = <A extends Action>(state: S, action: A) => S;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,8 +1,8 @@
 /**
- * An *action* is a plain object that represents an intention to change the
+ * An *action* is a plain object that represents an intention to change
  * state. Actions are the only way to get data into the store. Any data,
  * whether from UI events, network callbacks, or other sources such as
- * WebSockets needs to eventually be dispatched as actions.
+ * WebSockets need to eventually be dispatched as actions.
  *
  * Actions must have a `type` field that indicates the type of action being
  * performed. Types can be defined as constants and imported from another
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -26,7 +26,7 @@ export interface Action {
  * to reduce a collection of values down to a single value
  *
  * Reducers are not unique to Redux—they are a fundamental concept in
- * functional programming.  Even most non-functional languages, like
+ * functional programming.  Even many non-functional languages, like
```
adamdonahue: I'm not sure `most` is the right text; `many` seems more apt.  There are a lot of programming languages out there.

index.d.ts
```diff
@@ -37,10 +37,6 @@ export interface Action {
  * side-effects. This is what enables exciting features like hot reloading and
  * time travel.
  *
- * Reducers are the most important concept in Redux.
```
adamdonahue: Is this fair to say?  I feel it's misleading to say one concept or another are the "most important" in Redux.  Perhaps saying they are the primary contribution of, or new concept introduced by, Redux would be more apt.

index.d.ts
```diff
@@ -37,10 +37,6 @@ export interface Action {
...
- *
- * *Do not put API calls into reducers.*
```
adamdonahue: This is far too broad a statement.  As long as those APIs are also pure and side-effect free there is no reason not to use them in a reducer.  (For example, for handling complex data arithmetic one might want to use the `moment` API.)

index.d.ts
```diff
@@ -295,7 +291,7 @@ export function applyMiddleware(...middlewares: Middleware[]): GenericStoreEnhan
 
 /**
  * An *action creator* is, quite simply, a function that creates an action. Do
- * not confuse the two terms—again, an action is a payload of information, and
+ * not confuse the two terms—an action is a payload of information, and
```
adamdonahue: Saying "again" implies this was said before, which is wasn't.

adamdonahue:
Did I mention these are _extremely minor_?

index.d.ts
gaearon: Can change this to “side effects such as AJAX calls”.

adamdonahue: Made a small change.  What do you think?

gaearon:
Hi, did you mean to close this? I’m open to merging it.

adamdonahue:
I did. I'll reopen with the cleaner changes.

PR #1605, opened 2016-04-13
Define NODE_ENV on client-side in real world example

JeremyBernier:
Right now NODE_ENV is always undefined on the client-side. This fix allows NODE_ENV to be defined, thus giving us the option to run the app in production mode (currently it'll only run in dev mode - see configureStore.js).

### DIFF
--- a/examples/real-world/webpack.config.js
+++ b/examples/real-world/webpack.config.js
@@ -14,7 +14,12 @@ module.exports = {
   },
   plugins: [
     new webpack.optimize.OccurenceOrderPlugin(),
-    new webpack.HotModuleReplacementPlugin()
+    new webpack.HotModuleReplacementPlugin(),
+    new webpack.DefinePlugin({
+      'process.env': {
+        NODE_ENV: JSON.stringify(process.env.NODE_ENV)
+      }
+    })
   ],
   module: {
     loaders: [


examples/real-world/webpack.config.js
```diff
@@ -14,7 +14,12 @@ module.exports = {
   },
   plugins: [
     new webpack.optimize.OccurenceOrderPlugin(),
-    new webpack.HotModuleReplacementPlugin()
+    new webpack.HotModuleReplacementPlugin(),
+    new webpack.DefinePlugin({
+      'process.env': {
+        NODE_ENV: JSON.stringify(process.env.NODE_ENV)
```
mxstbr: Afaik this won't work since we don't define the `NODE_ENV` in the start command, so `process.env.NODE_ENV` will be undefined.

onbjerg: We could do this, then:

```js
NODE_ENV: JSON.stringify(process.env.NODE_ENV || 'development')
```

mxstbr: That won't work either, it'll never be in production. We don't even have a script to run this in production, see the package.json: https://github.com/JeremyBernier/redux/blob/patch-1/examples/real-world/package.json

onbjerg: 👍

gaearon:
These examples are meant as Redux examples, not as boilerplates you can run and bundle completely. I’d welcome an effort adding an `npm run build` script to *each of them* but just defining `NODE_ENV` is not enough: you also need to remove hot module replacement stuff, enable Uglify, etc.

gaearon:
I’ll close this specific PR but please feel free to submit another one that converts example to have a proper build.

examples/real-world/webpack.config.js
JeremyBernier: Yea it would be undefined if you ran `npm start`, but not if you ran `NODE_ENV=production npm start`

JeremyBernier:
Yea it's definitely not a complete solution, but I figured it would only help to add since the example currently checks for `process.env.NODE_ENV` on the client-side (in configureStore.js), so one would inevitably have to add this line of code for that variable to ever not be undefined on the client-side. But I totally understand if you'd prefer to only add in a full working solution.

gaearon:
>so one would inevitably have to add this line of code for that variable to ever not be undefined on the client-side

Not quite sure what you mean. It’s fine for it to be undefined in dev. And the examples are not production-ready anyway so just fixing this one issue might actually make it worse because it’ll give the *impression* that they are production-ready whereas in reality there is more work to do here.

PR #1608, opened 2016-04-13
Update Store.md

willnwhite:
Because I spent ages looking at middleware alternatives when I hadn't thought to do this: http://stackoverflow.com/questions/35305661/where-to-write-to-localstorage-in-a-redux-app.

### DIFF
--- a/docs/api/Store.md
+++ b/docs/api/Store.md
@@ -82,6 +82,8 @@ store.dispatch(addTodo('Read about the middleware'))
 
 Adds a change listener. It will be called any time an action is dispatched, and some part of the state tree may potentially have changed. You may then call [`getState()`](#getState) to read the current state tree inside the callback.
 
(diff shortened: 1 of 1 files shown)


gaearon:
To be honest I don’t think this is particularly discoverable here.

In 95% of the cases `subscribe()` is used to update the UI (although higher level libraries like React Redux tend to abstract over its usage) so saying this is why it exists might give the wrong impression.

Maybe rewording it to something like

>It is most often used to update the UI in response to the state changes. Libraries like [React Redux](https://github.com/reactjs/react-redux) provide higher level abstractions on top of it but you can always use it directly, too. One common use case for it is to persist some parts of the current state into the `localStorage` on every change.

willnwhite:
Yes, I agree about the discoverability. I've created a new PR for the gist code on the front page. I agree with your rewording for here.

gaearon:
Closing in favor of #1620.

PR #1610, opened 2016-04-13
Minor comment changes

adamdonahue:
Some minor changes to the comments in `index.d.ts`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,8 +1,8 @@
 /**
- * An *action* is a plain object that represents an intention to change the
+ * An *action* is a plain object that represents an intention to change
  * state. Actions are the only way to get data into the store. Any data,
  * whether from UI events, network callbacks, or other sources such as
- * WebSockets needs to eventually be dispatched as actions.
+ * WebSockets need to eventually be dispatched as actions.
  *
  * Actions must have a `type` field that indicates the type of action being
  * performed. Types can be defined as constants and imported from another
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1,8 +1,8 @@
 /**
- * An *action* is a plain object that represents an intention to change the
```
gaearon: Isn’t `the` necessary here?

index.d.ts
```diff
@@ -1,8 +1,8 @@
...
+ * An *action* is a plain object that represents an intention to change
  * state. Actions are the only way to get data into the store. Any data,
  * whether from UI events, network callbacks, or other sources such as
- * WebSockets needs to eventually be dispatched as actions.
+ * WebSockets need to eventually be dispatched as actions.
```
gaearon: To clarify, `needs` corresponds to `data` before. I think `needs` is correct here.

adamdonahue: Data is plural; datum is singular. 

Adam

index.d.ts
adamdonahue: Could go either way. Depends on whether you mean the definite article in the Russellean sense or more generally. I told you these were minor. :-)

gaearon:
On a second thought I’m inclined to not take it unless you also go over all the other places with identical sentences. Most of these are picked from the Markdown files with the docs.

adamdonahue:
Your call.  Totally cool with that. 

Adam

gaearon:
Yea, I’m sorry for wasting your time :disappointed: 
If you go over the docs in other places happy to take it though.

adamdonahue:
Totally no worries. I'll review the other docs but this was more a learning-the-code-and-noticed-some-stuff type change. 

Thanks for the feedback. 

Adam

PR #1615, opened 2016-04-14
Actions.md: swapped the order of flux and redux examples. Redux first.

lez:
As a casual reader the flux example might be confusing.
This is a Redux document, why not jump to the best part first?

### DIFF
--- a/docs/basics/Actions.md
+++ b/docs/basics/Actions.md
@@ -53,7 +53,20 @@ Finally, we’ll add one more action type for changing the currently visible tod
 
 **Action creators** are exactly that—functions that create actions. It's easy to conflate the terms “action” and “action creator,” so do your best to use the proper term.
 
-In [traditional Flux](http://facebook.github.io/flux) implementations, action creators often trigger a dispatch when invoked, like so:
+In Redux action creators simply return an action:
+
+```js
+function addTodo(text) {
+  return {
+    type: ADD_TODO,
+    text
+  }
+}
+```
+
+This makes them portable and easy to test.
+
+In Flux [traditional Flux](http://facebook.github.io/flux) action creators often trigger a dispatch when invoked, like so:
 
 ```js
 function addTodoWithDispatch(text) {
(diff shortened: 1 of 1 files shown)


adamdonahue:
Because it is a narrative. If not reviewing Flux first then it might be worth moving the Flux-specific notes to an Appendix. I'd argue it's fine as is though.

lez:
its fine by me, it just happened to trick me :)

docs/basics/Actions.md
```diff
@@ -53,30 +53,32 @@ Finally, we’ll add one more action type for changing the currently visible tod
...
     text
   }
-  dispatch(action)
 }
 ```
 
-By contrast, in Redux action creators simply return an action:
+This makes them portable and easy to test.
+
+In Flux [traditional Flux](http://facebook.github.io/flux) action creators often trigger a dispatch when invoked, like so:
```
gaearon: extra “Flux” here

docs/basics/Actions.md
```diff
@@ -53,30 +53,32 @@ Finally, we’ll add one more action type for changing the currently visible tod
...
+  const action = {
     type: ADD_TODO,
     text
   }
+  dispatch(action)
 }
 ```
 
-This makes them more portable and easier to test. To actually initiate a dispatch, pass the result to the `dispatch()` function:
+In Redux to actually initiate a dispatch, pass the result to the `dispatch()` function:
```
gaearon: I would highlight: In Redux this is not the case! On the contrary, to actuallly [...]

gaearon:
Nah, let’s do it. I left a few nits but I agree we should start de-emphasizing Flux.
This was written before Redux was popular and we needed to cater to Flux users.
Today the situation is often the opposite.

gaearon:
Thanks. I added my changes in 24ab45f2acb8434119998e77464237041571d758

lez:
great, thanks!

PR #1620, opened 2016-04-15
use subscribe to persist to localStorage

willnwhite:
more discoverable than https://github.com/reactjs/redux/pull/1608

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,11 @@ function counter(state = 0, action) {
 // Its API is { subscribe, dispatch, getState }.
 let store = createStore(counter)
 
-// You can subscribe to the updates manually, or use bindings to your view layer.
+/**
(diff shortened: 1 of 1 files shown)


gaearon:
The wording is not entirely what I wanted. Maybe something like this?

```
You can use subscribe() to update the UI in response to state changes.
Normally you’d use a view binding library (e.g. React Redux) rather than subscribe() directly.
However it can also be handy to persist the current state in the localStorage.
```

willnwhite:
Yep, yours is clearer.

gaearon:
Feel free to use it in your PR or tweak it. It’s your initiative, I’m just trying to be helpful. 😉

willnwhite:
Did I do it right? (I've never done more than one commit on a branch for a PR.)

gaearon:
Yep, thanks!

PR #1621, opened 2016-04-15
human readable compose.js

mondaychen:
A more readable compose.js with some lodash utils.

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -1,3 +1,5 @@
+import { head, tail, reverse, reduce } from 'lodash';
+
 /**
  * Composes single-argument functions from right to left. The rightmost
  * function can take multiple arguments as it provides the signature for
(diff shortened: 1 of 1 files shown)


gaearon:
Importing them from `lodash` top-level export rather than through individual modules is going to dramatically increase the UMD build size because it will bundle the whole Lodash with Redux. This is definitely not we want.

Even importing these modules directly, the utlities you use here will add a lot of code to Redux that we don’t actually need. You can verify it yourself by running `npm run build` and checking the size of the compiled bundle. I’m closing for this reason.

Thanks you for sending a PR!

gaearon:
I’m up for replacing `(...args) => args[0]` with `arg => arg` though!

mondaychen:
@gaearon understood. And I will only replace `...args` https://github.com/reactjs/redux/pull/1622

gaearon:
Thanks!

PR #1632, opened 2016-04-18
feat(store): add observable proposal interop to store

### LINKED ISSUE #1631: Proposal: Add support for observable spec interop point
Per the [stage 1 observable proposal](http://github.com/zenparsing/es-observable), there is an interop point that can be added for types that are "observable". It's as simple as adding a `[Symbol.observable]()` hook that returns a basic observable (it doesn't have to be Rx-based or anything).

### Pros

1. Adding this hook will enable interoperability with RxJS 5, and it should do the same for Kefir, Bacon, Most, and others.
2. People can stop asking me "what's the difference between RxJS and redux?"
3. People that want to use a more reactive methodology for delivering state to components can do so without forgoing the wonderful tooling this community has put together
4. It's a relatively minimal change

### Cons

Need to include a small `Symbol.observable` shim. Should probably add some integration test to a common library which adds one development-time dependency.

attn/ @gaearon @zenparsing

blesh:
- Adds `Symbol.observable` method to the store that returns a generic observable
- Adds tests to ensure interoperability. (rxjs 5 was used for a simple integration test, and
  is a dev only dependency)

closes #1631

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -1,4 +1,5 @@
 import isPlainObject from 'lodash/isPlainObject'
+import $$observable from 'symbol-observable'
 
 /**
  * These are private action types reserved by Redux.
@@ -198,6 +199,49 @@ export default function createStore(reducer, initialState, enhancer) {
     dispatch({ type: ActionTypes.INIT })
   }
 
+  /**
+   * Interoperability point for observable/reactive libraries.
+   * @returns {observable} A minimal observable of state changes.
+   * For more information, see the observable proposal:
+   * https://github.com/zenparsing/es-observable
+   */
+  function observable() {
+    var outerSubscribe = subscribe
+    return {
+      /**
+       * The minimal observable subscription method.
+       * @param {Object} observer Any object that can be used as an observer.
+       * The observer object should have a `next` method.
+       * @returns {subscription} An object with an `unsubscribe` method that can
+       * be used to unsubscribe the observable from the store, and prevent further
+       * emission of values from the observable.
+       */
+      subscribe(observer) {
+        if (typeof observer !== 'object') {
+          throw new TypeError('Expected observer to be an object')
+        }
+
+        var observeState = () => {
+          if (observer.next) {
+            observer.next(getState())
+          }
+        }
+
+        // send initial state to observer
+        observeState()
+
+        // send subsequent states to observer
+        var unsubscribe = outerSubscribe(observeState)
+
+        // return an unsubscribable
+        return { unsubscribe }
+      },
+      [$$observable]() {
+        return this
+      }
+    }
+  }
+
   // When a store is created, an "INIT" action is dispatched so that every
   // reducer returns their initial state. This effectively populates
   // the initial state tree.
@@ -207,6 +251,7 @@ export default function createStore(reducer, initialState, enhancer) {
     dispatch,
     subscribe,
     getState,
-    replaceReducer
+    replaceReducer,
+    [$$observable]: observable
   }
 }
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -2,6 +2,8 @@ import expect from 'expect'
 import { createStore, combineReducers } from '../src/index'
 import { addTodo, dispatchInMiddle, throwError, unknownAction } from './helpers/actionCreators'
 import * as reducers from './helpers/reducers'
+import * as Rx from 'rxjs'
+import $$observable from 'symbol-observable'
 
 describe('createStore', () => {
   it('exposes the public API', () => {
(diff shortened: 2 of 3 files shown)


blesh:
No more "RxJS vs Redux"!

Ducks for everyone! Ducks all around!

gaearon:
This looks great to me. I’d leave it open for a few days in case somebody else has any feedback.

tomenden:
This is great

src/utils/Symbol_observable.js
```diff
@@ -0,0 +1,18 @@
...
+    }
+    Symbol.observable = $$observable
+  } else {
+    $$observable = Symbol.observable
+  }
+} else {
+  $$observable = '@@observable'
+}
+
+export default $$observable
```
timdorr: Can this shim be maintained elsewhere? Much like we've swapped out built-in utilities for lodash, I don't think it's good if we try to maintain this shim code internally. Especially if the spec changes as it makes its way through the proposal process.

timdorr:
Only question I have is how stable the Observable proposal is going to be. Is it likely to remain unchanged? If so, then we should be A-OK here. If there is some potential for change in the future, that might be a headache to keep in mind on the horizon.

src/utils/Symbol_observable.js
blesh: I'm 50/50 on this, personally. On one hand, it's not a lot of code, and there's the ol' left-pad issue. On the other hand, I tend to agree it could be centralized somewhere. I know that @sindresorhus has a ponyfill for this, but I've seen a couple of bugs in it and I haven't had the time to reach out to collaborate with him (I don't actually know him at all). Perhaps later this afternoon I'll put in a PR on his repository and if that doesn't fly I can always publish something elsewhere, ReactiveX perhaps.

blesh: In the meantime, this _could_ ship as-is, if blessed. And I'll be happy to come back and make a change to use a separate module. I'll leave that up to @gaearon. 

FWIW, Falcor has similar concerns. But it was determined that it's not a *huge* deal.

gaearon:
I say let’s ship it but not document this until it shakes out a little?

timdorr:
A secret package of awesomeness 😄 

Let's see how [Ben's PR](https://github.com/reactjs/redux/pull/1632#discussion_r60105286) shakes out and then merge it in depending on the result there.

src/utils/Symbol_observable.js
sindresorhus: @blesh Regardless of it being included here, I would appreciate a issue/PR about the bugs 😀 We use the ponyfill in [AVA](https://github.com/sindresorhus/ava), so it's important that it works correctly.

acdlite:
Weeeee this is cool

blesh:
> Only question I have is how stable the Observable proposal is going to be.

This particular piece is likely to remain pretty solid. It's implemented in 3-4 major reactive libraries, and I don't see it changing any time soon. Even if the observable proposal flops at the TC39, I'd expect this piece to live on in libraries like RxJS, etc.

molily:
> Even if the observable proposal flops at the TC39, I'd expect this piece to live on in libraries like RxJS, etc.

Sorry to jump in here, I’m totally ignorant about these libraries. But as a user of Redux I do not expect Redux to modify global objects (`Symbol.observable = …`) via a utility file when the property is only defined in a TC39 stage 1 proposal. For the purpose of this PR, the line `Symbol.observable = $$observable` is negligible, isn’t it?

blesh:
> For the purpose of this PR, the line Symbol.observable = $$observable is negligible, isn’t it?

Actually no, it has real purpose. In the event that the engine has `Symbol`, but not `Symbol.for`, then a `Symbol` is made, which is completely unique. It then needs to be placed on `Symbol.observable` in order for it to be found and used by other libraries that care about it (Falcor, RxJS, Bacon, Kefir, etc).

>  I do not expect Redux to modify global objects 

This is a bit of a doom-and-gloom way to say it, the tone seems to want to incite fear. The reality is almost every library that you can use with a <script> tag is modifying at _least_ `window`. Is there a specific concern around this modification? Are you concerned that other libraries are leveraging `Symbol.observable` in some other way?

src/utils/Symbol_observable.js
timdorr: BTW, for those wondering (and observing :smile:), the aforementioned ponyfill is here: https://github.com/sindresorhus/symbol-observable

glenjamin:
Would it be feasible to ship this initially as a module, and then fold into core if it's popular / vital / stable / enough?

As far as I can tell it only depends on two redux API methods, `subscribe` and `getState`, both of which are part of the public API.

Something like this that could be referenced in the official docs should be doable?

```js
import { observeStore } from 'redux-observable';
observeStore(store);
```

gaearon:
Since the purpose is interop, I think shipping it as a separate module kinda defeats the purpose. It’s like if Immutable Maps weren’t Iterables and required a separate lib.

I do think we want to add this. However I would vastly prefer this be an external polyfill so discussions like this can happen on its issue tracker while we just declare compat 😄 .

molily:
@blesh 
> It then needs to be placed on `Symbol.observable` in order for it to be found and used by other libraries …
Thanks for the explanation. When this is about interoperability without explicitly passing the symbol to other libs, creating a global makes sense.
> This is a bit of a doom-and-gloom way to say it, the tone seems to want to incite fear.
That wasn’t my intention. :-D

blesh:
@gaearon FYI: I've updated this to use a "small module" (THANK YOU @sindresorhus!!!!) to get the reference to `Symbol.observable`.

gaearon:
y u no lint @blesh

src/createStore.js
```diff
@@ -198,6 +199,26 @@ export default function createStore(reducer, initialState, enhancer) {
     dispatch({ type: ActionTypes.INIT })
   }
 
+  /**
+   * Interop point for observable libraries
```
gaearon: Let’s format these a little more consistent with the rest of JSDoc in this file:

* End sentences with period.
* Start parameter descriptions with a capital letter.
* I would rather avoid the awkward hanging link and instead split this over two lines, i.e. `see the observable proposal:\nhttp://...`

test/createStore.spec.js
```diff
@@ -610,4 +611,65 @@ describe('createStore', () => {
...
+        key = Symbol.observable
+      } else if (typeof Symbol.for === 'function') {
+        key = Symbol.for('observable')
+      }
+    } else {
+      key = '@@observable'
+    }
+
+    expect(typeof store[key]).toBe('function')
+
```
gaearon: Nitpick: let’s make it a little denser. We generally follow this pattern:

```js
some setup

line 1
line 2
line 3
related expect call

line 4
line 5
related expect call

...
```

gaearon:
Thank you for your work on this, and special thanks to @sindresorhus.
Left a few minor nits and this is ready to go.

gaearon:
@blesh I’m happy to do nits myself if you don’t have the time. Just let me know. ✨

blesh:
@gaearon I can do them. I just got these messages. No problem.

test/createStore.spec.js
blesh: Here we can probably use `symbol-observable` now instead of having all of this code to figure out what to use.

blesh:
@gaearon I've made the requested changes (I think), and I've added some additional tests to ensure the functionality is solid.

I've also flattened the changes into a single commit for cleanliness.

gaearon:
Looking great. Will cut a release tomorrow.

gaearon:
Should this bump minor? I think so.

acdlite:
Yeah it's a new feature, so it should bump minor. Especially since extensions may depend on it.

gaearon:
Out in 3.5.0.

SimenB:
This explodes in IE8 with `Expected identifier`

blesh:
@SimenB there will be a new version of symbol-observable this weekend with the fix.

gaearon:
@SimenB I sympathize with your concerns but I don’t think any automated testing is worth the effort here. Technically we support IE8, and regressions aren’t good, but we try to fix them as they are reported. When ES3 plugins were broken in Babel 6 for three months, many libraries had ES3 issues, and I think at this point whoever cares about it should be wary of updating dependencies and take care of ES3-ifying the output themselves. I don’t mean to come across as dismissive—just saying there’s been a ton of IE8 breakage in the ecosystem, and we’re hardly the worst offenders. I agree we should fix it of course 😄 .

@blesh Thanks!

SimenB:
Always shrinkwrapping sucks, but it might make sense yeah. I'm out in a week, so not a huge bother for me personally, though!

blesh:
@SimenB `symbol-observable@0.2.3` is released with this change.

gaearon:
Redux 3.5.2 enforces the updated `symbol-observable`. Thanks @blesh.

src/createStore.js
```diff
@@ -198,6 +199,49 @@ export default function createStore(reducer, initialState, enhancer) {
     dispatch({ type: ActionTypes.INIT })
   }
 
+  /**
```
benjamingr: Hey @blesh , I wanted to add `Symbol.observable` to mobx and I was wondering if you felt like putting this code on npm somewhere because I'd feel stupid copy pasting it or rewriting it.

blesh: I would, but I don't know that it's really that globally applicable to everyone. At least the part that adapts some arbitrary type (in this case a redux store) into an Observable.

Is there an issue on MobX I can check out?

benjamingr: No, but I talked to @mweststrate directly and he was interested in the idea. mobx is based on observables but they are depedency tracking and very different (and much simpler) than rxjs ones.

I now write a lot of mobx -> rxjs -> mobx code and I figured it could be cool for Rx to consume mobx observables directly.

chicoxyzzy: @blesh @benjamingr  there is https://github.com/mobxjs/mobx/issues/169 but I don't experiment with mobx anymore so I didn't made any progress. Also there is an interop you @benjamingr might find useful https://github.com/chicoxyzzy/rx-mobx

Issue #1617, opened 2016-04-14
Document required default state in reducer


DaveStein:
Related to my pitfall here: http://stackoverflow.com/questions/36631728/why-isnt-combinereducers-receiving-state-in-redux/36633079#36633079

The docs show reducers having a default value for `state`, but it is never listed as required. The fact that an INIT action is fired before `createStore` takes advantage of `initialState` was confusing.

gaearon:
What can we do better? I think we’re already warning if you forgot to specify the initial state—is the warning not descriptive, or misleading?

DaveStein:
The warning simply tells me that nav is not returning state. I console logged the state at that point, being confused that I had passed initialState into createStore. It was being hit because of the INIT action from Redux, which passes state as undefined no matter what. So really the docs need to specify that default declaration is required when making a reducer (even if its just empty object, string, etc)

markerikson:
@DaveStein : you might want to read through http://stackoverflow.com/questions/33749759/read-stores-initial-state-in-redux-reducer , which covers how reducer default state and `createStore` initial state interact.

This probably should go in the FAQ.  I think I may have skipped it because I was considering putting it in a "Structuring Reducers", page, which I decided not to do as part of the initial FAQ work.

DaveStein:
@markerikson I would agree :) Going through Stack to find answers is unexpected. I did get the actual help via the IRC for #reactjs. I know the pain of trying to get docs write for people while also working on the thing you are doc'ing. But at least noobs like me point out what might need improvement haha

gaearon:
Yeah, `initialState` in `createStore()` is often tripping people. It was never meant as a way to “initialize” your application state manually. The only useful applications for it are:

* Booting up server rendered app from JSON state payload
* “Resuming” the app from a state saved into local storage

It is implied that you never write `initialState` manually and in most apps you don’t even use it. Instead, reducers must always specify their own initial state, and `initialState` is just a way to prefill that state when you have an existing serialized version of it.

Maybe we should rename it to `prefilledState` or `rehydratedState` to make a point..

gaearon:
This explains some reasoning behind it: http://stackoverflow.com/questions/33749759/read-stores-initial-state-in-redux-reducer/33791942#33791942

markerikson:
Added to the docs at http://redux.js.org/docs/recipes/reducers/InitializingState.html .

PR #1633, opened 2016-04-19
document assertReducerSanity behavior

ruggeri:
It is not clear from the documentation that it is insufficient to merely avoid ever passing in `undefined` as state by using initial state. `assertReducerSanity` will test that your reducer does the right thing when passed `undefined`, even if you never intend for it to be called this way.

This documentation change makes that behavior clearer, in the hopes of saving new users from needless frustration. It was very confusing to see my reducer passed undefined even though I never intended it to be used that way. I did not suspect `combineReducers` would do this.

Hope this helps!


gaearon:
Yeah. Ideally we should also explain *why* this is the case.
Some context here: https://github.com/reactjs/redux/issues/1617#issuecomment-210159600

I like your changes but please consider the broader picture—I’d be happy if you could take tweak `createStore.md` to better clarify `initialState` role while you’re at it. Let me know if it’s too much to ask though 😄 

Thanks!

timdorr:
We can add that in another PR, I suppose. Let's get this in for now. Thanks, @ruggeri!

PR #1640, opened 2016-04-20
Remove unneeded code

stevemao:
`console.error` can print the call site. Even if `console.error` didn't or doesn't exist, the `try...catch` block cannot print the call site. It does nothing here.

### DIFF
--- a/src/utils/warning.js
+++ b/src/utils/warning.js
@@ -10,11 +10,4 @@ export default function warning(message) {
     console.error(message)
   }
   /* eslint-enable no-console */
-  try {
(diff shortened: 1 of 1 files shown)


gaearon:
It’s not for printing, it’s for “break on all exceptions” that would pause the execution at this line.

stevemao:
Very interesting technique. Thanks :)

stevemao:
Although I think the comment is not very clear. It should be what you just said :)

gaearon:
Good point. Feel free to edit!

PR #1641, opened 2016-04-20
refactor combineReducers with a functional approach

5h1rU:
- Added new functions
- Decoupled the logic from the combineReducers function
- ~~Changed `for` loops to `map` in order to improve performance~~

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -39,9 +39,7 @@ function getUnexpectedStateShapeWarningMessage(inputState, reducers, action, une
     !unexpectedKeyCache[key]
   )
 
-  unexpectedKeys.forEach(key => {
-    unexpectedKeyCache[key] = true
-  })
+  unexpectedKeys.forEach(key => unexpectedKeyCache[key] = true)
 
   if (unexpectedKeys.length > 0) {
     return (
@@ -81,6 +79,46 @@ function assertReducerSanity(reducers) {
   })
 }
 
+function getReducers(config) {
+  var reducerKeys = Object.keys(config)
+  var reducers = {}
+  for (var i = 0; i < reducerKeys.length; i++) {
+    var key = reducerKeys[i]
+
+    if (process.env.NODE_ENV !== 'production') {
+      if (typeof config[key] === 'undefined') {
+        warning(`No reducer provided for key "${key}"`)
+      }
+    }
+
+    if (typeof config[key] === 'function') {
+      reducers[key] = config[key]
+    }
+  }
+  return reducers
+}
+
+function getNextState(keys, reducers, state, action) {
+  var hasChanged = false
+  var nextState = {}
+  for (var i = 0; i < keys.length; i++) {
+    var key = keys[i]
+    var reducer = reducers[key]
+    var previousStateForKey = state[key]
+    var nextStateForKey = reducer(previousStateForKey, action)
+
+    if (typeof nextStateForKey === 'undefined') {
+      var errorMessage = getUndefinedStateErrorMessage(key, action)
+      throw new Error(errorMessage)
+    }
+
+    nextState[key] = nextStateForKey
+    hasChanged = hasChanged || nextStateForKey !== previousStateForKey
+  }
+
+  return hasChanged ? nextState : state
+}
+
 /**
  * Turns an object whose values are different reducer functions, into a single
  * reducer function. It will call every child reducer, and gather their results
@@ -97,61 +135,24 @@ function assertReducerSanity(reducers) {
  * @returns {Function} A reducer function that invokes every reducer inside the
  * passed object, and builds a state object with the same shape.
  */
-export default function combineReducers(reducers) {
-  var reducerKeys = Object.keys(reducers)
-  var finalReducers = {}
-  for (var i = 0; i < reducerKeys.length; i++) {
-    var key = reducerKeys[i]
 
-    if (process.env.NODE_ENV !== 'production') {
-      if (typeof reducers[key] === 'undefined') {
-        warning(`No reducer provided for key "${key}"`)
-      }
-    }
-
-    if (typeof reducers[key] === 'function') {
-      finalReducers[key] = reducers[key]
-    }
-  }
-  var finalReducerKeys = Object.keys(finalReducers)
-
+export default function combineReducers(config) {
+  var reducers = getReducers(config)
+  var keys = Object.keys(reducers)
   if (process.env.NODE_ENV !== 'production') {
     var unexpectedKeyCache = {}
   }
 
-  var sanityError
-  try {
-    assertReducerSanity(finalReducers)
-  } catch (e) {
-    sanityError = e
-  }
-
   return function combination(state = {}, action) {
-    if (sanityError) {
-      throw sanityError
-    }
+    assertReducerSanity(reducers)
 
     if (process.env.NODE_ENV !== 'production') {
-      var warningMessage = getUnexpectedStateShapeWarningMessage(state, finalReducers, action, unexpectedKeyCache)
+      var warningMessage = getUnexpectedStateShapeWarningMessage(state, reducers, action, unexpectedKeyCache)
       if (warningMessage) {
         warning(warningMessage)
       }
     }
 
-    var hasChanged = false
-    var nextState = {}
-    for (var i = 0; i < finalReducerKeys.length; i++) {
-      var key = finalReducerKeys[i]
-      var reducer = finalReducers[key]
-      var previousStateForKey = state[key]
-      var nextStateForKey = reducer(previousStateForKey, action)
-      if (typeof nextStateForKey === 'undefined') {
-        var errorMessage = getUndefinedStateErrorMessage(key, action)
-        throw new Error(errorMessage)
-      }
-      nextState[key] = nextStateForKey
-      hasChanged = hasChanged || nextStateForKey !== previousStateForKey
-    }
-    return hasChanged ? nextState : state
+    return getNextState(keys, reducers, state, action)
   }
 }


markerikson:
Hmm.  Code changes look interesting, and certainly agree that `map()` is usually easier to read than a loop.  On the other hand, my understanding is that generally a raw `for` loop results in better performance than using `map()`.  Any particular reason why this would be faster?

5h1rU:
@markerikson You're right, metal lapsus there...

I don't know how fast is `for` with respect to `map`, would appreciate if you would share a recent comparision or some link in order to get more details about.

The only comparission that I found was from an old jsperf link that is not working anymore.

markerikson:
Some quick searching turns up the following links from within the last two years:

- http://stackoverflow.com/questions/22182454/javascript-performance-of-array-map
- https://www.quora.com/JavaScript-programming-language-Are-Javascript-functions-like-map-reduce-and-filter-already-optimized-for-traversing-array
- http://www.sitepoint.com/quick-tip-stop-writing-loops-start-thinking-with-maps/
- https://ryanpcmcquen.org/javascript/2015/10/25/map-vs-foreach-vs-for.html

The biggest factors, as far as I know, are that `map()` involves additional function calls, and that the browser-native functions like `map()` and `filter()` have to account for all possible irregularities like sparse arrays.

All that said, switching to `map()` for Redux may be an entirely valid approach, with no noticeable performance impact.  I just want to clarify that `map()` is unlikely to _improve_ performance as far as I know.

src/combineReducers.js
```diff
@@ -90,50 +140,20 @@ function assertReducerSanity(reducers) {
...
-    sanityError = e
-  }
+export default function combineReducers(reducers) {
+  var fr = finalReducers(reducers)
+  var keys = finalReducerKeys(fr)
 
   return function combination(state = {}, action) {
-    if (sanityError) {
-      throw sanityError
+    if (sanityError(reducers)) {
```
markerikson: This looks concerning.  The previous code only ran `assertReducerSanity()` once, but it looks like the modified code is actually calling this twice.

5h1rU: Valid concern, I'll change.

5h1rU:
Regarding the performance I'll strike out the point

src/combineReducers.js
```diff
@@ -74,6 +74,56 @@ function assertReducerSanity(reducers) {
   })
 }
 
+function finalReducers(reducers) {
```
gaearon: Not saying we want to get this in yet, but in any case the minimal requirement would be that we use verbs for function names.

ooflorent:
> Changed `for` loops to `map` in order to improve performance

While I understand the motivation behind this PR, I really don't like the `map` usage here since **the function is generated on every invocation**.

src/combineReducers.js
```diff
@@ -74,6 +74,52 @@ function assertReducerSanity(reducers) {
...
+    }
+  })
+
+  return functionReducers
+}
+
+function checkStateChange(keys, finalReducers, state, action) {
+  var hasChanged = false
+  var nextState = {}
+  keys.map(key => {
```
gaearon: Let’s use the `for` loop here. If you look at the history, it was `map` in the beginning. We landed at the current version after a few optimization passes so it’s unwise to go back now. 😉

src/combineReducers.js
```diff
@@ -74,6 +74,52 @@ function assertReducerSanity(reducers) {
...
+    hasChanged = hasChanged || nextStateForKey !== previousStateForKey
+  })
+  return hasChanged ? nextState : state
+}
+
+function checkSanityError(reducers) {
+  try {
+    assertReducerSanity(getFinalReducers(reducers))
+  } catch (error) {
+    throw error
```
gaearon: `catch` that rethrows immediately is as good as no `catch`. The error was rethrown later inside the generated reducer deliberately so that Redux DevTools can catch it. We should have added a test verifying this—our fault. But this is wrong and defeats the purpose of having that `try` block.

src/combineReducers.js
```diff
@@ -90,50 +136,18 @@ function assertReducerSanity(reducers) {
...
 
-  var sanityError
-  try {
-    assertReducerSanity(finalReducers)
-  } catch (e) {
-    sanityError = e
-  }
+export default function combineReducers(reducers) {
+  var finalReducers = getFinalReducers(reducers)
+  var keys = Object.keys(getFinalReducers(reducers))
```
suhaotian: why don't just this:
```js
var finalReducers = getFinalReducers(reducers)
var keys = Object.keys(finalReducers)
```

5h1rU: +1 nice catch thanks @suhaotian

src/combineReducers.js
```diff
@@ -74,6 +74,46 @@ function assertReducerSanity(reducers) {
...
+  Object.keys(reducers).map(key => {
+    if (typeof reducers[key] === 'function') {
+      functionReducers[key] = reducers[key]
+    }
+  })
+
+  return functionReducers
+}
+
+function checkStateChange(keys, finalReducers, state, action) {
```
gaearon: This name is misleading. We’re not just checking the state change, we’re calculating the next state. This is the important part. The fact that we check for changes is just an optimization.

src/combineReducers.js
```diff
@@ -74,6 +74,46 @@ function assertReducerSanity(reducers) {
   })
 }
 
+function getFinalReducers(reducers) {
```
gaearon: I’d rather have the argument be `config`, the variable be `reducers`, and the function be `getReducers`.

src/combineReducers.js
```diff
@@ -74,6 +74,46 @@ function assertReducerSanity(reducers) {
...
+    }
+
+    nextState[key] = nextStateForKey
+    hasChanged = hasChanged || nextStateForKey !== previousStateForKey
+  }
+
+  return hasChanged ? nextState : state
+}
+
+function environmentProductionWarning(state, finalReducers, action) {
```
gaearon: Should be a verb, e.g. `checkStateShape`.

src/combineReducers.js
```diff
@@ -90,50 +130,18 @@ function assertReducerSanity(reducers) {
...
-  var finalReducerKeys = Object.keys(finalReducers)
 
-  var sanityError
-  try {
-    assertReducerSanity(finalReducers)
-  } catch (e) {
-    sanityError = e
-  }
+export default function combineReducers(reducers) {
+  var finalReducers = getFinalReducers(reducers)
```
gaearon: Same here, let’s change `reducers` => `config`, and `finalReducers` => `reducers`.

src/combineReducers.js
```diff
@@ -74,6 +74,46 @@ function assertReducerSanity(reducers) {
...
+  Object.keys(config).map(key => {
+    if (typeof config[key] === 'function') {
+      reducers[key] = config[key]
+    }
+  })
+
+  return reducers
+}
+
+function checkStateShape(keys, reducers, state, action) {
```
gaearon: This function should be called `getNextState` because this is what it does.

src/combineReducers.js
```diff
@@ -74,6 +74,46 @@ function assertReducerSanity(reducers) {
...
+    }
+
+    nextState[key] = nextStateForKey
+    hasChanged = hasChanged || nextStateForKey !== previousStateForKey
+  }
+
+  return hasChanged ? nextState : state
+}
+
+function showWarningMessage(state, reducers, action) {
```
gaearon: This should be called `checkStateShape` because this is what it does. `showWarningMessage` is misleading because it feels like a warning would be called every time, whereas in practice it is only called if the state shape is unexpected.

src/combineReducers.js
5h1rU: @gaearon what do you think about the `map` here. Is better change to a `for` loop?

gaearon: Yep.

src/combineReducers.js
```diff
@@ -74,6 +74,48 @@ function assertReducerSanity(reducers) {
   })
 }
 
+function getReducers(config) {
+  var reducerKeys = Object.keys(config)
+  var reducers = {}
+
+  for (var i = 0; i < reducerKeys.length; i++) {
```
suhaotian: performance better with cache **reducerKeys.length**:
```js
for (var i = 0, l = reducerKeys.length; i < l; i++ {
    ...
}
```

ooflorent: While it is true in PHP, **it is not for JS**. Modern JS VMs optimize this case and are faster without this trick. @mraleph wrote about it: [How the Grinch stole array.length access](http://mrale.ph/blog/2014/12/24/array-length-caching.html).

src/combineReducers.js
```diff
@@ -90,50 +132,18 @@ function assertReducerSanity(reducers) {
...
-  }
+export default function combineReducers(config) {
+  var reducers = getReducers(config)
+  var keys = Object.keys(reducers)
 
   return function combination(state = {}, action) {
-    if (sanityError) {
-      throw sanityError
-    }
+    assertReducerSanity(getReducers(config))
```
gaearon: This is expensive. This calculates `getReducers` on every dispatch but they never change.

5h1rU:
@gaearon what do you think about it? I was thinking that if you are not thinking about merge, it is best close and avoid polute the open PR's.
Anyway thanks for the guidance and the advices for improve the PR. I learned a lot trying to contribute. 

Always glad to help

gaearon:
I think we’ll likely want to take this but I’ll need to run some manual testing first. I don’t currently have time to do it so I’d like to keep this open until I come around to reviewing the PRs. It’s not urgent, and the change rate to the source is very low, so there’s no harm in keeping it around. Thanks!

src/combineReducers.js
```diff
@@ -74,6 +74,47 @@ function assertReducerSanity(reducers) {
   })
 }
 
+function getReducers(config) {
```
dbrans: This function doesn't really do anything reducer-specific. It is just grabbing those keys off of `config` whose values are functions. You could just as well call this `getFunctionValues`.

gaearon: I like the current name because it expresses the intent.

timdorr:
I think I busted this with another PR merge. @5h1rU Can you update this branch? I think it's fine to get merged in.

5h1rU:
Ok, I will @timdorr

5h1rU:
@timdorr the PR is updated

timdorr:
Looking through this again, I'm not 100% sure I see the value in doing this. It breaks up the function into several sub-functions, but isn't really a functional approach. This is only separately a cohesive unit of code into groups for the supposed purposes of code organization. But now things have to be read out of order, as you must go back and forth between the functions. 

Unfortunately, I don't think this really nets us anything good and makes the code harder to read. Sorry for making you do the work of a rebase, but I'm going to have to close this out. Thank you for your efforts nonetheless, @5h1rU! Again, I'm sorry!

5h1rU:
no problem

PR #1651, opened 2016-04-21
Remove erroneous apostrophe

techniq:


gaearon:
This is intentional, as in “it’s a changed one”. Like in math you have A, A'.
Maybe replacing it with ’ will be more obvious?

techniq:
Understood, makes sense now that you explained it.

PR #1654, opened 2016-04-21
improve performance with cache length in a array loop

suhaotian:
Two points:
1. improve performance with cache length in a array loop;
2. rewrite **forEach** to **for**;

Hope to improve the performance 👍  of a little help


gaearon:
>cache length in a array loop;

This can be removed, it only has any effect in very old browsers.

suhaotian:
@gaearon  okay...

gaearon:
I'd rather keep the rest as it is for readability. The `forEach` cost there is not essential because this code only executes once as opposed to the inner loops.

PR #1657, opened 2016-04-23
Add devtools extension support in development

calebmer:
React works out of the box with its developer tools extension, Redux probably should to. This PR is in the hopes that there is a solution to get Redux working out of the box with its developer tools chrome extension.

Code was written in this way so that it can be minified away in production builds.

\cc @zalmoxisus

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -258,7 +258,7 @@ export default (() => {
   // property is exposed, let’s enhance our stores to support it.
   // https://github.com/zalmoxisus/redux-devtools-extension
   if (process.env.NODE_ENV === 'development') {
-    if (window.devToolsExtension) {
+    if (typeof window !== 'undefined' && window.devToolsExtension) {
       return window.devToolsExtension()(createStore)
     }
   }
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -36,7 +36,7 @@ export var ActionTypes = {
  * @returns {Store} A Redux store that lets you read the state, dispatch actions
  * and subscribe to changes.
  */
-export default function createStore(reducer, initialState, enhancer) {
+function createStore(reducer, initialState, enhancer) {
   if (typeof initialState === 'function' && typeof enhancer === 'undefined') {
     enhancer = initialState
     initialState = undefined
@@ -251,3 +251,16 @@ export default function createStore(reducer, initialState, enhancer) {
     [$$observable]: observable
   }
 }
+
+export default (() => {
+  // React works out of the box with its developer tools Chrome extension, we
+  // want Redux to as well. If we are in development, and a `devToolsExtension`
+  // property is exposed, let’s enhance our stores to support it.
+  // https://github.com/zalmoxisus/redux-devtools-extension
+  if (process.env.NODE_ENV === 'development') {
+    if (window.devToolsExtension) {
+      return window.devToolsExtension()(createStore)
+    }
+  }
+  return createStore
+})()


GGAlanSmithee:
If this gets accepted (which would be nice) can it please be changed to `if (typeof window !== 'undefined' && window.devToolsExtension)` instead? Depending on setup, it might crash on server side rendering othervise.

calebmer:
@GGAlanSmithee Oh of course! My bad that I forgot it. That's been fixed.

zalmoxisus:
React extension can be used in production and so does Redux extension if it is required.

I'm not against it, but what's the motivation? Are thete too much boilerplate if we use
```js
const store = createStore(reducer, initialState, 
   process.env.NODE_ENV === 'development' && typeof window !== 'undefined' &&
   window.devToolsExtension && window.devToolsExtension()
);
```
instead of
```js
const store = createStore(reducer, initialState);
```

calebmer:
The less friction there is for developer tools the more adoption there will be.

It's a little more complicated then your example above, what if the user has a global variable other than `NODE_ENV`, what if the user has other enhancers, what if the user wants Uglify to remove the enhancer in production, what if for whatever reason the user can't convince her team that the technical debt is worth it?

Also the `window.devToolsExtension` isn't documented in the Google Chrome Store, some users won't even check the repo documentation.

gaearon:
I don’t feel good about squatting `window.devToolsExtension` for our purposes. It’s fine when it’s opt-in, but this code breaks if anything else declares a global with this name.

I’m also not sure whether this would work with multiple enhancers (e.g. middleware and persistence). Their order matters but this doesn’t let the user specify the order. I’m also not sure what happens if the user has an extension but just integrated DevTools directly and wants to avoid the cost of talking to the extension.

calebmer:
Ultimately I'd like to see this happen, but I'm not the expert. I'll write the code required, but you should be really talking about these technical aspects with the extension's author @zalmoxisus. Again, I'm willing to write all the code but if he doesn't want to support it I'll close the issue.

To one of your points @gaearon ideally I'd like to see the extension be the only way to access the dev tools (like React). The extensibility argument I personally think is pointless, the community should agree on the best developer tool possible and add features to that instead of fragmenting on multiple extensions.

These views are my own and don't reflect those of @zalmoxisus and the extension's team. If @zalmoxisus is willing I think the next step is submitting a PR to deprecate the `devToolsExtension` name and change it to something more explicit (and has more underscores).

gaearon:
>I'd like to see the extension be the only way to access the dev tools (like React)

I think this goes against one of the design goals of Redux DevTools which is it's easy to create your own. The monitors exposed in extension would not have been created in the first place if you had to install and re-package a Chrome extension to work on them. Also I heard that the extension has some perf problems at scale (can't say when exactly) and in this case using DevTools directly helps.

calebmer:
@gaearon my point is if [this](https://twitter.com/dan_abramov/status/723509494291025922) is true and very few people are taking the time to work on tools like this, then why seperate the little work there is into multiple different projects? Why not combine all of the best and useful monitors into one single project, the chrome extension in this case. This is beneficial for two groups:

1. The monitor creators. They get to take advantage of the core project infrastructure and marketing.
2. The users. Removes the barrier of researching these monitors and the installing them and adding them to the project (with appropriate conditionals of course because you don't always want all of the monitors).

Combining Redux DevTools into one project and drastically lowering the friction (merging this PR and encouraging the Chrome Extension) also fosters competition with other developer tools (React, Ember, etc.) which means a higher quality end product for all developers.

Performance problems can be fixed (I think, again not an expert just a very vocal user 😊).

markerikson:
> Why not combine all of the best and useful monitors into one single project, the chrome extension in this case.

Because not everyone uses Chrome?

I do actually use it myself for development because I like the Chrome DevTools, but there's no reason to tie yourself to a single browser in order to use the Redux DevTools.

gaearon:
>Combining Redux DevTools into one project and drastically lowering the friction (merging this PR and encouraging the Chrome Extension)

I do not believe that we would have any of the monitors the community has developed if this was a Chrome extension first and foremost. I think that the state of Chrome extensions for some Flux libraries that existed before Redux proves that. It’s just *hard* to contribute to a Chrome extension because of the way they are set up and packaged. Also compare the feeling when you install a monitor from npm *knowing* it’s just a React component, to using an “off the shelf” vendor extension. I use React DevTools every day, and to be honest I’ve never even thought of contributing to it once. I bet a lot of people think this way.

I understand why the Chrome extension is great. But I disagree Redux DevTools should *become* one project with it and that we should remove support for using it directly.

As for your pull request. I think we can take something like this given that it satisfies a few conditions:

1. It uses something Redux-specific like `__REDUX_STORE_ENHANCER__`.
2. It handles scenarios where this particular enhancer is already applied, or when you already use vanilla Redux DevTools, or when you use other enhancers like `persistState` from DevTools repo, or Redux Loop, or Redux Router. None of the existing code should *break* because of this hook when the user has this extension installed.

zalmoxisus:
> I understand why the Redux Chrome extension is great. But I disagree Redux DevTools should become one project with it and that we should remove support for using DevTools as a component.

Definitely. Redux DevTools offers you the advantage of having a custom monitor specifically for your app, which is as easier as creating a React component, and yes it is faster when having a big amount of data, due to the fact that the extension has to serialize it (the serialization can be disabled, but not for all cases).

> I think we can take something like this given that it satisfies a few conditions:
> It uses something Redux-specific like `__REDUX_STORE_ENHANCER__`.

I'm fine with adding an alias like this. So we'll not break previous versions and can further call `window.devToolsExtension.open()`.

> It handles scenarios where this particular enhancer is already applied, or when you already use vanilla Redux DevTools, or when you use other enhancers like persistState from DevTools repo, or Redux Loop, or Redux Router. None of the existing code should break because of this hook when the user has this extension installed. How to ensure this? I don’t know but I’m open to ideas.

Keeping the extension's enhancer at the end of the `compose` (applying first) works well with other store enhancers ([even with some edge cases when the store is recreated inside the enhancer](https://github.com/zalmoxisus/redux-devtools-extension/issues/93)). The only case I think is `persistState`, which should be applied before the instrumentation, but till [we have it inside the extension](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/src/app/store/configureStore.js#L15) that's not a problem.

Not sure when having two DevTools instrumentation enhancers, if we should share the same `liftedStore` objects or disable previous instrumentation enhancer.

gaearon:
>Not sure when having two DevTools instrumentation enhancers, if we should share the same liftedStore object between them or disable previous instrumentation enhancer.

I’m not sure either. If somebody is willing to look into this and find a good compromise, maybe we can make this happen.

calebmer:
Opened https://github.com/zalmoxisus/redux-devtools-extension/pull/99 to rename `window.devToolsExtension`.

I'll see what I can do about the reducer not affecting other code, but again I'm no expert 😐

zalmoxisus:
@gaearon, I figure sharing `liftedStore` between instrumentation enhancers can lead to unexpected behaviours, and usually we don't need having multiple simultaneous Redux DevTools. So, disabling the previous instrumentation would be better. It will also allow us passing `window.devToolsExtension()` when needed to [provide additional parameters to the extension's enhancer](https://github.com/zalmoxisus/redux-devtools-extension#api).

Implemented it in https://github.com/zalmoxisus/redux-devtools-instrument/pull/1.

@calebmer, thanks! Will look into the changes.

calebmer:
Can we have the in browser DevTools move to the extension's window if both are simultaneously installed?

zalmoxisus:
@calebmer, the reason you apply vanilla DevTools is that you want it in the browser's page not in the extension's window. Anyway, it wouldn't be possible to render app's react components (monitors) on the extension side.

PR #1658, opened 2016-04-23
boundActionCreators can be nested

xialvjun:


gaearon:
Hi! This has been proposed before. Please see https://github.com/reactjs/redux/pull/1451 for why I’d rather not take it.

xialvjun:
Got it. Thanks for your time...By the way, es2015 spread attribute is usually enough...

Issue #774, opened 2015-09-22
does Redux support IE8 ?


FuGardenia:
react support IE8 with [react-ie8](https://github.com/xcatliu/react-ie8),

does Redux support IE8 ?

edge:
Redux intends not to support IE8.

https://github.com/rackt/redux/blob/master/webpack.config.production.js#L14

FredyC:
Yeah, I am wondering who might need that. Developing new application to support IE8? Why? I doubt you would be adding Redux to old running application just for the fun.

FuGardenia:
@FredyC In China， IE8 has 22.39%.   http://tongji.baidu.com/data/browser/  ， we cant abandon them

FuGardenia:
@edge tks~ If I want to continue using Redux, what should I do ？

FredyC:
@FuGardenia I guess you would have to make your own fork, but I am not sure how many incompatibilities are there, IE8 is really out of my interest since WinXP are gone.

aphillipo:
Redux really doesn't do that much.  Surely if you polyfill it should be fine?  I'd be interested in hearing how you get on FuGardenia.

gaearon:
There is no need for a fork. The `screw_ie8` line is from Webpack config, which is only relevant for UMD builds (the one on cdnjs and `dist` folder). If you use Redux from NPM as CommonJS modules, it's up to you to minify it any way you like (you get to pick Uglify settings in your own project obviously).

Redux should work fine in any browser supporting ES5. You can use [es5-shim](https://github.com/es-shims/es5-shim) to get the missing methods for really old browsers.

FuGardenia:
@gaearon tks~  Now it is clear for me.

tlightsky:
helpful for me too

gggin:
helpful for me too

gaearon:
Note that we plan to drop support for IE8 in the next major version. It is not supported neither by Microsoft nor by React anymore.

PR #1659, opened 2016-04-24
fix(Symbol.observable): update symbol-observable to fix IE8 issue

blesh:
This updates symbol-observable dependency to be 0.2.3 or higher in order to fix an issue where legacy browsers did not like `Symbol.for` statement inside of the ponyfill

related #1632
related #774


gaearon:
Out in 3.5.2.

PR #1663, opened 2016-04-25
chore(dependencies): update symbol-observable to 0.2.4

blesh:
This updates symbol-observable to 0.2.4 which resolves edge cases where a user may have another polyfill for Symbol.observable in place that results in redux store as not being identifiable as 'observable'. See http://github.com/blesh/symbol-observable/issues#7 and https://github.com/sindresorhus/is-observable/issues/1#issuecomment-214479549

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "lodash": "^4.2.1",
     "lodash-es": "^4.2.1",
     "loose-envify": "^1.1.0",
-    "symbol-observable": "^0.2.3"
+    "symbol-observable": "^0.2.4"
   },
   "devDependencies": {
     "babel-cli": "^6.3.15",


gaearon:
👍 

I won’t cut a release right now, as new `npm install`s should just work. If there aren’t more fixes this week, I can cut a patch in the end of the week.

blesh:
@gaearon yeah, this is such a minor issue I can't imagine it requiring an immediate release. Very edge-casey, IMO.

imranismail:
What does symbol-observable brings to redux? I don't see it documented anywhere yet.

gaearon:
>What does symbol-observable brings to redux? I don't see it documented anywhere yet.

Yeah, we probably won’t document this extensively because it’s a relatively niche feature. You can read about it in #1632.

PR #1664, opened 2016-04-25
Added support to combineReducers to make the coding of loading state easier.

thaggie:
With the existing implementation of `combineReducers` if you try to add an action to load
the state every reducer will get the entire state passed to them and would need to have
knowledge of where they are in the state tree to lookup their new state.

I've added a special HYDRATE action type (`@@redux/HYDRATE`) which combineReducers looks for and handles as a special case passing down only the state for each reducer.

This means that each reducer can opt in to the load and choose how it wants to handle it.
Usually this will just be returning `action.state`.

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -126,9 +126,13 @@ export default function combineReducers(reducers) {
       var key = finalReducerKeys[i]
       var reducer = finalReducers[key]
       var previousStateForKey = state[key]
-      var nextStateForKey = reducer(previousStateForKey, action)
+      var appliedAction = !action || action.type !== ActionTypes.HYDRATE ? action : {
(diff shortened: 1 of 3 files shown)


acdlite:
This strikes me as serving too specific a use case. We try to keep these features in userland, rather than modify core APIs that affect everybody.

Here's a way you could implement the same feature by composing the existing implementation of `combineReducers`:

```js
const makeHydratable = (reducer, key) => (state, action) => {
  let appliedAction = action.type !== HYDRATE
    ? action
    : { type: HYDRATE, state: action.state[key]}
  return reducer(state, appliedAction)
}

const combineReducersWithHydration = reducersMap =>
  combineReducers(mapValues(reducersMap, makeHydratable))
```

PR #1665, opened 2016-04-25
Use object literal property value shorthand consistently

joshburgess:
All action functions returned objects using the property value shorthand for the 'reddit' key & value except for 'receivePosts'. I updated it simply for consistency. I think it was just an oversight.


acdlite:
Sweet, thanks

Issue #1677, opened 2016-04-28
Production version


buzinas:
Looking at Redux, I can see that there are lots of expections handling, such as:

```javascript
if (!isPlainObject(action)) {
  throw new Error(
    'Actions must be plain objects. ' +
    'Use custom middleware for async actions.'
  )
}
```

I know that the minified version is already tiny enough (~2kb after gzip), but I think that most of those errors are for development use, and shouldn't exist in a production/minified version.

In React, if we set our `NODE_ENV` to `'production'`, they remove lots of those exceptions handling, and this decreases the code size considerably.

Wouldn't be good to do the same in Redux?

Aweary:
Redux is already pretty small, and I think most of these errors are essential. If the invariants they check are violated redux will likely throw a more cryptic error somewhere else in the code path.

buzinas:
@Aweary But those errors should **never** happen in production, right?

Most of them are `TypeError`, and if you use TypeScript, for example, you catch them before runtime. If you don't, you'll obviously catch that during development, and only if you're making an *enormous* mistake, there'll be an error like this in production.

Although it's pretty small (as I already said), it can be even smaller - **much** smaller, in fact, probably ~60% smaller, or something like that.

Aweary:
I'd assume they never *should* happen, and a typing system should prevent >95% of situations where it might, but I'd be interested in hearing what @gaearon or @timdorr think about it. There's been [some community support](https://github.com/facebook/react/issues/2686) for better production error handling in React, though redux is much simpler so I'd assume most people won't have a problem with this.

timdorr:
React and Angular have production error messages for much the same reason: Your application code isn't statically analyzable (at least not in JavaScript), so you're always going to have some error that slips through the test suite. It happens. 

Rather than producing unhelpful errors like the classic "Cannot read property 'x' of undefined", having some way to point to the source of the issue is going to be drastically more helpful to deciphering what's going on. 

They may seem wasteful, but so is something like Rollbar or another error tracker, since it doesn't do anything if your code works perfectly. But for the same reasons you add tools like that, you need some amount of error handling for when things go wrong. And the closer to the error point that you can get, the better.

buzinas:
OK, thanks!

PR #1678, opened 2016-04-28
Production version (#1677)

buzinas:
If the suggestion I made in #1677, this PR already handles it.


gaearon:
This is dangerous. We must never have behavior differences in production and development. We can’t throw in development and then not throw in production because this might cause hard to reproduce issues.

I’m open to consider a variation on this that changes some of these errors to be warnings. I would only do this for cases which are very perf-sensitive (e.g. inside `dispatch`). It is fine to disable warnings in production because they do not affect which code path gets executed—unlike errors.

buzinas:
OK, thanks!

PR #1683, opened 2016-04-30
Fix package.json main script path

jvanbrug:
Update `main` script path in `package.json` to reference `src/`, not `lib/`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "redux",
   "version": "3.5.2",
   "description": "Predictable state container for JavaScript apps",
-  "main": "lib/index.js",
+  "main": "src/index.js",
   "jsnext:main": "es/index.js",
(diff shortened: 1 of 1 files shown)


TrySound:
@jvanbrug `src` is es2015 code which doesn't work in all environments. `lib` contains code processed with babel, `es` contains processed code but with es2015-modules. Your PR doesn't make sense.

mxstbr:
The current behaviour is correct, as @TrySound mentioned, we want to give users access to the processed code, not the source! If users don't have an ES* transpiler and we only give them access to the source, they aren't able to use it, which we want to avoid.

Thanks for taking the time to submit this PR, we appreciate it a lot!

jvanbrug:
Noob PR, thanks for the insight!

gaearon:
No problem. Thanks for taking the time!

PR #1685, opened 2016-05-01
Fix real world to be case insensitive

### LINKED ISSUE #1684: Real world example si stuck loading due to upper cases
Hey there!

GitHub's api is case insensitive, which means that searching for the user "zhouzi" will result in "Zhouzi" and "zhouzi/theaterjs" in "Zhouzi/TheaterJS". Right now, it's causing the UI to be stuck loading because there's no "zhouzi" user in the entities (but there's "Zhouzi").

Zhouzi:
This PR fixes #1684

### DIFF
--- a/examples/real-world/containers/RepoPage.js
+++ b/examples/real-world/containers/RepoPage.js
@@ -73,7 +73,11 @@ RepoPage.propTypes = {
 }
 
 function mapStateToProps(state, ownProps) {
-  const { login, name } = ownProps.params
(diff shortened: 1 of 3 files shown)


Zhouzi:
That build's failure doesn't seem to be related to this PR. I just tried to clone the redux repository and `npm install` failed for the same reason (without any change to the code). My guess is that a dependency got updated and introduced a bug.

PR #1690, opened 2016-05-03
Change doc examples to use TOGGLE_TODO action consistently

peterxjang:
The examples in the Basics documentation use COMPLETE_TODO actions all the way up until 2.6 with the complete todo list source code, where it uses a TOGGLE_TODO action.


gaearon:
Thanks!

Issue #1300, opened 2016-01-28
Redux is great but major feature is missing...


born2net:
Redux is a great modeling pattern that I am sure will continue and grow...
I have been using a single source of truth type models for many years and I am happy yo see its becoming popular in the JS eco system.

However Redux is missing a critical part as currently it focuses more on the producer and less on the consumer.
Let me explain, I am talking about the:

```
let unsubscribe = store.subscribe(() =>
  console.log(store.getState())
)
```

currently when an action is produced, all subscribers will receive the callback, however if you ever developed a large application, or worst yet, had to jump into someone else's code you know that the biggest issue is finding your way around. And so if I look at a reducer which is switching on some list of action, I would like to be able to see all the places in the code base that "cares" about a particular (like we always do with events), just right click on the action in your favorite IDE and select find usage.

now I understand that with reducers you can divide the application to multiple smaller reducers to so its easier to find things... BUT, if you are going to refactor a line of code, you want to make sure all subscribers that rely on it will not explode... 

The cleanest way to get that is simply to pass the action to the subscriber as in:

```
let unsubscribe = store.subscribe((typeAction:string) =>
  if (typeAction === Consts.FOO)
    console.log('found foo'); 
 console.log(store.getState())
)
```
and we can find all places in the code that refer to Consts.FOO when we refactor...

It will be much easier to reason about how changes propagate in a large application if we can get the action to be passed to the subscribers.

The second point (which may or may not be valid) is why isn't the store reference passed as well just to save a single line of code: store.getState();

so instead wouldn't that be better:

```
let unsubscribe = store.subscribe((store:Redux.store, typeAction:string) =>
  if (typeAction === Consts.FOO) {
      console.log('I got my store);
  }
)
```

If we already have a subscriber we might as well use it to pass relevant parameters.

Thanks for a great framework and hope we can get it even better,

regards

Sean - founder of http://DigitalSignage.com

gaearon:
>It will be much easier to reason about how changes propagate in a large application if we can get the action to be passed to the subscribers.

Can you please explain on a specific scenario how that would make code easier to understand? In Redux you should re-render UI based on *current state*, not the action. By the time the state is returned from the reducer, the action doesn't matter anymore. For debugging, you can use [redux-logger](https://github.com/fcomb/redux-logger):

[...]

Issue #1243, opened 2016-01-15
I want to get the current action from store.


whatifif:
store has "subscribe" method to notify "the change of state" to all the subscribing clients.

It will be convenient if the current action is able to be sent to the clients in 'subscribe' method.

Current state can be obtained from store.getState().
But there is no such method to get the current action from store 
( for example, store.getAction() which will return the current action dispatched to the store ) 

Any idea is appreciated.

gaearon:
Please see https://github.com/rackt/redux/issues/1057, this is the same request.

gaearon:
Also https://github.com/rackt/redux/issues/1091, https://github.com/rackt/redux/pull/622.

whatifif:
< I found an exact solution to this issue. So I leave the link below.>

Strong advantage: Action can be controlled before acting on store.

# Sending Actions To The Server Using Redux Middleware
[http://teropa.info/blog/2015/09/10/full-stack-redux-tutorial.html]

<code>
-----------------------------------------------------
src/components/index.jsx

import {createStore, applyMiddleware} from 'redux';
import io from 'socket.io-client';
import remoteActionMiddleware from './remote_action_middleware';

const socket = io(`${location.protocol}//${location.hostname}:8090`);
socket.on('state', state =>
  store.dispatch(setState(state))
);

const createStoreWithMiddleware = applyMiddleware(
  remoteActionMiddleware(socket)
)(createStore);
const store = createStoreWithMiddleware(reducer);

----------------------------------------------------------
src/remote_action_middleware.js

export default socket => store => next => action => {
  if (action.meta && action.meta.remote) {
    socket.emit('action', action);
  }
  return next(action);
}

-----------------------------------------------------------
src/action_creators.js

export function vote(entry) {
  return {
    meta: {remote: true},
    type: 'VOTE',
    entry
  };
}
--------------------------------------------------------------
</code>

Issue #1091, opened 2015-12-02
Support to observe the relevant actions only


lnaie:
Hi,

In current implementation, all the subscribers are notified when an action has been performed.
Any plans to support notification of the relevant subscribers only? The use-case would be: I would like to subscribe for the results of a particular action only. Or provide the dispatched action as a param in the subscription callback?

It may not be relevant to reactjs, but it will be useful to the other worlds.

Thanks

just-boris:
As you can see in the [shopping-cart](https://github.com/rackt/redux/tree/master/examples/shopping-cart) example, you can the following reducer:

```
function addedIds(state, action) {
  switch (action.type) {
    case ADD_TO_CART:
      return handleCartUpdate(state, action);
    default:
      return state
  }
```

There is a lot of different actions, but you can change state only on some of them. If you get an action, which is not interesting for you, just skip add and go to `default` section.

Does it help for you?

lnaie:
That is clear. is the reducer part of the flow.
The use-case being: In this particular page I would like to be notified only be the action x, it probably relates to the way subscribers are being handled, more exactly this part of dispatch():
'listeners.slice().forEach(listener => listener())
return action'

just-boris:
This is the main concept of Redux. Listeners shouldn't know about recent action.
Here already was a discussion in https://github.com/rackt/redux/pull/622

If you want to handle specific action, just rely on *result* of that action, not the action itself

lnaie:
Did help. Thanks! :)

PR #1697, opened 2016-05-05
Add action as parameter to store.subscribe funcListeners 

slashhuang:
redux is awesome
when in use, I need to **add subscribe listeners with action as the parameter.** that doesn't really affect redux to be put in use and is only a patchy stuff.

the change is like the following:
`   listeners[i]()        == change to =>       listeners[i](action)`
**hope the minor change pull request can be merged**

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -176,6 +176,7 @@ export default function createStore(reducer, initialState, enhancer) {
     for (var i = 0; i < listeners.length; i++) {
       /**
        *  add action parameter to listeners,so that subscribers can dictate action
+       *  and use action to customize developer's code stuff
        */
 
       //listeners[i]()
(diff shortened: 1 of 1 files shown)


gaearon:
Hi, this has been proposed before. We don’t want to do this because Redux state listeners should depend on the state, not on the action. If you want to have logic specific to the action, middleware is the place to do that.

>that doesn't really affect redux to be put in use and is only a patchy stuff.

It’s a core API change, and it would affect Redux apps in a big way. If some API exists, people start to use it, and later we can’t remove it or change it because people depend on it. We don’t want people to depend on this pattern because it is contrary to ideas of Redux.

https://github.com/reactjs/redux/pull/622
https://github.com/reactjs/redux/issues/1243
https://github.com/reactjs/redux/issues/1091
https://github.com/reactjs/redux/issues/1300

Sorry I couldn’t be more helpful!

slashhuang:
thank u ,by the way I really appreciate the design of redux @gaearon

Issue #897, opened 2015-10-14
how to compose redux reducers with the same actions?


ccorcos:
I'm just getting started with redux and it looks amazing, but I'm a little worried about abstraction.

Suppose I have a really simple input component

```js
inputReducer = function(state='', action) {
  if (action.type === 'ON_CHANGE') {
    return action.value
  }
  return state
}

inputOnChangeAction = function(e) {
  return {
    type: 'ON_CHANGE',
    value: e.target.value
  }
}
```

Now its trivial to hook this up to a component with an input.

```js
reducer = combineReducer({
  username: inputReducer
})
```

But what happens when I have more than one input?

```js
reducer = combineReducer({
  username: inputReducer
  password: inputReducer
})
```

Ok, I suppose we could change the action type to specify which input its referring too.

```
inputReducer = name => (state='', action) => {
  if (action.type === 'ON_CHANGE_'+name.toUpperCase()) {
    return action.value
  }
  return state
}

inputOnChangeAction = name => (e) => {
  return {
    type: 'ON_CHANGE_'+name.toUpperCase(),
    value: e.target.value
  }
}

reducer = combineReducer({
  username: inputReducer('username')
  password: inputReducer('password')
})
```

 But now suppose I place two of the same views side-by-side? Now I need to modify the action yet again to specify which input in which view I'm referring to.

```
inputReducer = index => name => (state='', action) => {
  if (action.type === 'ON_CHANGE_'+name.toUpperCase()+'['+index+']') {
    return action.value
  }
  return state
}

inputOnChangeAction = index => name => (e) => {
  return {
    type: 'ON_CHANGE_'+name.toUpperCase()+'['+index+']',
    value: e.target.value
  }
}

reducer = function(state:[{username:'', password:''},{username:'', password:''}], action) {
  return state.map({username, password}, index) => {
    return {
      username: inputReducer(index)('username')(username)
      password: inputReducer(index)('password')(password)
    }
  }
}
```

Oy vey. Whats tough about this is that the input reducer somehow needs to know about how the rest of the app is structured. That just doesnt seem right. Ideally we would be able to abstract out the input's actions and reducers based on how the parent decides to arrange them. I suppose we could do this with a high-order function. But this is all getting pretty tedious. Am i missing something? How are you dealing with this?

Nicktho:
[redux-form](https://github.com/erikras/redux-form) is a great tool to use to handle forms. 

[...]

PR #1698, opened 2016-05-06
Compose Reducers with a common action

dustinmoorenet:
```javascript
function firstName(state = '', action = {}) {
  if (action.type === SET_FIRST_NAME) {
    return action.payload
  }

  return state
}

function lastName(state = '', action = {}) {
  if (action.type === SET_LAST_NAME) {
    return action.payload
  }

  return state
}

  firstName,
  lastName
})

function handleSetFullName(state, action) {
  if (action.type === SET_FULL_NAME) {
    const parts = action.payload.split(' ')

    return {
      ...state,
      firstName: parts[0],
      lastName: parts[1]
    }
  }

  return state
}

export default composeReducers(
  reducer,
  handleSetFullName
);
```

reactjs/redux#897

### DIFF
--- a/src/composeReducers.js
+++ b/src/composeReducers.js
@@ -1,0 +1,22 @@
+/**
+ * Composes reducer functions passing a new state object to the function on the
+ * right and each receiving the same action object. The composed function will
+ * have a function signature of (state, action) making itself a reducer.
+ *
+ * @param {...Function} funcs The reducer functions to compose.
+ * @returns {Function} A function obtained by composing the argument functions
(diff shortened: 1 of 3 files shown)


gaearon:
This is not very commonly requested so we will probably not include this in core public API. 

Something like this is available in https://github.com/acdlite/reduce-reducers.

gaearon:
I appreciate the effort though!

dustinmoorenet:
No problem, thanks for taking time to create redux and for looking at my pull request.

PR #1699, opened 2016-05-06
Use lodash-webpack-plugin.

jdalton:
I just released [lodash-webpack-plugin](https://github.com/lodash/lodash-webpack-plugin).

demo

This would ensure you continue to get smaller modular builds and would allow you to expand your Lodash module use without worrying about the overhead of things like [iteratee shorthands](https://github.com/lodash/lodash-webpack-plugin#feature-sets).

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,8 +61,8 @@
   },
   "homepage": "http://redux.js.org",
   "dependencies": {
-    "lodash": "^4.2.1",
-    "lodash-es": "^4.2.1",
+    "lodash": "^4.13.1",
+    "lodash-es": "^4.13.1",
     "loose-envify": "^1.1.0",
     "symbol-observable": "^0.2.4"
   },
@@ -102,6 +102,7 @@
     "gitbook-cli": "^0.3.4",
     "glob": "^6.0.4",
     "isparta": "^4.0.0",
+    "lodash-webpack-plugin": "^0.9.3",
     "mocha": "^2.2.5",
     "rimraf": "^2.3.4",
     "rxjs": "^5.0.0-beta.6",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var LodashModuleReplacementPlugin = require('lodash-webpack-plugin')
 var webpack = require('webpack')
 
 var env = process.env.NODE_ENV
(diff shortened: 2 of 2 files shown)


timdorr:
Neat! Been looking forward to this since I saw you posted about it on Twitter.

This gets a 👍  from me!

rgbkrk:
Oh. My.

gaearon:
>This would ensure you continue to get smaller modular builds and would allow you to expand your Lodash module use without worrying about the overhead of things like iteratee shorthands.

This would introduce a discrepancy between the CommonJS and the UMD builds because CommonJS builds wouldn’t work this way. So adding something like `mapValue` would then be cheap for UMD but expensive for CommonJS because we don’t use Webpack in the CommonJS build process.

To solve this, we could *always* bundle to a single file with Webpack, even for CommonJS. I’m not opposed to this, but it’s a breaking change, as we currently semi-support `redux/lib/*` style of imports. Now that Rollup and Webpack 2 are gaining traction, maybe we can remove support for `redux/lib/*` in the next breaking release as tree shaking solves the same problem more elegantly.

timdorr:
The problem is that people use `redux/lib/*`-style imports to access "private" APIs. We get this more on react-router, where there are more of them to be found. If there is anything like that in Redux, you're now removing access to that code.

gaearon:
>The problem is that people use redux/lib/*-style imports to access "private" APIs

They’re not quite private. I remember promising support for this use case some time during 2.x when tools like Rollup or Webpack 2 weren’t yet available, and some libraries that depended on Redux wanted smaller builds. So any top-level file under `/lib/` is currently considered public API. (`/lib/utils/*` is not.)

jdalton:
> So adding something like `mapValue` would then be cheap for UMD but expensive for CommonJS because we don’t use Webpack in the CommonJS build process.

I assume not all CommonJS use is to re-bundle. Couldn't you defer to the consumer handle to tackle how they want to bundle things. If they find the file size an issue they to could use `lodash-webpack-plugin`.

jdalton:
ping

jdalton:
I'm house cleaning lingering PRs. Since this is trivial to add on your own and this PR has gone cold I'm closing it.

gaearon:
Yea sorry, I’ve been busy with other projects and haven’t had a chance to take another look at this.
I still think that we can go with this only if:

>To solve this, we could always bundle to a single file with Webpack, even for CommonJS.

jdalton:
I'm sure you'll get there.

PR #1701, opened 2016-05-07
feat(compose): optimize one function case

julien-f:


src/compose.js
```diff
@@ -10,11 +10,16 @@
  */
 
 export default function compose(...funcs) {
-  if (funcs.length === 0) {
+  const { length } = funcs
+  if (!length) {
```
timdorr: I'd remove this variable extraction. You don't use it below. It's an unnecessary optimization.

timdorr:
Neat. The branching reads a bit better this way IMHO.

Can you add a test against the new return?

julien-f:
Done.

src/compose.js
```diff
@@ -12,9 +12,13 @@
 export default function compose(...funcs) {
   if (funcs.length === 0) {
```
dsomel21: Should I be using triple = ?

TrySound: Yep. It's strict equality.

PR #1704, opened 2016-05-09
May we speak about React Native instead of React

enguerran:
#I think React is in our head stamped with ReactDOM.
So it may be clearer to specifiate React Native, even if there is no difference between using Redux within React nor React Native.

What do you think?

_Please, tell me if I need to change the branch name, or if I need to open an issue._


mxstbr:
The tutorial actually talks about integrating Redux with React generally, there's nothing specifically about React Native in there. I think it's fine as-is…

gaearon:
Yea, technically this is about React Native, even though most of the content isn’t Native-specific.

PR #1705, opened 2016-05-09
Core enhancer

acdlite:
Following up on [my comment](https://github.com/reactjs/redux/pull/1702#issuecomment-217884624) from #1702, this PR makes `createStoreBase` less opinionated by moving certain invariants (actions must be plain objects; actions must have `type` field) into a `coreEnhancer` that is applied automatically by `createStore`. This allows for a future in which `createStoreBase` (or whatever we end up naming it) can be exported separately, providing a cleaner integration point for userland experiments and alternative createStore APIs that may want to opt-out of those assumptions that, while they may make sense for the majority of Redux consumers, aren't essential to the nature of Redux.

With this separation, `createStore` is now almost exclusively concerned with subscriptions. I like this because it should make all parties happier: tinkerers can opt-out of "conveniences" that they don't like (such as the INIT action) without having to rewrite any subscription logic, which in turn makes the barrier to adding more conveniences to `createStore` lower. This is similar in spirit to how we currently treat `combineReducers`: if you don't like how it works, you don't have to use it.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -1,64 +1,77 @@
 import isPlainObject from 'lodash/isPlainObject'
 import $$observable from 'symbol-observable'
 
-/**
- * These are private action types reserved by Redux.
- * For any unknown actions, you must return the current state.
- * If the current state is undefined, you must return the initial state.
- * Do not reference these action types directly in your code.
- */
 export var ActionTypes = {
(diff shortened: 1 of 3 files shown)


acdlite:
Oops submitted to the wrong branch. Closing and retrying.

PR #1708, opened 2016-05-09
Optimize Redux logo SVG code

cfleschhut:
* spotted & removed empty groups in the logo’s SVG markup first (which probably were leftover layers from the original Illustrator file)
* then fully optimized the code with [SVGO](https://github.com/svg/svgo)


gaearon:
Thanks!

PR #1715, opened 2016-05-11
Add redux-cut to middleware section

leonaves:
Is this really cheeky? 😜


gaearon:
Let’s wait for some more users, issues and versions?

leonaves:
Yeah fair I thought you might say something like that! No worries!

gaearon:
Don’t be shy to submit again. I just think some more real world usage would be needed so we don’t recommend something that might change the API soon.

markerikson:
FWIW, I do have it listed in my Redux addons catalog: https://github.com/markerikson/redux-ecosystem-links/blob/master/middleware.md#action-grouping-and-interception

PR #1720, opened 2016-05-12
Check ES3 syntax compatibility

SimenB:
Ref #1632 when a new dep messed up the syntax for ES3 compatibility.

This small tool won't ensure IE8 compat, but it will throw if the syntax is unsupported. In practice, this means consumers can exclude a module with a runtime flag, instead of messing with it at build time.

The current version failed for the minified umd build, as `default` was unquoted in property access, so I removed `screw_ie8` option.

A screenshot of sample output on mismatch (this is the minified file before the change to uglify). (the diff is just before and after `es3ify` with some color)

image

I added the check to `prepublish` and not the linting as the code has to have been built before the check can be run. Might be annoying if tags have been made and such, but unsure if it's better to have it somewhere else?


gaearon:
Looks good to me, thanks!

PR #1721, opened 2016-05-12
Add expect to devDependencies to run tests

jiayihu:
`expect` package is missing from devDependencies and it's required to run tests in this example.


gaearon:
Are any other examples missing dependencies? Can you delete `node_modules` in the root directory and verify that all other examples work fine with just their own `node_modules`?

jiayihu:
I didn't have node_modules installed in root because I was interested only in the examples :D Anyway I checked `npm start` and `npm test` whenever possible, tried also to play around with all the examples and none of them report errors in terminal or browser console ✅

gaearon:
Thank you!

PR #1722, opened 2016-05-12
Add github star buttons to ecosystem doc

ilyagelman:


gaearon:
Thanks for the effort!

I’m not sure if we want to take this though, it feels a bit noisy.
I’m also not sure that GitHub stars are relevant for the handpicked page.

markerikson:
Inclined to agree.  It does make more sense over in my ecosystem repo, where it might be useful as a way to gauge relevance of multiple libs on one page.  And, as it turns out, @ilyagelman did indeed submit a PR for that: https://github.com/markerikson/redux-ecosystem-links/pull/8 .  My concern there is trying to maintain those, since I'm just adding project links by hand.

gaearon:
Yeah. It’s up to you whether you’d like to do it there, but we’ll keep this as a simple list.
Thanks!

PR #1727, opened 2016-05-13
Data-first Navigation

grahammendick:
Dan, thanks again for [tweeting about my data-first router called Navigation](https://twitter.com/dan_abramov/status/730747596327813120). Navigation is very different from other routers so I thought you’d enjoy seeing it working beautifully with Redux. I’ve added it to your TodoMVC example because I know routing is something you’re looking at for your upcoming egghead tutorials.

You can see that, as well as adding browser history, Navigation simplifies your TodoMVC example. There’s no need for React state anymore because the router keeps track of the selected filter. Also, although it’s not shown here, Navigation plays nicely with your Redux optimisations because the Link components auto update after a transition. So, the active Link is never stale even if the parent component doesn’t update.

I want to give you an insight into Navigation’s data-first approach so I’ve intentionally left the route empty in `configureRouter.js`. That means that `filter` starts out as a query string parameter. All the functionality, like defaulting data values and highlighting the active Link, work with `filter` as a query string parameter. You can then change `filter` to a route parameter by setting the route to `{filter?}` and everything still works without changing any other code. Navigation operates on the data, not the URL.

Another thing that makes Navigation different is that you can test the actual transitions. You can see an example inside `Footer.spec.js` where I transition along the `href` prop of a filter Hyperlink. The transition’s executed by a call to `navigateLink`, after which I assert that the filter’s been set correctly.

### DIFF
--- a/examples/todomvc/components/Footer.js
+++ b/examples/todomvc/components/Footer.js
@@ -1,6 +1,6 @@
 import React, { PropTypes, Component } from 'react'
-import classnames from 'classnames'
 import { SHOW_ALL, SHOW_COMPLETED, SHOW_ACTIVE } from '../constants/TodoFilters'
+import { RefreshLink } from 'navigation-react'
 
 const FILTER_TITLES = {
   [SHOW_ALL]: 'All',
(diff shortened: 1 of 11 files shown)


timdorr:
This is neat! But I don't think we need an example that shows off two libraries at once (Navigation and Redux). Let's keep it simple with just one library at a time. Thanks for your efforts!

PR #1730, opened 2016-05-15
bindActionCreators refactor for a more functional approach

gillchristian:
Just a simple change to embrace a more functional approach.


gaearon:
Hi, thanks. We care about a functional API, not implementation details. In the implementation, it is more important to avoid unnecessary allocations.

PR #1741, opened 2016-05-17
fix(index.js): Add check for process object which might not be defined

natewallace:
If the redux module has been bundled outside of webpack and is running on a client browser the process object may not be defined.  This pull request adds a check to make sure the process object is ok to use.

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -113,7 +113,7 @@ export default function combineReducers(reducers) {
       throw sanityError
     }
 
-    if (process.env.NODE_ENV !== 'production') {
+    if (typeof process !== 'undefined' && process.env && process.env.NODE_ENV !== 'production') {
       var warningMessage = getUnexpectedStateShapeWarningMessage(state, finalReducers, action)
       if (warningMessage) {
         warning(warningMessage)
--- a/src/index.js
+++ b/src/index.js
@@ -12,6 +12,8 @@ import warning from './utils/warning'
 function isCrushed() {}
 
 if (
+  typeof process !== 'undefined' &&
+  process.env &&
   process.env.NODE_ENV !== 'production' &&
   typeof isCrushed.name === 'string' &&
   isCrushed.name !== 'isCrushed'


gaearon:
>If the redux module has been bundled outside of webpack and is running on a client browser the process object may not be defined

The CommonJS build is not intended to be used in browsers. (In fact, there is no way you could run it in a browser because it uses `module.exports` which also doesn’t exist there.)

The compiled UMD build does not contain `process` references because it is precompiled. So I don’t see why this would be necessary.

Thanks for the effort though!

natewallace:
I'm using browserify to bundle the redux code along with some other modules.  There is a browserify option that lets you omit builtins like process which trims down the bundle to only module loading support.

My thinking was that by checking for the existence of process the code would be less dependent on the type of module bundler used.

simplesmiler:
@natewallace I believe browserify [should](https://github.com/substack/browserify-handbook#browserifytransform-field) be checking [`browserify.transform`](https://github.com/reactjs/redux/blob/v3.5.2/package.json#L121) field in `package.json` and applying `loose-envify` automatically. Does not it do that for you?

natewallace:
I'm not creating the redux bundle independently but rather I'm pulling redux into a bundle that includes several other modules so I'm not using the browserify configuration found in this project.  

I already have a workaround for my scenario but thought my pull request would make the code less dependent on a particular module bundler or its setup.  

Just my 2 cents.

gaearon:
The problem with this is it’s a pain to remember and test all possible different configuration. Our advice is simple: if you use CommonJS build, it is up to you to envify it. If you’d rather not, you can use the UMD build as it is.

gaearon:
For the record, this is exactly what React does as well.

simplesmiler:
@natewallace when you use redux as a dependency in your project and bundle your project with browserify, it executes the `loose-envify` transform for redux code, because redux has `browserify.transform` field and `loose-envify` dependency specified. You don't need to manually `loose-envify` unless you are using `process.env` in your own code. Make sure you have `NODE_ENV` specified.

@gaearon please correct me if I'm wrong.

gaearon:
@simplesmiler The exception is when you manually tell Browserify *not* to read local transforms. I think it had some flag for that.

PR #1747, opened 2016-05-19
Add Redux Reset to Ecosystem md

abhiaiyer91:
As per @gaearon suggestion! Thanks!


gaearon:
I now see that we already have https://github.com/omnidan/redux-recycle there. How would you compare them in terms of features?

abhiaiyer91:
@gaearon you're right. we solved the same problem the same way. I just wasnt in favor of having multiple arrays of actions resetting a reducer. I thought it would be unwieldly. But you know, the recycle feature set handles that. So yeah we don't need both. Just keep that one I suppose.

Issue #1735, opened 2016-05-16
Add favicon to the website


gaearon:
This from https://learnredux.com looks rad:

<img width="205" alt="screen shot 2016-05-16 at 15 50 06" src="https://cloud.githubusercontent.com/assets/810438/15293123/6998b6ca-1b7e-11e6-80dc-2dfc19d5bd53.png">

We should do it too!

timdorr:
@wesbos Is it cool if we use that favicon.ico?

wesbos:
Yep - go nuts! 

https://learnredux.com/images/redux-logo-fav.png

preflight:
Gitbook doesn't appear to [support](https://github.com/GitbookIO/gitbook/issues/1254) custom favicons yet for the default theme. If anyone knows how to hack the `<head>` in the generated `_book/index.html` file, I can provide the icon files (for iOS, Android, Safari, Windows, etc.) and the `<link>`/`<meta>` tags.

gaearon:
Done via #1761.

PR #1761, opened 2016-05-23
Use Redux logo as favicon on GitBook docs

jtmthf:
Added favicon and apple-touch-icon to docs as requested in #1735. 

I added the files to the logos directory as that seemed the most logical place to put them. Upon the docs:build step, the image files are copied into the built GitBook.


gaearon:
Thanks! I couldn’t find a way to bust the cache and change the URL to something like `favicon.ico?v2` so it’ll only be visible to those who haven’t visited the docs yet.

jtmthf:
Another more flexible option would be to fork the default theme. As of GitBook 3.0 that's now an option that can be specified in the book.json. The documentation is sparse, but I think I could figure it out. Only consequence is that the theme would be behind on commits from the upstream theme.

gaearon:
I don’t think it’s worth it, tbh. Let’s keep it simple!

PR #1763, opened 2016-05-23
Docs: remove semicolons according to code style

frooeyzanny:
Hi!

Just removed semicolons from docs code, according to code style.


gaearon:
Fair enough, this is the style we use for now.
We’ll probably add them back with #1559 later.

PR #1766, opened 2016-05-24
Docs: add whitespace

frooeyzanny:


gaearon:
The current style is correct.

>The em dash is typically used without spaces on either side.

http://www.thepunctuationguide.com/em-dash.html

PR #1771, opened 2016-05-27
Update index.js

andyhu:
Better ES6 module syntax


SimenB:
This is a breaking change, as IE8 support would be dropped by it.

https://phabricator.babeljs.io/T7322

PR #1781, opened 2016-05-30
Use destructuring for one-line var swapping

mAAdhaTTah:
This is just a stylistic thing, but this can be simplified to one line. [More info](http://www.2ality.com/2016/05/six-nifty-es6-tricks.html) (#4 in the list). Up to you if you want it.


gaearon:
This doesn’t seem to pass lint. In general, we prefer to avoid stylistic changes because they don’t bring much benefit, and sometimes introduce subtle errors. But thanks for the PR!

PR #1789, opened 2016-06-03
Add a warning for undefined properties passed to combineReducers

borkxs:
This is easy to accidentally do with import statements and `combineReducers` doesn't currently give any indication of anything going wrong, it just ignores properties that aren't functions.

```js
// forgot to export
const thing = (state, action) => ...
```

```js
import {thing} from "../actions/thing"

combineReducers({
  thing
})
```

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -95,6 +95,11 @@ export default function combineReducers(reducers) {
   var finalReducers = {}
   for (var i = 0; i < reducerKeys.length; i++) {
     var key = reducerKeys[i]
+    if (process.env.NODE_ENV !== 'production') {
+      if (typeof reducers[key] === 'undefined') {
+        warning(`No reducer provided for key "${key}"`)
+      }
+    }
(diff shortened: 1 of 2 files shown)


src/combineReducers.js
```diff
@@ -95,6 +95,11 @@ export default function combineReducers(reducers) {
   var finalReducers = {}
   for (var i = 0; i < reducerKeys.length; i++) {
     var key = reducerKeys[i]
+    if (process.env.NODE_ENV !== 'production') {
+      if (typeof reducers[key] === 'undefined') {
+        warning(`No reducer provided for key "${key}"`)
+      }
+    }
```
timdorr: Nit: Do you mind adding newlines before and after this new block? It makes it easier to spot the dev-only code.

timdorr:
LGTM once we can get the style fixed up.

borkxs:
Newlines added. 👍

timdorr:
Thanks! I'll tweak some of the messaging after-the-fact.

PR #1795, opened 2016-06-08
Expose action to listener

sandrinodimattia:
Today if you want to know if something changed you have to compare your previous state to the new state and in that case you might need to compare different values with one another.

With this PR we now also expose the action to the listener, making it easier to do something when an action was executed.

```js
store.subscribe((action) => {
  if (action.type === 'SOMETHING') {
     // do something else
  }
});


TrySound:
@sandrinodimattia Was discussed tons of times. https://github.com/reactjs/redux/pull/1697

sandrinodimattia:
Whoops, thanks for pointing that out @TrySound ☺️ 

Closing...

PR #1799, opened 2016-06-11
set default state in todo reducer

Jaxwood:


gaearon:
Hi, thanks for the PR! This is unnecessary in this particular example because we are explicitly initializing it with `ADD_TODO` action. I would rather not take it because there is no code path with which this would run.

PR #1800, opened 2016-06-11
(WIP) Update examples

gaearon:
Just letting everyone I decided to start working on updating all examples.
I’m starting with work @xulien did in #1403 and adding stuff on top.

There is no cohesive plan yet, but I want to:

* Update versions of React, Redux, and React Router
* Remove React Transform
* Add some selectors instead of reading state directly
* Use string literals instead of constants (at least, in simple examples)
* Use more functional components
* Add some comments
* Add object spread transform
* Whatever misc work I might decide to do

I have no idea how long it’s going to take (I’ll work on this in my free time). I’ll likely try to cherry-pick some tests in the open PRs, and use them here.

PR #1802, opened 2016-06-12
Example on replaceReducers() and combineReducer()

ev1stensberg:
######  Added an example to show the usage of how to implement reducers on the fly/ on demand.

**Remember to add this to the homepage if you decide to implement.**

#1800 
@gaearon  please look over so I've not done anything stupid 🎱


ev1stensberg:
@timdorr Can we simplify/revise the example, so we would have an example of this in the documentation?

PR #1805, opened 2016-06-13
Add DefinePlugin for NODE_ENV on Real World Example

willrstern:
`NODE_ENV=production npm start` in the Real World Example always loads `Root.dev` component (which includes dev tools), because `NODE_ENV` is never set on the client:
- add webpack DefinePlugin for `process.env.NODE_ENV`

### DIFF
--- a/examples/real-world/webpack.config.js
+++ b/examples/real-world/webpack.config.js
@@ -14,7 +14,10 @@ module.exports = {
   },
   plugins: [
     new webpack.optimize.OccurrenceOrderPlugin(),
-    new webpack.HotModuleReplacementPlugin()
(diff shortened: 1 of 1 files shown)


gaearon:
This is partly intentional, because just adding `DefinePlugin` will not be enough for a real production setup, and we might give a false impression by providing a half-working setup. This is why I opted for providing no production setup in examples at all.

I plan to fix this as part of #1800 but it will take some time. In the meantime, I would rather not introduce inconsistencies.

Thank you for taking time!

willrstern:
Ah, didn't know that!  Thanks @gaearon for taking the time to explain.  Sorry to take up repo-time on that.

PR #1806, opened 2016-06-13
preloadedState is optional

raineorshine:


gaearon:
Thank you for the PR!
I don’t know TypeScript so leaving this for @aikoven and @ulfryk to review.
(It should indeed be optional, and we should support omitting both `preloadedState` and `enhancer`.)

aikoven:
Isn't the case of omitted `preloadedState` already covered by the first overload of `StoreCreator`?

raineorshine:
I learned Typescript four days ago, so bear with me :).

I needed to create a store with [redux-socket.io](https://github.com/itaylor/redux-socket.io) middleware. The key line is here:

```
const store = applyMiddleware(socketIoMiddleware)(createStore)(reducer)
```

The typescript compiler was erroring out (no matching type definition for call target) until I added the peloadedState argument:

```
const store = applyMiddleware(socketIoMiddleware)(createStore)(reducer, {})
```

I thought I tracked the type definition down to the right place, but it is quite possible that I am off a bit. I am pretty new at this.

Let me know if there is something else I can do to verify the change that needs to be made here without it being too much work for you.

I am using Redux v3.5.2 and Typescript v1.8.10.

aikoven:
@raineorshine You are right, your change to `StoreEnhancerStoreCreator` fixes exactly your issue.

But change to `StoreCreator` is unnecessary and creates a bit of ambiguity for which overload to use when calling it with exactly one argument.

aikoven:
Thanks!

PR #1807, opened 2016-06-14
Correct first argument name into mapStateToProps

scmx:
Having it named mapDispatchToProps can be confusing, since connect takes
a mapStateToProps function as first argument.


markerikson:
Thanks for the fix!

PR #1811, opened 2016-06-16
complete example 'action' and 'filter'，separate component

liuxuewei:
complete example 'action' and 'filter'，separate component，make example more easy to understand


timdorr:
Filtering doesn't belong in the store. It's a local component state. This is intentional to show that not everything goes in your Redux store.

Issue #959, opened 2015-10-27
Multiple Actions one dispatch?


kswope:
Is there currently a way to dispatch multiple actions atomically (is that the right term?).  I suppose I mean two or more actions sent in succession which only triggers one component update (but of course multiple mutations to the store).

Normally it looks like middleware would be the place for this to happen, but I don't see how the middleware chain can handle more than one action passing through.

I know I can just call multiple dispatches in succession, but I'm not sure of the race condition implications of this - but I do it anyway :)

Thanks!

nhagen:
This is probably better handled by using a more generic action type and
just having more reducers listen to it. If there are mutations that would
also trigger after more specific actions are dispatched you can call
functions that do that mutation in multiple reducers.

jasongonzales23:
@nhagen is correct, but if you really need to chain multiple actions make them "thenable"

dispatch(doThis())
.then(() => dispatch(doThat())

it is also true this would need be in the context of some middleware such as redux-thunk

kswope:
Let me come at this from another direction, why is this not a good feature?

```
store.dispatch([
  {type: PREPOP_DIALOG, value:'me@example.com'}, 
  {type:OPEN_DIALOG, value:true}
])
```

jasongonzales23:
because the dialog may not truly be open yet, I believe you want a promise like thing in this case

kswope:
OPEN_DIALOG is a signal to open the dialog box.

Something like this works and has proved effective for me
```
store.dispatch({type: PREPOP_DIALOG, value:'me@example.com'})
store.dispatch({type: OPEN_DIALOG, value:true})

```

But the problem is it probably triggers an update between the two commands - which goes unnoticed.

mindjuice:
See:

    https://github.com/acdlite/redux-batched-updates

and also this thread:

    https://github.com/rackt/redux/issues/542

jasongonzales23:
[...]

Issue #1342, opened 2016-02-01
Redux 4 breaking changes


gaearon:
Creating this issue to keep a pile of breaking changes we want to release at once. **This doesn’t mean we plan to do much breaking changes! Mostly they are cosmetic but since we’re using semver, we want to batch them. This is why this issue exists.**
- [x] Updated TypeScript bindings (for TS 2.0+)
- [x] Remove `ActionTypes` from `createStore` exports. It was never intended to be a part of public API. It is not visible of you import `redux` but since we now support importing `redux/lib/*` officially, we should stop exporting `ActionTypes` from `redux/lib/createStore`.
- [x] Make sure `INIT` name is random (https://github.com/reactjs/redux/issues/1736#issuecomment-219484915)
- [x] Bundle the cjs and es modules (#2358)
- [x] Switch to babel-preset-env and drop support for IE < 11.
- [x] Drop lodash dependency #2599
- [x]  Fix dispatching inside middleware setup. Or document this behavior. #1240.
- [x] Forbid `subscribe()` and `getState()` while reducer is running: https://github.com/reactjs/redux/issues/1568

timdorr:
>  we now support importing `redux/lib/*`

We do? I always viewed any import from outside of `main` (basically any import with a slash) as accessing private APIs. Your `main` should be the list of public APIs provided by the library. The only exception is for reducing your bundle size (until ES6 module builders become more readily available), but you would be doing this as an optimization step after importing something from the public API.

gaearon:
>The only exception is for reducing your bundle size (until ES6 module builders become more readily available), but you would be doing this as an optimization step after importing something from the public API.

Yes, [since 3.0.6](https://github.com/rackt/redux/releases/tag/v3.0.6). `/lib/` is a practical way to do it today and I see no reason to disallow it until ES6-aware bundlers are commonplace. Our top-level API is stable so I don’t expect supporting it to be a pain.

>Then we can import #1123 finally? 

I’m looking into getting this in with `es3ifyPlugin` for now. But future-wise, yeah.

timdorr:
> Our top-level API is stable so I don’t expect supporting it to be a pain.

Yeah, I don't think it's a pain thing either, just more of a philosophical issue. How do we want to go about [importing them for `combineReducers`](https://github.com/rackt/redux/blob/master/src/combineReducers.js#L1)? Move them to `utils` and import from there?

timdorr:
[...]

Issue #1070, opened 2015-11-25
Have any good way to implement multi promise control like Async?


hank7444:
example1, I have 3 apis need to call, if they all resolve, return success and get me all data which return by those apis. If one of them are rejected, return failure to me.

example2, I have 2 apis need to call, if one of them resolve, return success and get me data.
If all of them are rejected, return failure to me.

Have any possible do such like these sophisticated async control on redux?

I have already used apiMiddleware, but it only use on 1 ajax call..

just-boris:
It is not a Redux thing. You can do it with native promise

**example 1**
```
Promise.all([callAPI1(), callAPI2(), callAPI3()]).then(onSucces, onFailure)
```

**example 2**
```
Promise.race([callAPI1(), callAPI2(), callAPI3()]).then(onSucces, onFailure)
```

then you can wrap it code in action using https://github.com/gaearon/redux-thunk or https://github.com/acdlite/redux-promise and it will work

hank7444:
@just-boris  thx!  I will try it~

itsmepetrov:
Also, for combining async actions you can use [redux-combine-actions](https://github.com/itsmepetrov/redux-combine-actions).

gaearon:
Yes, we suggest using middleware like [redux-thunk](https://github.com/gaearon/redux-thunk) or [redux-promise](https://github.com/acdlite/redux-promise) for this. You can then use `Promise.all` and other Promise composition functions. Note that you need to include a polyfill for Promises to use them safely—for example, you may use the polyfill that ships with Babel by importing `babel-core/polyfill`.

Issue #911, opened 2015-10-18
Add 'redux-skip-by-action' to Ecosystem


tshelburne:
I needed to be able to mark specific actions as "skipped" so that a larger action could encapsulate smaller ones without notifying subscribers of incoherent state. I created [redux-skip-by-action](https://github.com/tshelburne/redux-skip-by-action) for this purpose - unit tests are pending, QA tests passing, and any feedback would be really appreciated.

Thanks.

gaearon:
In my view this hurts predictability.
What is the use case compared to just firing one action instead of many?

gaearon:
Instead, the “Redux approach” to this would be to create a batching action creator:

```js
function batchActions(...actions) {
  return {
    type: 'BATCH_ACTIONS',
    actions: actions
  };
}

// usage
store.dispatch(
  batchActions(
    doSomething(),
    doSomethingElse()
  )
);
```

and a higher order reducer:

```js
function enableBatching(reducer) {
  return function batchingReducer(state, action) {
    switch (action.type) {
    case 'BATCH_ACTIONS':
      return action.actions.reduce(reducer, state);
    default:
      return reducer(action, state);
    }
  }
}

// usage
let store = createStore(enableBatching(reducer));
```

gaearon:
We might want to add this to the docs..

fubhy:
I solved this with a middleware that acts on action yielding generators.

tshelburne:
@gaearon I have a situation where I need to set a property (action creator 1), validate via external service the new value of the property (action creator 2), get the initial state via external service of another object based on that property, and then set that object in the state as well (action creator 3), and then finally validate the entirety of the new state via another external service (action creator 4). The place where things fell apart for me was the external service parts - if I allow subscribers to receive a notification after action 1, they receive an incoherent state, but I kept finding myself pushed to make the async calls in the reducers in order to move in the direction you described.

If I'm reading your suggestion right, I basically need to create a list of actions to run at the end of a promise chain that gets dispatched all at once. Rather than dispatching actions to the store to get updated state, I use `getState` initially and run my reducer manually against that "copy" to build the list of simple object actions that will ultimately be run in the store. Does this sound right?

This all felt a bit manual to me before, but I think that *does* pretty drastically improve the predictability of things.

tshelburne:
@gaearon That's a much better solution - thanks for the direction, feel free to close this.

tshelburne:
[...]

PR #1813, opened 2016-06-16
Addresses #959 without affecting the library's existing functionality…

robertleeplummerjr:
….  Ultimately tries to prevent re-renders when multiple states are needed in quick succession.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -150,7 +150,7 @@ export default function createStore(reducer, preloadedState, enhancer) {
    */
   function dispatch() {
     var actions = []
-    var i;
+    var i
     for (i = 0; i < arguments.length; i++) {
       var action = arguments[i]
 
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -122,7 +122,7 @@ export default function createStore(reducer, preloadedState, enhancer) {
   }
 
   /**
-   * Dispatches an action. It is the only way to trigger a state change.
+   * Dispatches action(s). It is the only way to trigger a state change.
    *
    * The `reducer` function, used to create the store, will be called with the
    * current state tree and the given `action`. Its return value will
@@ -139,45 +139,55 @@ export default function createStore(reducer, preloadedState, enhancer) {
    * a good idea to keep actions serializable so you can record and replay user
    * sessions, or use the time travelling `redux-devtools`. An action must have
    * a `type` property which may not be `undefined`. It is a good idea to use
-   * string constants for action types.
+   * string constants for action types. Can handle multiple dispatches so to
+   * keep re-renders at a minimum.
    *
-   * @returns {Object} For convenience, the same action object you dispatched.
+   * @returns {Object|Array} For convenience, the same action object you dispatched.
+   * If multiple actions are supplied, then will be returned as an Array.
    *
    * Note that, if you use a custom middleware, it may wrap `dispatch()` to
    * return something else (for example, a Promise you can await).
    */
-  function dispatch(action) {
-    if (!isPlainObject(action)) {
-      throw new Error(
-        'Actions must be plain objects. ' +
-        'Use custom middleware for async actions.'
-      )
-    }
+  function dispatch() {
+    var actions = []
+    var i;
+    for (i = 0; i < arguments.length; i++) {
+      var action = arguments[i]
 
-    if (typeof action.type === 'undefined') {
-      throw new Error(
-        'Actions may not have an undefined "type" property. ' +
-        'Have you misspelled a constant?'
-      )
-    }
+      if (!isPlainObject(action)) {
+        throw new Error(
+          'Actions must be plain objects. ' +
+          'Use custom middleware for async actions.'
+        )
+      }
 
-    if (isDispatching) {
-      throw new Error('Reducers may not dispatch actions.')
-    }
+      if (typeof action.type === 'undefined') {
+        throw new Error(
+          'Actions may not have an undefined "type" property. ' +
+          'Have you misspelled a constant?'
+        )
+      }
 
-    try {
-      isDispatching = true
-      currentState = currentReducer(currentState, action)
-    } finally {
-      isDispatching = false
+      if (isDispatching) {
+        throw new Error('Reducers may not dispatch actions.')
+      }
+
+      try {
+        isDispatching = true
+        currentState = currentReducer(currentState, action)
+      } finally {
+        isDispatching = false
+      }
+
+      actions.push(action)
     }
 
     var listeners = currentListeners = nextListeners
-    for (var i = 0; i < listeners.length; i++) {
+    for (i = 0; i < listeners.length; i++) {
       listeners[i]()
     }
 
-    return action
+    return actions.length > 1 ? actions : actions[0]
   }
 
   /**
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -77,6 +77,20 @@ describe('createStore', () => {
     ])
   })
 
+  it('can handle multiple dispatches', () => {
+    const store = createStore(reducers.todos)
+    store.dispatch(addTodo('Hello'), addTodo('World'))
+    expect(store.getState()).toEqual([
+      {
+        id: 1,
+        text: 'Hello'
+      }, {
+        id: 2,
+        text: 'World'
+      }
+    ])
+  })
+
   it('applies the reducer to the initial state', () => {
     const store = createStore(reducers.todos, [
       {


Aweary:
This seems like an issue that's already resolved in userland. Per #959 The current [recommended approach](https://github.com/reactjs/redux/issues/911#issuecomment-149361073) is to use redux-batched-actions.

robertleeplummerjr:
I did see that approach but felt ultimately that if one simply wants to dispatch multiple actions, it would be a whole lot easier to allow for that, and iterate through them giving redux the ability to solve it.  Also, I noticed that there were a lot of people asking for this feature (including me and my team) and I just felt like it made sense as a natural progression of the library.

robertleeplummerjr:
As for it being resolved, you are right it was resolved.  My solution requires much less knowledge and is more performant.

To illustrate: Suppose there were two free cars, and a Tesla Roadster and a coal power 1950's VW Bug.  Both get the job done, getting one from point a to point b, but one is just... cooler.

😛 

The above is not meant as a zing to any library or work, just that redux is cool.

robertleeplummerjr:
Thinking further about the issue, I actually don't think [the recommended approach](https://github.com/reactjs/redux/issues/911#issuecomment-149361073) sovles the issue.  If I search for "redux multiple dispatches" (or I reference #959) I get people asking something to the likeness of: "Can redux do multiple actions with one dispatch"?  The above solution is _not_ done in redux, _it is done in react_, and then after it is done, the middleware/addon tells redux to essentially fire a dispatch with the same state it already has, which afterwards calls listeners.  So the recommended solution to get redux to handle multiple state changes isn't even in redux.  This pr does answer #959 directly, inside redux, specifically allowing for multiple actions, and then after the state is satisfied, it calls listeners.

robertleeplummerjr:
@kswope Since you originally opened #959, would you be so kind to offer input?

Aweary:
> I get people asking something to the likeness of: "Can redux do multiple actions with one dispatch"? The above solution is not done in redux, it is done in react

redux-batched-actions provides a higher-order reducer and batching function to use with dispatch. It's not coupled to React in any way.

Redux has purposefully maintained a small API surface area and left enhancements up to third party libraries. I don't think making dispatch polymorphic brings enough advantages; it may or may not be more performant, but it seems like it would be marginally so and at the expense of expanding the API.

robertleeplummerjr:
Could you point me in the direction of the code you are referencing?

robertleeplummerjr:
@Aweary & @markerikson do you feel that a single loop, iterating through the arguments of dispatch is the tipping point in complexity for redux?

markerikson:
"Tipping point"?  Dunno about that.  But, we _do_ have plenty of overloads and such already, and I'm hesitant to support further changes to the API for something that's a relatively niche use case (especially given that there's already a "blessed" userland approach).

Aweary:
>Could you point me in the direction of the code you are referencing?

If you're referring to redux-batched-actions you can [see the source here](https://github.com/tshelburne/redux-batched-actions)

robertleeplummerjr:
@markerikson I understand your hesitation, and I would be hesitant too.  A lot of people use this tool and it is very powerful.  @Aweary thank you for clarification on the recommended approach, I got it mixed up with another approach that was found in userland, and saw that I was ultimately wrong in saying that.  The solution that I was thinking of was: https://github.com/acdlite/redux-batched-updates/blob/master/src/index.js, which uses a now redacted addon in react to accomplish batch actions.

As for this being "niche", lets us assess the numbers so that we have an accurate view of what this adds:

- This week at least one other of my fellow engineers ran into an issue where this would solve the problem of multiple rerenders.  This is probably niche.
- There are a few closed, related, not handled directly in redux, unresolved issues on this project, and perhaps a closed pr (this one looks like it may soon be).  (#544, #797, #959, #1070) Still probably niche.
- If I search for "redux multiple dispatches", I get 58,000 results on the most popular search engine, wow!  

Let us break these result down a bit: 
- at least a couple stackoverflow questions with ansers: (http://stackoverflow.com/questions/35493352/can-i-dispatch-multiple-actions-without-redux-thunk-middleware, http://stackoverflow.com/questions/34799677/orchestrating-multiple-actions)
- There are few blog or blog like posts (http://jlongster.com/Two-Weird-Tricks-with-Redux, https://egghead.io/lessons/react-testing-the-redux-store-multiple-actions, https://www.reddit.com/r/reactjs/comments/3lk6gt/how_to_do_many_tasksactions_from_a_single_ui/, http://ricostacruz.com/cheatsheets/awesome-redux.html, https://twitter.com/sh_mck/status/670777740866359297, http://www.initcode.info/reactredux_implementing_multiple_actions_with_redux_thunk)
Niche, you say?

Seems like niche may not be the right word here, I'm 3 pages into the search!  What if there were a small number of these that were directly related to dispatching multiple actions are limited to 10,000, a relatively big number.  A similar number (http://www.skincancer.org/skin-cancer-information/skin-cancer-facts) will die this year from skin cancer, is that niche?  I'm getting too philosophical, let's shrink the number.  Let's say there are just 100, of the 58,000 results that are related to dispatching multiple actions (yes that is right 0.01%) that is still 100 people who have taken the time (at least 4 to 20 hours) to learn about redux, start to use it, and realize that they need to dispatch multiple actions with redux.  Let us do the math, shall we!


The average salary per year for a software engineer (as I'm sure you are aware) is $99,530 (according to the most popular search engine).  Broken down by hour that is `$99,530 / (52 - 2) / 40 = $49.765` hourly.  We'll round down to $49 hourly.  If 100 people spent 10 hours to finally arrive at that they needed to dispatch multiple actions and spent another two hours looking for the solution, I believe that is `(10 + 2) * 100 = $58,800` worth of niche, and that is humbly without adding myself.

But Han, I must!
There are as well at least 4 solutions dispatching multi actions (https://github.com/acdlite/redux-batched-updates, https://github.com/tshelburne/redux-batched-actions, https://github.com/stephan83/redux-smart-action, https://github.com/acdlite/redux-promise) let us say each of these had a good 50 or so hours under their belt to create their proposed solution: `(4 * 50) * $49 = $9,800`, add that to the `$58,800` and you arrive at `$58,800 + $9,800 = $68,600`.

Now guys, that is a lot of cheddar.


I can think of a few things I'd buy if I had it...


In the humblest way, it just feels like this tiny improvement adds what a lot of people have taking time to both communicate about and to try and find an answer to.  There are many people involved on many levels who would like to see this feature implemented directly into redux.  If you are satisfied with not having this in the api, that is fine and I am fine with that, but my humble and non-argument is that based off sheer numbers, it would find a very cozy home amongst the other bits that are of representation of your fine work.

But please do not feel that I am negative in adding to the already sterling work here.  Please don't think:


I like simplicity and saw that it could be added here for the greater good.  Also, I got to use a few gifs, which is always fun.  Thank you for your hard work, and fully expect this pr to be closed by someone directly in the redux team.  But in the unlikely chance that it makes it into redux, even if there are requests made to refine said pr:

markerikson:
Erm.  Speaking for myself: arguing via gifs is not inclined to make me support _anything_.

markerikson:
Having said that, let me try to respond more directly.

First, I want to be clear: I have been given commit access to the ReactJS org repos, including the Redux repo.  However, I was given that access because I've spent a bunch of time working on the docs, and I've kept my actual commit actions limited to that side of things.  I've participated in discussions regarding code changes, such as this one, but I don't regard myself in any way as an arbiter of what actually makes it into the ReactJS org codebases.

Having said that, I do also have opinions on where the APIs should go, such as this one.

First, there's a couple main conceptual ways someone may want to do "multiple dispatches".  They may simply want to do a number of things in sequential order, or they may be looking to combine multiple actions into one dispatch so that there's only a single subscription notification emitted.  Yours appears to be the latter. 

Regarding the appeal to popularity based on search results: the search approach really isn't valid.   However, of the eight search result links you provided, all but one of them appear to be talking about sequential dispatches, and the other one simply has a link to the "redux-multi" middleware, which _also_ does sequential dispatches.  In addition, when I ran that search myself, I did see roughly 58K search results claimed, but it stopped listing any after 10 pages, and many of the search results appeared to be Stack Overflow clone spam results.  So, as an appeal to "everyone wants this", that doesn't really work.

As for the whole "time is money" argument: I see your point, but I'm basically not going to waste _my_ time coming up with a further response to that aspect.

Ultimately, my opinion is that while the concept _is_ useful, and there _are_ reasons why someone would want it, and it _is_ a relatively small change... I don't think there's enough overall usefulness or value to justify including it in the core.  If someone wants to simply dispatch multiple actions sequentially with a change event for each as normal, all they need are thunks or even just multiple dispatches from a component.  If someone is actually in a situation where they need to batch dispatches so that there's only one change event, there's options out there, and the FAQ points to them.

robertleeplummerjr:
Did I argue with gifs?  I humbly thought I did it with numbers.

robertleeplummerjr:
Hestation on single for loop, lol.  It has been real fellas.  Adios.

thinkloop:
I too think this is an easy and important change to make without the heavy-handed recommended solutions:

`dispatch([action1, action2])`

is pretty clean and straightforward solution.

Is it possible to achieve this using enhancers without forking?

robertleeplummerjr:
This solution isn't: `dispatch([action1, action2])`, it is actually: `dispatch(action1, action2)` (no array), which keeps the entire existing solution intact, and adds the functionality you mention.

thinkloop:
also nice, the array solves the ordering issue as well tho, and an array as input is an unused/unsupported object - this makes a very nice interface:

`dispatch(action)` core

`dispatch(fn)` thunk

`dispatch(array)` batch

Whether the array is in redux or middleware doesn't matter, but I don't think it's possible with middleware, is it? If it is, I'd be happy with that.

gaearon:
It's not about the loop, it's about adding something to the core API. Anything we add we basically have to support forever.

This is not an “easy” change because it would affect any extension to Redux (hundreds of userland middlewares would have to be adapted to support this).

robertleeplummerjr:
This is open source, you can rely on the world to help a little, after all we use this stuff to feed our families and to build a better planet.

Also, I disagree that it would not affect hundreds of userland middlewares directly, it would only affect them if they wanted it.  IE: if they'd like to continue using `dispatch(oneAction)` that is perfectly fine, this pr doesn't change that.  However for those that want to live a little on the wild side and decide at some point to pass in another action to dispatch `dispatch(one, and, you, can, keep, em, comin, boss)` can do so and render will only be called once at the end.

robertleeplummerjr:
Too, it actually is easy, it is a single loop that doesn't affect anything currently.  Doubt me?  I added a unit test, and your existing unit tests don't break.

robertleeplummerjr:
Sorry, the only thing that is affected, is (ironically) the unit test.  Because it is the only thing that currently takes advantage of it.  My bad.

robertleeplummerjr:
I feel like I'm hearing negative about supporting open source.  Come on guys, chins up!  Open source rocks!  The ideas are always bigger than the code that runs them, and this library is quite the idea!

markerikson:
Open source is great, but just a project is open source doesn't mean that every PR must automatically be accepted.  And, the more popular the project, the more changes have to be carefully considered.

gaearon:
First of all I would like to ask everybody to remain calm and respectful to each other. Let's remember each of us has the best intentions.

I'm sorry I haven't addressed this thread timely. I'm currently busy with other projects, and I can't dedicate as much time to Redux—also I'm deliberately trying to move slow here because of the size of the ecosystem. Any decisions are not taken lightly, especially when concerning the core APIs.

Right now it's 3am so I won't dive into details of what I think about this change. I will try to write up my thinking tomorrow. Hopefully we can avoid the dismissive tone, respect each other's choices and opinions, and avoid bikeshedding as we discuss this.

Thanks for your time!

robertleeplummerjr:
Where is that github dance emoji when you need it?

robertleeplummerjr:
💃 <- nailed it

markerikson:
Hmm. Actually, now that I think about it... I'm pretty sure the multi-argument approach is flat-out dead.  That's because enhancers and middleware are expecting a single action, and the given `for` loop change is only executed once the action has made it through all of those.  For example, check out the implementation of the `applyMiddleware` enhancer:

```js
export default function applyMiddleware(...middlewares) {
    var dispatch = store.dispatch
    var chain = []

    var middlewareAPI = {
      getState: store.getState,
    }

    return {
      ...store,
      dispatch
    }
  }
}
```

I'm still very slightly hazy on the exact flow of execution, but I believe this right here will not work with more than one argument to `dispatch`.  The single-argument-array approach _might_ still be technically feasible, but I don't think the multi-argument approach is.

markerikson:
In other words, it's  not a question of how many arguments a middleware itself is trying to dispatch, it's that every single middleware would now have to correctly propagate all arguments along the pipeline.

thinkloop:
If the right hooks are not available to create a middleware/enhancer for the single-argument-array approach, can we investigate adding those hooks to core.

markerikson:
All the hooks needed for a middleware or enhancer to support an array of actions already exist, which is why there are several implementations and variations on this approach already: the ones linked earlier in this thread, as well as the ones at https://github.com/markerikson/redux-ecosystem-links/blob/master/middleware.md#action-grouping-and-interception and https://github.com/markerikson/redux-ecosystem-links/blob/master/store.md#store-change-subscriptions.

thinkloop:
I don't believe any of those solutions do the equivalent of:

```
[action1.action2].forEach(action => dipatch(action));

// trigger subscribers once
```

markerikson:
- https://github.com/tappleby/redux-batched-subscribe : a store enhancer that lets you specify a way to handle each notification, allowing you to do things like debouncing notifications or using React's batched updating.  should allow effectively equivalent behavior
- https://github.com/tshelburne/redux-batched-actions : wrapper action creator + reducer that takes an array of multiple actions, dispatches a new action with those attached to "meta", and reduces the array of actions in a single "dispatch". should _definitely_ be equivalent behavior.
- https://github.com/mrydengren/redux-batch-middleware : looks like an almost identical approach as `redux-batched-actions`
- https://github.com/stephan83/redux-smart-action : a replacement for `applyMiddleware` that adds a bunch of extra abilities on top. Specifically shows an example of a `pushMutiple()` (sic) function in the README that it says would only notify once.


Not quite the same behavior, but related enough to mention:

- https://github.com/tshelburne/redux-skip-by-action : store enhancer to allow skipping notification for certain actions
- https://github.com/nakamura-to/redux-debounce-listener : store enhancer that automatically debounces subscribers.

So there's at least four libs that I think _should_ fit your exact use case, or at least be close enough to work for you.  And, failing that, you could probably just fork `redux-batched-subscribe` and swap out the current dispatch handling configuration for your specific use case.

thinkloop:
Thanks for the effort. These wrap actions into a larger `batch` action, rather than run each action natively. So if you were logging action types, for example, every action would be of type `batch`. Additionally they require importing the special wrapper everywhere you want to use it. They are similar, but not the same, and break the redux paradigm.

robertleeplummerjr:
Unless my logic is wrong, this (https://github.com/tappleby/redux-batched-subscribe/blob/master/src/index.js#L54) if n is actiones, it does `n + 1

robertleeplummerjr:
Sorry for the premature comment, finger slipped... My bad.

Unless my logic is wrong:

- This (https://github.com/tappleby/redux-batched-subscribe/blob/master/src/index.js#L54), if n is actions, it does `n + 1`.
- This solution as well does `n + 1` (https://github.com/tshelburne/redux-batched-actions/blob/master/src/index.js#L10 😜) because it has an extra action.
- This solution (https://github.com/mrydengren/redux-batch-middleware/blob/master/modules/index.js#L6) is the same as `store.dispatch(one); store.dispatch(and); store.dispatch(you) ...`, which causes `n` and `n renders`, which is the exact problem this pr resolves.
- This solution (https://github.com/stephan83/redux-smart-action/blob/master/src/applySmartMiddleware.js#L58) composes a chain (I think.  Wow, there is a lot going on in there, amazing) and then calls `store.dispatch`, it too is `n + 1`.

This pr does `n` actions and one render.  I could be wrong, and probably am, but I believe the above links handle the problem indirectly, whereas this pr handles the problem directly.

This pr is performant and light, and does not cripple in any way the above solutions.

robertleeplummerjr:
Too, I would like to address the changes in architecture mentioned above.  While they are simple, and I wholeheartedly feel that a pr that considers them would be nice to have for review, I ultimately feel that it is outside the realm of this pr.

robertleeplummerjr:
As for what is mentioned here (https://github.com/reactjs/redux/pull/1813#issuecomment-226916853) and "enhancers and middleware are expecting a single action" if there is interest in this paradigm (multi actions, one render) I can open a pr there that provides a solution if there is interest.  I would think though that that would not come after this pr is either rejected or would come if this pr is accepted.

markerikson:
@robertleeplummerjr : I'm afraid you've misunderstood how each of those libraries works.  Let me try to explain how each of those examples would indeed allow dispatching multiple actions while only producing a single subscription notification.  (Not trying to play https://xkcd.com/386/ , just want to make sure we're all on the same page regarding what is and is not possible with Redux, and how things work.)

So first, remember that a "store enhancer" _wraps_ around a lower-level store definition, whether it be the true Redux base store API, or the API returned from another enhancer.  That means that a store enhancer does need to call the lower-level store's `dispatch` function to handle the action, but  it can present its own versions of the store API functions to the outside world.

If you look at the source for the `applyMiddleware` enhancer I pasted above, we can see this in action.  Normally, `dispatch` is of course the original Redux `store.dispatch`.  `applyMiddleware` sets up the middleware API argument (`getState` + `dispatch`), calls each of the middleware factory functions, and creates a pipeline of the middlewares.  This itself _becomes the new `dispatch`_, and `applyMiddleware` returns _that_ function to the outside world.  _That_ function becomes what the application code would see as `store.dispatch` entirely.

So, if we look at each of the lines you've pointed to in turn:

- https://github.com/tappleby/redux-batched-subscribe/blob/master/src/index.js#L54 : On lines 53-57, the enhancer defines its own brand-new `dispatch` function.  When the app calls that new `dispatch`, it _will_ immediately pass the action to the wrapped "real" `store.dispatch`.  However, the enhancer also defines its own version of `subscribe.  All calls to `store.subscribe` get handled by the enhancer instead of the real store, so when the real store tries to notify listeners, there _are no listeners_.  Meanwhile, the enhancer is running its own `notifyListenersBatched` function, which executes whatever `batch` callback the developer gave it during the store setup process.  So, per the examples on the lib's front page, that could be something like a debounce function, and so in the end the listeners would only get a single notification after a quick series of actions were dispatched.

- https://github.com/stephan83/redux-smart-action/blob/master/src/applySmartMiddleware.js#L58: this is another store enhancer, and is doing significantly more work internally.  I'm skimming the code and not quite following exactly how it works, but again, as a store enhancer, it's substituting its own implementation of `dispatch` for the outside world, and only using the "real" `dispatch` internally when it's ready to.

- https://github.com/tshelburne/redux-batched-actions/blob/master/src/index.js#L10: if you pass multiple actions to the `batchActions` action creator, it creates a new action that puts the array of actions as the `payload`.  The higher-order reducer wraps around your "real" reducer, looks for that `type : BATCH` action, and calls your "real" reducer repeatedly with each action in the array.  Since the higher-order reducer is only being executed once, from the store's perspective it's all one action being handled by one call to the reducer, so there's only one notification.

- https://github.com/mrydengren/redux-batch-middleware/blob/master/modules/index.js#L6: exact same batched reducer concept as the previous one

So yes, overall, each of those examples should indeed allow some form of ensuring that only a single subscriber notification is sent after some series of actions is dispatched.

markerikson:
@thinkloop , @robertleeplummerjr :

> This is a beautiful api:

>` dispatch(obj) // core`
> `dispatch(fn) // thunk`
> `dispatch(array) // batch`

Sure, I see some progression there.  And with that, remember that even thunks were split out of the Redux package early on.  The _only_ thing that core Redux itself supports -  the original `createStore` implementation - is dispatching a single plain object.  Anything beyond that requires a store enhancer, and Redux includes the `applyMiddleware` enhancer in the main library because of its obvious usefulness.  So, based on that, there's no reason to put this behavior directly into the core `createStore` implementation, and the precedent would be to handle it as an enhancer.

robertleeplummerjr:
@markerikson Did you see https://github.com/reactjs/redux/pull/1813#issuecomment-226943127?  That should be discussed elsewhere.

robertleeplummerjr:
@markerikson
> @robertleeplummerjr : I'm afraid you've misunderstood how each of those libraries works. Let me try to explain how each of those examples would indeed allow dispatching multiple actions while only producing a single subscription notification.

Lol, you are totally right, thanks for the clarification!  It was early and I had not had any coffee yet.

markerikson:
> Lol, you are totally right, thanks for the clarification! It was early and I had not had any coffee yet.

Heh, no problem.  Totally with you there :)

robertleeplummerjr:
💯

markerikson:
So just for comparison purposes, I tried copying the `redux-batched-subscribe` implementation and modifying it to take an optional array of actions.  Code's over at https://gist.github.com/markerikson/0383775096f2b32e48ad41499cf8a1fb#file-arraybatchedactionsenhancer-js-L53-L60 .  It's totally untested / proof-of-concept, but hopefully illustrates the basic principle.  The biggest question I had was how to handle the return value from `dispatch`.  Opted to try to match the format of the incoming action.

robertleeplummerjr:
Does it have to return the actions?  Is there a specific reason that we return what we passed in?

markerikson:
That's the contract of `dispatch`.  It returns the original incoming action.  However, as wrapper layers we can't (or shouldn't) just return the incoming value directly, as the return value may have been modified by later layers.  For example, if this enhancer was sitting in front of `applyMiddleware`, each of the middlewares would have chance to modify or replace the return value along the way.  That's how you can do things like `dispatch(somePromise).then(doSomethingElse);`.  The promise middleware would handle the promise, eventually dispatch a plain object, but return the original promise for chaining.

See http://redux.js.org/docs/api/Store.html#dispatch and https://github.com/reactjs/redux/blob/master/src/createStore.js#L144 for reference.

markerikson:
(I do have to say that this whole discussion has forced me to dig pretty deep into the internals of Redux and these other libs, so if nothing else it's been a learning experience and helped solidify my own understanding of what's going on.)

robertleeplummerjr:
very cool, same here, but rather just a few months in.

robertleeplummerjr:
That last value passed in could be passed out, so if you wanted to pass in a specific set of values, the last of those could be used.  That way, if you pass in a single anything that is what you get back, but if you pass in a bunch of arguments, you get the last one back.

gaearon:
Eh, still can’t find the time to do a proper response. Maybe tomorrow 😄 .

But I’d say any decisions need to wait for #1702 because it significantly changes how things are structured internally and may make something like this possible without big downsides.

But you might want to consider: how does this work with `redux-thunk`, `redux-saga`, and `redux-observable`? Can you `dispatch()` arrays from within them? If so, how would you implement that? If not, why not?

robertleeplummerjr:
@gaearon on:
> But you might want to consider: how does this work with `redux-thunk`, `redux-saga`, and `redux-observable`? Can you dispatch() arrays from within them? If so, how would you implement that? If not, why not?

I would say if we were to reach a consensus here, that multiple arguments could be passed in, a pr must be opened to handle `redux-thunk`, `redux-saga`, and `redux-observable`'s.  Please note as well, that this pr isn't about dispatching arrays, but rather dispatching multiple arguments.

I will say, though, that my initial thought was to handle arrays, but I found this is so much simpler:
`dispatch(arg1)` or `dispatch(arg1, arg2)`.  The reason: it doesn't change the existing architecture so that it needs to handle different types, but rather allows you to handle more of the same types.

Aweary:
>  a pr must be opened to handle redux-thunk, redux-saga, and redux-observable
Another PR on Redux or a PR for each of those projects? I believe @gaearon is asking what kind of changes, specifically, would be required to ensure full backwards compatibility?

markerikson:
@robertleeplummerjr : That's sort of the point.  Any changes to the Redux core itself _must_ either be compatible with the rest of the ecosystem as-is, or have an _extremely_ strong justification for breaking compatibility.  There's certainly been discussion of other possible breaking changes (see #1342 ), but most of those are more about internals than the public API.

As far as I can see at the moment, the multi-arg dispatch approach would be incompatible with every existing middleware and enhancer out there.  So, that's not going to be viable.  Dispatching an array full of actions might at least be theoretically feasible, but would require a lot of thought and consideration (and still potentially be too incompatible).

robertleeplummerjr:
@Aweary @markerikson This pr is already backwards compatible, unless either the unit tests do not check for that, or I am mistaken that `argument[0]` will be something other than `action`.  The change specifically would be to check for multiple arguments and simply iterate over them, nothing more or less.  

@markerikson do all the libraries you mentioned override the existing `dispatch`?

Aweary:
i believe they all (or most) assume that dispatch is only passed one action, which would mean they'd require an update to support this new API

markerikson:
The issue is not whether or not the core implementation of `dispatch` can accept multiple arguments.  It's that effectively every library so far, both middlewares and enhancers, assumes that `dispatch` will only take one argument.  So, enhancers like `applyMiddleware`, the Redux DevTools, or any other enhancer that does provide its own version of `dispatch`, are only looking for one argument to their own `dispatch`, and only passing that one argument onwards to whatever version of `dispatch` they wrap (which may be the real Redux store `dispatch`, or the version provided by the next enhancer in the sequence).  Not all enhancers override `dispatch` (some, like `redux-localstorage`, might override other functions like `subscribe` instead), but many do.

So, if I am using an enhancer like `applyMiddleware`, and I call `store.dispatch(action1, action2)`, action2 will be completely ignored.  Only action1 will be passed onwards to the real store, and the `for` loop in the core `dispatch` implementation will be effectively useless.

Without looking, I would guess that the current unit tests are simply checking the core store's behavior, not the store's behavior with enhancers applied.  Probably a situation where more tests could be used, but again, that's because the core contract being both provided and consumed right now is that there's only one argument to `dispatch`.

robertleeplummerjr:
We could always just iterate over the arguments here to maintain backwards compatibility: https://github.com/robertleeplummerjr/redux/blob/master/src/applyMiddleware.js#L27

robertleeplummerjr:
Also, @Aweary and @markerikson, thank you very much for your insight.

markerikson:
*sigh* 

I'm trying to be patient and explain things clearly, but it doesn't seem like you're grasping what I'm saying:

The problem is _not_ the core version of `dispatch` itself.  The problem is _not_ the `applyMiddleware` enhancer.  The problem is that there's an ***entire ecosystem*** built around this contract, and your current proposal, as best as I understand it, ***will.  not.  work.*** with the rest of the ecosystem as written.  _Any_ user with _any_ enhancer added to their store would be unable to use the proposed multi-argument syntax for `dispatch`, and that effectively means _everyone_.  

Breaking changes to library APIs do happen.  The concept of dispatching multiple actions in and of itself is not bad, outside proposals are generally worth discussing, and there are times when it's worth breaking backwards compatibility to move forward by dropping outdated approaches or allow new uses cases that weren't possible previously.  But, it's rarely as simple as "add two lines and one unit test", especially when it's a root library with so many people depending on it, and decisions have to be made with both the maintainability of the root library and the interaction with the rest of the ecosystem in mind.

robertleeplummerjr:
Here is the rub, the original need for this I ended up simplifying into a single action.  That isn't always going to be the case, but yea, it is what it is.  To begin with, I naively didn't think this change was as large of a change as am now seeing it is.  I was mistaken.  It doesn't seem like this is needed here and I'm fine with closing this issue if there you guys are in agreement.  I'd hate to take more dev time here, if it isn't needed.

robertleeplummerjr:
I went ahead and closed it.  Thanks everyone for your insight and putting our heads together and thinking about what this truly meant for not just the lib but the entire ecosystem, especially @markerikson & @Aweary!  Thanks again guys.  Hopefully my next pr will be one that _is_ needed 😆.

markerikson:
Sure.

Again, I'm not trying to slam you, or say that the PR isn't wanted at all, or that it's inherently a bad idea.  There's been some good discussion, I've learned a fair amount from digging through the code so I can explain things, and as you said, OSS projects do grow from their community.  It's just an issue of maintenance, library philosophy, and ecosystem compatibility.

thinkloop:
I'm not sure if it is ready for closing yet, as "there are times when it's worth breaking backwards compatibility to move forward".

For the record this wouldn't "break" implementations in the traditional sense, it's just that they wouldn't be making use of the additional actions - somewhat of a break, but not a "my code imploded break".

There are major version numbers for a reason. I think this is worth thinking about.

robertleeplummerjr:
How about I close it for the time being, until it becomes something actually needed?

thinkloop:
Additionally I think @robertleeplummerjr api is better than the array-based one I proposed, for a couple of reasons. One is that we would be consuming a whole data type that enhancers now can't make use of for other purposes. And secondly this change is very meta, in the sense that I'd still want all the enhancers to work with it: `dispatch(action, thunk, thunk, action)`.

robertleeplummerjr:
@thinkloop ah snap! 
> `dispatch(action, thunk, thunk, action)`
Sophisticated, yet simple.


@thinkloop this is for you bud.

Aweary:
> especially @markerikson & @Aweary! Thanks again guys.

Always happy to help in any way possible! 😄 


> For the record this wouldn't "break" implementations in the traditional sense, it's just that they wouldn't be making use of the additional actions - somewhat of a break, but not a "my code imploded break".

Right, backwards compatibility was the wrong term here, as current implementations should work. The main issue is that this new API will likely be broken for the majority of users until all the major enhancers were updated to support this new variadic API.

>I'm not sure this is ready for closing yet, as "there are times when it's worth breaking backwards compatibility to move forward".

The question is, does this provide enough of a benefit to disrupt the ecosystem like that? As discussed this behavior is already available in userland (for the most part) in a way that is fully compatible with the current ecosystem. 

> There are major version numbers for a reason, and enhancers can define which they are dependent on. I think this is worth talking about.

Creating a division like that is not ideal. You *could* release a new major version that supports this and ask the entire community to support the new API, but that's asking a lot.

Ultimately it's up to @gaearon and the other maintainers, but from my perspective, this provides minimal benefits over the current options and would require a non-trivial amount of aggregated development time to bring the community enhancers and tools to a place where this change has any value in the first place.

thinkloop:
Weighing the costs/benefits is definitely important here, and I am not taking the position that this is absolutely certainly something that should make it, but I would like to make the benefits clear so we can think through it properly.

Redux is a very simple lib, it basically does two things: dispatch actions, and respond to them. This feature is at the very essence and purpose of redux, and not some secondary extraneous nice-to-have functionality. If redux is meant to solve state management for all/most apps, some degree of sophistication/nuance is to be expected from these core activities.

The feature being proposed is not esoteric or edge-case either, it's the standard one v. many situation that almost everyone eventually upgrades their code to: `getOrder(orderID)` always becomes `getOrders([orderID])` soon after.

The userland solutions are contorting themselves to severe levels to bypass the existing limitation, and push up a lot of unnecessary cruft and complexity to the dev and app, breaking the raw, clean redux paradigm we are used to. Additionally none of them solve the issue perfectly, each having their own costs and problems.

Finally, and most importantly, if this feature were introduced *everyone* would use it. There will be an extreme minority of people who will never have a need to process two actions to complete a task. As useful as shouldComponentUpdate is, nothing beats omitting the request in the first place.

> You could release a new major version that supports this and ask the entire community to support the new API, but that's asking a lot.

Redux has the big advantage of being very simple and unchanging, devs could stay on this current version probably forever without issue, if their favorite middleware didn't upgrade. That is, the issue of handling breaking changes is probably most easily handled by a lib like redux than almost any other.

markerikson:
Per my earlier comment, there's a distinction between "wanting to dispatch multiple actions sequentially", and "wanting to dispatch multiple actions with only one notification".  Dispatching sequentially is easily achieved using thunks - just `dispatch(firstAction); dispatch(secondAction);`.  Dispatching with only one notification is a much less common use case.

Aweary:
> Finally, and most importantly, if this feature were introduced **everyone** would use it. Literally everyone. There will be an extreme minority of people who will never have a need to process two actions to complete a task. As useful as shouldComponentUpdate is, nothing beats omitting the request in the first place.

That's the thing, people already do this. You can dispatch two actions from a parent action using something like redux-thunk. The only advantage this brings is that it will only produce one subscription listener call for parallelized actions, which people already handle using techniques like debouncing or batching.


> Redux has the big advantage of being very simple and unchanging, devs could stay on this current version probably forever without issue

But you just established that

> if this feature were introduced everyone would use it.

So how is that a reasonable concession? If the point of this is to introduce a feature that everyone would want to use, we can't just ignore that they'd be unable to use it until every related enhancer/tool they use has been updated to support it. 

Let me be clear, I'd love to see this feature implemented in the way you guys are asking for. I understand current batching enhancers come with some overhead. I just don't think this API specifically is the answer given the immense requirements it brings to the community in terms of compatibility.

If this was implemented it should be implemented in a way that wouldn't require hundreds (or thousands) of packages to update.

thinkloop:
> So how is that a reasonable concession?

The issue with breaking changes is that usually you have a deprecation period, creating a situation where you temporarily maintain 2 lines of code for some period for bug fixes and security patches. With redux, there are so few vectors for bugs and security holes, and few changes in general, that the effort required to maintain a long deprecation period is minimal/non-existent.

> if this feature were introduced everyone would use it.

I hope this isn't turning into a con lol. The point was that people will have the choice to remain exactly as they are, or have the ability to make use a fancy new feature. Nothing lost.

> which people already handle using techniques like debouncing or batching

Debouncing the main render adds a delay for all actions reducing the quality of the app, and causing other problems, I would never recommend this. Batching has the aforementioned high costs in ugliness, complexity, etc.

Good discussion!

P.S. I definitely know you can dispatch more than one action in many ways, thunks being one, everything I wrote was in the context of not having subscribers trigger.

robertleeplummerjr:
We could easily have a method for forward compatibility that would allow for graceful degrade.

robertleeplummerjr:
actually, even simpler:
```js
if (this.noMultiAction && arguments.length > 1) {
  throw new Error('multiple arguments not supported');
}
```

Aweary:
> I hope this isn't turning into a con lol. The point was that people will have the choice to remain exactly as they are, or have the ability to make use a fancy new feature. Nothing lost.

But they **won't** have the ability to use the new fancy feature because this requires all their middleware to be updated to support this. That's my whole point. I'm all for offering a better more robust API that involves batched dispatches with one native subscription notification, but I don't think this is the way to do it given the huge overhead it requires for forward compatibility.

> I doubt it, I'd guess 99% would use it.
We should try to avoid hyperbole as much as possible here. I watch the redux repo pretty closely and I rarely see this being requested as native behavior. redux-batched-subscribe and efficient subscription listeners tend to cover this use case fairly well. @markerikson is constantly engaged with the community and as a maintainer I would give a certain weight to his observations in regards to the demand for this.

@markerikson @gaearon do you think there is an alternative way to address this, such as offering a way to enhance subscription notification behavior?

thinkloop:
> We should try to avoid hyperbole as much as possible here. I watch the redux repo pretty closely and I rarely see this being requested as native behavior. redux-batched-subscribe and efficient subscription listeners tend to cover this use case fairly well. @markerikson is constantly engaged with the community and as a maintainer I would give a certain weight to his observations in regards to the demand for this.

There is no need for this comment. I'm actually good friends with @markerikson and have been chatting with him more than I have anyone else over the last year. He made a guess that "Dispatching with only one notification is a much less common use case." and I am making the opposite guess. Simple as that. Please don't passively aggressively divide us or insinuate I was insulting him in any way.

My guess is also very reasonable - regardless of whether people have been using a non-existent feature or not, or going out of their way to add plugins to make it a feature, just think of basic application needs. It is coming to want to do more than one thing before rendering.

robertleeplummerjr:
To further the `noMultiAction` property, it should not be a property, it should be a variable.  It could be false by default, so that only those who know specifically that they are wanting to use it can switch it on.  Perhaps a better proposal would be:

```
let _isMultiAction = false;

function useMultiAction() {
  _isMultiAction = true;
}

get isMultiAction() {
  return _isMultiAction;
}
```

markerikson:
Need to try to get some more actual work done before I leave for the day.  I'll come back to this discussion later this evening.

Aweary:
> There is no need for this comment. I'm actually good friends with @markerikson and have been chatting with him more than I have anyone else over the last year. He made a guess that "Dispatching with only one notification is a much less common use case." and I am making the opposite guess. Simple as that. Please don't passively aggressively divide us or insinuate I was insulting him in any way.

All I implied was that comments like "everyone would use it" and "99% would use it" aren't necessarily accurate justifications here. If you're interpreting any passive-aggressive intentions on my end then you're imbuing them into my comments yourself, I honestly don't have any interest in who is friends with who or dividing anyone. I'm interesting in making sure Redux progresses in the most universally accessible and beneficial way possible.

>My guess is also very reasonable - regardless of whether people have been using a non-existent feature or not, or going out of their way to add plugins to make it a feature, just think of basic application needs. It is common for an app to want to do more than one thing before rendering.

Right and we've covered that this is possible right now and agreed that this approach brings a certain advantage. We seem to diverge at the point where we make value judgements regarding the pros of this approach and the cons of its requirements.

robertleeplummerjr:
The other thing we could do is simply provide an alternate multi-dispatch function.

```js
dispatchAll(action1, thunk1, thunk2, action2)
```

The last of the arguments (`action2` here) could be returned from the function, so that you may continue as is the current api.

This leaves the existing function intact and provides those who'd like to use it a means that nicely matches up to, for example, `Promise.all`, and solves the multi actions in succession, and one render issue.

To keep the code simple, we could put the in a single location, and use it in both places.  This would provide all the benefits of multi actions, not affect the existing api and those who use it.  A win win...win scenario?

Aweary:
That would still be incompatible with existing enhancers though, right? All middleware/enhancers used would be required to know about and support `dispatchAll` as well as `dispatch`

robertleeplummerjr:
@Aweary that is correct, they would need to know about dispatchAll, but I would gladly write those all in a single pr.

robertleeplummerjr:
ur, uh, what I mean is I'd expose internally not for all middleware/enhancers all written.  They don't know about them, so simply won't use them.

Aweary:
OK, but the enhancers would still need to be updated to support this right? It may make more sense internally but the main issue that I've been reiterating is that this API change doesn't work with current enhancers. That's my main issue with this API that I'm hoping we can consider and find a way around.

robertleeplummerjr:
@Aweary good points!  The enhancers would need to updated to take advantage of a new feature, yes, but it won't break them, because the `dispatch` function would still be usable and.  Future enhancers could use `dispatchAll`.

> "That's my main issue with this API that I'm hoping we can consider and find a way around."
Everything using `dispatch(action)` will continue just as is.
A new function `dispatchAll(action1, thunk1, thunk2, action2)` would be available for any library to take advantage of.

I believe this is one possible way around that doesn't harm anything, yet resolves the issue/feature entirely.

markerikson:
Earlier in this thread, I linked to a version of `redux-batched-subscribe` that I'd modified to support accepting a single array of actions.  I'm busy at the moment, but I'd interested to know if that modified version works, and reasonably solves this use case.  Any chance someone could try it out?

robertleeplummerjr:
@markerikson So you are dismissing my idea with no input?

robertleeplummerjr:
@markerikson The other thing you could do is simply write a unit test, as I have done, that proves your code works, rather than have someone else use their time to unit test your code.

robertleeplummerjr:
@thinkloop would care to offer some input?

markerikson:
Uh... I'm busy with work at the moment.  I've spent a bunch of time answering questions in this thread over the last couple days, in my free time.  I've dug through Redux's code and the code of a half-dozen addon libs to try to explain how things actually work.  I've pretty clearly expressed the pros and cons of various ways that the idea of multi-dispatch/single-notification might be implemented.  I even took the time to proof-of-concept how a potential single-notification enhancer that needs no no changes to the core code _might_ work.  

I'm certainly not in favor of the current set of changes, but I've absolutely responded to them in depth.  I'll also point out that your one unit test merely proves that a modification to the core `dispatch` implementation works, but without any consideration for the larger ecosystem as a whole.

I'm not done with this thread completely, but I definitely have better things to do at the moment.  Like actual work.

gaearon:
>I would say if we were to reach a consensus here, that multiple arguments could be passed in, a pr must be opened to handle redux-thunk, redux-saga, and redux-observable's. Please note as well, that this pr isn't about dispatching arrays, but rather dispatching multiple arguments.

This pushes very complex semantics onto companion packages, and increases surface area for bugs in these packages. Doesn’t seem like a good solution.

>I will say, though, that my initial thought was to handle arrays, but I found this is so much simpler:
dispatch(arg1) or dispatch(arg1, arg2). The reason: it doesn't change the existing architecture so that it needs to handle different types, but rather allows you to handle more of the same types.

Can you show a complete proposal including how this works with `redux-thunk`, one way or another?

thinkloop:
> Any chance someone could try it out?

I've been looking forward to trying it out, haven't had a chance yet myself, likely tonight. Would you expect it to work with thunks and stuff:

`dispatch([action1, thunk1, action2]`

> This pushes very complex semantics onto companion packages, and increases surface area for bugs in these packages.

That's true, but saves similar complexity from userland - I'd argue that professional package maintainers are better suited to handle said complexity than devs.

markerikson:
I'll point out that we still have two different proposals being tossed around in here.  The actual PR from @robertleeplummerjr tries to accept `dispatch(action1, action2, action3)`, and @thinkloop is suggesting `dispatch([action1, action2, action3])`.  They may be conceptually the same idea, but semantically and implementation-wise are very different.

Also, the point of the whole core-vs-userland thing is that _all_ "companion packages" _are_ "userland" - enhancers, middleware, etc.  In other words, it's not a question of whether a typical app dev is having to make changes, it's a question of whether all extension/addon authors are having to make changes.

markerikson:
And yes - again, the gist I threw together was simply copying the implementation of `redux-batched-subscribe` and swapping out the guts of its own version of `dispatch`, but I would _expect_ that if you set up your store with `compose(myNewBatchedEnhancer, applyMiddleware(thunk, somethingElse))`, it _should_ accept `dispatch([action1, action2])` and return an array itself.

thinkloop:
> A new function dispatchAll(action1, thunk1, thunk2, action2) would be available for any library to take advantage of.

@robertleeplummerjr I think this is still the same general problem. What happens if a user uses dispatchAll but thunk hasn't implemented it - it will just do weird different stuff than regular dispatch. In other words it won''t be just more dispatch, as we want, it will be different dispatch. Might as well just version and deprecate.

thinkloop:
> @thinkloop is suggesting dispatch([action1, action2, action3])

I actually prefer @robertleeplummerjr's but in case there are insurmountable issues with it, threw that out there for consideration.

thinkloop:
> Also, the point of the whole core-vs-userland thing is that all "companion packages" are "userland" - enhancers, middleware, etc. In other words, it's not a question of whether a typical app dev is having to make changes, it's a question of whether all extension/addon authors are having to make changes.

This is a bit of semantics, there's a clear difference between redux, middlewares, and the app - especially in practice, middlewares are generally re-used, shared, focused and maintained by 3rd parties.

gaearon:
Anything that requires middleware authors to add support won’t work for us.
The new proposed API should work out of the box with most of the existing middleware.

thinkloop:
OK, I think there are still ways to make this work. The first problem to solve is that listeners are triggered at the end of the dispatch function: https://github.com/reactjs/redux/blob/master/src/createStore.js#L175-L178

To begin the thought process, would it break stuff if a 'skipListeners' param were added to `dispatch`?

`dispatch(action, skipListeners)`

Aweary:
It's the same issue, enhancers would have to be aware of the new `skipListeners` argument, so we'd still have the same problem. I think any solution that requires a change to the signature of `dispatch` is going to cause the same issue.

thinkloop:
It doesn't look like thunk would break, any example of one that would break?

gaearon:
If we were to add an argument to the dispatch function, usage like `<div onClick={dispatch.bind(store, myAction)}` would break: the previously ignored `event` argument suddenly gains a meaning (and significantly changes the behavior).

robertleeplummerjr:
We could expose a function that alters how dispatch works for one iteration that reverts afterwards.  This way we would always know that what comes next will be a multi dispatch that will revert to a single dispatch after or not be a multi dispatch afterall.

`multi(dispatch(one, two, and, you, can, keep, em, comin, boss))` <- works amazing and only triggers render once
`dispatch(one, two, and, you, can, keep, em, comin, boss)` <- throws exception because multi was not called first

I'm just trying to think outside the box here, feel free to shoot down.

gaearon:
This sounds like a lot of accidental complexity to me. Yes, we can theoretically devise ways to make it work but I’m afraid we’re not going to understand all possible edge cases until we release this, and we might find ourselves unable to fix it, thus making it painful for everyone around. On the other hand, explicit batching on reducer level like `redux-batch-actions` completely solves the problem and doesn’t introduce complexity for anyone.

Don’t want to install that package? [Copy and paste it.](https://github.com/tshelburne/redux-batched-actions/blob/master/src/index.js) It’s literally 16 lines.

markerikson:
Hah.  So.  I just tried testing out the proof-of-concept enhancer I linked earlier.  Borrowed the source for the "counter-vanilla" Redux example, transpiled the enhancer to ES5 for good measure, and put them together.  And other than the one capitalization typo I'd left in there... ***it worked***.  I was able to run `dispatch([{type : "INCREMENT"}, {type : "INCREMENT"}])`, see the number go up by 2, and only have one notification.  I was even able to put the batched dispatch enhancer after `applyMiddleware`, and dispatch an array of actions from a thunk.  I've updated the gist appropriately: https://gist.github.com/markerikson/0383775096f2b32e48ad41499cf8a1fb

So.  Based on all that, it appears that a simple enhancer solves this use case sufficiently. It requires no change to the core API, it's entirely opt-in, and it works.  I don't see any reason to make any changes to the core `dispatch` itself at this time.

gaearon:
After #1702 this enhancer should get considerably less hairy.

robertleeplummerjr:
Good job @markerikson! Shall I close this then?

markerikson:
Yeah.  I don't see any changes being made to the core at this time.  It _may_ be worth revisiting after #1702 is done just to see where things stand, but for now it seems like that enhancer solves the use case easily enough.

robertleeplummerjr:
Thanks everyone who got involved!  This was fun.  I hope to have more fun in the future.

arcanis:
I'm just sharing my experience: in a decent-sized project, we currently have a redux-saga that handle the resource management logic. Basically, when it "sees" a RESOURCE_UPDATE action, it will first push the update resource into the store (optimistically) via dispatching a RESOURCE_SET action, send the data, then push the updated resource again - either to apply what the server answered, or to rollback if an error happened.

In some cases, a single RESOURCE_UPDATE action actually needs to update multiple resources. A prime example is when adding resources to a to-many relationship: we need to update both ends of the relationship.

In such a case, we need to dispatch multiple RESOURCE_SET actions, one for each resource that needs to be updated. We started doing this by simply dispatching multiple times, but issues arised when react-redux would trigger a rerender in-between those RESOURCE_SET actions. In such a case, the store was still in an incomplete state: some data would say something that wasn't baked by the others. We eventually implemented a simple BATCH_ACTION in our primary reducer so that we could dispatch multiple RESOURCE_SET actions at once, and everything just worked. Our tree stopped being rerendered when the state wasn't yet ready to be read.

My concern is that we fixed this behaviour via an application-specific piece of code. Redux enhancers have no idea that BATCH_ACTION is a batched action, and will treat it like any other one. I could use a third-party packages to add this action, but the result would be the same: most enhancers would have no idea what to do with the third-party action, unless they explicitely write code to support such an action. But at this point, why not add it to the core, since they will have to support it anyway? I could also use @markerikson's enhancer, or a similar one, but then what if I want to extract this resource management code into its own library (which is what I'm doing)? Should I tell users that they need to manually add this enhancer for everything to work?

On the plus side, this issue has a few easy workarounds, so it's not the most urgent one to solve. But I still think it should be seriously considered for inclusion in the next major release. The more soon you publish an Intent-to-implement, the more time the third-party enhancers have to adapt.

markerikson:
@arcanis : Thanks for your comment.  Some thoughts:

First, creating a "batched reducer" that wraps up other reducers is an entirely valid approach.  In fact, it very much mirrors the common approach seen in Elm.  Remember that "reducer composition" doesn't just mean use of `combineReducers`.  It can be things like:

```js
const combinedReducer = combineReducers(bunchOfReducers);
const batchedReducer = createBatchedReducer(combinedReducer);
const hydratableReducer = createHydratableReducer(batchedReducer);

const store = createStore(hydratableReducer);
```

Each layer of wrapping is independent of the others, can be composed together.  So yes, that's a great approach right there, and the fact that your `BATCH_ACTION` is "unknown to other enhancers` is a feature and not a bug.

Another approach would be to modify your `RESOURCE_SET` handler itself to accept multiple possible values in one action (conceptually different from `BATCH_ACTION` in that you're not re-running `RESOURCE_SET` multiple times, but rather that `RESOURCE_SET` can apply to multiple things at once).  

A third approach, particularly if you're dealing with managing relational entities, would be to use a library like Redux-ORM, and write a case reducer that handles that specific scenario rather than a generic `RESOURCE_SET` event.  You can use Redux-ORM to look up multiple entities, queue updates, and apply them in a single step.

Beyond that, there's the variety of possible batching strategies already discussed in this thread, all of which have their own tradeoffs.

So yes, my take is that this is _still_ entirely solveable in userland, the proper approach is still specific to each user's use cases, and there is still no need to add this to the Redux core.

markerikson:
@arcanis : As one more reference point, Dan demonstrated exactly that `BATCH_ACTION` approach in a tweet last year:

https://twitter.com/dan_abramov/status/656074974533459968

arcanis:
@markerikson Yep, that's what we're currently using  :)

I guess my point is that while it is possible to implement this feature userland, any such implementation will eventually fall into one of those two buckets:

- Either it won't work with third-party middlewares (unless they add a specific support for a common "batch action", at which point the arguments in favor of not breaking the enhancers become moot)

- Or it will require a particular setup from the user, requiring them to add a specific enhancer before any other one. So more boilerplate code, with the risk that it needs to be done multiple times if multiple third-parties use multiple different implementations.

As for expanding actions in order to take more responsibilities (such as my `RESOURCE_SET` that could take an array of resources), while it would work fine at the beginning, it would start to come short once we reach the point were we need to do different things all at once. To keep on the resource example, in some case (deleting a resource and updating its parent relationship) we need do to both a `RESOURCE_SET` on the resource A and a `RESOURCE_UNSET` on a resource B. Supporting this use case as a single action would requiring duplicating code, which I would try to avoid.

Issue #1792, opened 2016-06-06
RFC: Make combineReducers() shape-agnostic


gaearon:
I propose we should make `combineReducers()` work with ImmutableJS and any other libraries. This way all the checks it makes would work out of the box, but the actual state transformation for associating the values would be swappable.

I’m thinking something like `combineReducers(reducers, [options])`.
By default `options` is something like

``` js
{
  create: (obj) => obj,
  get: (state, key) => state[key]
}
```

but you could supply something like

``` js
{
  create: (obj) => (obj instanceof Immutable.Map) ? obj : new Immutable.Map(obj),
  get: (state, key) => state.get(key)
}
```

Does this make sense to anyone?
If you’d like to work on a PR to make a proof of concept, please comment here and submit it!

If you think it’s a bad idea, please write here as well.

bvaughn:
I know it's just an example listed, but I'd suggest changing to use `fromJS` (since it does a deep conversion). :)
```js
{
  create: (obj) => new Immutable.fromJS(obj),
  get: (state, key) => state.get(key)
}
```

markerikson:
Ooo.  I ***like*** this idea.  :+1: :100:

gaearon:
Deep conversion is not necessarily what you want (you might have nested plain object reducers). Also it would be much slower in the common case.

bvaughn:
Fair enough. In my experience, shallow conversion commonly leads to errors (eg methods like `getIn` and `setIn` don't work as expected, you can accidentally mutate nodes in your immutable tree, etc). But I'm sure you've thought about this particular use case much more than I have. :)

markerikson:
To expand on my comment: this seems useful, simplifies interop, and is consistent with the "sensible default with optional customization" approach used by `mapState` and `mapDispatch`.  Seems like all-upsides on this one.

gaearon:
@bvaughn 

Since it's pluggable you would be able to customize it. But it would be super confusing if applying it at the root level broke third-party reducers from another packages, for example.

Usually combineReducers() has no effect on the state of the tree below. It just manages one level deep. It would be strange if suddenly it took over the state returned by nested reducers and transformed it. So while the approach you suggest is possible, it isn't something we'd encourage.

markerikson:
Quick question: How well does this tie in with the other open PRs to `combineReducers`, #1768 and #1789 ?

kennetpostigo:
Don't use fromJS it's really expensive operation in immutable lib.

[...]

PR #1817, opened 2016-06-17
#1792 candidate: Make combineReducers shape-agnostic

borkxs:
Candidate for #1792. Adds options parameter to `combineReducers` which can include `get`, `create`, and `keys` methods. There is no `set` method because we can just use `create` instead.

Adds a few new warnings and errors relevant to these methods without altering the old warnings and errors. The `isPlainObject` warning now just checks for `typeof state === 'object'`, so it still warns when something like a `Number` is passed in.

The keys methods is for iterating over the state keys, which is needed to check for unexpected keys. It can be excluded and will simply warn if the result of the default `Object.keys(...)` doesn't match the reducer shape.

Also added a param to `getUnexpectedStateShapeWarningMessage` to pass in the already computed `reducerKeys` instead of recalculating them from the reducer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,7 @@
     "expect": "^1.8.0",
     "gitbook-cli": "^0.3.4",
     "glob": "^6.0.4",
+    "immutable": "3.8.1",
     "isparta": "^4.0.0",
     "mocha": "^2.2.5",
     "rimraf": "^2.3.4",
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -1,5 +1,4 @@
 import { ActionTypes } from './createStore'
-import isPlainObject from 'lodash/isPlainObject'
 import warning from './utils/warning'
 
 function getUndefinedStateErrorMessage(key, action) {
@@ -12,8 +11,7 @@ function getUndefinedStateErrorMessage(key, action) {
   )
 }
 
-function getUnexpectedStateShapeWarningMessage(inputState, reducers, action) {
-  var reducerKeys = Object.keys(reducers)
+function getUnexpectedStateShapeWarningMessage(inputState, keysMethod, reducers, reducerKeys, action) {
   var argumentName = action && action.type === ActionTypes.INIT ?
     'preloadedState argument passed to createStore' :
     'previous state received by the reducer'
@@ -25,7 +23,7 @@ function getUnexpectedStateShapeWarningMessage(inputState, reducers, action) {
     )
   }
 
-  if (!isPlainObject(inputState)) {
+  if (typeof inputState !== 'object') {
     return (
       `The ${argumentName} has unexpected type of "` +
       ({}).toString.call(inputState).match(/\s([a-z|A-Z]+)/)[1] +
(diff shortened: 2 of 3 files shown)


borkxs:
It can't find the immutable module I added to devDependencies, I'm guessing node_modules are cached?

timdorr:
I'm :-1: on this. combineReducers is a convenience function, not a requirement for Redux to operate. If you want to use another library to store the results of your reducers, I would build bindings for it and implement your own version of this function. This adds too many abstractions.

timdorr:
Closing due to lack of input. I'll reopen if some support for this can be drummed up (of course, I'm still against it, sorry :( )

markerikson:
@borkxs , @timdorr , @jimbolla , @gaearon : I'm seriously interested in revisiting this idea.  Tim, I see you closed this a year ago as a "eh, we don't want this" decision.  On the other hand, in November you said at https://github.com/reactjs/redux/pull/1768#issuecomment-257911890 : 

> I'm a fan of composition vs. expanding an API. If we could break up combineReducers and let users put it back together however they want (along with maintaining the existing API), that would be the most scalable option.

Is the approach of defining `get/set/create` functions sufficient for "breaking things up" in your view, or did you have something else in mind?

I personally am in favor of adding this, and also adding options to make the key warnings and dropping of unhandled slices configurable (related to #2427 ).

timdorr:
I didn't have any specific API or structure in mind, just the overall goal of a composable combineReducers. I'm happy to examine it, but we should have scalability in mind. If it can solve the issues of #1768 at the same time, that's even better.

eddyw:
I had an idea some time ago how to implement some kind of 'utils' in the reducers, so they would work either with ImmutableJS or without it.
I wrote something similar but not exactly to this:
```javascript
const redux = require('redux')
const ini = {
	count: 0,
	nested: {
		sample: 1
	}
}
const countReducer = (state = ini, action) => {
	const u = action.utils
	switch(action.type) {
		case 'increment':
			console.log('Current state tree:', u.getState())
			return u.merge(state, {
				count: u.get(state, 'count') + action.payload,
				nested: u.merge(u.get(state, 'nested'), {
					newPropInNested: 2
				})
			})
		default:
			return u.object(state)
	}
}

const addUtils = api => {
	const utils = {
		get: (obj, prop) => obj[prop],
		object: obj => obj, // Use: obj => Immutable.Map(obj) for ImmutableJS, for instance
		merge: (...objs) => Object.assign({}, ...objs),
		getState: () => api.getState() // Additionally allow access to entire state
		// other helpers or utils
	}
	return dispatch => action => {
		const a = Object.assign({}, action)
		a.utils = utils
		dispatch(a)
	}
}
const store = redux.createStore(countReducer, redux.applyMiddleware(addUtils))
store.subscribe(() => {
	console.log('State tree changed:', store0.getState())
})
store.dispatch({ type: 'increment', payload: 10 })
```
However, it relays in the fact that I will never send a 'utils' prop in the action object to the dispatch method.

My proposal would be to make the middlewareApi available as third parameter in the reducers, so it can be extended instead of changing anything in the combineReducers fn:
```javascript
const countReducer = (state = null, action, middlewareApi) => { // middlewareApi.utils = {...}
	switch(action.type) {
		default:
			return state
	}
}
const middleware = middlewareApi => {
	middlewareApi.utils = {
		get: (obj, prop) => obj[prop],
		merge: (...objs) => Object.assign({}, ...objs)
	}
	return dispatch => action => dispatch(action)
}
const store = redux.createStore(countReducer, redux.applyMiddleware(middleware))```
This way, it even exposes `getState()` method to the reducer, so it can access the state tree (for any reason somebody would like to do that)

**Comment by SeeThruHead** [comment]
would love to see this
```
reduceReducers(
  combineReducers({
    meta,
    contacts,
    content
  }),
  combineReducers({
      blah,
      foo,
      baz
  })
);
```
i've completely defined my state tree, but i have no state, this is really odd to me

PR #1818, opened 2016-06-17
Only warn for unexpected key once per key

### LINKED ISSUE #1748: combineReducers does not remove unexpected keys when reducers have no changes
For example in the following, I would expect `newState` to equal `{ id: 1234, name: 'hello' }` but it instead returns `currentState`.
```javascript
const currentState = { id: 1234, name: 'hello', unexpected: 4567 };
const reducer = combineReducers( { id, name } );
const newState = reducer( currentState, {} );
```

The additional key is harmless in code, but it is annoying since an error warning may be printed to console repeatedly. This case is pretty easy to run into if the redux tree is persisted and the data shape changes.

Aweary:
Resolves #1748 

Pretty simple, just keeps a caches and filters keys that have already been warned about.

I had to update one of the related tests as it expected some keys to be warned about multiple times.

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -12,7 +12,7 @@ function getUndefinedStateErrorMessage(key, action) {
   )
 }
 
-function getUnexpectedStateShapeWarningMessage(inputState, reducers, action) {
+function getUnexpectedStateShapeWarningMessage(inputState, reducers, action, unexpectedKeyCache) {
   var reducerKeys = Object.keys(reducers)
   var argumentName = action && action.type === ActionTypes.INIT ?
     'preloadedState argument passed to createStore' :
@@ -34,7 +34,14 @@ function getUnexpectedStateShapeWarningMessage(inputState, reducers, action) {
     )
   }
 
-  var unexpectedKeys = Object.keys(inputState).filter(key => !reducers.hasOwnProperty(key))
+  var unexpectedKeys = Object.keys(inputState).filter(key =>
+    !reducers.hasOwnProperty(key) &&
+    !unexpectedKeyCache[key]
+  )
+
+  unexpectedKeys.forEach(key => {
+    unexpectedKeyCache[key] = true
+  })
 
   if (unexpectedKeys.length > 0) {
     return (
@@ -101,6 +108,10 @@ export default function combineReducers(reducers) {
   }
   var finalReducerKeys = Object.keys(finalReducers)
 
+  if (process.env.NODE_ENV !== 'production') {
+    var unexpectedKeyCache = {}
+  }
+
   var sanityError
   try {
     assertReducerSanity(finalReducers)
@@ -114,7 +125,7 @@ export default function combineReducers(reducers) {
     }
 
     if (process.env.NODE_ENV !== 'production') {
-      var warningMessage = getUnexpectedStateShapeWarningMessage(state, finalReducers, action)
+      var warningMessage = getUnexpectedStateShapeWarningMessage(state, finalReducers, action, unexpectedKeyCache)
       if (warningMessage) {
         warning(warningMessage)
       }
--- a/test/combineReducers.spec.js
+++ b/test/combineReducers.spec.js
@@ -210,9 +210,9 @@ describe('Utils', () => {
         /Unexpected key "bar".*createStore.*instead: "foo", "baz"/
       )
 
-      createStore(reducer, { bar: 2, qux: 4 })
+      createStore(reducer, { bar: 2, qux: 4, thud: 5 })
       expect(spy.calls[1].arguments[0]).toMatch(
-        /Unexpected keys "bar", "qux".*createStore.*instead: "foo", "baz"/
+        /Unexpected keys "qux", "thud".*createStore.*instead: "foo", "baz"/
       )
 
       createStore(reducer, 1)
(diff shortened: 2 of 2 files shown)


src/combineReducers.js
```diff
@@ -34,7 +36,9 @@ function getUnexpectedStateShapeWarningMessage(inputState, reducers, action) {
     )
   }
 
-  var unexpectedKeys = Object.keys(inputState).filter(key => !reducers.hasOwnProperty(key))
+  var unexpectedKeys = Object.keys(inputState).filter(key => !reducers.hasOwnProperty(key) && !unexpectedKeyCache[key])
```
gaearon: Let’s split this into several lines so they format more nicely?

src/combineReducers.js
```diff
@@ -34,7 +36,9 @@ function getUnexpectedStateShapeWarningMessage(inputState, reducers, action) {
...
+
+  unexpectedKeys.forEach(key => unexpectedKeyCache[key] = true)
```
gaearon: Let’s use curly braces to make it clear it’s a side effect.

gaearon:
Looks good to me after fixing the nits.

Aweary:
@gaearon the nits have been fixed 👍

src/combineReducers.js
```diff
@@ -2,6 +2,8 @@ import { ActionTypes } from './createStore'
 import isPlainObject from 'lodash/isPlainObject'
 import warning from './utils/warning'
 
+const unexpectedKeyCache = {}
```
gaearon: Let’s use `var` for consistency (we’ll codemod ’em all over later)

src/combineReducers.js
```diff
@@ -34,7 +36,12 @@ function getUnexpectedStateShapeWarningMessage(inputState, reducers, action) {
     )
   }
 
-  var unexpectedKeys = Object.keys(inputState).filter(key => !reducers.hasOwnProperty(key))
+  var unexpectedKeys = Object.keys(inputState).filter(key =>
+  !reducers.hasOwnProperty(key) && !unexpectedKeyCache[key])
```
gaearon: I meant something like

```js
var unexpectedKeys = Object.keys(inputState).filter(key =>
  !reducers.hasOwnProperty(key) &&
  !unexpectedKeyCache[key]
)
```

Aweary:
@gaearon changes made 👍

test/combineReducers.spec.js
```diff
@@ -237,5 +237,26 @@ describe('Utils', () => {
 
       spy.restore()
     })
+
+    it('only warns for unexpected keys once', () => {
```
gaearon: Do I undestand correctly that tests now depend on global shared state and run order? This doesn’t look very good.

src/combineReducers.js
```diff
@@ -2,6 +2,8 @@ import { ActionTypes } from './createStore'
 import isPlainObject from 'lodash/isPlainObject'
 import warning from './utils/warning'
 
+var unexpectedKeyCache = {}
```
gaearon: Can/should we keep a cache per `combineReducers()` call?

Aweary: Yes, we definitely should, good call.

test/combineReducers.spec.js
Aweary: Right, that should be resolved by using individual cache instances per `combineReducers` call

src/combineReducers.js
Aweary: I guess we could just create the cache within `combineReducers` and pass that to `getUnexpectedStateShapeWarningMessage`. Does that work for you @gaearon?

gaearon: Yea

test/combineReducers.spec.js
gaearon: Yep

Aweary:
@gaearon I updated the PR so that it creates and passes a new cache to `getUnexpectedStateShapeWarningMessage` whenever `combineReducers` is called. I didn't see any existing code that indicated how you'd want to format these functions now that the argument list is rather long, so let me know if any changes are required 👍

src/combineReducers.js
```diff
@@ -101,6 +99,7 @@ function assertReducerSanity(reducers) {
  */
 export default function combineReducers(reducers) {
   var reducerKeys = Object.keys(reducers)
+  var unexpectedKeyCache = {}
```
gaearon: Probably no need to declare it unless we’re in DEV.

src/combineReducers.js
```diff
@@ -137,7 +136,7 @@ export default function combineReducers(reducers) {
       var previousStateForKey = state[key]
       var nextStateForKey = reducer(previousStateForKey, action)
       if (typeof nextStateForKey === 'undefined') {
-        var errorMessage = getUndefinedStateErrorMessage(key, action)
+        var errorMessage = getUndefinedStateErrorMessage(key, action, unexpectedKeyCache)
```
gaearon: I don’t think `getUndefinedStateErrorMessage` uses this?

Aweary: Sorry that was accidental

Aweary:
@gaearon updated again. Fourth try's the charm.

gaearon:
Have you tested this on a real project to check that it behaves as expected?

Aweary:
I don't typically use the master branch of a library in real projects, I can run it on some of the example projects and verify the results if you like.

Aweary:
@gaearon 

I tested the real-world example branch and verified it's working as expected. I changed the `rootReducer` call so it passed an invalid `foobar` value to the `pagination` reducer

```js
  entities,
  pagination: (state, action) => {
    return pagination(
      Object.assign({}, state, {foobar: false}),
      action
    )
  },
  errorMessage,
  routing
})
```

With the current release of redux this caused multiple warnings. After running `npm run build` and replacing the `lib` folder in the example project's `node_modules/redux` folder the error was only logged once no matter what actions were performed.

gaearon:
👍

PR #1829, opened 2016-06-26
Avoiding confusion with "mutating state"

atufkas:
Hi Dan and the whole redux Dev Team,

I'd kindly ask you to review this little documentation change affecting the introduction/Three Principles chapter.

As proposed in the commit message I'd suggest to use the word 'change' instead of 'mutate' to avoid confusion with the concept of 'immutable state'. As we all know: While not enforced, making state objects immutable is a strongly recommended way of using redux (and only then redux plays well together with pure reducers, time traveling and so on).

I found myself felt a bit confused when reading the introduction for the first time. Furthermore I remember to have seen similar comments elsewhere here.

Maybe this little change helps people not getting on the wrong path in the very beginning.

Cheers and thanks for checking this,
Matthias

### DIFF
--- a/docs/introduction/ThreePrinciples.md
+++ b/docs/introduction/ThreePrinciples.md
@@ -30,9 +30,9 @@ console.log(store.getState())
 
 ### State is read-only
 
-**The only way to mutate the state is to emit an [action](../Glossary.md#action), an object describing what happened.**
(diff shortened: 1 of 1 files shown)


markerikson:
Seems pretty reasonable.  Thanks!

PR #1834, opened 2016-06-28
Add Symbol.observable interop point to mock store provided to middleware

### LINKED ISSUE #1833: store provided to middleware doesn't implement Symbol.observable
https://github.com/reactjs/redux/blob/master/src/applyMiddleware.js#L25-L29

Because it isn't a _real_ store, this can cause issues with middleware that need some of the real store's features. Particularly the one we care about right now is the [`Symbol.observable` interop](https://github.com/reactjs/redux/blob/master/src/createStore.js#L202-L238). We need the ability to receive a stream of state changes in [redux-observable](https://github.com/redux-observable/redux-observable/issues/56)

I imagine the reason for providing a mock store is to not bleed abstractions like `replaceReducer` right? Any objections to including the `Symbol.observable` interop in that mock store too? (or just not mocking it)

jayphelps:
fixes #1833

### DIFF
--- a/src/applyMiddleware.js
+++ b/src/applyMiddleware.js
@@ -1,4 +1,5 @@
 import compose from './compose'
+import $$observable from 'symbol-observable'
 
 /**
  * Creates a store enhancer that applies middleware to the dispatch method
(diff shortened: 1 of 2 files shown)


jayphelps:
Use store enhancer instead, redux core does not wish to expose this ability at the moment ##1833

PR #1835, opened 2016-06-28
Directs to babel-plugin-transform-object-assign instead

siakaramalegos:
babel-plugin-object-assign is deprecated


ellbee:
Yep, good catch, looks like you are right.

Thanks!

PR #1842, opened 2016-07-08
Mention redux-saga

paldepind:
This PR is documentation changes only. It changes two small things:

* ~~It removes the words "for most common needs" because `fetch` replaces `XMLHttpRequest` in its entirety. It is a common misunderstanding that `fetch` only replaces some high level aspects of `XHR`. But actually `fetch` is more low level than `XMLHttpRequest` and `XHR` can be defined entirely in terms of `fetch`—while the reverse is not true. For more info see: https://jakearchibald.com/2015/thats-so-fetch/~~
* Mention @yelouafi's redux-saga because it is definitely worth doing in the section about alternative side effect middleware.


chicoxyzzy:
> `XMLHttpRequest` and `XHR` can be defined entirely in terms of `fetch`

That's not true because `fetch` can't be canceled while `XHR` can

sompylasar:
@chicoxyzzy :+1: although they think about adding it: https://github.com/whatwg/fetch/issues/27

markerikson:
Mentioning redux-saga is worthwhile, but the new text has some typos, and the description of redux-saga isn't accurate.  Returning side effect descriptions from reducers is what redux-loop does.

paldepind:
> That's not true because fetch can't be canceled while XHR can

You are indeed correct. It turns out I was the one with a misunderstanding. Thank you.

> Mentioning redux-saga is worthwhile, but the new text has some typos, and the description of redux-saga isn't accurate. Returning side effect descriptions from reducers is what redux-loop does.

Yes. I've modified the text.

mxstbr:
I'm not sure

> that preserves actions as plain object

really describes the benefits of `redux-saga` :wink:

docs/advanced/AsyncActions.md
```diff
@@ -490,7 +490,7 @@ store.dispatch(fetchPostsIfNeeded('reactjs')).then(() =>
...
+[Thunk middleware](https://github.com/gaearon/redux-thunk) isn’t the only way to orchestrate asynchronous actions in Redux. You can use [redux-promise](https://github.com/acdlite/redux-promise) or [redux-promise-middleware](https://github.com/pburtchaell/redux-promise-middleware) to dispatch Promises instead off unctions. You can use the side effects middleware [redux-saga](https://github.com/yelouafi/redux-saga/) that preserves actions as plain objects. You can dispatch Observables with [redux-rx](https://github.com/acdlite/redux-rx). You can even write a custom middleware to describe calls to your API, like the [real world example](../introduction/Examples.md#real-world) does. It is up to you to try a few options, choose a convention you like, and follow it, whether with, or without the middleware.
```
sompylasar: `instead off unctions.` -> `instead of functions.`

paldepind:
It certainly dosen't—only a small part of its advantages. I wrote it because it is one aspect that should be readily understandable to people who have only read up to "Async Actions" in the Redux documentation. Explaining more about redux-saga, I think, would end requiring many more words. But ideas are welcome :smile:

docs/advanced/AsyncActions.md
paldepind: Thank you. Fixed.

sompylasar:
> I'm not sure
> 
> > that preserves actions as plain object
> 
> really describes the benefits of redux-saga 😉

It describes `redux-saga` compared to other middlewares in this list. Having actions as plain objects is important for reproducability which Redux architecture provides.

markerikson:
Personally, I've described sagas as "stateful background threads for Javascript".  You could also say that they enable "advanced asynchronous flow control" or something.

timdorr:
I'm going to fix this after merging to be less descriptive, since these are are a set of small sentences anyways. Just "You can use the redux-saga middleware to handle more complex asynchronous actions."

docs/advanced/AsyncActions.md
```diff
@@ -490,7 +490,7 @@ store.dispatch(fetchPostsIfNeeded('reactjs')).then(() =>
...
+[Thunk middleware](https://github.com/gaearon/redux-thunk) isn’t the only way to orchestrate asynchronous actions in Redux. You can use [redux-promise](https://github.com/acdlite/redux-promise) or [redux-promise-middleware](https://github.com/pburtchaell/redux-promise-middleware) to dispatch Promises instead of functions. You can use the side effects middleware [redux-saga](https://github.com/yelouafi/redux-saga/) that preserves actions as plain objects. You can dispatch Observables with [redux-rx](https://github.com/acdlite/redux-rx). You can even write a custom middleware to describe calls to your API, like the [real world example](../introduction/Examples.md#real-world) does. It is up to you to try a few options, choose a convention you like, and follow it, whether with, or without the middleware.
```
gaearon: We should also recommend redux-observable instead of redux-rx now.

PR #1852, opened 2016-07-18
Update ExampleTodoList.md

janhenrikoverland:
Calling todo(undefined, action) does not work without the default state value.


timdorr:
Thanks!

PR #1853, opened 2016-07-19
Correct Link For Redux Website ;)

lourenco:


markerikson:
The original link actually _is_ correct.  Since there's a PHP framework also named "Redux", that link is intended to redirect people who are actually looking for the PHP framework instead of the JS library.

Issue #1855, opened 2016-07-19
Universal example has xss vulnerability


brigand:
Say the initial state is `{"foo":"some string"}`, you're fine, but if it's is `{"foo":"</script><script>alert('xss')</script><script>"}` then you're screwed.

Not an issue if you're sure it's an integer, but I've seen people use this exact pattern in other places where it could cause serious issues, so best to get it right in a common point of reference.

A simple solution is to escape the `<`s, but not sure if there's a better way to do it.

```js
window.__PRELOADED_STATE__ = ${JSON.stringify(preloadedState).replace(/</g, '\\x3c')}
```

steida:
Use serialize-javascript. Example: https://github.com/este/este/blob/443b78e894d0aa5c666b3df8a60b6bfd2397c7d3/src/server/frontend/render.js#L48

PR #1856, opened 2016-07-19
fixes xss vulnerability in universal example, #1855

brigand:
See https://github.com/reactjs/redux/issues/1855 for details.


timdorr:
I'm not sure this is actually sufficient. It certainly removes a whole class of attacks, but there are further things that can be done. I would instead use a combo of `encodeURI` and `decodeURI` or something like [`serialize-javascript`](https://github.com/yahoo/serialize-javascript).

timdorr:
I guess for now, just for the people copypastaing this code, this works well enough.

PR #1868, opened 2016-07-25
Improve typings for `compose` function

aikoven:
Introduce more strict typings for `compose` function to allow smarter inference, e.g.:

```ts
import {compose} from "redux";
import {connect} from "react-redux";
import {withRouter} from "react-router";

compose(
  withRouter,
  connect(
    (state, ownProps) => {
      return {
        foo: 'bar',
      };
    }
  )
)(props => {
  // `props` type is inferred to `{foo: string, router: Router}`
});
```

cc @ulfryk, @Igorbek

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -357,6 +357,11 @@ export function bindActionCreators<
 
 /* compose */
 
+type Func0<R> = () => R;
+type Func1<T1, R> = (a1: T1) => R;
+type Func2<T1, T2, R> = (a1: T1, a2: T2) => R;
+type Func3<T1, T2, T3, R> = (a1: T1, a2: T2, a3: T3, ...args: any[]) => R;
+
 /**
  * Composes single-argument functions from right to left. The rightmost
  * function can take multiple arguments as it provides the signature for the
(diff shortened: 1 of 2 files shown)


ulfryk:
@aikoven - this PR is a proof that React and Redux where designed for dynamically typed world…

Anyway it looks GOOD 👍

aikoven:
@ulfryk Things will get better with support for [variadic kinds](https://github.com/Microsoft/TypeScript/issues/5453) in TS :)

index.d.ts
```diff
@@ -367,27 +372,54 @@ export function bindActionCreators<
  *   to left. For example, `compose(f, g, h)` is identical to doing
  *   `(...args) => f(g(h(...args)))`.
  */
-export function compose(): <R>(a: R, ...args: any[]) => R;
+export function compose(): <R>(a: R) => R;
```
Igorbek: This overload will hide the next one. Is it allowed for compose?

Igorbek:
Thank you for adding me as a reviewer. Besides of minor concern (see comment), it looks good to me.

index.d.ts
Igorbek: Ah! My bad, didn't notice it has no parameters.

timdorr:
> @ulfryk Things will get better with support for variadic kinds in TS :)

Will this have any backwards-compatibility concerns?

Sounds like everyone who knows TS is in agreement on this. In she goes! 

(side note: I *really* need to learn TS...)

aikoven:
@timdorr 
> > @ulfryk Things will get better with support for variadic kinds in TS :)

> Will this have any backwards-compatibility concerns?

I guess so, older versions of TS compiler won't be able to consume new typings. However, as far as I can see, new versions are adopted very fast by the community.

If the problem get serious (e.g. there would be a lot of projects not being able to migrate quickly to the new TS compiler), there are some workarounds. For example, we can have a separate repo for older version typings. It's very easy to substitute official typings coming with NPM package by some external ones.

PR #1871, opened 2016-07-26
Add module entry point for webpack 2

taion:
This is what webpack 2 prefers. Ideally Rollup will follow along as well, at which point we can drop `jsnext:main`.


taion:
Here's the corresponding issue on rollup-plugin-node-resolve: rollup/rollup-plugin-node-resolve#48.

timdorr:
👍

PR #1872, opened 2016-07-28
Removed extra comma in package.json example

diegofu:
There's an extra comma in package.json on WritingTests example recipe page that would cause some of the npm commands to fail


markerikson:
Thanks!

PR #1874, opened 2016-07-31
Switch out non-ASCII quotes for ASCII versions.

timdorr:
I was going to just commit this on master, but that would probably bust every PR currently in play that touches the docs 😄 

But the idea here is to use standard ASCII quotes everywhere. They smart quotes or angled quotes are causing some display issues in places (namely on GH Pages).


markerikson:
Yeah, uh... Dan's kinda picky on that one :)  I never managed to properly get them entered or displayed while working on the FAQ page, but Dan really seems to want that bit of typography, and he took the time to edit them in once I published my initial text.

timdorr:
@gaearon Is this an issue? ASCII quotes are more accessible (namely when screenreaders vocalize the HTML character code as literal text instead of an ignored quote) and generally have less problems with tooling.

gaearon:
Meh, whatever. I think ASCII quotes look super ugly but seems like nobody cares about this so I'm fine doing what most people do. :-)

>namely when screenreaders vocalize the HTML character code as literal text instead of an ignored quote

Weird, I didn't known that. Do you mean that every single serious magazine on the web has this problem? I've never seen ASCII quotes on any serious website with text except tech docs but maybe I'm wrong.

timdorr:
> Weird, I didn't known that. Do you mean that every single serious magazine on the web has this problem? I've never seen ASCII quotes on any serious website with text except tech docs but maybe I'm wrong.

It's more of an issue when the tooling gets in the way and converts a quote symbol to an HTML entity or a mangled UTF-8 entity ("donâ€™t").

gaearon:
Do we have tooling doing that?

gaearon:
To be honest, I'm happy to stop caring about this, it's been an uphill battle for as long as I can remember. So I'm ok with this PR.

markerikson:
Heh.  That makes _my_ job a bit easier, but I can understand the desire to make things like kinda nice.  Thanks, Dan :)  And sorry, Dan :(

gaearon:
OK let's do it. 😄

PR #1877, opened 2016-08-02
Fix(dependencies) Update symbol-observable requirement to 1.0.2

btd:

PR #1880, opened 2016-08-04
Update symbol-observable version

asiniy:
Hey!

Unfortunately some kinds of webpack builds causes an exception with outdated `symbol-observable` polyfill. There was a fix for that situation: https://github.com/blesh/symbol-observable/pull/8

Could you update the version for this case, please?

Alex


btd:
Because of the same issue i already created https://github.com/reactjs/redux/pull/1877

asiniy:
@btd I'll close this PR

PR #1883, opened 2016-08-07
Port examples to use Create React App

gaearon:
Work in progress.
This is a good opportunity to dogfood the tool!

Issue #1866, opened 2016-07-24
Example with flow types


sandysaders:
It would be nice to see an example of using flow with redux. I've seen a few articles that start to explain how to use flow with redux and it just seems like it would be useful for people like myself who know how to use redux but not flow and want to see both be used in an example app. [This](http://frantic.im/using-redux-with-flow) article starts to show the benefits really quickly and It would be dope if someone familiar with flow could make add an example app in the examples folder.

jimbolla:
See #1887

timdorr:
Oh, this was merged in. Great! 😄

PR #1887, opened 2016-08-09
add examples with Flowtype support + libdefs for redux and react-redux

gcanti:
**Using Flow for statically analyzing state and actions**

This PR is a proof of concept and is not meant to be merged

Added examples (ordered by difficulty)

- `counter-flow`
- `todos-flow`
- `todomvc-flow`

Added libdefs (in the `ROOT/flow-typed` directory)

- `redux`
- `react-redux`

**Observations**

1) the best type safety is achieved by avoiding the following helpers

- `combineReducers` (you can miss a field and / or assign a wrong reducer to a field)
- `bindActionCreators` (<del>all action creator signatures are erased</del>) <= this is now type safe

`combineReducers` may be replaced by combining the reducers by hand (see the todos-flow example). `bindActionCreators` is unsafe when an object is passed in. <del>However `bindActionCreators` also accepts a single function which is safer (see the todomvc-flow example)</del>

2) `react-redux`'s libdef was pretty hard to write. Should cover the most common cases (at least)

3) In order to please Flow I've done some refactorings but I tried to keep them as small as possible

4) libdefs need community feedback, then we might consider to push them to flow-typed

### DIFF
--- a/flow-typed/react-redux.js
+++ b/flow-typed/react-redux.js
@@ -1,0 +1,73 @@
+import type { Dispatch, Store } from 'redux'
+
+declare module 'react-redux' {
+
+  /*
+
+    S = State
+    A = Action
+    OP = OwnProps
+    SP = StateProps
+    DP = DispatchProps
+
+  */
+
+  declare type MapStateToProps<S, OP: Object, SP: Object> = (state: S, ownProps: OP) => SP | MapStateToProps<S, OP, SP>;
+
+  declare type MapDispatchToProps<A, OP: Object, DP: Object> = ((dispatch: Dispatch<A>, ownProps: OP) => DP) | DP;
+
+  declare type MergeProps<SP, DP: Object, OP: Object, P: Object> = (stateProps: SP, dispatchProps: DP, ownProps: OP) => P;
+
+  declare type StatelessComponent<P> = (props: P) => ?React$Element<any>;
+
+  declare class ConnectedComponent<OP, P, Def, St> extends React$Component<void, OP, void> {
+    static WrappedComponent: Class<React$Component<Def, P, St>>;
+    getWrappedInstance(): React$Component<Def, P, St>;
+    static defaultProps: void;
+    props: OP;
+    state: void;
+  }
+
+  declare type ConnectedComponentClass<OP, P, Def, St> = Class<ConnectedComponent<OP, P, Def, St>>;
+
+  declare type Connector<OP, P> = {
+    (component: StatelessComponent<P>): ConnectedComponentClass<OP, P, void, void>;
+    <Def, St>(component: Class<React$Component<Def, P, St>>): ConnectedComponentClass<OP, P, Def, St>;
+  };
+
+  declare class Provider<S, A> extends React$Component<void, { store: Store<S, A>, children?: any }, void> { }
+
+  declare type ConnectOptions = {
+    pure?: boolean,
+    withRef?: boolean
+  };
+
+  declare function connect<A, OP>(
+    options?: ConnectOptions
+  ): Connector<OP, { dispatch: Dispatch<A> } & OP>;
+
+  declare function connect<S, A, OP, SP>(
+    mapStateToProps: MapStateToProps<S, OP, SP>,
+    options?: ConnectOptions
+  ): Connector<OP, SP & { dispatch: Dispatch<A> }>;
+
+  declare function connect<A, OP, DP>(
+    mapStateToProps: null,
+    mapDispatchToProps: MapDispatchToProps<A, OP, DP>,
+    options?: ConnectOptions
+  ): Connector<OP, DP & OP>;
+
+  declare function connect<S, A, OP, SP, DP>(
+    mapStateToProps: MapStateToProps<S, OP, SP>,
+    mapDispatchToProps: MapDispatchToProps<A, OP, DP>,
+    options?: ConnectOptions
+  ): Connector<OP, SP & DP>;
+
+  declare function connect<S, A, OP, SP, DP, P>(
+    mapStateToProps: MapStateToProps<S, OP, SP>,
+    mapDispatchToProps: MapDispatchToProps<A, OP, DP>,
+    mergeProps: MergeProps<SP, DP, OP, P>,
+    options?: ConnectOptions
+  ): Connector<OP, P>;
+
+}
--- a/flow-typed/react-redux.js
+++ b/flow-typed/react-redux.js
@@ -45,7 +45,7 @@ declare module 'react-redux' {
   declare type Null = null | void;
 
   declare function connect<A, OP>(
-    ...rest: Array<void>
+    ...rest: Array<void> // <= workaround for https://github.com/facebook/flow/issues/2360
   ): Connector<OP, $Supertype<{ dispatch: Dispatch<A> } & OP>>;
 
   declare function connect<A, OP>(
--- a/flow-typed/react-redux.js
+++ b/flow-typed/react-redux.js
@@ -42,26 +42,39 @@ declare module 'react-redux' {
     withRef?: boolean
   };
 
+  declare type Null = null | void;
+
   declare function connect<A, OP>(
-    options?: ConnectOptions
-  ): Connector<OP, { dispatch: Dispatch<A> } & OP>;
+    ...rest: Array<void>
+  ): Connector<OP, $Supertype<{ dispatch: Dispatch<A> } & OP>>;
+
+  declare function connect<A, OP>(
+    mapStateToProps: Null,
+    mapDispatchToProps: Null,
+    mergeProps: Null,
+    options: ConnectOptions
+  ): Connector<OP, $Supertype<{ dispatch: Dispatch<A> } & OP>>;
 
   declare function connect<S, A, OP, SP>(
     mapStateToProps: MapStateToProps<S, OP, SP>,
+    mapDispatchToProps: Null,
+    mergeProps: Null,
     options?: ConnectOptions
-  ): Connector<OP, SP & { dispatch: Dispatch<A> }>;
+  ): Connector<OP, $Supertype<SP & { dispatch: Dispatch<A> } & OP>>;
 
   declare function connect<A, OP, DP>(
-    mapStateToProps: null,
+    mapStateToProps: Null,
     mapDispatchToProps: MapDispatchToProps<A, OP, DP>,
+    mergeProps: Null,
     options?: ConnectOptions
-  ): Connector<OP, DP & OP>;
+  ): Connector<OP, $Supertype<DP & OP>>;
 
   declare function connect<S, A, OP, SP, DP>(
     mapStateToProps: MapStateToProps<S, OP, SP>,
     mapDispatchToProps: MapDispatchToProps<A, OP, DP>,
+    mergeProps: Null,
     options?: ConnectOptions
-  ): Connector<OP, SP & DP>;
+  ): Connector<OP, $Supertype<SP & DP & OP>>;
 
   declare function connect<S, A, OP, SP, DP, P>(
     mapStateToProps: MapStateToProps<S, OP, SP>,
--- a/flow-typed/redux.js
+++ b/flow-typed/redux.js
@@ -1,0 +1,54 @@
+declare module 'redux' {
+
+  /*
+
+    S = State
+    A = Action
+
+  */
+
+  declare type Dispatch<A: { type: $Subtype<string> }> = (action: A) => A;
+
+  declare type MiddlewareAPI<S, A> = {
+    dispatch: Dispatch<A>;
+    getState(): S;
+  };
+
+  declare type Store<S, A> = {
+    // rewrite MiddlewareAPI members in order to get nicer error messages (intersections produce long messages)
+    dispatch: Dispatch<A>;
+    getState(): S;
+    subscribe(listener: () => void): () => void;
+    replaceReducer(nextReducer: Reducer<S, A>): void
+  };
+
+  declare type Reducer<S, A> = (state: S, action: A) => S;
+
+  declare type Middleware<S, A> =
+    (api: MiddlewareAPI<S, A>) =>
+      (next: Dispatch<A>) => Dispatch<A>;
+
+  declare type StoreCreator<S, A> = {
+    <S, A>(reducer: Reducer<S, A>, enhancer?: StoreEnhancer<S, A>): Store<S, A>;
+    <S, A>(reducer: Reducer<S, A>, preloadedState: S, enhancer?: StoreEnhancer<S, A>): Store<S, A>;
+  };
+
+  declare type StoreEnhancer<S, A> = (next: StoreCreator<S, A>) => StoreCreator<S, A>;
+
+  declare function createStore<S, A>(reducer: Reducer<S, A>, enhancer?: StoreEnhancer<S, A>): Store<S, A>;
+  declare function createStore<S, A>(reducer: Reducer<S, A>, preloadedState: S, enhancer?: StoreEnhancer<S, A>): Store<S, A>;
+
+  declare function applyMiddleware<S, A>(...middlewares: Array<Middleware<S, A>>): StoreEnhancer<S, A>;
+
+  declare type ActionCreator<A> = (...args: Array<any>) => A;
+  declare type ActionCreators<K, A> = { [key: K]: ActionCreator<A> };
+
+  declare function bindActionCreators<A, C: ActionCreator<A>>(actionCreator: C, dispatch: Dispatch<A>): C;
+  declare function bindActionCreators<A, K, C: ActionCreators<K, A>>(actionCreators: C, dispatch: Dispatch<A>): C;
+
+  // unsafe (you can miss a field and / or assign a wrong reducer to a field)
+  declare function combineReducers<S: Object, A>(reducers: {[key: $Keys<S>]: Reducer<any, A>}): Reducer<S, A>;
+
+  declare function compose<S, A>(...fns: Array<StoreEnhancer<S, A>>): Function;
+
+}
--- a/examples/todos-flow/.flowconfig
+++ b/examples/todos-flow/.flowconfig
@@ -1,0 +1,9 @@
+[ignore]
+<PROJECT_ROOT>/node_modules/fbjs
+
+[include]
+
+[libs]
+../../flow-typed
+
+[options]
--- a/examples/todos-flow/.gitignore
+++ b/examples/todos-flow/.gitignore
@@ -1,0 +1,11 @@
+# See http://help.github.com/ignore-files/ for more about ignoring files.
+
+# dependencies
+node_modules
+
+# production
+build
+
+# misc
+.DS_Store
+npm-debug.log
--- a/examples/todos-flow/README.md
+++ b/examples/todos-flow/README.md
@@ -1,0 +1,34 @@
+# Redux Todos Example
+
+This project template was built with [Create React App](https://github.com/facebookincubator/create-react-app).
+
+## Available Scripts
+
+In the project directory, you can run:
+
+### `npm start`
+
+Runs the app in the development mode.<br>
+Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
+
+The page will reload if you make edits.<br>
+You will also see any lint errors in the console.
+
+### `npm run build`
+
+Builds the app for production to the `build` folder.<br>
+It correctly bundles React in production mode and optimizes the build for the best performance.
+
+The build is minified and the filenames include the hashes.<br>
+Your app is ready to be deployed!
+
+### `npm run eject`
+
+**Note: this is a one-way operation. Once you `eject`, you can’t go back!**
+
+If you aren’t satisfied with the build tool and configuration choices, you can `eject` at any time. This command will remove the single build dependency from your project.
+
+Instead, it will copy all the configuration files and the transitive dependencies (Webpack, Babel, ESLint, etc) right into your project so you have full control over them. All of the commands except `eject` will still work, but they will point to the copied scripts so you can tweak them. At this point you’re on your own.
+
+You don’t have to ever use `eject`. The curated feature set is suitable for small and middle deployments, and you shouldn’t feel obligated to use this feature. However we understand that this tool wouldn’t be useful if you couldn’t customize it when you are ready for it.
+
--- a/examples/todos-flow/index.html
+++ b/examples/todos-flow/index.html
@@ -1,0 +1,21 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>Redux Todos Example</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <!--
+      This HTML file is a template.
+      If you open it directly in the browser, you will see an empty page.
+
+      You can add webfonts, meta tags, or analytics to this file.
+      The build step will place the bundled scripts into the <body> tag.
+
+      To begin the development, run `npm start` in this folder.
+      To create a production bundle, use `npm run build`.
+    -->
+  </body>
+</html>
--- a/examples/todos-flow/package.json
+++ b/examples/todos-flow/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "todos",
+  "version": "0.0.1",
+  "private": true,
+  "devDependencies": {
+    "enzyme": "^2.4.1",
+    "react-addons-test-utils": "^15.3.0",
+    "react-scripts": "^0.4.0"
+  },
+  "dependencies": {
+    "react": "^15.3.0",
+    "react-dom": "^15.3.0",
+    "react-redux": "^4.4.5",
+    "redux": "^3.5.2"
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "eject": "react-scripts eject",
+    "test": "react-scripts test"
+  },
+  "eslintConfig": {
+    "extends": "./node_modules/react-scripts/config/eslint.js"
+  }
+}
--- a/examples/todos-flow/src/actions/index.js
+++ b/examples/todos-flow/src/actions/index.js
@@ -1,0 +1,26 @@
+// @flow
+import type { Id, Text, VisibilityFilter, Action } from '../types'
+
+let nextTodoId: Id = 0
+
+export const addTodo = (text: Text): Action => {
+  return {
+    type: 'ADD_TODO',
+    id: nextTodoId++,
+    text
+  }
+}
+
+export const setVisibilityFilter = (filter: VisibilityFilter): Action => {
+  return {
+    type: 'SET_VISIBILITY_FILTER',
+    filter
+  }
+}
+
+export const toggleTodo = (id: Id): Action => {
+  return {
+    type: 'TOGGLE_TODO',
+    id
+  }
+}
--- a/examples/todos-flow/src/actions/index.spec.js
+++ b/examples/todos-flow/src/actions/index.spec.js
@@ -1,0 +1,25 @@
+import * as actions from './index'
+
+describe('todo actions', () => {
+  it('addTodo should create ADD_TODO action', () => {
+    expect(actions.addTodo('Use Redux')).toEqual({
+      type: 'ADD_TODO',
+      id: 0,
+      text: 'Use Redux'
+    })
+  })
+
+  it('setVisibilityFilter should create SET_VISIBILITY_FILTER action', () => {
+    expect(actions.setVisibilityFilter('active')).toEqual({
+      type: 'SET_VISIBILITY_FILTER',
+      filter: 'active'
+    })
+  })
+
+  it('toggleTodo should create TOGGLE_TODO action', () => {
+    expect(actions.toggleTodo(1)).toEqual({
+      type: 'TOGGLE_TODO',
+      id: 1
+    })
+  })
+})
--- a/examples/todos-flow/src/components/App.js
+++ b/examples/todos-flow/src/components/App.js
@@ -1,0 +1,15 @@
+// @flow
+import React from 'react'
+import Footer from './Footer'
+import AddTodo from '../containers/AddTodo'
+import VisibleTodoList from '../containers/VisibleTodoList'
+
+const App = () => (
+  <div>
+    <AddTodo />
+    <VisibleTodoList />
+    <Footer />
+  </div>
+)
+
+export default App
--- a/examples/todos-flow/src/components/Footer.js
+++ b/examples/todos-flow/src/components/Footer.js
@@ -1,0 +1,23 @@
+// @flow
+import React from 'react'
+import FilterLink from '../containers/FilterLink'
+
+const Footer = () => (
+  <p>
+    Show:
+    {" "}
+    <FilterLink filter="SHOW_ALL">
+      All
+    </FilterLink>
+    {", "}
+    <FilterLink filter="SHOW_ACTIVE">
+      Active
+    </FilterLink>
+    {", "}
+    <FilterLink filter="SHOW_COMPLETED">
+      Completed
+    </FilterLink>
+  </p>
+)
+
+export default Footer
--- a/examples/todos-flow/src/components/Link.js
+++ b/examples/todos-flow/src/components/Link.js
@@ -1,0 +1,27 @@
+// @flow
+import React from 'react'
+
+export type Props = {
+  active: boolean,
+  children?: React$Element<any>,
+  onClick: () => void
+};
+
+const Link = ({ active, children, onClick }: Props) => {
+  if (active) {
+    return <span>{children}</span>
+  }
+
+  return (
+    <a href="#"
+       onClick={e => {
+         e.preventDefault()
+         onClick()
+       }}
+    >
+      {children}
+    </a>
+  )
+}
+
+export default Link
--- a/examples/todos-flow/src/components/Todo.js
+++ b/examples/todos-flow/src/components/Todo.js
@@ -1,0 +1,22 @@
+// @flow
+import React from 'react'
+import type { Text } from '../types'
+
+export type Props = {
+  onClick: () => void,
+  completed: boolean,
+  text: Text
+};
+
+const Todo = ({ onClick, completed, text }: Props) => (
+  <li
+    onClick={onClick}
+    style={{
+      textDecoration: completed ? 'line-through' : 'none'
+    }}
+  >
+    {text}
+  </li>
+)
+
+export default Todo
--- a/examples/todos-flow/src/components/TodoList.js
+++ b/examples/todos-flow/src/components/TodoList.js
@@ -1,0 +1,23 @@
+// @flow
+import React from 'react'
+import Todo from './Todo'
+import type { Todos, Id } from '../types'
+
+export type Props = {
+  todos: Todos,
+  onTodoClick: (id: Id) => void
+};
+
+const TodoList = ({ todos, onTodoClick }: Props) => (
+  <ul>
+    {todos.map(todo =>
+      <Todo
+        key={todo.id}
+        {...todo}
+        onClick={() => onTodoClick(todo.id)}
+      />
+    )}
+  </ul>
+)
+
+export default TodoList
--- a/examples/todos-flow/src/containers/AddTodo.js
+++ b/examples/todos-flow/src/containers/AddTodo.js
@@ -1,0 +1,38 @@
+// @flow
+import React from 'react'
+import { connect } from 'react-redux'
+import { addTodo } from '../actions'
+import type { Dispatch } from '../types'
+import type { Connector } from 'react-redux'
+
+type Props = {
+  dispatch: Dispatch
+};
+
+const AddTodo = ({ dispatch }) => {
+  let input
+
+  return (
+    <div>
+      <form onSubmit={e => {
+        e.preventDefault()
+        if (!input.value.trim()) {
+          return
+        }
+        dispatch(addTodo(input.value))
+        input.value = ''
+      }}>
+        <input ref={node => {
+          input = node
+        }} />
+        <button type="submit">
+          Add Todo
+        </button>
+      </form>
+    </div>
+  )
+}
+
+const connector: Connector<{}, Props> = connect()
+
+export default connector(AddTodo)
--- a/examples/todos-flow/src/containers/FilterLink.js
+++ b/examples/todos-flow/src/containers/FilterLink.js
@@ -1,0 +1,32 @@
+// @flow
+import { connect } from 'react-redux'
+import { setVisibilityFilter } from '../actions'
+import Link from '../components/Link'
+import type { Props } from '../components/Link'
+import type { State, Dispatch, VisibilityFilter } from '../types'
+import type { Connector } from 'react-redux'
+
+type OwnProps = {
+  filter: VisibilityFilter
+};
+
+const mapStateToProps = (state: State, ownProps) => {
+  return {
+    active: ownProps.filter === state.visibilityFilter
+  }
+}
+
+const mapDispatchToProps = (dispatch: Dispatch, ownProps) => {
+  return {
+    onClick: () => {
+      dispatch(setVisibilityFilter(ownProps.filter))
+    }
+  }
+}
+
+const connector: Connector<OwnProps, Props> = connect(
+  mapStateToProps,
+  mapDispatchToProps
+)
+
+export default connector(Link)
--- a/examples/todos-flow/src/containers/VisibleTodoList.js
+++ b/examples/todos-flow/src/containers/VisibleTodoList.js
@@ -1,0 +1,40 @@
+// @flow
+import { connect } from 'react-redux'
+import { toggleTodo } from '../actions'
+import TodoList from '../components/TodoList'
+import type { State, Dispatch } from '../types'
+import type { Connector } from 'react-redux'
+import type { Props } from '../components/TodoList'
+
+const getVisibleTodos = (todos, filter) => {
+  switch (filter) {
+    case 'SHOW_COMPLETED':
+      return todos.filter(t => t.completed)
+    case 'SHOW_ACTIVE':
+      return todos.filter(t => !t.completed)
+    case 'SHOW_ALL':
+    default :
+      return todos
+  }
+}
+
+const mapStateToProps = (state: State) => {
+  return {
+    todos: getVisibleTodos(state.todos, state.visibilityFilter)
+  }
+}
+
+const mapDispatchToProps = (dispatch: Dispatch) => {
+  return {
+    onTodoClick: (id) => {
+      dispatch(toggleTodo(id))
+    }
+  }
+}
+
+const connector: Connector<{}, Props> = connect(
+  mapStateToProps,
+  mapDispatchToProps
+)
+
+export default connector(TodoList)
--- a/examples/todos-flow/src/index.js
+++ b/examples/todos-flow/src/index.js
@@ -1,0 +1,17 @@
+// @flow
+import React from 'react'
+import { render } from 'react-dom'
+import { createStore } from 'redux'
+import { Provider } from 'react-redux'
+import App from './components/App'
+import reducer from './reducers'
+import type { Store } from './types'
+
+const store: Store = createStore(reducer)
+
+render(
+  <Provider store={store}>
+    <App />
+  </Provider>,
+  document.getElementById('root')
+)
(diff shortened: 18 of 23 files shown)


timdorr:
I don't think we need the examples to be run with Flow. We don't have any other examples that use TypeScript, by comparison.

I don't think there is value in the Flow definitions, though. Can you open that up as a separate PR?

gcanti:
@timdorr I added the examples (and I do believe they are very important) for several reasons:

- was suggested in (https://github.com/reactjs/redux/issues/290) or specifically requested (https://github.com/reactjs/redux/issues/1866)
- to check the libdefs: you can't really say they are working without some running examples. Examples are like test unit for libdefs
- in order to get feedback from the community: we need some code people can play with, submitting the libdefs alone is not feedback-friendly
- these libdefs are written with the goal of maximizing the type safety and work better if coupled with some coding styles and idioms which are present in the submitted code. For example:
  - this type annotation for the store is mandatory https://github.com/reactjs/redux/pull/1887/files#diff-e1a29e2479017e72e237ecf03baab261R9
  - this "connector tecnique" https://github.com/reactjs/redux/pull/1887/files#diff-ef2fbfc16613d42f52b86d75581efc9aR27 is not easily guessable by just looking at the libdef alone
- there are some `.flowconfig` real world examples (e.g. how to handle `import 'todomvc-app-css/index.css'` in todomvc)
- they show that, in the context of flow, some idiomatic redux code like using constants, `combineReducers` or `bindActionCreators`, may be less useful or even unsafe

But even more important, while libdefs alone seem the main asset, examples are crucial for beginners: adding type annotations (but I'd say more in general adding type safety) is not always a straightforward process. Moreover, these examples are valuable because show the transition from an untyped codebase to a typed one.

timdorr:
[We test our TypeScript defs](https://github.com/reactjs/redux/blob/master/test/typescript.spec.js). Is there no equivalent testing infrastructure for Flow?

All of what you're saying describes Flow-specific issues and idioms. Those could apply to any other library, not just Redux. I don't want us to take on a language extension that will be unfamiliar to beginners or indicate some pattern they should be following. In fact, this is what Dan is trying to do in #1883, where create-react-app will let us get rid of a lot of the tooling infrastructure that distracts from the important part: the code. 

Also, none of regular contributors are Flow experts (otherwise, we would have done this a year ago when Dan brought it up :smile:), so adding code that no one can really own isn't going to serve us well.

gaearon:
I would like to have a Flow example.
Not for *every* single example, but having something like `flow-todos` is something I want.
However it would need to be rebased on top of #1883 and use the same infra.

>In fact, this is what Dan is trying to do in #1883, where create-react-app will let us get rid of a lot of the tooling infrastructure that distracts from the important part: the code.

I want Flow to eventually be included by default with Create React App. We might also start using it more in React docs as it becomes better suited to common workflows / stacks in the community.

I would like to reopen and re-scope this in the following way:

* Just a single `todos-flow` example.
* It should build on top of #1883.
* I don’t understand how `flow-typed` folder works: does this mean we’ll ship Flow definitions? Or would users have to copy-paste it? I want `todos-flow` to work out of the box if possible. If we need to ship definitions, I’d like to do that.

timdorr:
Well, my bad then. Sorry for closing!

alexeygolev:
@gaearon `flow-typed` folder within a project is automatically used by flow as a libdef folder. To make it work out of the box the definitions should be shipped (Draft.js team has made some progress on that as far as I know — they ship their definitions with the lib). Another possibility is to add it to (flow-typed)[https://github.com/flowtype/flow-typed/] repo. There is a community created libdef for redux there, but having an official one is always better.

gaearon:
If someone is willing to lead the effort of official Flow types I’m happy to assist in any way I can.

alexeygolev:
@gaearon I guess we just need to decide between flow-typed repo or shipping with redux

gaearon:
>I guess we just need to decide between flow-typed repo or shipping with redux

@thejameskyle Any opinions on this?

gcanti:
@gaearon if there's something I can do I'm glad to help. I worked on this for a few weeks, when you find the right tradeoff, flow is awesome

timdorr:
@gcanti Do you mind splitting out the single commit into one for each example and the libdefs separately? Even better would be a separate PR for the definitions and the examples, but you don't have to do that. Reading through 2,675 new lines makes my browser crawl 😄

timdorr:
Thank you so much! 💯

timdorr:
So, I'm with @gaearon on keeping this to just one single example. Duplicating examples means pain whenever one is updated and you have to synchronize the changes to the other. 3 examples mean that happens 3x as often. 

Given the size of the `todomvc-flow` example (1,560 lines of code), that one should go. I know it has some things to help with dealing with CSS imports, but that's a Flow issue, not a Redux one. So, I don't think we're in any position here to be establishing patterns for another library.

I'd be happy with either the `todos-flow` or `counter-flow` examples being added. It looks like `todos-flow` has more substantial Flow-specific additions (namely, a separate types definition for the app), so it seems like the better choice. What are your thoughts, @gcanti?

gcanti:
I agree, one single example is enough. counter-flow is too simple while todomvc-flow is unnecessary big.

todos-flow seems perfect, is fairly small but contains all the juice. It shows how to:

- configure a minimal .flowconfig file
- type check the action creators
- type check the reducers
- write a type safe combined reducer
- type check stateless components
- type check the containers and connect them via react-redux with the connector pattern
- group the domain model in a separate file
- tie all together by annotating the store

kristian-puccio:
The other thing that I've been doing for this is typed check tests to go
with it.

The pattern I've used is each reducer file exports it's own state +
initialState and each corresponding test file exports some populated typed
test data.
That way in my root reducer I can combine states and in my root reducer
test file I have access to typed test state from other reducers so I can
test the selectors that select across more than one reducer.

I've never been able to get typing working for the connect HOC though so
I've been getting the mapStateToProps to export StateProps and
mapDispatchToProps to export DispatchProps.
Then my props type for the component is `type Props = StateProps &
DispatchProps & IncomingProps;`. This works but feels super hacky.

flow-typed/redux.js
```diff
@@ -0,0 +1,74 @@
...
+  declare type Dispatch<S, A> = (action: A) => A;
+
+  declare type MiddlewareAPI<S, A> = {
+    dispatch: Dispatch<S, A>;
+    getState(): S;
+  };
+
+  declare type Store<S, A: { type: $Subtype<string> }> = {
+    // rewrite MiddlewareAPI members in order to get nicer error messages (intersections produce long messages)
+    dispatch: Dispatch<S, A>;
```
kasperpeulen: Wouldn't `Dispatch<A>` do the job as well?

kasperpeulen:
> 4) libdefs need community feedback, then we might consider to push them to flow-typed

Should I give the feedback in this PR?

flow-typed/redux.js
gcanti: Ah yes, that `S` was an experiment to retain type info in favour of react-redux (I mean the `dispatch` function injected by `connect` in some cases) but I guess is unnecessary. Better to discard it and simplify both libdefs, thanks for pointing out.

kasperpeulen:
@gcanti I have one problem with your definition, so basically when you have this:

`const connector: Connector<OwnProps,Props> = connect(...)`

It now works like that Props are the Props of the component. And OwnProps are the Props that the connected Component exposes. It assumes that `Props = SP & { dispatch: Dispatch<S, A> }` where `SP` are the props it gets from the redux State. But this is not true right? The props can also come from the view, it doesn't have to come from the redux state.

I get a bunch of errors of code that work, because of that (verryyy looongg error messages). But I think this is the reason.

kasperpeulen:
For example, say I have:
```
    <Counter style={{color: 'red'}}/>
```

where Counter is

```
type Props = {
  show: boolean,
  value: number,
  dispatch: Dispatch<State, Action>,
  style: Object
}
const Counter = ({show, value, dispatch, style}: Props) =
```

Now `show` and `value` is coming from the redux state like this:

```
const connector: Connector<{style: Object}, Props> = connect(
  ({Counter}: State) => {
    return {...Counter}
  }
);
```

Where Counter has type 
```
{
  value: number,
  show: boolean
}
```

So everything works if I run it, but I get the following errors:

```
src/Counter/index.js:0
inconsistent use of library definitions
 33:   declare type Connector<OP, P> = {
                                       ^ intersection. This type is incompatible with. See lib: flow-typed/react-redux.js:33
 34:     (component: StatelessComponent<P>): ConnectedComponentClass<OP, P, void, void>;
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See lib: flow-typed/react-redux.js:34
  Member 1:
   34:     (component: StatelessComponent<P>): ConnectedComponentClass<OP, P, void, void>;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See lib: flow-typed/react-redux.js:34
  Error:
   38: const connector: Connector<{style: Object}, Props> = connect(
                                                   ^^^^^ object type. This type is incompatible with. See: src/Counter/Counter.js:38
   52:   ): Connector<OP, SP & { dispatch: Dispatch<S, A> }>;
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ intersection. See lib: flow-typed/react-redux.js:52
    Member 1:
     38: const connector: Connector<{style: Object}, Props> = connect(
                                                              ^ type parameter `SP` of function call. See: src/Counter/Counter.js:38
    Error:
     38: const connector: Connector<{style: Object}, Props> = connect(
                                                     ^^^^^ property `style`. Property not found in. See: src/Counter/Counter.js:38
     40:     return {...Counter}
                    ^^^^^^^^^^^^ object literal. See: src/Counter/Counter.js:40
    Member 2:
     52:   ): Connector<OP, SP & { dispatch: Dispatch<S, A> }>;
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ object type. See lib: flow-typed/react-redux.js:52
    Error:
     38: const connector: Connector<{style: Object}, Props> = connect(
                                                     ^^^^^ property `style`. Property not found in. See: src/Counter/Counter.js:38
     52:   ): Connector<OP, SP & { dispatch: Dispatch<S, A> }>;
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ object type. See lib: flow-typed/react-redux.js:52
  Member 2:
   35:     <Def, St>(component: Class<React$Component<Def, P, St>>): ConnectedComponentClass<OP, P, Def, St>;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ polymorphic type: function type. See lib: flow-typed/react-redux.js:35
  Error:
  inconsistent use of library definitions
   21:   declare type StatelessComponent<P> = (props: P) => ?React$Element<any>;
                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ statics of function type. Callable signature not found in. See lib: flow-typed/react-redux.js:21
   16: declare class React$Component<DefaultProps, Props, State> {
       ^ statics of React$Component. See lib: /private/tmp/flow/flowlib_377a32d6/react.js:16

src/Counter/index.js:0
inconsistent use of library definitions
 33:   declare type Connector<OP, P> = {
                                       ^ intersection. This type is incompatible with. See lib: flow-typed/react-redux.js:33
 35:     <Def, St>(component: Class<React$Component<Def, P, St>>): ConnectedComponentClass<OP, P, Def, St>;
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ polymorphic type: function type. See lib: flow-typed/react-redux.js:35
  Member 1:
   34:     (component: StatelessComponent<P>): ConnectedComponentClass<OP, P, void, void>;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See lib: flow-typed/react-redux.js:34
  Error:
  inconsistent use of library definitions
   34:     (component: StatelessComponent<P>): ConnectedComponentClass<OP, P, void, void>;
                       ^^^^^^^^^^^^^^^^^^^^^ function type. Callable signature not found in. See lib: flow-typed/react-redux.js:34
   35:     <Def, St>(component: Class<React$Component<Def, P, St>>): ConnectedComponentClass<OP, P, Def, St>;
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^ statics of React$Component. See lib: flow-typed/react-redux.js:35
  Member 2:
   35:     <Def, St>(component: Class<React$Component<Def, P, St>>): ConnectedComponentClass<OP, P, Def, St>;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ polymorphic type: function type. See lib: flow-typed/react-redux.js:35
  Error:
   38: const connector: Connector<{style: Object}, Props> = connect(
                                                   ^^^^^ object type. This type is incompatible with. See: src/Counter/Counter.js:38
   52:   ): Connector<OP, SP & { dispatch: Dispatch<S, A> }>;
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ intersection. See lib: flow-typed/react-redux.js:52
    Member 1:
     38: const connector: Connector<{style: Object}, Props> = connect(
                                                              ^ type parameter `SP` of function call. See: src/Counter/Counter.js:38
    Error:
     38: const connector: Connector<{style: Object}, Props> = connect(
                                                     ^^^^^ property `style`. Property not found in. See: src/Counter/Counter.js:38
     40:     return {...Counter}
                    ^^^^^^^^^^^^ object literal. See: src/Counter/Counter.js:40
    Member 2:
     52:   ): Connector<OP, SP & { dispatch: Dispatch<S, A> }>;
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ object type. See lib: flow-typed/react-redux.js:52
    Error:
     38: const connector: Connector<{style: Object}, Props> = connect(
                                                     ^^^^^ property `style`. Property not found in. See: src/Counter/Counter.js:38
     52:   ): Connector<OP, SP & { dispatch: Dispatch<S, A> }>;
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ object type. See lib: flow-typed/react-redux.js:52


Found 2 errors
```

gcanti:
@kasperpeulen type parameter `S` removed from `Dispatch`.

> It assumes that Props = SP & { dispatch: Dispatch<S, A> } where SP are the props it gets from the redux State. But this is not true right?

Yes, theorically you are right. Technically, in order to type check as much as possible, I need that property (I tried for hours other approaches but I failed. I'm open to alternative implementation though)

> I get a bunch of errors of code that work

Seems an easy fix

```js
const connector: Connector<{style: Object}, Props> = connect(
  ({Counter}: State, {style}) => {
    return {...Counter, style}
  }
)
```

> verryyy looongg error messages

Yeah, alas there are (5 overloadings of `connect`) x (2 ways of defining a component), i.e. stateless and `React$Component`, thus the error messages are long and awkward

kasperpeulen:
> Yes, theorically you are right. Technically, in order to type check as much as possible, I need that property (I tried for hours other approaches but I failed. I'm open to alternative implementation though)

I see, I tried a bit as well, but I also failed getting it to work like I wanted.

> Seems an easy fix

That works indeed. Strangely enough, this did not work:

```
const connector: Connector<{style: Object}, Props> = connect(
  ({Counter}: State, ownProps) => {
    return {...Counter, ...ownProps}
  }
)
```

If this will be chosen as the way to go, I think it is important to have good documentation. 

`
type Props = {
  show: boolean,
  value: number,
  dispatch: Dispatch<State, Action>,
  style?: Object
}
`

That seems to work as well. In this case this is some sense better, as it seems natural to make a style property optional.

gcanti:
> Strangely enough, this did not work

Yeah, kind of surprising. The spread works if you add a type annotation

```js
type OwnProps = {
  style: Object
};

const connector: Connector<OwnProps, Props> = connect(
  ({Counter}: State, ownProps: OwnProps) => {
    return {...Counter, ...ownProps}
  }
)
```

agentcooper:
@gcanti Thank you for working on this, proper definitions for react-redux are the missing piece for full Flow coverage on our project.

I'm trying to compile a simple example with your definitions: https://github.com/agentcooper/react-redux-flow-example.

`App` [renders](https://github.com/agentcooper/react-redux-flow-example/blob/master/src/App.js#L18) `Counter` without `text` prop and Flow does not seem to pick up the error. Is this solvable?

gcanti:
@agentcooper

- move the `flow-typed` dir to the project root (otherwise flow doesn't see it)
- add the `// @flow` pragma to `types/index.js`

you'll see a few errors

agentcooper:
@gcanti just did that and found out that I have the same issue as @kasperpeulen. It is actually working now, however merging ownProps with stateProps looks little bit awkward: https://github.com/agentcooper/react-redux-flow-example/blob/master/src/Counter.js#L30.

flow-typed/redux.js
```diff
@@ -0,0 +1,74 @@
+declare module 'redux' {
+
+  /*
+
+    S = State
+    A = Action
+
+  */
+
+  declare type Dispatch<S, A> = (action: A) => A;
```
rsolomon: I think this definition is a bit more rigid than the middleware API allows. For example, `redux-actions` would allow a `Dispatch` object to take a param `action` of type `{type: T, payload: Promise<P>}`, but would then return `Promise<{type: T, payload: P}>`.

gcanti: Good point. Is there a better option than returning `any`?

Note. The typescript definition file should be modified as well

rsolomon: I can't think of anything better at the moment, since any middleware could modify this in all sorts of ways. It might be possible to pass through the definition of `Dispatch` from `StoreEnhancer`. However, I think I remember reading that a reason Flow typedefs for Redux were put on hold last year was because this was difficult to achieve.

aaronjensen:
How would one augment `Dispatch<A>` with the thunk middleware? You can't just include a thunk in `A` because it doesn't have a `type` property. Would you have to edit the types directly to do this or is there another way?

agentcooper:
@aaronjensen take a look at https://github.com/agentcooper/react-redux-flow-example/blob/master/src/types/index.js#L20

aaronjensen:
@agentcooper thanks, that worked. I didn't know you could intersect functions like that.

flow-typed/redux.js
```diff
@@ -0,0 +1,74 @@
+declare module 'redux' {
+
+  /*
+
+    S = State
+    A = Action
+
+  */
+
+  declare type Dispatch<A: { type: $Subtype<string> }> = (action: A) => A;
```
aaronjensen: Why does this need to be `$Subtype<string>` instead of `string`?

gcanti: Bacause otherwise you get the following error (todos-flow example)

```
types/index.js:28
 28:     { type: 'ADD_TODO', id: Id, text: Text }
                 ^^^^^^^^^^ string literal `ADD_TODO`. Expected string literal `ADD_TODO`
 10:   declare type Dispatch<A: { type: string }> = (action: A) => A;
                                        ^^^^^^ string. See lib: ../../flow-typed/redux.js:10

types/index.js:29
 29:   | { type: 'TOGGLE_TODO', id: Id }
                 ^^^^^^^^^^^^^ string literal `TOGGLE_TODO`. Expected string literal `TOGGLE_TODO`
 10:   declare type Dispatch<A: { type: string }> = (action: A) => A;
                                        ^^^^^^ string. See lib: ../../flow-typed/redux.js:10

types/index.js:30
 30:   | { type: 'SET_VISIBILITY_FILTER', filter: VisibilityFilter }
                 ^^^^^^^^^^^^^^^^^^^^^^^ string literal `SET_VISIBILITY_FILTER`. Expected string literal `SET_VISIBILITY_FILTER`
 10:   declare type Dispatch<A: { type: string }> = (action: A) => A;
                                        ^^^^^^ string. See lib: ../../flow-typed/redux.js:10
```

examples/todos-flow/components/Link.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+import React from 'react'
+
+export type Props = {
+  active: boolean,
+  children?: React$Element<any>,
```
cr0cK: According to http://www.saltycrane.com/blog/2016/06/flow-type-cheat-sheet/, React$* types are private.

It's not better like this ?
```
import type { Element } from 'react';
// ...
children?: Element<any>,
// ...
```
?

gcanti: @cr0cK yes, thanks. That declaration is weird though, unlike `Component` and `PureComponent` there's no "real" named export called `Element`. 

> I **guess** names with a $ are supposed to be private

Honestly I don't know what's the official advice, I might update the PR if there is a consensus on this

ryyppy: @cr0cK To clear some things up...

> I guess names with a $ are supposed to be private

I think what you should consider as private are those special "magic" types like `$Shape<T>` ... `React$` is a namespace to prevent collisions with other loaded libdefs, it doesn't mean that this is private.

ccorcos:
@gcanti I've been playing around with Flow and Redux and this is by far the best example yet. 🏆🏆🏆 

In some of the other implementations I've seen, I'm not sure they completely understand the concept of "strongly typed". Even the [flow-typed](https://github.com/flowtype/flow-typed/blob/master/definitions/npm/react-redux_v4.x.x/flow_%3E%3Dv0.30.x/react-redux_v4.x.x.js#L3) definitions are pretty crappy if you ask me -- state is annotated as an Object!

Anyways, I had some feedback to give you, but I answered all my own questions as I went back through to try find the answers. This looks awesome. I'm going to play with it today. Let me know if I can help.

Also, it would be great if these lib defs were in `.js.flow` format so we dont have mess with the flowconfig file

ccorcos:
I've gotten this to work pretty well for me. The only thing that isnt working is validating component props after `connect`.

For example:

```js
// @flow

import React from 'react'
import classnames from 'classnames'
import UnderlineSrc from 'skeleton/src/images/underline.png'
import 'skeleton/src/components/counter/style.css'
import * as actions from 'skeleton/src/actions'
import type { State, Dispatch, Action } from 'skeleton/src/types'

type Props = {
  color: 'red' | 'blue',
  pending: boolean,
  count: number,
  decrement: Function,
  increment: Function,
}

const Counter = (props : Props) => {
  const className = classnames({
    counter: true,
    disabled: props.pending,
  })
  return (
    <div className={className} style={{color: props.color}}>
      <div>
        <button className="decrement" onClick={props.decrement}>{'-'}</button>
        <span className="count">{props.count}</span>
        <button className="increment" onClick={props.increment}>{'+'}</button>
      </div>
      <img alt="underline" className="underline" src={UnderlineSrc} />
    </div>
  )
}

const mapStateToProps = (state: State) => ({
  count: state.get('count'),
  pending: state.get('pending').includes('decrement'),
})

const mapDispatchToProps = (dispatch: Dispatch) => ({
  increment: () => dispatch(actions.increment()),
  decrement: (count) => () => dispatch(actions.decrement(count)),
})

const mergeProps = (stateProps, dispatchProps, ownProps) => ({
  ...stateProps,
  ...dispatchProps,
  ...ownProps,
  decrement: dispatchProps.decrement(stateProps.count),
})

export default connect(mapStateToProps, mapDispatchToProps, mergeProps)(Counter)
```

And this doesnt throw an error:

```js
ReactDOM.render((
  <Provider store={store}>
    <Counter/>
  </Provider>
), root)
```

gcanti:
@ccorcos try this pattern https://github.com/reactjs/redux/pull/1887/files#diff-ef2fbfc16613d42f52b86d75581efc9aR27

ccorcos:
hmm. @gcanti don't you think its possible somehow to deduce OwnProps?

ccorcos:
I'm toying around with little success. I was thinking something like this:

```js
  declare type Connect<State,Action,Props> = (
    mapStateToProps: (state: State, ownProps: $Diff<Props,ResultProps>) => StateProps,
    mapDispatchToProps: (dispatch: Dispatch<Action>, ownProps: $Diff<Props,ResultProps>) => DispatchProps,
    mergeProps: (stateProps: StateProps, dispatchProps: DispatchProps, ownProps: $Diff<Props,ResultProps>) => ResultProps
  ) => Connector<$Diff<Props,ResultProps>, Props>
```

ccorcos:
So even when I do this:

```js
const connector : Connector<{color: 'red' | 'blue'}, Props>= connect(mapStateToProps, mapDispatchToProps, mergeProps)

export default connect(Counter)
```

This checks out fine:

```js
ReactDOM.render((
  <Provider store={store}>
    <Counter/>
  </Provider>
), root)
```

gcanti:
@ccorcos please put your code in a repo, it's hard to tell why from a fragment

ccorcos:
@gcanti I put it all here: https://github.com/gcanti/redux/pull/1

flow-typed/redux.js
```diff
@@ -0,0 +1,74 @@
...
+  declare function bindActionCreators<A, B1, B2, B3, B4, B5>(actionCreators: Fn5<A, B1, B2, B3, B4, B5>, dispatch: Dispatch<A>): Fn5<A, B1, B2, B3, B4, B5>;
+  declare function bindActionCreators<A, B1, B2, B3, B4, B5, B6>(actionCreators: Fn6<A, B1, B2, B3, B4, B5, B6>, dispatch: Dispatch<A>): Fn6<A, B1, B2, B3, B4, B5, B6>;
+  declare function bindActionCreators<A, B1, B2, B3, B4, B5, B6, B7>(actionCreators: Fn7<A, B1, B2, B3, B4, B5, B6, B7>, dispatch: Dispatch<A>): Fn7<A, B1, B2, B3, B4, B5, B6, B7>;
+  declare function bindActionCreators<A, B1, B2, B3, B4, B5, B6, B7, B8>(actionCreators: Fn8<A, B1, B2, B3, B4, B5, B6, B7, B8>, dispatch: Dispatch<A>): Fn8<A, B1, B2, B3, B4, B5, B6, B7, B8>;
+  declare function bindActionCreators<A, B1, B2, B3, B4, B5, B6, B7, B8, B9>(actionCreators: Fn9<A, B1, B2, B3, B4, B5, B6, B7, B8, B9>, dispatch: Dispatch<A>): Fn9<A, B1, B2, B3, B4, B5, B6, B7, B8, B9>;
+  // unsafe (all action creator signatures are erased)
+  declare function bindActionCreators<A, K, C: ActionCreators<K, A>>(actionCreators: C, dispatch: Dispatch<A>): ActionCreators<$Keys<C>, A>;
```
aaronjensen: I think you can make this safe by just returning `C`

https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgAiAlgM44CGGAxgBYA8AggFxgDeY2+rAJAMoBXAEZc89UhgBOxAHYBzAHxgAvkoC8YABQVqGYnBmtGASjBqljTLgKNd+mQGFJeKnEml6AaQA0YRuvYwAG0AazwsVk8AXVZNADoEikk5UiNJSQosegoZLAVTcz8VAG5UKAEZOwMwIVkAE1s9AycXDDcPRl8fMAcjKsdnV3cvX38FbX6WodSe3zqyShpaVhJyKjomfNYHdlQwMGcMAUkZMB0mgdb21GVSq3w-frNA0VYAckYHABUASQB5AByb18lCw8AodVYMgEAFshHhJCp9siUaiUahqAYJGB5mslisFusGI17AEKGYlBQMViMGcnhpNMQobD4ZICkpNBxXmAPt9-kCQZlwZCwMQVMZ0JiZNjzvZSM9ajIGpNBm13Jy6RcVHNCUsJahZVi4oaZJo3rRiG9jEA

flow-typed/redux.js
```diff
@@ -0,0 +1,74 @@
...
+  declare function bindActionCreators<A, B1, B2, B3>(actionCreators: Fn3<A, B1, B2, B3>, dispatch: Dispatch<A>): Fn3<A, B1, B2, B3>;
+  declare function bindActionCreators<A, B1, B2, B3, B4>(actionCreators: Fn4<A, B1, B2, B3, B4>, dispatch: Dispatch<A>): Fn4<A, B1, B2, B3, B4>;
+  declare function bindActionCreators<A, B1, B2, B3, B4, B5>(actionCreators: Fn5<A, B1, B2, B3, B4, B5>, dispatch: Dispatch<A>): Fn5<A, B1, B2, B3, B4, B5>;
+  declare function bindActionCreators<A, B1, B2, B3, B4, B5, B6>(actionCreators: Fn6<A, B1, B2, B3, B4, B5, B6>, dispatch: Dispatch<A>): Fn6<A, B1, B2, B3, B4, B5, B6>;
+  declare function bindActionCreators<A, B1, B2, B3, B4, B5, B6, B7>(actionCreators: Fn7<A, B1, B2, B3, B4, B5, B6, B7>, dispatch: Dispatch<A>): Fn7<A, B1, B2, B3, B4, B5, B6, B7>;
+  declare function bindActionCreators<A, B1, B2, B3, B4, B5, B6, B7, B8>(actionCreators: Fn8<A, B1, B2, B3, B4, B5, B6, B7, B8>, dispatch: Dispatch<A>): Fn8<A, B1, B2, B3, B4, B5, B6, B7, B8>;
+  declare function bindActionCreators<A, B1, B2, B3, B4, B5, B6, B7, B8, B9>(actionCreators: Fn9<A, B1, B2, B3, B4, B5, B6, B7, B8, B9>, dispatch: Dispatch<A>): Fn9<A, B1, B2, B3, B4, B5, B6, B7, B8, B9>;
```
aaronjensen: likewise these can probably be simplified to a single clause that accepts an action creator and returns the same type

gcanti: So something like

```js
type ActionCreator<A> = (...args: Array<any>) => A;
type ActionCreators<K, A> = { [key: K]: ActionCreator<A> };

declare function bindActionCreators<A, K, C: ActionCreator<A>>(actionCreators: C, dispatch: Dispatch<A>): C;
declare function bindActionCreators<A, K, C: ActionCreators<K, A>>(actionCreators: C, dispatch: Dispatch<A>): C;
```

right? It's a great improvement, thanks a lot @aaronjensen

aaronjensen: Yes, though I don't know what the purpose of K is, they'll all be `string`, right?

gcanti: Yeah, the `K` type parameter is almost useless. A contrived use case would be to restrict the set of the keys. Say you have all the action creators in a big dictionary

```js
const allActionCreators = {
  a: () => {...},
  b: () => {...},
  c: () => {...}
}

// I just want a and b
const x: { [key: 'a' | 'b']: * } = bindActionCreators(allActionCreators, dispatch)

x.c // <= error
```

Anyway, keeping the `K` type parameter doesn't seem harmful, what do you think?

aaronjensen: Hmm. It looks like it actually errors on the assignment--which is what I would expect. I could see this being handy if it somehow aided in errors around prop types. I'm too new to flow to know if it's a good idea or not. Maybe we could leave it in and loosen if if it is too strict for some real use cases?

flow-typed/redux.js
```diff
@@ -0,0 +1,54 @@
+declare module 'redux' {
```
SimenB: Should this use/replace https://github.com/flowtype/flow-typed/tree/master/definitions/npm/redux_v3.x.x?

aaronjensen: imo yes

ccorcos:
@gcanti @gaearon what are we waiting for to merge this?

timdorr:
I've been following along with this thread (this is our biggest open PR by number of comments), but have been waiting for the dust to settle before approaching it. I'm not a Flow expert (I'm not even a typings expert), so I'm reliant on the comments from others to determine if this should be merged in or not. 

I think I just need a consensus at this point. So, any participants or onlookers want to comment on the suitability of these typings currently? 

I also need remove the todomvc-flow and counter-flow examples, as discussed before, and update todos-flow to use Create React App (if possible?). But those things can happen separately right after the merge process.

gcanti:
So

1. just a single `todos-flow` example
2. should build on top of the last master (Create React App integration)
3. wrt libdefs we need to decide between flow-typed repo or shipping with redux / react-redux

1, 2 seem easy, I'm not sure what to do for 3 though. Once we agree on 3, I can close this PR and open a new one meant to be merged. @gaearon @thejameskyle any advices?

timdorr:
I think I'd lean towards flow-typed only because they have a testing infrastructure in place and building that out here looks like a lot of work, which would be a burden on those contributing who aren't concerned with Flow-related code.

timdorr:
@gcanti Can you `git rebase -i master` and drop the todosmvc-flow and counter-flow commits? I think only cc0b9ae8 might give you some trouble. If you don't want to wrestle with git stuff, let me know. I can take that on after-the-fact once we get things merged in appropriately.

And good job everyone for the comments on and subsequent fixes to the definitions!

gcanti:
@timdorr rebased and squashed (2 commits now):

- libdefs
- todos-flow example

ccorcos:
This is looking great! I'm curious what @thejameskyle thinks about whether these typedefs should be here or in flow-typed.

aaronjensen:
@ccorcos my understanding is that until shadow files are a fleshed out thing, flow-typed is the place they're meant to be.

alexeygolev:
Playing around with the typedefs. Quite amazing. I'm not sure how would I go about typing my own middleware if I want it to be generic. Another thing (which is probably not possible with flow, at least yet) — if my app is structured in this way:
```
module1/
  component.js
  container.js
  reducer.js
  actions.js
  types.js
  index.js
module2/
  ...
```
I can import my actions and have a union type `type Action = ModuleOneAction | ModuleTwoAction` so my actions file is decoupled from the app. But container and reducer are still coupled because the reducer will be called with the app `Action` type and the container uses Dispatch that in its turn uses `ReduxDispatch<Action>` as well as the `Store` type.
I wonder what one would do if his module depends only on its own Actions and Store property.
``

flow-typed/react-redux.js
```diff
@@ -0,0 +1,73 @@
...
+    mapStateToProps: null,
+    mapDispatchToProps: MapDispatchToProps<A, OP, DP>,
+    options?: ConnectOptions
+  ): Connector<OP, DP & OP>;
+
+  declare function connect<S, A, OP, SP, DP>(
+    mapStateToProps: MapStateToProps<S, OP, SP>,
+    mapDispatchToProps: MapDispatchToProps<A, OP, DP>,
+    options?: ConnectOptions
+  ): Connector<OP, SP & DP>;
```
MarcoPolo: Shouldn't this be something like `SP & DP & OP` since that's what the default merge function does?

flow-typed/react-redux.js
```diff
@@ -0,0 +1,73 @@
...
+  };
+
+  declare function connect<A, OP>(
+    options?: ConnectOptions
+  ): Connector<OP, { dispatch: Dispatch<A> } & OP>;
+
+  declare function connect<S, A, OP, SP>(
+    mapStateToProps: MapStateToProps<S, OP, SP>,
+    options?: ConnectOptions
+  ): Connector<OP, SP & { dispatch: Dispatch<A> }>;
```
MarcoPolo: Similar to below shouldn't this also have `& OP` because of the default merge fn?

MarcoPolo: Also could dispatch be optional? You could connect a component whose props don't care about dispatch. With this you'd get an error saying Props doesn't have dispatch in it. Maybe something like this would work? `  ): Connector<OP, SP & OP & ({dispatch: Dispatch<A>} | {})>`

flow-typed/react-redux.js
```diff
@@ -0,0 +1,73 @@
...
+  declare class Provider<S, A> extends React$Component<void, { store: Store<S, A>, children?: any }, void> { }
+
+  declare type ConnectOptions = {
+    pure?: boolean,
+    withRef?: boolean
+  };
+
+  declare function connect<A, OP>(
+    options?: ConnectOptions
+  ): Connector<OP, { dispatch: Dispatch<A> } & OP>;
```
jimbolla: Am I reading this correctly that this is stating there's an overload of `connect` that will accept `options` as the first parameter? If so, that's incorrect.  `options` will only be recognized as the 4th parameter. If you wanted to pass `options` but nothing else. you'd have to do `connect(null, null, null, { withRef: true })`, for example.

flow-typed/react-redux.js
```diff
@@ -0,0 +1,73 @@
...
+    (component: StatelessComponent<P>): ConnectedComponentClass<OP, P, void, void>;
+    <Def, St>(component: Class<React$Component<Def, P, St>>): ConnectedComponentClass<OP, P, Def, St>;
+  };
+
+  declare class Provider<S, A> extends React$Component<void, { store: Store<S, A>, children?: any }, void> { }
+
+  declare type ConnectOptions = {
+    pure?: boolean,
+    withRef?: boolean
+  };
```
jimbolla: react-redux v5 will add several more options for customizing equality checks. I'm not sure the best way to coordinate updating this file with those changes, as they move from beta to release. Would it be better to host the react-redux type specifications inside its repository?

gcanti:
@kasperpeulen @MarcoPolo @jimbolla Thanks for your comments, I'm trying a new approach using the `$Supertype` magic type:

```js
declare function connect<S, A, OP, SP>(
  mapStateToProps: MapStateToProps<S, OP, SP>,
  mapDispatchToProps: Null,
  mergeProps: Null,
  options?: ConnectOptions
): Connector<
    OP, 
    $Supertype< // <= this should allow the `& OP` intersection below and "optional" field declarations in the connected component (@MarcoPolo)
      SP & 
      { dispatch: Dispatch<A> } & 
      OP> // <= this should account for the default merge function (@kasperpeulen and @MarcoPolo)
  >;
```

Also the wrong `connect` overloading should be fixed (@jimbolla comment)

Could you please try / review my last commit?

flow-typed/react-redux.js
aaronjensen: Until https://github.com/facebook/flow/issues/1996 is addressed it's hard to do that afaik. flow-typed seems to be the only reasonable way at the moment.

flow-typed/react-redux.js
```diff
@@ -0,0 +1,86 @@
...
+
+  declare type Null = null | void;
+
+  declare function connect<A, OP>(
+    ...rest: Array<void>
```
aaronjensen: What is this for? Is it necessary to ensure that the other clauses are used if any params are passed?

gcanti: It's a workaround for https://github.com/facebook/flow/issues/2360

aaronjensen: Huh. I had to remove them from redux-saga as they caused issues. I wonder if I introduced more issues... also is a better workaround to reorder them? Maybe they're the same...

jimbolla: I would add a comment about why it's necessary to do this, with a link to the issue.

gcanti: @aaronjensen moving it to the last position doesn't work, you get "Case 4 may work: But if it doesn't, case 6 looks promising too: Please provide additional annotation(s) to determine whether case 4 works"

@jimbolla done

timdorr:
OK, we've got @MarcoPolo, @aaronjensen, and @alexeygolev all saying this looks good. So, it's on you 3 if the community gets out their pitchforks 🔥  😈  

But seriously, thanks for the awesome work on this @gcanti and the help from everyone who commented. Couldn't have done it without you all!

gnoff:
@gcanti what is the best way to consume this today. I have the redux and react-redux libdef from flow-typed installed right now. Are you going to publish there? Or will this be distributed as part of the npm dep on install and we should point flow to it?

aaronjensen:
@gnoff they're pull requested https://github.com/flowtype/flow-typed/pull/318

jcreamer898:
I'm presently trying to grok all of this information. So, as of right NOW, there's a `flow-typed` directory that automatically allows me to do this...

```js
import type { Dispatch as ReduxDispatch } from "redux";
```

BUT, as of https://github.com/flowtype/flow-typed/pull/318, when it gets merged, the `flow-typed` repo will contain the declarations, so they'll at that point be pulled from there?

Loving the work you guys are all doing!

gcanti:
type safe `combineReducers` with the new `$ObjMap` magic type (Flow v0.33+) https://gist.github.com/gcanti/1e4494371c46f7f7ef4fb919778a32e9

aaronjensen:
Awesome. I'd say we should incorporate it.

gcanti:
@aaronjensen in my last version (sent to flow-typed) I also improved `bindActionCreators`

```js
declare type ActionCreator<A, B> = (...args: Array<B>) => A;
declare type ActionCreators<K, A> = { [key: K]: ActionCreator<A, *> };

declare function bindActionCreators<A, C: ActionCreator<A, *>>(actionCreator: C, dispatch: Dispatch<A>): C;
declare function bindActionCreators<A, K, C: ActionCreators<K, A>>(actionCreators: C, dispatch: Dispatch<A>): C;
```

> I'd say we should incorporate it

What about users of Flow 0.32-? As @jcreamer898 pointed out, it's not yet clear to me how to handle these things properly

rsolomon:
@gcanti flow-typed allows different versions for a lib via subdir. For example, redux currently has a `0.30.x` subdir: https://github.com/flowtype/flow-typed/tree/master/definitions/npm/react-redux_v4.x.x

Any libs utilizing `$ObjMap` would presumably live under `0.33.x`

Where did you send your updated definitions to? I can't find any related PRs.

gcanti:
@rsolomon I know, I was talking about redux which  at the moment contains a copy of the definition files

The PR is here https://github.com/flowtype/flow-typed/pull/318

mull:
@gcanti did you ever attempt to use this with react-redux@5.0.x?

gcanti:
@mull no, I didn't

flow-typed/redux.js
faceyspacey: wow this looks amazing. I don't see it in the `react-redux` repository--what's the recommended way to annotate the `export default connect(...` line until then? 

I've found it real funny that there's no idiomatic way to bet the benefit of a typed connected component where you use the resulting react component. There's very little information around about it--this is in fact the only information. What I've started to do is:

```
type HamburgerType = ({ name: string }) => React$Element<*>
export default (connect(null, { toggleSidebar })(Hamburger): HamburgerType)
```

i.e. a typecast but that doesn't seem right. Using the `Connector` type seems right. Does that properly trigger flow validation on usage of the resulting react component? I.e. if it's missing props (`ownProps`) or if they are incorrect?

azaharakis:
@ccorcos did you ever come to a solution to the problem mentioned above:
https://github.com/reactjs/redux/pull/1887#issuecomment-243937316
I tried clicking through to the repo you linked but it is no longer there.

If I understood you correctly Prop validation on a component isn't occurring when wrapped with connect. Particularly when it is getting some of it's props from it's parent. Which is what i'm seeing. 

In my case my connect is partially applied so it can take different react component's. Which I can't see a good example for in this PR

AndrewSouthpaw:
I'm having the same problem as @azaharakis, did anyone come to a solution? It's worrisome that I lose prop validation as soon as I connect my component. 🙁

rsolomon:
@AndrewSouthpaw I've been using [this](https://gist.github.com/rsolomon/397292b51fff7a66f8ff23a4afa91a60) for class-based components and it maintains type safety as expected. I've been meaning to submit it for PR to `flow-typed`, but it needs to be cleaned up to support stateless components.

It is largely lifted from the existing definitions there, but it provides much more sound type safety for me. Hopefully it can help you out.

AndrewSouthpaw:
Wow, that's some bonafide magic right there @rsolomon, it works like a charm (I don't use stateless components), thanks for sharing. Did you come up with that yourself?

AndrewSouthpaw:
Wow, that's some bonafide magic right there @rsolomon, it works like a charm (I don't use stateless components), thanks for sharing. Did you come up with that yourself? I'm trying to get better at flow typing and maybe contribute to the library, but I get so lost when looking at a file like that one. What was your key change(s)?

rsolomon:
Nah I didn't completely come up with it. It's largely based on [this definition](https://github.com/flowtype/flow-typed/blob/master/definitions/npm/react-redux_v5.x.x/flow_v0.30.x-/react-redux_v5.x.x.js), but as written that file provided me no type safety.

I had used similar strategies for typing `CSSModules` and other HoCs, so I figured it could probably be implemented here. Seemed to work like a charm. Glad you found it useful :)

rsolomon:
> What was your key change(s)?

In particular, avoiding catch-all types like `Object` and using `$Diff` were the keys to the magic.

PR #1888, opened 2016-08-10
Update todos example

rtryoda:
Store could be read-only.


timdorr:
`const` doesn't make the store read-only, just the instance variable for it.

PR #1895, opened 2016-08-12
Throw error when dispatching boundAction or dispatch

CoryDanielson:


gaearon:
Looks reasonable to me!

src/createStore.js
```diff
@@ -148,6 +148,21 @@ export default function createStore(reducer, preloadedState, enhancer) {
    */
   function dispatch(action) {
     if (!isPlainObject(action)) {
+      if (action) {
+        if (action.__boundContext__) {
+          throw new Error(
+            'You have called dispatch with a boundAction. ' +
+            'Bound actionCreators are arleady dispatched when called.'
```
BerkeleyTrue: found a typo: already not arleady

test/createStore.spec.js
```diff
@@ -420,6 +422,30 @@ describe('createStore', () => {
...
+      foo: (state = 0) => state,
+      bar: (state = 0) => state
+    }))
+    const actionCreator = actionCreators.addTodo
+    const boundActionCreator = bindActionCreators(actionCreator, store.dispatch)
+    expect(() => {
+      store.dispatch(boundActionCreator)
+    }).toThrow(
+      'You have called dispatch with a boundAction. ' +
+      'Bound actionCreators are arleady dispatched when called.'
```
BerkeleyTrue: same typo as above

src/createStore.js
```diff
@@ -148,6 +148,21 @@ export default function createStore(reducer, preloadedState, enhancer) {
    */
   function dispatch(action) {
     if (!isPlainObject(action)) {
+      if (action) {
+        if (action.__boundContext__) {
+          throw new Error(
+            'You have called dispatch with a boundAction. ' +
```
BerkeleyTrue: Should this be `boundActionCreator` or `Bound actionCreator`?

src/bindActionCreators.js
```diff
@@ -1,5 +1,7 @@
 function bindActionCreator(actionCreator, dispatch) {
-  return (...args) => dispatch(actionCreator(...args))
+  const boundActionCreator = (...args) => dispatch(actionCreator(...args))
+  boundActionCreator.__boundContext__ = dispatch
```
Aweary: Is there a reason we're referencing `dispatch` instead of just setting a `__bound__` property to `true`?

src/createStore.js
CoryDanielson: Good catches. I've updated them to use "bound action creator" which seems to match the docs and another string in the unit tests. 
```
it('wraps the action creators with the dispatch function', () => {
```

src/bindActionCreators.js
CoryDanielson: Originally, I was checking to see if the `__boundContext__ === dispatch` but I think I left out the right-hand side of that check while moving things around.

I've updated it to use `__isBoundActionCreator__` instead, because `__boundContext__` was a misnomer, since there isn't actually a bound context on the function via `Function.prototype.bind`.

CoryDanielson:
I've been run into this problem a few times today after some furious refactoring of action creator code. A connected components is unknowingly dispatching bound actions which has been overlooked behind a trail of imports. The current error `Actions must be plain objects. Use custom middleware for async actions.` is technically correct, but not explicit to my particular problem. I debug for a while only to discover that the action that I'm importing is now "bound", so my code is is essentially:
```
store.dispatch(store.dispatch(ActionCreator()))
```
The first dispatch call (the inner one) receives an action and my reducer state is updated as expected. The second dispatch call receives a function (dispatch!) which throws an error because it's not a plain object... but the state has updated as expected?! Bewildered, I track down the bug with no side-effect, until I realize that I'm dispatching an actionCreator that is now bound. Oops. This helps with that 😄

hmillison:
we've recently run into this problem again. any updates on getting this merged?

timdorr:
Looks like this needs to be updated. @CoryDanielson If you want to update it, I can reopen and we can look into getting it merged in. Thanks!

PR #1897, opened 2016-08-15
small wording fix

da-vaibhav:
from ”headlines for a select subreddit” to: => ”headlines for a selected subreddit”.


markerikson:
Thanks!

Issue #1784, opened 2016-05-30
Add a "Structuring Reducers" recipe


markerikson:
We really, _really_ need a page talking about approaches, guidelines, and suggestions for organizing reducer logic.  I've been saying I want to write this for a while, but have been too busy so far.  I'm still up for it, but further suggestions and/or offers of help would be appreciated.

**WIP page**: https://github.com/markerikson/redux/blob/structuring-reducers-page/docs/recipes/StructuringReducers.md

Initial sketch of possible topics:

- [x] Understanding that you really only have _one_ reducer function, which is just subdivided for maintainability and simplicity
- [x] Understanding "reducer composition" and that you can break things down using plain ol' functions
- [x] The basic rules of reducers: `(state, action) -> newState`, immutable updates, and one other item I've said before but I'm not remembering at the moment
- [x] Emphasis that `combineReducers` is simply a utility function for the common use case of delegating update logic based on state domain organization, and not a requirement
- [x] That defining object keys when using `createReducers` is, effectively, defining the name/shape of your state (not always clear when using ES6 object literal shorthand - naming of imported reducer functions matters!)
- [x] That "actions are passed to all reducers" _only_ if you're using `combineReducers`
- [x] Passing different chunks or all of the state to a sub-reducer based on need for that action
- [x] Initializing state
- [x] Normalizing data
- [x] Updating normalized data
- [x] Updating data immutably in general, particularly nested data and arrays, and how just using a variable assignment doesn't mean you've "made a copy"
- [x] Reusing logic and creating reducers that can be "targeted"

- A whole bunch of other stuff that has crossed my mind at various points and that I'm not thinking of right now, but hopefully will remember later

Random related links:

- https://github.com/reactjs/redux/issues/601
- https://github.com/reactjs/redux/issues/644
- https://github.com/reactjs/redux/issues/749
- https://github.com/reactjs/redux/issues/822
- https://github.com/reactjs/redux/issues/857
- https://github.com/reactjs/redux/issues/1014
- https://github.com/reactjs/redux/issues/1018
- https://github.com/reactjs/redux/issues/1171
- https://github.com/reactjs/redux/issues/1617
- https://github.com/reactjs/redux/issues/1628
- https://github.com/reactjs/redux/issues/1653
- https://github.com/reactjs/redux/issues/1726
- https://github.com/reactjs/redux/issues/1602#issuecomment-208987442
- https://github.com/reactjs/redux/issues/1824
[...]

PR #1904, opened 2016-08-21
[Proposal] Make `combineReducers` pass extra arguments to child reducers

aikoven:
This can be useful a couple cases:

## 1. Reducers with dependencies

In my opinion the most explicit and convenient way to have one branch of reducers dependent on another branch is to make them have common parent that calculates dependency branch first and then passes its to the dependent reducer:

```js
const dependencyReducer = (state = ..., action) => {...};

const dependentReducer = (state = ..., action, dependency) => {...};

const parentReducer = (state = {}, action) => {
  const {dependency, dependent} = state;

  const nextDependency = dependencyReducer(state, action);
  const nextDependent = dependentReducer(state, action, nextDependency);

  if (nextDependency !== dependency || nextDependent !== dependent)
    return {dependency: nextDependency, dependent: nextDependent};

  return state;
};
```

The change in this PR allows `dependentReducer` to be built from smaller parts, each one accepting dependency.

## 2. Multiple branches of same shape

Suppose we are developing an instant messenger. We can switch between chat views for different contacts and we want to store some data for each view, e.g. entered text, scroll position etc.

Let's structure state like this:

```js
{
  alice: {
    enteredText: '...',
    scrollPosition: ...,
  },
  bob: {
    enteredText: '...',
    scrollPosition: ...,
  },
  ...
}
```

Actions:

```js
const chatOpened = (contact) => ({
  type: 'CHAT_OPENED',
  contact,
});

const textEntered = (contact, text) => ({
  type: 'TEXT_ENTERED',
  contact,
  text,
});
```

My first approach to this was to have parent reducer look into action's `contact` prop and pass action to `chatViewReducer` for particular branch. But that couples parent to child, parent has to know of all actions that child accepts etc.

So my current approach is to pass all actions to `chatViewReducer` for all child branches with `contact` passed as third argument:

```js
const chatsReducer = (state = {}, action) => {
  if (action.type === 'CHAT_OPENED') {
    return {
      ...state, 
      [action.contact]: chatViewReducer(undefined, action, action.contact),
    };
  }

  const nextState = {};
  let hasChanged = false;

  for (let contact of Object.keys(state)) {
    nextState[contact] = chatViewReducer(state[contact], action, contact);
    hasChanged = hasChanged || nextState[contact]  !== state[contact];
  }

  return hasChanged ? nextState : state;
};
```

Once again, this change allows `chatViewReducers` to be built from smaller parts each accepting `contact` as third argument, allowing them to filter incoming actions by `contact` to see if they correspond to that particular chat view.

Right now I have a custom `combineReducers` version that is almost the same as original.

Also, the above pattern can be generalised as reducer factory:

```js
/**
 * @param childReducer reducer for state branches
 * @param shouldAddKey function that accepts action and returns undefined or 
 *   string key to be added to state
 * @param shouldRemoveKey function that accepts action and returns undefined or
 *   string key to be removed from state
 */
function createMappingReducer(childReducer, shouldAddKey, shouldRemoveKey) {
  ...
}

const chatsReducer = createMappingReducer(chatViewReducer, (action) => {
  if (action.type === 'CHAT_OPENED')
    return action.contact;
});
```

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -126,7 +126,7 @@ export default function combineReducers(reducers) {
     sanityError = e
   }
 
-  return function combination(state = {}, action) {
+  return function combination(state = {}, action, ...args) {
     if (sanityError) {
       throw sanityError
     }
@@ -144,7 +144,7 @@ export default function combineReducers(reducers) {
       var key = finalReducerKeys[i]
       var reducer = finalReducers[key]
       var previousStateForKey = state[key]
-      var nextStateForKey = reducer(previousStateForKey, action)
+      var nextStateForKey = reducer(previousStateForKey, action, ...args)
       if (typeof nextStateForKey === 'undefined') {
         var errorMessage = getUndefinedStateErrorMessage(key, action)
         throw new Error(errorMessage)


markerikson:
This has been proposed _numerous_ times, and was generally rejected.  The last time this was proposed was #1768 , which appears to have stalled due to questions about what exactly the new behavior should involve.  

Remember that `combineReducers` is simply a utility function for the most common use case, and is not intended to handle all situations.  If you need more specific behavior, you can always write your own variation on the concept.

You may also be interested in my WIP addition to the Redux docs on the topic of [Structuring Reducers (#1784)](https://github.com/reactjs/redux/issues/1784), which gives some further suggestions for ways to handle things.

Given that #1768 is already around, I'm going to close this for now.  If @gaearon happens to think that further changes to `combineReducers` are warranted, this can be reopened.

aikoven:
OK, no probs!

This is a bit different from #1768, because it doesn't add any semantics or implicit logic. Whatever you pass as third argument is up to you, it just allows to implement many useful patterns without requiring users to use custom `combineReducers` function.

PR #1911, opened 2016-08-25
Fix Generating Reducers example in Reducing Boilerplate doc

ValentinH:
I'm quite new with Redux and ES6, but I think that the current example for the "Generating Reducers" helper usage is not correct.
Here's my suggested change.


SimenB:
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_2015

This is a combination of "Shorthand method names" and "Computed property names "

ValentinH:
OK my bad, I'm closing it. I'll try to understand it but it's quite complex!

PR #1926, opened 2016-09-02
Add fetch import to async creators testing example

### LINKED ISSUE #1925: Testing Async Action Creators: 'fetch is not defined'
Hi,

I've followed closely the [guide on testing async action creators](http://redux.js.org/docs/recipes/WritingTests.html#async-action-creators). Unfortunately, I get the error 'fetch is not defined'. I know that this is not really a bug or a feature request but there seem to be other people which had the same problem. Are there any requirements besides what is noted in the docs? Like Node 6? Is Mocha a requirement or rather a recommendation?

I use the latest Redux version (3.5.2) with Jasmine 2.5.0 and Jest 15.1.1 as the task runner on Node 4.4.7.

Thank you!

P.S.: I hope that this issue doesn't fall under the Should-go-to-Stackoverflow category.

timkraut:
Fixes #1925

### DIFF
--- a/docs/recipes/WritingTests.md
+++ b/docs/recipes/WritingTests.md
@@ -81,6 +81,8 @@ For async action creators using [Redux Thunk](https://github.com/gaearon/redux-t
 #### Example
 
 ```js
+import fetch from 'isomorphic-fetch';
+
 function fetchTodosRequest() {
   return {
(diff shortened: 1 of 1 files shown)


timdorr:
Doesn't hurt. Thanks!

PR #1928, opened 2016-09-03
Don't lint examples

gaearon:
I broke linting of examples in #1883 because they are now linted via CRA, and it has incompatible configuration with the ancient version of ESLint we’re using. Our lint setup is a mess right now, with `rackt` config being used which is not maintained anymore. For now, I’ll just disable linting of examples altogether to try to get Travis to pass. In the future we should pick a lint config and update ESLint to a modern version.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,5 @@
 **/dist/**
+**/examples/**
 **/node_modules/**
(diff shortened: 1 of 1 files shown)


gaearon:
Yay passing!

PR #1931, opened 2016-09-06
Update README.md

Dzhakhar:
Code optimization. Instead of looping call function by key. In this structure you can store many actions and it's more readable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,15 +89,14 @@ import { createStore } from 'redux'
  * follows a different convention (such as function maps) if it makes sense for your
  * project.
  */
+ 
+actionTypes = {
+  "INCREMENT": function(state){ return state + 1 },
+  "DECREMENT": function(state){ return state - 1}
+}
+ 
 function counter(state = 0, action) {
-  switch (action.type) {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,15 +89,14 @@ import { createStore } from 'redux'
...
 function counter(state = 0, action) {
-  switch (action.type) {
-  case 'INCREMENT':
-    return state + 1
-  case 'DECREMENT':
-    return state - 1
-  default:
-    return state
-  }
+  return actionTypes(action.type)(state);
```
andreieftimie: You are missing the default branch where the reducer doesn't match any action type and returns the original state.

There's also a typo here. You want square brackets: `actionTypes[action.type]`.

But ultimately this makes the code harder to understand, and since this is an example that should be the first priority.

markerikson:
Thanks for the suggestion.  However, the lookup table approach is explicitly described already in the docs, at http://redux.js.org/docs/recipes/ReducingBoilerplate.html .  Also, since this is intended as a trivial example in the project README, there's no reason to try to make it shorter.

PR #1933, opened 2016-09-06
Update ServerRendering.md

chanhokim:
dist/bundle.js -> static/bundle.js
the HTML portion says it's in "static" directory.


timdorr:
Thanks!

PR #1936, opened 2016-09-07
Fix typings for `compose` function to accept single rest argument

### LINKED ISSUE #1935: Typescript: compose function should be able to accept single rest argument
As for now, it is not possible to call `Resux.compose` like this:
```ts
declare const middlewares: Redux.Middleware[];
declare const enhancers: Redux.GenericStoreEnhancer[];
const enhancer = Redux.compose(Redux.applyMiddleware(...middlewares), ...enhancers);
```
This is broken since https://github.com/reactjs/redux/pull/1868 was merged.
I think we should have an overloading for `compose`, which covers this case.
/cc @aikoven

aikoven:
Fixes #1935.

Cc. @ulfryk @Igorbek

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -419,7 +419,8 @@ export function compose<A, B, C, T1, T2, T3, R>(
 ): Func3<T1, T2, T3, R>;
 
 /* rest */
-export function compose<A, B, C, R>(
-  f1: (b: C) => R, f2: (a: B) => C, f3: (a: A) => B,
-  ...funcs: Function[]
+export function compose<R>(
+  f1: (b: any) => R, ...funcs: Function[]
 ): (...args: any[]) => R;
+
+export function compose<R>(...funcs: Function[]): (...args: any[]) => R;
--- a/test/typescript/compose.ts
+++ b/test/typescript/compose.ts
@@ -33,3 +33,7 @@ const t10: string = compose(numberToString, stringToNumber,
 
 const t11: number = compose(stringToNumber, numberToString, stringToNumber,
   multiArgFn)('bar', 42, true);
+
+
+const funcs = [stringToNumber, numberToString, stringToNumber];
+const t12 = compose(...funcs)('bar', 42, true);


timdorr:
Thanks!

stephtr:
Without this commit the following code works:
`const finalCreateStore = compose(
  applyMiddleware(m1),
  createDevTools(null).instrument()
)(createStore)

const store = finalCreateStore(reducer)`

Since the commit TypeScript now automatically selects `function compose<{}>(...funcs: Function[]): (...args: any[]) => {};` instead of `function compose<A, T1, R>( f1: (b: A) => R, f2: Func1<T1, A> ): Func1<T1, R>;`, which results in _Type '{}' has no compatible call signatures._ But shouldn't TypeScript select the first appropriate overloaded function?

aikoven:
@stephtr Filed an issue in TypeScript repo: https://github.com/Microsoft/TypeScript/issues/11343

stephtr:
Would it maybe be wise to revert this commit until the TypeScript issue is somehow solved? Because code written with the definition of Redux 3.6.0 (like the one above) needs to be changed to `compose<Redux.StoreEnhancerStoreCreator<...>>(`. However this line does not work with the 3.6.0 version of index.d.ts.

aikoven:
@stephtr How about `compose(...) as Redux.StoreEnhancerStoreCreator<...>`?

stephtr:
In that case it has to be `compose as Redux.StoreEnhancer<...>`. I just tested it, works in 3.6.0 **and** master (even though using no typecast was nicer).

Igorbek:
I just wanted to make a note what a technical challenge is to properly type `compose` or, more importantly, a store enhancing mechanism in TypeScript.
In general, an enhancer is just a function which may have it's own assumptions of what is the store and what will be a new enhanced store it would construct. I'll list just a very obvious aspects:
- an enhancer can expect a specific state type (or its specific capabilities) or work with any generic type
- an enhancer can modify the dispatcher's signature (add a new overload or even completely substitute the signature) - it's not just additive (`&` type operator cannot be used in general)
  - a new signature can be constructed in very generic way by using type combinators

I was thinking about this and tried to prototype type flowing (state type, dispatch type) through enhancers composition and came up with a suggestion for the TypeScript team which seems very complicated even for just starting discussion :) I described the challenge and possible solution here Microsoft/TypeScript#10247 please come there and bug TS team to pick it up for a design meeting.

PR #1942, opened 2016-09-12
Fat arrow functions in the counter example

ChrisJamesC:
With this PR I would like to propose to add more ES6-like syntax in the examples. 

This is an opinionated subject since we might want to stick with some more ES5-like syntax in order for new comers to not be intimidated by the syntax on top of redux. 

Please let me know what you think.

### DIFF
--- a/examples/counter/src/components/Counter.js
+++ b/examples/counter/src/components/Counter.js
@@ -1,19 +1,14 @@
 import React, { Component, PropTypes } from 'react'
 
 class Counter extends Component {
-  constructor(props) {
-    super(props)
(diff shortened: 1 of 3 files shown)


timdorr:
This fits in well with what was leftover from #1800. Thanks!

ChrisJamesC:
Thanks @timdorr for the quick review and for the reference to the issue!

alexeyraspopov:
What's the purpose of using arrow function instead of class method? It seems to be noisy.

timdorr:
@alexeyraspopov It auto binds the function. It removes the need to do manual binding like the example had before in the constructor.

PR #1943, opened 2016-09-12
Suggest Jest as default

cpojer:
The examples now use CRA so it makes sense to also recommend Jest in the docs. 

See https://twitter.com/flashpunk/status/775350612879749121


gaearon:
Let’s also use built-in `expect` in examples? It is currently imported from `expect` library.

timdorr:
All done!

PR #1945, opened 2016-09-12
Use more ES6 syntaxes in the async example

ChrisJamesC:
Convert the async example files to using fat arrows and stateless components.

This is a follow up to what I started in the counter example in #1942. 

I’ll be ok if we decide to stick to the old notation on this example since the fat arrows don’t add any feature in itself, except reduce the number of lines in files.

References #1800


timdorr:
Thank you so much!

PR #1947, opened 2016-09-12
Use more ES6 syntaxes in the shopping cart example

ChrisJamesC:
Convert the shopping cart example files to using fat arrows, spread operators and stateless components.

References  #1800

PR #1950, opened 2016-09-13
Use more ES6 syntaxes in the todomvc example

ChrisJamesC:
Convert the shopping cart example files to using fat arrows, spread operators, static class members and stateless components.

References #1800


ChrisJamesC:
I'll wait for this PR, #1950 and #1947 to merge before finishing all the other examples. 

Don't hesitate to let me know if there is anything to change or if it's going too far in your opinion.

ellbee:
Thank you!

Issue #1944, opened 2016-09-12
Convert test suite to Jest


timdorr:
Since we now [recommend using Jest](https://github.com/reactjs/redux/pull/1943), we should practice what we preach 😄 

Jest 15 introduces some major improvements and makes the TDD process a heck of a lot easier and more fun. It's also faster and and has more helpful error messages. It's really awesome and there's good reason for its recommendation in the Writing Tests docs.

This is a request for a PR from the community. If you want to help out with open source development, this is a great chance to do so. Just remove `mocha` (the current test runner) and `expect` (Jest provides an `expect` function built-in), and switch out the test suite to run through Jest instead. There are hopefully only minimal changes needed to the actual test code, so this should mostly be just simple infrastructure work.

It *should* be good for beginners, but let us know if anything is particularly challenging. We can pass that along to the Jest team, who is always looking for ways to make the tool as friendly and fun as possible.

jimbolla:
I was going to suggest the same thing

HeinrichFilter:
I am keen to give it a try.

It will be my first contribution to a popular project. Please feel free to point out mistakes or anything that can be done better/differently.

timdorr:
Thanks @HeinrichFilter!

PR #1951, opened 2016-09-13
#1944 converting test suite from Mocha to Jest

HeinrichFilter:
Just wanted to get something going where the tests are passing on my machine.

The replacement for `expect.spyOn()` is not ideal. It is quite verbose and the linter doesn't like that `console` is accessed. It might make sense to use https://github.com/facebook/jest/issues/1592#issuecomment-244577592 when it becomes available.

Please let me know what else I can do.


docs/recipes/WritingTests.md
```diff
@@ -304,7 +304,7 @@ import Header from '../../components/Header'
 
 function setup() {
   const props = {
-    addTodo: expect.createSpy()
+    addTodo: jest.fn(() => {})
```
Florian-R: Could be wrong but I think 

```
addTodo: jest.fn()
```

is sufficient for a noop.

HeinrichFilter: @Florian-R indeed, according to the [documentation](https://facebook.github.io/jest/docs/api.html#jest-fn-implementation) that should work.

But weirdly, when I run jest with a no-args [`jest.fn()` in applyMiddleware.spec.js](https://github.com/reactjs/redux/pull/1951/commits/927bf45f1c7666e42ae8c4c390cc726a75cab903#diff-0c67d0d5b7357141efeb1d7e87115773R15) I get the following error which, to me, seems totally unrelated:
```
 FAIL  test/applyMiddleware.spec.js
  ● Test suite failed to run

    /Users/heinrichfilter/dev/src/redux/test/applyMiddleware.spec.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import { createStore, applyMiddleware } from '../src/index';
                                                                                             ^^^^^^
    SyntaxError: Unexpected token import

      at transformAndBuildScript (../../../.nvm/v6.5.0/lib/node_modules/jest/node_modules/jest-runtime/build/transform.js:284:10)
```

I'm a bit confused by it and thought I'd leave it in until someone can shed some light on it or I figure it out. Can someone maybe run the tests and see if they get the same error?

Florian-R: Yeah I give this issue a quick try too and noticed this kind of errors (my branch is roughly the same than yours). Didn't had much time to dig further though.

Florian-R: @HeinrichFilter Played with your branch a bit, you might want to change your package.json scripts to 

```
"test": "cross-env BABEL_ENV=commonjs jest",
```
otherwise you wont use the `transform-es2015-modules-commonjs` plugin.

Also, make sure to clear Jest cache on first launch, e.g. 

```
npm run test -- --no-cache
```

Having just `jest.fn()` seems OK with this setup.

HeinrichFilter:
@Florian-R Thanks for the suggestions. I made the changes and updated the PR

package.json
```diff
@@ -105,6 +106,7 @@
     "gitbook-cli": "^2.3.0",
     "glob": "^6.0.4",
     "isparta": "^4.0.0",
+    "jest": "^15.1.1",
     "mocha": "^2.2.5",
```
ellbee: I guess we can remove mocha, expect and isparta from the package.json now.

HeinrichFilter: Thanks @ellbee 

I removed the dependencies from package.json and updated `test:cov` to use jest

test/typescript.spec.js
```diff
@@ -2,7 +2,7 @@ import * as tt from 'typescript-definition-tester'
 
 
 describe('TypeScript definitions', function () {
-  this.timeout(0)
+  //this.timeout(0)
```
gaearon: Can this just be removed?

HeinrichFilter: According to the [Mocha homepage](https://mochajs.org/):
> Use `this.timeout(0)` to disable timeouts all together

On Travis (and my machine) the new Jest test passes without hitting the default timeout. Thus, in my opinion, the line should be removed.

But if I'm missing anything, let me know.

gaearon: Let's kill it then

gaearon:
Thanks!

PR #1957, opened 2016-09-16
Update .eslintrc

ibbatta:
typo

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,5 @@
 {
-  "extends": "rackt",
+  "extends": "react",
   "rules": {
     "valid-jsdoc": 2,
     // Disable until Flow supports let and const


alexeyraspopov:
It's [not](https://github.com/reactjs/redux/blob/master/package.json#L102) a typo.

ellbee:
Thank you for taking the trouble to open a PR. This is not a typo, but it is understandable you might think so. The rackt lint rules were introduced when Redux joined the now defunct Rackt org (the idea was that all the projects in the org should have consistent stylings). There are plans to change the rules to something else, possibly `airbnb` or `eslint-plugin-react:recommended`

ibbatta:
My fault, I miunderstood.
I had linter problems with "rackt" and I supposed wrong

PR #1961, opened 2016-09-17
Fixes wrong file name in docs

Sparragus:
The guide is making reference to `containers/VisibleTodoList.js` but it currently says `components/App.js`. This change renames the section to the correct filename.


ellbee:
Thank you!

PR #1963, opened 2016-09-18
Added new features to Redux Counter example

bndrml:


timdorr:
I don't think we really need this. The important part of the counter example is it's simplicity so you can focus on Redux, rather than the features of the app.

PR #1969, opened 2016-09-19
Provides "how"

quavmo:
http://ell.stackexchange.com/questions/49448/using-that-or-how-after-describe/49477#49477


timdorr:
Thanks!

PR #1977, opened 2016-09-21
Add isomorphic-fetch and lodash

mdulac:


mdulac:
The real-world project does not works due to a lack of those two librairies : isomorphic-fetch and lodash. I had to execute extra "npm install" for these two librairies to make this project work

gaearon:
Interesting. I wonder why Travis doesn't fail on master. Can you take a look at previous build logs?

timdorr:
Might be caching on Travis's end. Maybe try busting the cache and rebuilding?

gaearon:
Thanks. I’ll remove `isomorphic-fetch` separately because Create React App already includes a polyfill for it.

PR #1991, opened 2016-09-26
Typo fix

vasanthk:
NIT: Fixed to use proper spelling for "trade-offs".
I was too anxious to read through this one as it was just merged. Great work!
Just noticed this super minor typo. Feel free to disregard if it's too much of a NIT :)


markerikson:
A quick search says both uses are valid-ish, but sure.  Thanks!

PR #2001, opened 2016-09-30
Add missing [enhancers] argument to API reference

mxstbr:


docs/api/README.md
```diff
@@ -6,7 +6,7 @@ This section documents the complete Redux API. Keep in mind that Redux is only c
 
 ### Top-Level Exports
 
-* [createStore(reducer, [preloadedState])](createStore.md)
+* [createStore(reducer, [preloadedState], [enhancers])](createStore.md)
```
Florian-R: Should be plural no?

gaearon: I think it should be `[enhancer]`. You can’t pass an array, you’d have to `compose` them into a single enhancer.

PR #2005, opened 2016-10-03
In Flow types, make Reducer's state parameter optional

cameron-martin:
According to the [docs](http://redux.js.org/docs/basics/Reducers.html#handling-actions), the reducer initially gets called with the state parameter as undefined. This means this type signature is incorrect, since the state parameter is optional.

### DIFF
--- a/flow-typed/redux.js
+++ b/flow-typed/redux.js
@@ -22,7 +22,7 @@ declare module 'redux' {
     replaceReducer(nextReducer: Reducer<S, A>): void
   };
 
-  declare type Reducer<S, A> = (state: S, action: A) => S;
+  declare type Reducer<S, A> = (state?: S, action: A) => S;
 
   declare type Middleware<S, A> =
     (api: MiddlewareAPI<S, A>) =>


timdorr:
How does TS handle default parameters? Yes, it does pass through undefined, but it's expected that you handle that case specifically. Sorry for being pedantic, I just want to make sure this is typed correctly.

timdorr:
We also should update [the corresponding flow types](https://github.com/reactjs/redux/blob/master/flow-typed/redux.js#L25)

cameron-martin:
This PR *is* for the flow types. The TS types might have the same problem when TS is configured to have non nullable types.

timdorr:
Nevermind this dummy... :smile:

timdorr:
Of course, the question still remains for the equivalent Typescript types: https://github.com/reactjs/redux/blob/master/index.d.ts#L46

aikoven:
@cameron-martin You are right, TypeScript definition should be updated too. 

However, in TS, it is an error if optional argument is followed by required one.
I also think that it's too early to have `state: S | undefined`, because it will break builds for TS < 2.0.

I still don't know solution for this. Maybe we should migrate definitions to 2.0 but keep old definitions in a separate repo to use with [Typings](https://github.com/typings/typings).

gcanti:
@cameron-martin Flow understands ES6 default arguments syntax

```js
function f(n: number = 1): void {}
f() // <= ok
function g(n: number): void {}
g() // <= error
```

timdorr:
OK, sounds like this isn't needed. The optional-ness of the arg is defined in the code rather than the type definition.

cameron-martin:
@gcanti Sorry, I don't understand your point, can you elaborate?

@timdorr in the example gcanti gave, the type of f is `(n?: number) => void`, which is what I'm proposing changing the type signature to.

gcanti:
> can you elaborate?

@cameron-martin yes, sorry, I've been too terse.

There are 2 cases:

- 1) using `createStore(reducer, preloadedState)`
- 2) using `createStore(reducer)`

In the first case `state` is always valued, hence the signature `(s: S, a: A) => S` is ok.

In the second case `s` will be `undefined` so the signature could be `(s?: S, a: A) => S`, but AFAIK the idiomatic way to handle this case in redux is to set a default. Now Flow understands ES6 default arguments syntax so it allows for

```js
(s: S = initialState, a: A) => S
```

instead of

```js
(s?: S, a: A) => S
```

hence, again, we can keep the former.

There are 2 problems though:

- P1) if you use 2), we keep `(s: S, a: A) => S` and the user forgets to set a default, likely you get a runtime error
- P2) on the other hand, if we change to `(s?: S, a: A) => S` (safer) users which use 1) will be in trouble since they don't know how to return a valid state

```js
function reducer(s?: S, a: A): S {
  if (typeof s !== 'undefined') {
    return s
  }
  return ??? // <= how can I return a sensible state here?
}
```

Both situations are not ideal, but P2) seems a no go. 

TL;DR I agree with you that `(s?: S, a: A) => S` will be safer but then I don't know how to solve P2

dignifiedquire:
Sorry for commenting on this old issue, but this is currently making it very hard for me to test the default state of a reducer, as if do `reducer(undefined)` in my tests flow will complain that `undefined` is not a valid type for `State`. So I effectively have to either not test that or add flow escape hatches around it

PR #2006, opened 2016-10-03
Docs gh 1

GaryH21:
Updated to say "if you're not" instead of "if you don't" as the previous line due to the "are using" in the previous line.


timdorr:
Thanks!

PR #2007, opened 2016-10-03
Simplify Promise nesting

ForbesLindesay:
If you need to keep past promise results in scope, you can just nest.   There's no need for creating and de-structuring objects.  This makes the code look more complex than it would be in callbacks, when in reality the pyramid approach of callbacks works just as well if you accept small amounts of promise nesting.

Promises are about fixing error propagation and zalgo, not fixing the pyramid of doom.  The pyramid of doom is almost never a real/significant problem, so we should just accept it existing here.

### DIFF
--- a/examples/real-world/src/middleware/api.js
+++ b/examples/real-world/src/middleware/api.js
@@ -25,20 +25,20 @@ const callApi = (endpoint, schema) => {
 
   return fetch(fullUrl)
     .then(response =>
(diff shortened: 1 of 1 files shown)


ForbesLindesay:
The diff is a lot clearer if you ignore whitespace: https://github.com/reactjs/redux/pull/2007/files?w=1

timdorr:
Thanks!

Issue #1972, opened 2016-09-20
Can not test async action base on redux.org tutorial


craigcosmo:
I just wonder is there demo code of the async action test on redux.org?

I tried to remake the tutorial, here is my demo: https://github.com/craigcosmo/react-redux-test

action.js

``` js
import axios from 'axios'

export function getGoodDataStart(){
    return{
        type: "GOOD_DATA_START"
    }
}
export function getGoodDataSuccess(payload){
    console.log('success', payload)
    return {
        type: "GOOD_DATA_SUCCESS",
        payload: payload
    }
}
export function getGoodDataFail(){
    return{
        type: "GOOD_DATA_FAIL"
    }
}
export function getGoodData(){
    return (dispatch) => {
        dispatch( getGoodDataStart() )
        return  axios.get('http://www.google.com/list')
            .then( response => {
                console.log('fake res',response)
                dispatch(getGoodDataSuccess (response) )
            })
            .catch( err => {
                console.log('fake err',err)
            })
    }   
}
```

register.test.js

``` js
import nock from 'nock'
import React from 'react'
import {expect} from 'chai'
import {getGoodData} from 'registerAction'
import configureMockStore from 'redux-mock-store'
import thunk from 'redux-thunk'

const middlewares = [ thunk ]
const mockStore = configureMockStore(middlewares)

describe('Register component', () => {

    it('async action', function () {

        nock('http://www.google.com')
        .get('/list')
        .reply(200,'ok!' )

        const store = mockStore({ 
            myData: '' ,
        })
        const expected = [
            {type: "GOOD_DATA_START"},
            {type: "GOOD_DATA_SUCCESS", payload: 'ok!'}
        ]

        return store.dispatch(getGoodData())
            .then( () => { 
                expect(store.getActions()).to.equal(expected)
            })
    })
})
```

The problem is, nock does not block the request as we think it would. Therefore, action actually makes the request to remote server, therefore, the result of the request is not the same as `expectedResult`

timdorr:
This is an issue with nock or axis, and not Redux. You'll have better luck reaching out on Stack Overflow or those projects' appropriate support channels.

craigcosmo:
but it's redux that put together this example right? My code is a copy the redux async action tutorial. You guys have been there so I think you know best what might have went wrong.

gaearon:
Somebody should take a look at that tutorial and make sure it’s solid and up to date. It was put up by different people over many months, and I think it might be broken/outdated in some places by now.

[...]

PR #2008, opened 2016-10-03
Add simpler async actions with tested example

oalmali:
According to [related issue](https://github.com/reactjs/redux/issues/1972) async actions needed to be re-tested and I have created a simpler self-working example which doesn't require any additional code to start and test.

### DIFF
--- a/docs/recipes/WritingTests.md
+++ b/docs/recipes/WritingTests.md
@@ -80,24 +80,24 @@ For async action creators using [Redux Thunk](https://github.com/gaearon/redux-t
 #### Example
 
 ```js
-import fetch from 'isomorphic-fetch';
+import fetch from 'isomorphic-fetch'
 
 function fetchTodosRequest() {
   return {
-    type: FETCH_TODOS_REQUEST
+    type: 'ADD_TODOS_REQUEST'
   }
 }
 
 function fetchTodosSuccess(body) {
   return {
-    type: FETCH_TODOS_SUCCESS,
-    body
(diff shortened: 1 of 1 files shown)


wdhorton:
I'm not sure I understand what this PR changes, besides renaming some things and moving lines of code around. Also it's recommended best practice not to use strings for action types, so that seems like a step in the wrong direction. Could you explain a bit more?

oalmali:
@wdhorton as it can be seen in the issue that I've mentioned, people wanted that example to be tested and verified. But the current version is demanding to import actions and types which makes difficult to verify tutorial. What I've aimed is anyone could be able to copy paste the tutorial code and will be ready to go. But as a starter I didn't know that strings are not good for action types. Thanks for helping and I hope my explanation is descriptive enough. There might be another version which places actions in the tutorial and doesn't require any additional code from counter example?

timdorr:
@oalmali The previous code does function, as far as I know. I think the bigger question is whether it needs a complete revamp. Let's keep that discussion on #1972 until we have a definitive answer.

markerikson:
@wdhorton , @oalmali : I'm very confused by something.  Why do you say that it's "best practice to not use strings for action types"?  That's _exactly_ the recommended practice.  Are you just trying to say that it's recommended to define your action types as constants rather than writing strings inline, or is there something I'm missing here?

wdhorton:
@markerikson that's what i meant, sorry for any confusion

PR #2014, opened 2016-10-05
Adding the case for nested state objects

sabahang:
Immutability helpers are mentioned elsewhere in the docs but I thought a reference here would be nice as I myself couldn't find the reason for my issues by looking at the troubleshooting page

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -119,7 +119,7 @@ return state.map((todo, index) => {
 
 Note that experimental language features are subject to change.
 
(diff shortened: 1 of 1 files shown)


brigand:
I don't know that "deep copied" is the right term here. In this example, to update `.a.b` you don't need to clone `.a.c`, and you shouldn't clone it for performance reasons.

```js
{
  a: {
    b: {},
    c: {},
  }
}
```

timdorr:
Fixed up the wording to be a little less awkward and make the linked article correct. Thanks!

PR #2026, opened 2016-10-10
check for initialization call of redux in "Reuing Reducer Logic" example

timse:
Reducers get called initialy to return their initial state, this action will probably not pass a filter.
Therefor it is necessary to check for undefined as the given state and return a call to the reducer,
in order to get the reducers state initialzed


markerikson:
Yeah, I could see that working.  Thanks!

PR #2027, opened 2016-10-10
Mark suggested adding this article about structuring app state to Prerequisite Concepts

talkol:
https://twitter.com/acemarke/status/782956839935893504
I will be honored if you decide to add this in the official docs. I hope the writeup will help people tackle this issue


markerikson:
Was actually considering putting it up in the "Reducer Basics" section, but this works pretty well too.

Thanks for a great article!

PR #2029, opened 2016-10-11
Cutted applyMiddlewares args (#2028)

### LINKED ISSUE #2028: Always undefined enhancer argument
`applyMiddleware` Enhancer [takes three](https://github.com/reactjs/redux/blob/master/src/applyMiddleware.js#L20) arguments now.
But `createStore` pass [two arguments](https://github.com/reactjs/redux/blob/master/src/createStore.js#L50) to enhancer.

`applyMiddleware` should take only two arguments.

owanturist:
Resolve #2028


timdorr:
https://github.com/reactjs/redux/issues/2028#issuecomment-252968176

PR #2030, opened 2016-10-11
Enable support for Mozilla Addons

qzb:
Redux seems to be great solution for managing state of browser extensions. Unfortunately it doesn't work properly with Firefox addons, because it requires global *process* object which is not available inside addon environment. This applies to both web-extensions and jetpack-based addons. In case of web-extensions it is possible make polyfill for *process* object, but it is not possible with jetpack (at least without [using custom modules loader][1]).

I've fixed this issue by checking if *process* object exists before using it. There was also issue with one of dependencies - *symbol-observable* - but it was [fixed yesterday][2].

[1]: https://gist.github.com/empyrical/f1cca44da1f81cb09a26
[2]: https://github.com/blesh/symbol-observable/pull/23

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "lodash": "^4.2.1",
     "lodash-es": "^4.2.1",
     "loose-envify": "^1.1.0",
-    "symbol-observable": "^1.0.2"
+    "symbol-observable": "^1.0.3"
   },
   "devDependencies": {
     "babel-cli": "^6.3.15",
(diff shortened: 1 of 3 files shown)


timdorr:
Ah, that's a good point. I think it's less about supporting the FF addon environment and more about supporting browser environments without assuming polyfills are provided. Good call!

gaearon:
I'm confused by this. Does this still strip out the DEV code in production? Is Uglify smart enough to figure out what's going on here?

gaearon:
Please see https://github.com/reactjs/redux/issues/2244.

PR #2035, opened 2016-10-14
Remove eslintConfig from package.json in examples

keik:
At react-scripts@^0.6.0 (as dev-dependencies of some examples), `eslintConifg` was removed from package.json default. (facebookincubator/create-react-app/pull/649)

At least, current value `./node_modules/react-scripts/config/eslint.js` is invalid in this version!


gaearon:
Thanks.

PR #2039, opened 2016-10-16
Use eslint-config-react-app instead of rackt for linting

alexanderjsx:
Changed to use eslint-config-react-app 
[eslint-config-react-app](https://github.com/facebookincubator/create-react-app/tree/master/packages/eslint-config-react-app) from create-react-app repo instead of `rackt`.
Enabled linting for the examples which was disabled because of linting mismatch between examples created by create-react-app and redux linter https://github.com/reactjs/redux/pull/1928.
Fixed a few lint errors according to the new config.
`babel-eslint` still used as peer dependency for `eslint-config-react-app`


timdorr:
Any reason we can't get rid of babel-eslint? Are we using decorators anywhere? That's normally what I've had to use it for.

timdorr:
I went ahead and did this by hand since there were merge conflicts. Thanks for the PR for me to copy from!

PR #2042, opened 2016-10-17
Update Motivation.md for grammar

alextsg:
Fix for parallelism:
**BEFORE:** "...we need to manage the active route, the selected tab, whether to show a spinner or not, should pagination controls be displayed, and so on."
**AFTER:** "...we need to manage active routes, selected tabs, spinners, pagination controls, and so on."


timdorr:
Thanks!

PR #2055, opened 2016-10-27
combineReducers doc amendment

Leonid-Yeromin:
Important to understand that if you move from custom combine function to combineReducers during refactoring, code: `myLastState === store.getState()` will have different behavior in some cases.


timdorr:
Fixed up the grammar a bit more. Thanks!

PR #2066, opened 2016-10-29
Small improvement

marcelaraujo:
Store the key length into a variable before iterate the object.


chicoxyzzy:
This optimisation is useless. JS engines are smart enough to make it under hood.

markerikson:
Appreciate the PR, but I don't think there's enough reason to actually make a change here.  I'll close this for now.  If you can provide specific benchmarks that prove this is a meaningful improvement, it may be wroth reconsidering.

PR #2071, opened 2016-10-31
add return to state.map()

maltepet:


jimbolla:
A lambda function without curly braces will return the result of its body. Adding a return keyword makes it invalid syntax.

PR #2073, opened 2016-11-02
Add support to compose only functions in utils/compose

evilj0e:
In our project we found some case that can brake your application. 
It can be broken when you've tried to pass not a function in enhance array to compose function.

```
const toEnhance = [
    router.enhancer,
    window && window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__() // false in this case will cause a bug
];
const enhancer = compose(...toEnhance);
```

Of course I can filter array before passing it to enhancer, but I think it will be useful to manage it in compose function.

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -14,7 +14,7 @@ export default function compose(...funcs) {
     return arg => arg
   }
 
-  funcs = funcs.filter(func => (typeof func === 'function'))
+  funcs = funcs.filter(func => typeof func === 'function')
 
   if (funcs.length === 1) {
     return funcs[0]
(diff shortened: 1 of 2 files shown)


chevsky:
You still can avoid this "bug" in one line if you want.
```
((window || {}).__REDUX_DEVTOOLS_EXTENSION__ || Function.prototype)()
```

markerikson:
I could see some value in this.  I know that @jimbolla added a somewhat similar check to React-Redux's `connect` in v5.  Maybe gate the check behind a `NODE_ENV` check so it only happens in dev?

/cc @timdorr

timdorr:
This is basically what I was thinking here. LGTM. Thanks!

PR #2075, opened 2016-11-03
rename Reducers to Seducers

mannynotfound:
Hey there. 

I have found the name Reducer a bit misleading because sometimes they create more than what was previously there, thus hardly "reducing" anything. "Seducers" is a much more romantic and ambiguous name that makes me feel like a poet as I code. What do you all think?


Koleok:
Gotta hand it to you, this is thorough 😜

umidbekkarimov:
So this lib should be renamed to `sedux`?

sompylasar:
Is it April Fools day today? 0_o

PR #2080, opened 2016-11-07
Docs: Flux Store snippet is outdated

alexeyraspopov:
[FluxUtils](http://facebook.github.io/flux/docs/flux-utils.html) to the rescue.

### DIFF
--- a/docs/recipes/ReducingBoilerplate.md
+++ b/docs/recipes/ReducingBoilerplate.md
@@ -439,24 +439,25 @@ Redux reduces the boilerplate of Flux stores considerably by describing the upda
 Consider this Flux store:
 
 ```js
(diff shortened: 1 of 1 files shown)


timdorr:
That counter example is meant to be as simple and generic as possible. That way you can be coming from another flux implementation and understand what's being discussed. I'd rather keep this library-agnostic.

alexeyraspopov:
Current example isn't really library-agnostic since it relies on `AppDispatcher` which has Facebook's Flux signature.

alexeyraspopov:
Also, this quote:

> the boilerplate of Flux stores

and a lot of others quotes about Flux implies to Facebook Flux implementation.

PR #2085, opened 2016-11-08
enable the keyword "this" in actioncreators when it is an object

doving:
when actioncreators is an object , i always need to use 'this' in an actioncreator function, but after using bindActionCreator , the keyword 'this' doesn't work


markerikson:
Not really seeing a reason to change anything here.  The recommended approach is to define action creators as standalone functions, which would have no need to reference `this`.

Given the relative simplicity of `bindActionCreators`, it should be easy enough to copy and paste into your own project and tweak it for your own needs.

PR #2086, opened 2016-11-09
const => let

zaxlct:


Florian-R:
Why? Unless I miss something, this ones are never reused.

vaclavbohac:
@zaxlct I don't understand why would you change these to `const` as there's no reassign.

jimbolla:
Should be `const` unless they're being reassigned, which does not appear to be the case.

zaxlct:
You're right

PR #2098, opened 2016-11-12
Replace react-addons-update

datenreisender:
Because react-addons-update is legacy and the official doc on https://facebook.github.io/react/docs/update.html also references [immutability-helper](https://github.com/kolodny/immutability-helper) as a replacement.


markerikson:
Yeah, that's a good point.  Thanks!

PR #2100, opened 2016-11-13
Added chapter in docs intro

thedgbrt:
Following this thread : https://twitter.com/thedgbrt/status/797779637573128193

I've put the simplified explanation from @gaearon where it seemed to fit the most, just before the dive into the three principles in more detail.

It adds a bit of repetition to the flow of the whole section, but overall makes things clearer imo.

Let me know what you think.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -3,7 +3,7 @@
 * [Read Me](/README.md)
 * [Introduction](/docs/introduction/README.md)
   * [Motivation](/docs/introduction/Motivation.md)
-  * [The Redux Way](/docs/introduction/TheReduxWay.md)
+  * [Core Concepts](/docs/introduction/CoreConcepts.md)
   * [Three Principles](/docs/introduction/ThreePrinciples.md)
   * [Prior Art](/docs/introduction/PriorArt.md)
   * [Ecosystem](/docs/introduction/Ecosystem.md)
--- a/docs/README.md
+++ b/docs/README.md
@@ -3,6 +3,7 @@
 * [Read Me](/README.md)
 * [Introduction](/docs/introduction/README.md)
   * [Motivation](/docs/introduction/Motivation.md)
+  * [The Redux Way](/docs/introduction/TheReduxWay.md)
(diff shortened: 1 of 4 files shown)


markerikson:
Thanks, appreciate the help!

A couple thoughts:

- Is there maybe a better title we can come up with besides "The Redux Way"?  I don't immediately have a better idea, but might want to consider other options.
- I've used a description a few times that we might want to include in here somewhere.  Quoting from a couple of my comments on Reddit:

> - Conceptually, Redux itself is a pretty simple pattern: it encapsulates a single value; uses a single function to update that value; calls the update function with the previous value and the descriptive object you just passed in; replaces the previous value with the new value; and uses a simple change emitter to notify subscribers of an update. ( https://www.reddit.com/r/reactjs/comments/5664xx/redux_why_are_reducers_useful/d8gmoc8/ )

> - At a conceptual level, Redux is really a very simple pattern: it stores a value, calls one function to create an updated value when you ask it to, and has a single change emitter to notify subscribers that there's been an update. There's a number of aspects that wrap around that further, but that's really all that it's doing internally.  ( https://www.reddit.com/r/programming/comments/543b4w/help_could_anyone_help_annotating_redux_i_really/d7yvrkw/?context=3 )

@timdorr , @jimbolla : thoughts?

thedgbrt:
For the title, what I think is important is to convey the idea of a generic and straightforward explanation: "Core purpose" or "Concept" might be good alternatives.

I found the first of your two comments easier to understand, I can it being helpful at the very end of the page, as a way to go over the whole concept one last time before digging deeper in the next section.

Phoenixmatrix:
yeah, core concept, basics, core architecture, etc would all be a bit better.

As for the explanations, I think Dan Abramov does a very good job here at explaining. Adding terminology like "storing a value" or "change emitter" just distract from the main idea and make it out to be more complicated than it actually is (and is arguably inaccurate anyway).

thedgbrt:
I've updated with a new title : "Core concepts". Is it good for merging?

markerikson:
LGTM.  Thanks!

PR #2106, opened 2016-11-15
improved comment, was a bit unclear

DanielRuf:
I am not 100% sure if this is what you meant but this makes most sense to me.


markerikson:
Thanks!

PR #2113, opened 2016-11-21
docs(Ecosystem): add redux-define

smeijer:


markerikson:
Hi.  Thanks for the submission, but I'm not sure it's something we want to add to the main "Ecosystem" list at the moment.  That said, I've bookmarked it and will add it to my [Redux addons catalog](https://github.com/markerikson/redux-ecosystem-links) in the next round of updates.

Issue #2114, opened 2016-11-21
CombineReducers flow-type throws error with valid arguments


nickarora:
**Do you want to request a *feature* or report a *bug*?**  
Bug.

**What is the current behavior?**
Currently, the combineReducers function is flow-type checked in such a way that the resulting reducer _always_ expects State to contain the same keys used to create the reducer.

For example:
```
const rootReducer = 
  combineReducers({
    foo: fooReducer,
    bar: barReducer,
  })

/* this works fine */
rootReducer({
  foo: 'some-value',
  bar: 'some-value',
}, someAction);

/* this causes an error (foo and bar are missing) */
rootReducer({}, someAction);

/* this causes an error (bar is missing!) */
rootReducer({
  foo: 'some-value',
}, someAction);
```

This can become problematic if you are trying to completely wipe out your store.

**What is the expected behavior?**
```
/* this no longer causes an error */
rootReducer({}, someAction);
```

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**
Redux 3.6.0; no.

PR #2115, opened 2016-11-21
Update argument type of reducer produced by combineReducers

nickarora:
Addresses #2114 

Updates CombineReducers flow-type

### DIFF
--- a/flow-typed/redux.js
+++ b/flow-typed/redux.js
@@ -22,7 +22,9 @@ declare module 'redux' {
     replaceReducer(nextReducer: Reducer<S, A>): void
   };
 
-  declare type Reducer<S, A> = (state: $Shape<S>, action: A) => S;
+  declare type Reducer<S, A> = (state: S, action: A) => S;
+
+  declare type CombinedReducer<S, A> = (state: $Shape<S> & {} | void, action: A) => S;
 
   declare type Middleware<S, A> =
     (api: MiddlewareAPI<S, A>) =>
@@ -46,7 +48,7 @@ declare module 'redux' {
   declare function bindActionCreators<A, C: ActionCreator<A, any>>(actionCreator: C, dispatch: Dispatch<A>): C;
   declare function bindActionCreators<A, K, C: ActionCreators<K, A>>(actionCreators: C, dispatch: Dispatch<A>): C;
 
-  declare function combineReducers<O: Object, A>(reducers: O): Reducer<$ObjMap<O, <S>(r: Reducer<S, any>) => S>, A>;
+  declare function combineReducers<O: Object, A>(reducers: O): CombinedReducer<$ObjMap<O, <S>(r: Reducer<S, any>) => S>, A>;
 
   declare function compose<S, A>(...fns: Array<StoreEnhancer<S, A>>): Function;
 
--- a/flow-typed/redux.js
+++ b/flow-typed/redux.js
@@ -46,7 +46,7 @@ declare module 'redux' {
   declare function bindActionCreators<A, C: ActionCreator<A, any>>(actionCreator: C, dispatch: Dispatch<A>): C;
   declare function bindActionCreators<A, K, C: ActionCreators<K, A>>(actionCreators: C, dispatch: Dispatch<A>): C;
 
-  declare function combineReducers<O: Object, A>(reducers: O): Reducer<$ObjMap<O, <S>(r: Reducer<S, any>) => S>, A>;
+  declare function combineReducers<O: Object, A, Z: $Shape<O>>(reducers: O): Reducer<$ObjMap<Z, <S>(r: Reducer<S, any>) => S>, A>;
 
   declare function compose<S, A>(...fns: Array<StoreEnhancer<S, A>>): Function;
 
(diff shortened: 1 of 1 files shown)


nickarora:
tagging @gcanti  and @aaronjensen

gcanti:
Using `$Shape` makes `combineReducers` unsafe:

```js
// test_combineReducers.js in flow-typed
type State = {
  a: AState,
  name: string,
  age: number
};


const reducer1: Reducer<State, Action> = combineReducers({
  a: reducerA,
  name: reducerName
  // missing age reducer but no error
})
```

> This can become problematic if you are trying to completely wipe out your store.

What do you mean? The root reducer is not supposed to be called "by hand"

```js
rootReducer({}, someAction);
```

nickarora:
> The root reducer is not supposed to be called by hand

@gcanti See [Dan Abramov's response here](http://stackoverflow.com/questions/35622588/how-to-reset-the-state-of-a-redux-store/35641992#35641992) where he details a recommended approach to resetting the store.  That would be the use case.  In his case, he names it "appReducer"  it is the same idea.

gcanti:
@nickarora note that in the linked approach `appReducer` is no more the root reducer, hence it can be called by hand. Also `appReducer` is called with `undefined` rather than `{}`. Maybe a better option, without losing type safety, would be

```js
declare function combineReducers<O: Object, A>(reducers: O): Reducer<$ObjMap<O, <S>(r: Reducer<S, any>) => S> | void, A>;
```

nickarora:
@gcanti I've pushed up a change that should address your type-safety concerns while still allowing state to be reset using their default initialValues -- as dan explains.

aaronjensen:
@nickarora I think I may have misled you on this a little when we talked earlier. If `Reducer<S, A>` is meant to be a general `Reducer` function then we shouldn't do this. It should take `(S | void, A)` as its arguments.

The result of `combinedReducer`, however, is different. It is actually less strict than `Reducer<S, A>` in that it allows the `$Shape<S> | void` as an argument.

Basically, if you pass `undefined` into the combined reducer, [it defaults to `{}`](https://github.com/reactjs/redux/blob/master/src/combineReducers.js#L131). It then grabs the previousStateForKey (which can be `undefined`) and passes that to the child reducer which should handle `undefined` with a default.

Adding `void` to the first argument of the `Reducer<S, A>` type is a good change to make, but to fix this actual problem you'll want something like this:

```js
declare type CombinedReducer<S, A> = (state: $Shape<S> | void, action: A) => S
declare function combineReducers<O: Object, A>(reducers: O): CombinedReducer<$ObjMap<O, <S>(r: Reducer<S, any>) => S>, A>;
```

nickarora:
Sounds reasonable to me -- I'll let @gcanti weigh in before making any other changes

gcanti:
Not sure why `$Shape` is involved, do you want to *partially* reset the store?

Beware: `$Shape` is lenient with both `null` and `undefined`

```js
type O = { name: string };
const x1: $Shape<O> = undefined // <= ok
const x2: $Shape<O> = null // <= ok
```

Maybe we should exclude `null` explicitly

```js
declare type CombinedReducer<S, A> = (state: $Shape<S> & {} | void, action: A) => S;
declare function combineReducers<O: Object, A>(reducers: O): CombinedReducer<$ObjMap<O, <S>(r: Reducer<S, any>) => S>, A>;
```

Then you can write

```js
import type { Store } from 'redux'
import { createStore, combineReducers } from 'redux'

type State = {
  a: number,
  b: number
};

type Action = { type: 'RESET' };

const appReducer = combineReducers({
  a: (s = 1) => s,
  b: (s = 2) => s
})

const reducer = (s, a) => {
  if (a.type === 'RESET') {
    // return appReducer(null, a) // <= error
    // return appReducer(undefined, a) // <= ok
    return appReducer({}, a) // <= ok
    // return appReducer({ a: s.a }, a) // <= ok, partially reset the store
    // return appReducer({ a: 'a string' }, a) // <= ok BUG :( see https://github.com/facebook/flow/issues/2674
  }
  return appReducer(s, a)
}

const store: Store<State, Action> = createStore(reducer, { a: 3, b: 4 })
```

aaronjensen:
> Not sure why $Shape is involved, do you want to partially reset the store?

In our case, yes.



> Beware: $Shape is lenient with both null and undefined

Ah, thanks.

> Maybe we should exclude null explicitly

Agreed.

> ```
>  // return appReducer({ a: 'a string' }, a) // <= ok BUG :( see https://github.com/facebook/flow/issues/2674
> ```

Ugh. That was already a problem though, yeah? So this wouldn't introduce a new bug afaict.

Ok, so it sounds like the `| {}` tweak is needed to exclude `null`. Other than that, I think we're good to go on this.

gcanti:
> Ugh. That was already a problem though, yeah? So this wouldn't introduce a new bug afaict.

Yes, I agree, that was already a problem. It's not critical if you use the result of `combineReducers` as root reducer, it's only dangerous in this use case where you call the combined reducer by hand, though I'm afraid there's nothing we can really do.

> Other than that, I think we're good to go on this

I ran this change against the tests in the flow-typed repo and they are still green

nickarora:
@gcanti @aaronjensen updated and opened a PR in the flow-typed repo as well

timdorr:
Thanks for the fix!

PR #2118, opened 2016-11-22
Update ServerRendering to highlight security issue

ThisIsMissEm:
This adds a comment directly in the code which most people copy and paste to do server rendering, which actually has potential to be an XSS vulnerability.


markerikson:
Good idea.  Thanks!

PR #2124, opened 2016-11-25
Provide example of initial state SSR scrub

wmertens:
This short piece of code renders all html tags in JSON harmless, by replacing the `<` character with its unicode escape `\u003c`. Since the only place where JSON can have `<` is in strings, this is sufficient for protecting JSON from XSS injection.

Providing an easy cut'n'paste solution in the documentation ensures that more people will be protected from this problem.


markerikson:
Not an XSS expert, but sounds like a good thing.  Thanks!

PR #2132, opened 2016-11-30
Add optional 'immediate' argument to unsubscribe.

itbrandonsilva:
Setting `immediate` to true ensures that the given listener will not be called during the current dispatch (unless it was called before unsubscribing). Useful in scenarios where you'd like to safely clean up resources along with unsubscribing listeners that depend on those resources simultaneously.

Given the proposition is accepted, I'll update the relevant docs asap.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -108,7 +108,7 @@ export default function createStore(reducer, preloadedState, enhancer) {
     ensureCanMutateNextListeners()
     nextListeners.push(listener)
 
-    return function unsubscribe() {
+    return function unsubscribe(immediate) {
       if (!isSubscribed) {
         return
       }
@@ -118,6 +118,11 @@ export default function createStore(reducer, preloadedState, enhancer) {
       ensureCanMutateNextListeners()
       var index = nextListeners.indexOf(listener)
       nextListeners.splice(index, 1)
+
+      if (immediate) {
+        index = currentListeners.indexOf(listener)
+        currentListeners[index] = function () {}
+      }
     }
   }
 
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -317,6 +317,19 @@ describe('createStore', () => {
     expect(listener3.mock.calls.length).toBe(1)
   })
 
+  it('supports immediate unsubscriptions', () => {
+    const store = createStore(reducers.todos)
+
+    const listener1 = jest.fn(() => unsubscribe(true))
+    const listener2 = jest.fn()
+    store.subscribe(listener1)
+    var unsubscribe = store.subscribe(listener2)
+
+    store.dispatch(unknownAction())
+    expect(listener1.mock.calls.length).toBe(1)
+    expect(listener2.mock.calls.length).toBe(0)
+  })
+
   it('delays subscribe until the end of current dispatch', () => {
     const store = createStore(reducers.todos)
 
--- a/test/createStore.spec.js
+++ b/test/createStore.spec.js
@@ -322,12 +322,15 @@ describe('createStore', () => {
 
     const listener1 = jest.fn(() => unsubscribe(true))
     const listener2 = jest.fn()
+    const listener3 = jest.fn()
     store.subscribe(listener1)
     var unsubscribe = store.subscribe(listener2)
+    store.subscribe(listener3)
 
     store.dispatch(unknownAction())
     expect(listener1.mock.calls.length).toBe(1)
     expect(listener2.mock.calls.length).toBe(0)
+    expect(listener3.mock.calls.length).toBe(1)
   })
 
   it('delays subscribe until the end of current dispatch', () => {


timdorr:
This sounds fine to me. Can you update the docs too? Also, this should go on the `next` branch for now.

gaearon:
Can I see the usage code before and after?
If get this in we'll effectively never get rid of it and it'll stay in public API forever.
Do we really need this?

itbrandonsilva:
My specific case involved a set of game entity instances all having their own subscription to the store, allowing each entity to have their own space in which they can read the state and update graphics. When the app first initializes (before any entities are created), the very first subscription that is made to the store would be responsible for diffing the old state with the new state to determine entities that need be created or removed. The entity removal logic was along the lines of `entity.destroy()`, where resources would be destroyed and (most importantly) _unsubscribe from the store_. This lead to an issue where listeners that were unsubscribed on destroyed entities would still be called an additional time. Of course some bandage code was added to the entity logic so that the listener can check if the entity had already been destroyed, but that was where I felt that this functionality should naturally be available.

Simple example:

```
let game: any;
...

// Learns of new entities, entity changes, or entities
// that are no longer needed (calling entity.destroy()).
store.subscribe(() => {
    state = store.getState();
    ...
});

class Entity() {
    constructor() {
        ...
        this.unsubscribe = store.subscribe(this.update.bind(this));
        ...
    }

    update() {
        let state = store.getState();
        ...
        this.sprite.x = state.f.o.o.x;
        this.sprite.y = state.f.o.o.y;
        ...
    }

    destroy() {
        ...
        this.sprite = null;
        this.unsubscribe();
        ...
    }
}

// Note that a subscription is made in the constructor of Entity.
let entity = new Entity(); 
game.add(entity);

store.dispatch({type: 'MAKE_STUFF_HAPPEN', ...});
// Subscribers would then called after the action has been
// processed, where the subscription created at the top of this
// example is called first.
...
// That same handler determines that an entity should be destroyed.
entity.destroy()
...
// Shortly after we get:
//     TypeError: Cannot read property 'x' of undefined 
// The subscription of the destroyed entity is called once more,
// referencing the member sprite that no longer exists.
```

itbrandonsilva:
@gaearon Having a few days to consider this, I can't think of anything else one would want `unsubscribe` to do. I'm on the boat of really hoping that no other functionality would ever be needed on `unsubscribe`.

itbrandonsilva:
@timdorr Will do so shortly.

jimbolla:
Couldn't you achieve the same effect by modifying `destroy`:
```js
    destroy() {
        ...
        this.sprite = null;
        this.update = () => {}; // <-- basically the same thing you're doing in your PR.
        this.unsubscribe();
        ...
    }
```

gaearon:
Okay, here is a different question. If this behavior is more sensible should we make it the new default instead (in a major version)?

I really really don't like introducing options to API. In my experience it seems like a good idea at first and then bites you forever later.

itbrandonsilva:
I _definitely don't think_ it should be the typical behavior. At least not so quickly.

gaearon:
To be clear, does the workaround of setting `this.isDestroyed = true` and checking that flag in change handler work as a workaround for you?

I know it's annoying and it's been bothering me too.

What is the downside of doing this by default?

itbrandonsilva:
That's actually what I'm already doing. 

Downside is that it feels unnatural. That's about it. The upside/downside is fairly insignificant given the size of the feature.

jimbolla:
It's not very intuitive that a subscriber could still get a notification post unsubscribe because the unsubscription happened during the notify loop. And that's dependent on the ordering of subscriptions; if you change the order in which your subscribers subscribed, that could change whether or not that extra callback happens. I'm in favor of making it the default behavior.

itbrandonsilva:
@jimbolla You don't feel that some users may already be dependent on the current behavior?

gaearon:
I'm opposed to making this an option.

I'm good with making it the default behavior if you do this:

* Go through all past changes in subscrube behavior in the past and collecting use cases / problems that we tried to solve by making it work this way.
* Try this change on large open source app examples to see what would break.
* Write up an analysis showing that this change is always possible to migrate to (doesn't lock users out of an important use case) and does not make performance worse.

jimbolla:
For a reference, I made a C# program to see how it handles this scenario, since it has event handlers built-in as a first-class language feature:  
```c#
    internal class Program
    {
        public static void Main() { new Program().Run(); }

        public event EventHandler SubTest;

        public void Run()
        {
            EventHandler a = (o, e) => { Console.Write('a'); };

            EventHandler b = (o, e) =>
            {
                SubTest -= a;
                Console.Write('b');
            };

            SubTest += b;
            SubTest += a;
            SubTest(this, EventArgs.Empty);
            SubTest(this, EventArgs.Empty);
            Console.WriteLine();
        }
    }
```
It's output was `bab`. So in c#, the behavior is that all handlers fire, even if one of them unsubscribes in the middle of the loop. This is consistent with how Redux currently behaves. I tested the same idea for adding a handler as well. I think we should leave Redux's behavior as is. This could be implemented as a store enhancer, provided either in the docs or as an additional API.

PR #2137, opened 2016-12-06
Looking into for, forEach abstractions

felquis:

I want to suggest an edit in the source

Currently it's working just fine

It's not a bug

Tests must pass after this change

## Objective

Make it cleaner what is going on, the function's name can help developers
understand what is going on

For example

before
```js
for (var i = 0; i < keys.length; i++) {
  var key = keys[i]
  var actionCreator = actionCreators[key]
  if (typeof actionCreator === 'function') {
    boundActionCreators[key] = bindActionCreator(actionCreator, dispatch)
  }
}
```

after
```js
var walkKeys = (key) => {
  var actionCreator = actionCreators[key]
  if (typeof actionCreator === 'function') {
    boundActionCreators[key] = bindActionCreator(actionCreator, dispatch)
  }
}

forEach(keys, walkKeys)
```

## build size Ramda#forEach

First I run `npm run build` from master, then zip the files with

tar -zcvf lib.tar.gz lib/ 
tar -zcvf es.tar.gz es/
tar -zcvf dist.tar.gz dist/

Then I used [gzip-size](https://www.npmjs.com/package/gzip-size-cli) to see the
size of the `.gz` files, which the output is in the right side of the file name

Size difference per build before and after, with Ramda#forEach

| Build Output  | Before        | After         | size diff     |
| ------------- | ------------- | ------------- | ------------- |
| dist          | 13K or 12997  | 15K or 15038  | 2041 bytes    |
| es            | 6,7K or 6822  | 6,7K or 6873  | 51 bytes      |
| lib           | 6,9K or 7072  | 7,1K or 7202  | 130 bytes     |

There may be a alternative to [ramda#forEach](http://ramdajs.com/docs/#forEach)
but its implementation looks very optimized

## utils/forEach

| Build Output  | Before        | After         | size diff     |
| ------------- | ------------- | ------------- | ------------- |
| dist          | 13K or 12997  | 13K or 13209  | 212 bytes    |
| es            | 6,7K or 6822  | 6,9K or 7006  | 184 bytes      |
| lib           | 6,9K or 7072  | 7,2K or 7340  | 268 bytes     |

## jsperf

Today's for loop is pretty fast, you can benchmark it in your machine
in this link https://jsperf.com/for-each-try-5, I used the Ramda and Lodash
forEach, and the faster from this [for loop jsperf](https://jsperf.com/for-vs-foreach/75) to compare

These tests make it clear for me, that using the simple forEach abstraction is the best fit between code readability + performance.

I suggest you to run the benchmark on your computer or smartphone, and put here the printscreen

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -1,6 +1,6 @@
 import isPlainObject from 'lodash/isPlainObject'
 import $$observable from 'symbol-observable'
-import forEach from 'ramda/src/forEach'
+import forEach from './utils/forEach'
 
 /**
  * These are private action types reserved by Redux.
@@ -176,7 +176,7 @@ export default function createStore(reducer, preloadedState, enhancer) {
     var listeners = currentListeners = nextListeners
 
     var callListener = (listener) => listener()
-    forEach(callListener, listeners)
+    forEach(listeners, callListener)
 
     return action
   }
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -1,5 +1,6 @@
 import isPlainObject from 'lodash/isPlainObject'
 import $$observable from 'symbol-observable'
+import forEach from 'ramda/src/forEach'
 
 /**
  * These are private action types reserved by Redux.
@@ -173,10 +174,9 @@ export default function createStore(reducer, preloadedState, enhancer) {
     }
 
     var listeners = currentListeners = nextListeners
-    for (var i = 0; i < listeners.length; i++) {
-      var listener = listeners[i]
-      listener()
-    }
+
+    var callListener = (listener) => listener()
+    forEach(callListener, listeners)
 
     return action
   }
--- a/package.json
+++ b/package.json
@@ -68,7 +68,6 @@
     "lodash": "^4.2.1",
     "lodash-es": "^4.2.1",
     "loose-envify": "^1.1.0",
-    "ramda": "^0.22.1",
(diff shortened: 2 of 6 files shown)


felquis:
screenshot from 2016-12-06 14-06-45

felquis:
I could't test the perf in a nodejs env

felquis:
Firefox 47 on Linux, Simple Loop and Simple Loop Abstraction wins

screenshot from 2016-12-06 14-12-52

felquis:
Chrome 54 on Linux, Simple Loop Reverse and Simple Loop Abstraction wins 
screenshot from 2016-12-06 14-14-50

jimbolla:
It was slower on my machine...

untitled

markerikson:
Hi.  Thanks for the PR, but I'm not seeing enough reason to change the looping behavior at the moment.

src/createStore.js
```diff
@@ -173,10 +174,9 @@ export default function createStore(reducer, preloadedState, enhancer) {
     }
 
     var listeners = currentListeners = nextListeners
-    for (var i = 0; i < listeners.length; i++) {
-      var listener = listeners[i]
-      listener()
-    }
+
+    var callListener = (listener) => listener()
+    forEach(listeners, callListener)
```
derekstavis: `forEach(listeners, listener => listener())`

PR #2139, opened 2016-12-07
Update AsyncActions.md - missing `export` keywords

nateq314:
Actions in 'actions.js' missing `export` keywords in a few key locations, these have been added here.


nateq314:
Afterwards noticed that these functions are intentionally not exported in the final version. If so then it seems like they should also not be shown as exported in the initial versions of Actions.js shown towards the beginning of the file? Opening new PR with those deleted instead.

PR #2140, opened 2016-12-07
Update AsyncActions.js

nateq314:
Remove `export` keyword from requestPosts, receivePosts, and fetchPosts actions, as these are not exported in the final version. Also suggest adding brief explanation as to why they are not exported.


timdorr:
Thanks!

Issue #2033, opened 2016-10-13
Redux compose "TypeError: undefined is not an object (evaluating 'last.apply')"


igorrfc:
Hey guys. I'm just building the dependencies in a project made with react/redux. And when webpack bundle the main file, the error: "TypeError: undefined is not an object (evaluating 'last.apply')" has been shown on console log.

```javascript
function compose() {
  for (var _len = arguments.length, funcs = Array(_len), _key = 0; _key < _len; _key++) {
    funcs[_key] = arguments[_key];
  }

  if (funcs.length === 0) {
    return function (arg) {
      return arg;
    };
  }

  if (funcs.length === 1) {
    return funcs[0];
  }

  var last = funcs[funcs.length - 1];
  var rest = funcs.slice(0, -1);
  return function () {
    return rest.reduceRight(function (composed, f) {
      return f(composed);
    }, last.apply(undefined, arguments));
  };
}
```

Is there any solution for this error?

markerikson:
That sounds as if you're passing in something invalid when creating your store.  Please check your store creation code and verify that the call to `compose` has valid enhancers as arguments.

JoeCortopassi:
@igorrfc were you able to figure out steps to reproduce? Wondering if related to this https://github.com/reactjs/redux/pull/2145

aniddan:
In my case caused because of Redux Dev Tools. By passing undefined to compose it generates this error
```JavaScript
createStore(
	combineReducers({ auth, missions, artworks, artists }),
	compose(...[
		applyMiddleware(promiseMiddleware),
		__REDUX_DEVTOOLS_EXTENSION__ && __REDUX_DEVTOOLS_EXTENSION__()
	]),
);
```

ballerabdude:
I got this error from safari and I was able to fix it with this.

``` 
// Only chrome can handle the redux dev tool
// redux compose cannot handle a null or undefined middleware
if (window.navigator.userAgent.includes('Chrome')) {
  var store = createStore(
    reducer,
    initialState,
    compose(
      applyMiddleware(
        routerMiddleware(browserHistory)
      ),
      window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()
    )
  );
} else {
  var store = createStore(
    reducer,
    initialState,
    compose(
      applyMiddleware(
        routerMiddleware(browserHistory)
      )
    )
  );
}
```

Barlog-M:
[...]

Issue #632, opened 2015-08-27
`compose` is not a normal compose


jlongster:
`compose` is a function that normally doesn't eagerly evaluate it's arguments, but returns a new function that, when called, does the right-to-left evaluation.

It's easy if the composition returns a function that takes a single argument (I think this is the proper mathematical composition):

```js
function compose(...funcs) {
  return x => {
    return funcs.reduceRight((composed, f) => f(composed), x);
  };
}
```

But usually we want to seed the application with multiple args, but it's just a little more code:

```js
function compose(...funcs) {
  return (...args) => {
    const initialValue = funcs[funcs.length - 1].apply(null, args);
    funcs = funcs.slice(0, -1);
    return funcs.reduceRight((composed, f) => f(composed),
                            initialValue);
  };
}
```

This changes the pattern of wrapping `createStore` though. One option would be to rename this to `eagerCompose` or something. Not sure if you all are interested in fixing this, but it is somewhat confusing that acts different than most (all?) other `compose` functions.

gaearon:
I agree, let's fix it somehow. It confused the hell out of me too on one occasion. :-P

jlongster:
Haha. If you're not open to changing how higher-order stores work (they return a function that takes a "createStore" function), seems like the only option is the change the name. That's the main problem right now: they impose a contract on the composed functions when normally the functions don't even know they are being composed.

Even changing the name will cause breakage though, so I dunno.

gaearon:
>Haha. If you're not open to changing how higher-order stores work (they return a function that takes a "createStore" function)

I'm open to it as long as we can make dev tools and hot reloading work with some other contract.

gaearon:
See also https://github.com/rackt/redux/issues/376 #350 

If you can figure out a way to make it simpler while keeping devtools and hot reloading working, please do!

gaearon:
(I don't mind breakage, we'll jump to 2.0 and that'll be it, but we need to make sure new API is better.)

jlongster:
Here's a simple change that keeps the current interface for middlewares and higher-order stores. Make `compose` work like normal:

```js
function compose(...funcs) {
  return (...args) => {
    const initialValue = funcs[funcs.length - 1].apply(null, args);
    const leftFuncs = funcs.slice(0, -1);
    return leftFuncs.reduceRight((composed, f) => f(composed),
                                 initialValue);
  };
}
```

[...]

PR #2145, opened 2016-12-10
Simplify composer

JoeCortopassi:
cleaned up the composer so that it doesn't rely on slicing and reversing the array of functions

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -20,7 +20,5 @@ export default function compose(...funcs) {
     return funcs[0]
   }
 
-  const last = funcs[funcs.length - 1]
-  const rest = funcs.slice(0, -1)
-  return (...args) => rest.reduceRight((composed, f) => f(composed), last(...args))
+  return funcs.reduce((a, b) => (...args) => a(b(...args)))
 }


toddw:
Nice! That definitely makes the code a lot cleaner and easier to understand

markerikson:
My FP-fu is admittedly kinda weak.  First issue I have here is I don't know if the changes actually keep the order of operations correct.  Second issue is that even if the existing code isn't the absolute prettiest, it's there and it works.  I'd need a pretty compelling reason to make changes to it at this point.  

Now, I do see a `compose.spec.js` test file, and it looks like all existing tests do pass.  That's definitely a starting point.  I'm just not sure that there's enough of a reason to make changes.

gaearon:
Might be a good idea to look through the revision history to see how it got to be this way.
I don't remember but we changed it a few times.

markerikson:
Yeah. As I've said a couple times, given Redux's current status and my own relative newness as a maintainer, my default attitude towards code change requests is "No unless it's really obvious it's a good thing".

JoeCortopassi:
I think the main argument for the change is this: if something broke concerning the current implementation would you be able to a) understand what it’s doing, b ) explain it to someone else, c) be sure it wasn’t the source of the error. The update makes it easier to parse, and actually makes the composition function more easily testable.

@gaearon history doesn't really show a great reason for it to exist that way. First commit has it using reduceRight with the head/rest chopping as well. Seems like the main changes that were made over time were things like guarding against non-functions being passed or short-circuits for stuff like single item arrays

markerikson:
Did some digging through the history.  Looks like the current iteration was made in https://github.com/reactjs/redux/pull/1050 .

JoeCortopassi:
@markerikson that actually makes me feel like this change is even more important. The current implementation isn't a true compose. Give >1 arguments, the first function will be _evaluated_, before being composed with the rest. That seems broken

JoeCortopassi:
I mean, the whole thing can be boiled down to 
```
funcs.filter(f=>typeof f === "function").reduce((a, b) => (...args) => a(b(...args)), i=>i)
```
but I wanted to be as low impact as possible

markerikson:
As I said, my own FP-fu is fairly weak. I'm going to ask you to simplify and clarify the issues involved :)

- What is a "true compose", and how does the current implementation differ?  What about that difference is "broken"?  What impact does it have?
- What about the current implementation is "evaluating the first function"?

src/compose.js
```diff
@@ -20,7 +20,5 @@ export default function compose(...funcs) {
     return funcs[0]
   }
 
-  const last = funcs[funcs.length - 1]
-  const rest = funcs.slice(0, -1)
-  return (...args) => rest.reduceRight((composed, f) => f(composed), last(...args))
```
JoeCortopassi: the last part `last(...args)` in `reduceRight` is evalutating, not composing. That is broken

JoeCortopassi: @gaearon

JoeCortopassi:
A true compose is binding many functions together (think chaining) such that it reproduces one function that is callable at a later time. 

semi-bad example that should help demonstrate:
```
function add10 (num) { return num+10; }
function multiply10 (num) { return num*10; }
function subtract10 (num) { return num-10; }

const compose = (...funcs) => funcs.filter(f=>typeof f === "function").reduce((a, b) => (...args) => a(b(...args)), i=>i);

const addMultiplySubtract = compose(add10, multiply10, subtract10);
const subtractMultiplyAdd = compose(subtract10, multiply10, add10);

addMultiplySubtract(5) === subtractMultiplyAdd(5) // false. Order of operations matter
```
this is why getting order of operations right important for redux/middleware/etc

markerikson:
Sorta following the concern here.  Does the fact that this is normally used with store enhancers affect the usage/behavior in any way?

gaearon:
Please add tests that fail with master.

JoeCortopassi:
@gaearon there isn't really a test against it. It's just not really composing. It's using closures to make it ok to evaluate the last function as `reduceRight`'s initial value.

I didn't really expect this to be a big deal, I just noticed it while playing with some stuff and was just trying to help out. Thought it would be pretty obvious that it had some rust by stuff like doing `const last = funcs[funcs.length - 1]` rather than `const last = funcs.pop()`. It could even be changed to 
```
  if (funcs.length === 0) {
    return arg => arg
  }

  funcs = funcs.filter(func => typeof func === 'function')

  if (funcs.length === 1) {
    return funcs[0]
  }

  const last = funcs.pop()
  return (...args) => funcs.reduceRight((composed, f) => f(composed), last(...args))
}
```
I just figured that if I was fixing that, I might as well update the `reduceRight` as well. If we wanted to get more rigorous and have something that added a test, I guess I could add a check in case a method was trying to be composed as well, in which case we'd get a type error when it was eventually called whether it was bound or not

gaearon:
I’m having a hard time following this because I’m not really a smart programer.
So I am asking you to help me by providing unit tests that demonstrate the observable difference in behavior.

If it’s calling functions too early, we could test this by setting a flag inside and checking if the flag gets set during `compose` call or during a call on the result function. Would that make sense?

Thank you for your effort! I appreciate your help but my concerns are coming from a good intent, not from being stubborn or self-righteous about a broken implementation (see also https://github.com/reactjs/redux/pull/2146#issuecomment-266221061).

It’s just that, if this is an observable change and people might rely on it, it should go into the next major.

JoeCortopassi:
@gaearon  https://github.com/reactjs/redux/issues/2033 looks like it's caused by a transpiler being confused by the current implementation as well. I assume a non-trivial portion of Redux users are using babel or something, which would mean that you've likely seen errors caused by this, but not clearly traceable back to this. 

You're clearly a smart and accomplished programmer, this isn't about ability. Ask my coworkers, I'm nothing special. We just use and love redux and I wanted to help out. I saw something I thought I could make better, and I'm sure someone will eventually see glaring problems with my patch and make it better as well. That's the benefit/nature of open source. It's an area that already has good tests, and it's a low impact change, if it causes side effects that are completely unknown (it shouldn't), then now we know a new case to test for and it's easy to revert. Admittedly, I'm a big fan of undocumented features/implementation being open game to be changed out from under people (i.e. don't play out of bounds), but I don't believe that's what's being done here

JoeCortopassi:
Interesting, this might be the conversation about why this happened that you were thinking of @gaearon https://github.com/reactjs/redux/issues/632#issue-103392532

Looks like the `reduceRight` and `last`/`rest` were more about being able to hack in multiple args being accepted

markerikson:
Isn't code history spelunking fun? :)

jridgewell:
This PR and the original code are equivalent, minus the performance.

- The old code creates `2` closures (1 could be hoisted, avoiding the allocation every time) and calls apply `1` time.
- The new code creates `n - 2` closures, and calls apply `n` times.

There's no observable eager evaluation happening, unless you monkeypatch `Array#reduceRight`, and that's unlikely.

JoeCortopassi:
@jridgewell timed over 10,000 iterations for each situation. It's doubly faster in the most common situation, and ranges between 3x faster and simply matching in other situations
screen shot 2016-12-10 at 9 42 31 pm

markerikson:
Interesting. Granted, the typical use case is "call once while setting up the store", but I suppose it's possible that someone out there might be reusing it for something more complicated.  Thanks for the benchmarks.

JoeCortopassi:
@markerikson 10000 iterations was just to prevent random issues with the computer from skewing the average. By common use case, I simply meant it being passed in between 1-10 items for composition.

jridgewell:
> It's doubly faster in the most common situation, and ranges between 3x faster and simply matching in other situations

It depends on what you want to prioritize for, the current code is by no means optimized. `#reduceRight` and all the array iteration methods are terribly slow compared for simple for loops, creating functions slows down startup time, etc. It may be that readability is more important than speed. But even then, I'd argue the current code is easier to understand than the composing closure functions (though yours is extremely beautiful, I love reading these solutions I'd never come up with).

http://justin.ridgewell.name/browser-benchmark/redux-compose/

gaearon:
I’m good with making this change.
#yolo

jimbolla:
I like this change. But also, shouldn't the filter: `funcs = funcs.filter(func => typeof func === 'function')` be before the `length===0` check? If someone calls
```js
compose(
  isDev ? someEnhancer : null,
  isDev ? anotherEnhancer : null
)
```
it would miss out on that check.

Also, recompose has the [same function](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/compose.js). I'd suggest also submitting this code change there. Ideally the implementations would be exactly the same to aid code minification.

JoeCortopassi:
@jimbolla @gaearon I can add the `filter` move to this pr or put it in a separate one. Personally I'd rather close out this pr and put that change in a separate pr, but I'm really good with whatever, just let me know :-)

JoeCortopassi:
Filter change pr
https://github.com/reactjs/redux/pull/2147/files

JoeCortopassi:
so is this good to go?

markerikson:
Sorry, been focused on other tasks the last couple days.  I see a :+1: from Dan, so let's do this.

markerikson:
Thanks for the assistance, @gaearon and @jimbolla , and thank you @JoeCortopassi for the fixes!

PR #2147, opened 2016-12-11
fixed condition where multiple bad values would be filtered out and h…

JoeCortopassi:
fixed condition where multiple bad values would be filtered out and have nothing to return. 

The test added would break with current production code


JoeCortopassi:
@markerikson good to merge?

markerikson:
Per discussion in #2145 , looks useful.

Issue #2150, opened 2016-12-13
The shopping-cart/containers/App.js is a container component?


kjirou:
**Do you want to request a *feature* or report a *bug*?**  

One of the examples looks inappropriate.

**What is the current behavior?**

The [shopping-cart's App.js](https://github.com/reactjs/redux/blob/master/examples/shopping-cart/src/containers/App.js) is defined as a container component.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar.**

**What is the expected behavior?**

The App.js is not `connect` ed, so it is a presentational component?

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**

timdorr:
It's arguably a container because it contains the entirety of the application's components. I would leave it there because, while it doesn't take on any of the functionality of a container, it would normally scale to add that kind of thing once the app becomes sufficiently complex. At that point, you would need a refactor if you had been thinking of it as purely presentational. So, this is mostly about setting up a good skeleton for others to build on in the future.

khirayama:
@timdorr Thanks.
But I think this `App.js` is presentational component. According to this issue.
https://github.com/reactjs/redux/issues/1466

How do you think it?

mattschmitz:
This behavior is inconsistent between shopping cart and todos_with_undo; in the todos_with_undo App.js is in components. Should it not be consistent?

PR #2151, opened 2016-12-13
Move the App component to under the components dir

kjirou:
Ref) #2150


timdorr:
https://github.com/reactjs/redux/issues/2150#issuecomment-269737631

PR #2159, opened 2016-12-20
Correct reference for recipe documentation.

kdavh:
This is more a suggestion for more general fixes to recipe docs links.  I'm relatively new to redux and was trying to follow this server rendering recipe but saw incorrect files (according to current master branch) referenced when trying to complete recipe steps.  Looking back in the project history, it looks like v3.0.6 may be the last tag that is compatible with the suggested changes. 

Maybe someone who knows the project better can confirm that this is the correct tag to reference for these examples?

I assume this is preferred over rewriting all of the recipes right now.  I just want accurate docs. What do others think?

### DIFF
--- a/docs/recipes/ServerRendering.md
+++ b/docs/recipes/ServerRendering.md
@@ -20,7 +20,7 @@ Redux's **_only_** job on the server side is to provide the **initial state** of
 
 ## Setting Up
 
-In the following recipe, we are going to look at how to set up server-side rendering. We'll use the simplistic [Counter app](https://github.com/reactjs/redux/tree/master/examples/counter) as a guide and show how the server can render state ahead of time based on the request.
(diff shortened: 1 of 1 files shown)


markerikson:
Ah... the "master/examples/counter" link is a valid link within the repo right now.  Can you clarify what the issue is?

kdavh:
@markerikson Yeah, the current link is valid, but that server rendering example refers to, for example a file called `./server.js` or `./containers/App.js` in the counter app example, which don't exist on current master.  The last time those files existed in the counter app example was at release v3.0.6

timdorr:
I would update the references to the right files. Linking to a tag might get people looking at old code in the long term.

mikeboyle:
I'm curious if any PRs were ever made or merged to address this issue. I don't see any references to the issue noted here. The file and directory references in the server-side recipe in the docs still do not match the counter example on master. The example on master also now uses Create React App and the recipe in the docs doesn't address how to handle the different build scripts, package.json, etc. In general, Redux documentation is fantastic; this is just one area in need of maintenance.

PR #2160, opened 2016-12-21
Changed sub-headings so that the IDs are unique

chadfurman:
Having unique IDs on the sub-headings allows linking to sections of the documentation easier.  Without this change, the following link is ambiguous:

http://redux.js.org/docs/basics/UsageWithReact.html#container-components


markerikson:
Thanks!  I think I actually noted that the other day, then it skipped my mind.  Appreciate the fix!

PR #2162, opened 2016-12-22
Remove not necessary argument

NE-SmallTown:
The 'enhancer' argument is not necessary,beacuse there is:
`return enhancer(createStore)(reducer, preloadedState)` at ' createStore.js'

We can see that the function call doesn't pass enhancer as the arguments.

So this is my thought.


markerikson:
Thanks for the PR, but this is a duplicate of #2128 .  Please see my comment at https://github.com/reactjs/redux/pull/2128#issuecomment-263581319 for more info.

PR #2164, opened 2016-12-22
Fix filename typo in UsageWithReactRouter docs

axelclark:
* Footer component filename is listed as `containers/Footer.js` in docs
* Directory in the example app is `components/Footer.js`
* Update docs to match example app


markerikson:
Thanks!

PR #2167, opened 2016-12-24
Remove filtering inside compose 

istarkov:
1) With filtering `compose` does not have the well know definition of this function as like as in your comments https://github.com/reactjs/redux/blob/master/src/compose.js#L7-L9
2) Then I accidentally pass something what is not a function into compose, the only thing I could expect is an error not silent code execution. (there is no need at all to pass anything which is not a function)
3) Typed definition of compose (even you does not use it) will be really hard to create with filtering.

So I suggest you to remove this line.

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -10,8 +10,6 @@
  */
 
 export default function compose(...funcs) {
-  funcs = funcs.filter(func => typeof func === 'function')
-
   if (funcs.length === 0) {
     return arg => arg
   }
--- a/test/compose.spec.js
+++ b/test/compose.spec.js
@@ -21,15 +21,15 @@ describe('Utils', () => {
       expect(compose(c, a, b)(final)('')).toBe('cab')
     })
 
-    it('composes only functions', () => {
+    it('throws at runtime if argument is not a function', () => {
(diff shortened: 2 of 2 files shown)


chicoxyzzy:
As a developer I expect my code to throw an Error not eat it. Also I strongly opposite to extra useless computations in redux. Why not to filter functions outside of `compose` if you really need this?
So big 👍 to this PR

alexeyraspopov:
This breaks the @evilj0e's use case described in https://github.com/reactjs/redux/pull/2073

alexeyraspopov:
But throwing errors explicitly is definitely a better thing.

jimbolla:
What if it were to only filter out falsey?

chicoxyzzy:
@jimbolla why filter something at all?
The problem described in #2073 solves in many ways by one-liner. But the fix from #2073 is a fix of a particular codebase and it leads to ommited errors and performance regressions in others people code

chicoxyzzy:
One can filter everything he wants to outside of compose. It's so easy I can't even

timdorr:
Good points being made. Thanks!

PR #2169, opened 2016-12-27
Update variables to ES6

### LINKED ISSUE #2168: Update variable assignment to ES6 usage
Currently redux uses pre-ES6 variable assignment, as shown [here](https://github.com/reactjs/redux/blob/master/src/applyMiddleware.js#L21). Per @gaearon it was [due to flow failing on let/const a long time ago](https://twitter.com/dan_abramov/status/811368841964097536). The issue is no longer relevant now.

Will be working on a PR to update all variables today. It also seemed that there is interest in adding in semicolons. I could do that as well, any thoughts?

dfrownfelter:
Fixes #2168. Updated all variable declarations to respective ES6 variants when applicable. All tests passed. No other code changed.

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -2,11 +2,11 @@ import { ActionTypes } from './createStore'
 import isPlainObject from 'lodash/isPlainObject'
 import warning from './utils/warning'
 
-var NODE_ENV = typeof process !== 'undefined' ? process.env.NODE_ENV : 'development'
+const NODE_ENV = typeof process !== 'undefined' ? process.env.NODE_ENV : 'development'
 
 function getUndefinedStateErrorMessage(key, action) {
-  var actionType = action && action.type
-  var actionName = actionType && `"${actionType.toString()}"` || 'an action'
+  const actionType = action && action.type
+  const actionName = actionType && `"${actionType.toString()}"` || 'an action'
 
   return (
     `Given action ${actionName}, reducer "${key}" returned undefined. ` +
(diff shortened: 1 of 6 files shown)


markerikson:
Looks pretty good to me.  My only real question is whether we want to put this straight into master, or into the `next` branch that @jimbolla has been playing with.  Jim, @timdorr , thoughts?

jimbolla:
I don't see any problem with either. Should we also add the eslint rules to forbid `var`, and to prefer `const` over `let` when possible?

markerikson:
Sure, might as well.

kennetpostigo:
I remember seeing a chrome benchmark of let and const vs var a couple weeks ago and var seemed to be optimized by chrome. Not sure if that matters much

src/combineReducers.js
```diff
@@ -115,13 +115,13 @@ export default function combineReducers(reducers) {
       finalReducers[key] = reducers[key]
     }
   }
-  var finalReducerKeys = Object.keys(finalReducers)
+  const finalReducerKeys = Object.keys(finalReducers)
 
   if (NODE_ENV !== 'production') {
     var unexpectedKeyCache = {}
```
dfrownfelter: This is one spot where I couldn't change to let, without errors failing.

dfrownfelter: A fix I had was declaring `let unexpectedKeyCache` above the conditional check. But I don't know if you guys want an undefined variable floating around in production.

dfrownfelter: On second thought, the `let` fix should be fine, because if it's using variable hoisting as it is. Then defining let before the block is no problem. Let me know what you guys think.

jimbolla:
You could disable the rule on that line with `// eslint-disable-line (rule-name)` and add a comment why it's not changeable right now.

dfrownfelter:
@jimbolla I think defining `let unexpectedKeyCache` above the conditional is fine. It seems the original `var` declaration is just using hoisting, So defining with `let` in the main scope works the same. All tests passed when I did it.

dfrownfelter:
Everything is good to go now. I can also add the eslint rules too. Just let me know what level you want them at.

jimbolla:
Error is fine I guess.

dfrownfelter:
Eslint rules added. All lint rules passed. All tests passed. Should be good to go now.

jimbolla:
LGTM!

PR #2171, opened 2016-12-28
Update normalizr in example to 3.0.0

paularmstrong:
# Problems

1. There are old references to the old Normalizr location
2. Normalizr is at 3.0.0 now and has a slightly different API. The "real world" example is now out of date.

# Solutions

1. Update all references
2. Update usage of normalizr and package version according to https://github.com/paularmstrong/normalizr/blob/master/CHANGELOG.md

# ~~Thoughts~~

~~This could wait until after normalizr is out of Beta, but I'm seeing issues of new people coming to redux and normalizr and getting confused about usage, since `npm install normalizr` and `yarn add normalizr` always install the newest version, regardless of alpha/beta status.~~


timdorr:
I would consider doing what we do on both react-redux and react-router, which is push any prerelease version (alpha, beta, rc) under a `next` dist-tag on npm and keep `latest` for only final, production-quality releases. Note that you can fix this with `npm dist-tag rm` and `npm dist-tag add` without having to unpublish anything or bump versions.

I would rather do that than encourage people to use a beta version at the moment. Looks like you're working hard to make the final release happen soon. So, unless it's going to be a few months, let's wait until you go final. We can coordinate this PR with that release too.

paularmstrong:
@timdorr We've just published v3.0.0 (non-beta). I've updated this PR with the appropriate package reference as well.

timdorr:
That was quick! Congrats! This all looks good then.

Issue #2130, opened 2016-11-30
Typescript definition does not allow for enhancer composition


santiagoaguiar:
**What is the current behavior?**
```typescript
export type StoreEnhancerStoreCreator<S> = (reducer: Reducer<S>, preloadedState?: S, enhancer?: StoreEnhancer<S>) => Store<S>;
```

Does not allow an enhancer argument, therefore the 

This doesn't let you create a enhancer like `https://github.com/reactjs/redux/blob/master/src/applyMiddleware.js` with typescript, since you can't chain the enhancers.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar.**

```typescript
import * as Redux from "redux";
interface State {}
export const enhancer: Redux.StoreEnhancer<State> = createStore => (reducer, preloadedState, enhancer?) => {
    return createStore(reducer, preloadedState, enhancer) ;
}
```
Does not compile.

**What is the expected behavior?**
Should compile.

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**
3.6.0

Given https://github.com/reactjs/redux/issues/1648 I may be missing something, but I think changing it to
```typescript
export type StoreEnhancerStoreCreator<S> = (reducer: Reducer<S>, preloadedState?: S, enhancer?: StoreEnhancer<S>) => Store<S>;`
``` 
would be enough until that issue is resolved.

aikoven:
I guess you're right, however as you can see in [`createStore`](https://github.com/reactjs/redux/blob/52910a3e8ce5996c5910e00250f7273d9875b320/src/createStore.js#L50) code, `enhancer` isn't passed to the enhanced store creator:
```js
function createStore(reducer, preloadedState, enhancer) {
  // ...
  return enhancer(createStore)(reducer, preloadedState)
  // ...
}
```
Thus, in `applyMiddleware` the value of `enhancer` is almost always `undefined`, so you can't rely on it inside the enhancer. You can use `compose` instead.

begincalendar:
@santiagoaguiar I'm not sure this issue is dependent on #1648, because as Dan has labelled it, that issue just appears to be an enhancement, not a bug (like this issue).

Having said that, I'm confused by this...

To me it appears the TypeScript definition for `StoreEnhancer` is correct and that the problem is actually that the enhancer API (and implementation) do not allow for composition.

Yet that doesn't seem right, because I have seen no other mention of this train of thought.

[...]

PR #2175, opened 2016-12-30
allow StoreEnhancerStoreCreator to have 3rd param

TotooriaHyperion:
createStore with middleware should allow a third param to pass enhancers in.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -218,7 +218,7 @@ export interface StoreCreator {
  */
 export type StoreEnhancer<S> = (next: StoreEnhancerStoreCreator<S>) => StoreEnhancerStoreCreator<S>;
 export type GenericStoreEnhancer = <S>(next: StoreEnhancerStoreCreator<S>) => StoreEnhancerStoreCreator<S>;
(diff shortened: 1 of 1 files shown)


aikoven:
See https://github.com/reactjs/redux/issues/2130

aikoven:
The code looks good, but is it correct? Is there a need for enhanced store creator to accept another enhancer as third argument? Consider this:
```js
const myEnhancer = createStore => (reducer, preloadedState, enhancer) => {
  let store = createStore(reducer, preloadedState, enhancer);
  // do something with store
  return store;
};
```
When using `myEnhancer` like this:
```js
const store = createStore(reducer, preloadedState, myEnhancer);
```
this results in a call `myEnhancer(createStore)(reducer, preloadedState)`, so enhanced store creator doesn't get `enhancer` argument.

The only way I see when it could be useful is to somehow modify `enhancer` argument, which would be possible if we used it like this:
```js
const enhancedCreateStore = myEnhancer(createStore);
const store = enhancedCreateStore(reducer, preloadedState, someOtherEnhancer);
```
But I don't think that implementation of store enhancer should be dependent on the way it's used. Am I wrong?

timdorr:
Closing

PR #2181, opened 2017-01-05
Added stricter function type checking for create store

andersontr15:
Added stricter function type checking for create store.

E.g.: Running typeof Number will return Function.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -37,20 +37,20 @@ export const ActionTypes = {
  * and subscribe to changes.
  */
 export default function createStore(reducer, preloadedState, enhancer) {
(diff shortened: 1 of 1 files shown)


brigand:
@andersontr15 I don't think this changes the behavior except in edge cases (@@toStringTag, overriding Object.prototype.toString). Number is a valid function, and `Object.prototype.toString.call(Number)` returns `"[object Function]"`.

Btw, you have lint errors.  You can run `npm run lint:src` to see them locally.

timdorr:
The `typeof` checks are more than sufficient here. If there's something I'm missing, a test would be the best way to explain it.

PR #2182, opened 2017-01-05
Mapped type for combineReducers in index.d.ts

mkusher:
From #2177 

1. Added generic interface for Dispatch(from master)
2. Updated reducer type to support strict null checks
3. Added mapped type for combineReducers, so that result reducer type can be inferred from arguments


timdorr:
Thanks!

drewnoakes:
This looks great and I'd love to use it. As it's merged to `next`, is there an npm package containing this definition file somewhere?

aikoven:
@drewnoakes For now you can use updated typings by adding this to your `package.json` dependencies:
```js
"redux-next": "reactjs/redux#next"
```
And telling TypeScript to look for Redux typings there:
```js
// tsconfig.json
{
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "redux": ["node_modules/redux-next"]
    }
  }
}
```

drewnoakes:
@aikoven thanks very much, this looks great.

To confirm, `redux-next` is a different NPM package. It should be used in place of `redux`, or are you suggesting using `redux-next` in `dev-dependencies` and only taking its type declarations? Is that even possible?

aikoven:
It is not an NPM package, `"reactjs/redux#next` means the files from Redux GitHub repository on the `next` branch. So you'll end up with two versions of Redux in your `node_modules`, but the `next` one will be called `redux-next`. Then you're just telling TS where the typings for `redux` are.

MichaelTontchev:
@aikoven sorry to necro this old thread, but when can we expect this change to go through to @types/redux ?

aikoven:
@MichaelTontchev Redux typings are bundled with `redux` NPM package so you don't have to install `@types/redux` separately.

Unfortunately, we can't release this change without bumping major version of `redux`, so we have to wait indefinitely until a critical mass of breaking changes forms v4 of Redux.

I'd be happy to put Redux typings to `@types`, but currently, it would mean that they must be kept in DefinitelyTyped which is not convenient for multiple reasons. Hopefully one day we can use `@types` while maintaining typings in a separate repo. And I guess this will happen before v4.

MichaelTontchev:
Got it.

Just curious: any idea why there is a separate `@types/redux`, then? Is it different from the typings maintained in this repo?

PR #2186, opened 2017-01-06
Deletes no-longer-valid applyMiddleware test.

jimbolla:
Behavior has been replaced with the "it warns when dispatching during middleware setup" test in the 'next' branch.


jimbolla:
I just gonna assume nobody has any objections to deleting a failing test that is explicitly commented as not-valid for 'next' and already has a replacement test for the changed behavior.

PR #2187, opened 2017-01-06
Update combineReducers.js

slorber:
This may not seem obvious to many of us, but we can write

```javascript
const bootstrapDataReducer = (state = null, event) => {
  switch (event.name) {
    case APPLICATION_STARTED:
      return event.data.bootstrapData;
    default:
      return state;
  }
};
```
instead of 

```javascript
const bootstrapDataReducer = (state = {}, event) => {
  switch (event.name) {
    case APPLICATION_STARTED:
      return event.data.bootstrapData;
    default:
      return state;
  }
};
```

Assigning an empty object as initial state looks to me an anti-pattern, yet many people tend to do this to satisfy Redux

I think it would give a better developer experience to document the ability to use "null" instead of "undefined" as reducer state in error messages

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -10,7 +10,8 @@ function getUndefinedStateErrorMessage(key, action) {
 
   return (
     `Given action ${actionName}, reducer "${key}" returned undefined. ` +
-    `To ignore an action, you must explicitly return the previous state.`
+    `To ignore an action, you must explicitly return the previous state. ` +
+    `If you want this reducer to hold no value, you can return null instead of undefined.`
   )
 }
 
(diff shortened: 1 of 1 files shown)


markerikson:
Erm... not sure I follow what you're saying here.  Can you give some more details on the background and intent?

slorber:
@markerikson I mean that sometimes you want to create a reducer, but you don't want it to have an initial value assigned. And current error message does not really explain that "null" can be used for such usecase, it only explains that undefined is an illegal value

timdorr:
More detailed error messaging isn't a bad thing. Thanks!

PR #2189, opened 2017-01-06
Add ReduxInitAction interface to type definitions

merrywhether:
In attempting to model of all of the action types in our TypeScript app, the `@@redux/INIT` action is the odd one out, as it originates from the redux code itself. It'd be great if that were part of the redux type definitions themselves instead.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -17,6 +17,13 @@ export interface Action {
   type: any;
 }
 
+/**
+ * This is the action type that combineReducers() sends to initialize
+ * reducers.
+ */
+export interface ReduxInitAction extends Action {
+  type: "@@redux/INIT";
+}
 
 /* reducers */
 


jimbolla:
I believe INIT is considered a private and not intended to be exposed to the outside. There's a work item to make the type random.

mfollett:
@jimbolla If you wanted to do a discriminated union of actions what would you use to model the internal actions, then?

jimbolla:
I blieve the idea is that a reducer can handle an *unknown* action.

mfollett:
But, in TS you'd either need to model that unknown action with a specific shape or have a catchall that has no shape and then you lose your type safety, unless there's something I'm missing. There are gradients, you could end with `any` or `{type: string}` which would give you various levels of safety loss, but that doesn't seem great.

mfollett:
Actually, if the type is going to be random, the best you can go with is just the `{type: any}` anyway.

TrySound:
Actions shoudn't be random, but init action is internal one which do not required to be public. Typescript is for static analysis of your code, not libraries code.

merrywhether:
Cool, sounds like a bail then.

PR #2194, opened 2017-01-10
Moved data request into componentDidMount

BTMPL:
As far as I know Fiber will introduce changes to the mounting pipeline which can result in componentWillMount being called and the component not mounting (aborted, stalled) I think we should suggest to perform data fetching in componentDidMount.

### DIFF
--- a/examples/real-world/src/containers/UserPage.js
+++ b/examples/real-world/src/containers/UserPage.js
@@ -22,7 +22,7 @@ class UserPage extends Component {
     loadStarred: PropTypes.func.isRequired
   }
 
-  componentWillMount() {
+  componentDidMount() {
     loadData(this.props)
   }
 


brigand:
Also server rendering.

The downside, is that this causes an extra render if the store updates to be in a loading state, right?

BTMPL:
Yes, but same thing will happen with a long running request started in `componentWillMount()`

The better solution would be to have the request sent before trying to mount the component at all (for example by implementing a loader function in react-router) but this might be out of scope of this simple use case.

timdorr:
Let's not try to future-proof against something that hasn't even come to pass yet. While they've hinted at this change, the React team hasn't yet committed to it and there will likely be mitigation for this in the future.

acdlite:
FWIW, side-effects in `componentWillMount` are definitely not supported, regardless of whatever other churn the API experiences.

Issue #2200, opened 2017-01-12
Possible error in applyMiddlewares/createStore


fourlastor:
I was checking out the source code for https://github.com/reactjs/redux/blob/master/src/applyMiddleware.js#L20 and https://github.com/reactjs/redux/blob/master/src/createStore.js#L50 and noticed one thing:

```javascript
// createStore.js
return enhancer(createStore)(reducer, preloadedState);
// applyMiddleware.js
return (createStore) => (reducer, preloadedState, enhancer) => {
```

So as far as I understood, `applyMiddleware`'s (and any store enhancer's) signature is `(createStore) => createStore`, while `createStore` is assuming `(createStore) => (reducer, preloadedState)`.

Which one should be the correct one? I would assume `(createStore) => createStore`.

jimbolla:
See https://github.com/reactjs/redux/pull/2128#issuecomment-263581319

Issue #2131, opened 2016-11-30
an unnecessary argument in applyMiddleware.js


HQidea:
Since there is no `enhancer` argument passed to an enhancer, so this argument can be omitted.

see: https://github.com/reactjs/redux/blob/master/src/applyMiddleware.js#L20

If this is ok, I'll send a pr later.

jimbolla:
Duplicate: https://github.com/reactjs/redux/pull/2128#issuecomment-263581319

PR #2201, opened 2017-01-12
Refactors applyMiddleware to reduce confusion around createStore args.

jimbolla:
Because #2200 #2131 https://github.com/reactjs/redux/pull/2128 #2028...

### DIFF
--- a/src/applyMiddleware.js
+++ b/src/applyMiddleware.js
@@ -17,8 +17,8 @@ import compose from './compose'
  * @returns {Function} A store enhancer applying the middleware.
  */
 export default function applyMiddleware(...middlewares) {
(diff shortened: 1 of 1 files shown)


markerikson:
Heh.  Sure, why not.  

I'm really rather surprised this is such an ongoing thing.  Must be a number of people digging through the source for some reason.

jimbolla:
Right? It's interesting that that's the thing people keep fixating on. I'm not sure what the lesson to learn from that is.

aikoven:
Also https://github.com/reactjs/redux/issues/2130, https://github.com/reactjs/redux/pull/2175

PR #2202, opened 2017-01-13
Update AsyncActions doc

shanemcgraw:
As the example shows in `index.js` right below, it runs: 

`import { selectSubreddit, fetchPosts } from './actions'`

...but up until that point, `fetchPosts` is not exported from the `actions.js` file and would be unavailable to the `index.js` module. Changing such that `fetchPosts` is exported.


markerikson:
Thanks!

Issue #2184, opened 2017-01-05
Switch React tests to react-test-renderer


timdorr:
Since we've switched over to Jest and [they recommend using `react-test-renderer`](https://facebook.github.io/jest/docs/tutorial-react.html#content), we should stay in sync with that guidance. 

The previous plan (#1481) was to use enzyme, but that doesn't support Jest's snapshot tests. I would like to add some of those as well, since they're a nice quick way to ensure you haven't broken anything unintentionally (or have missed a non-deterministic test that is falsely passing). 

Lastly, we'll need to make sure any docs are updated for this as well. More and more tooling and projects are using Jest, so converging community focus on it is a good thing. It's pretty swell!

Hemko:
@timdorr

Hi Tim. I don't have a lot of experience with react and jest, but i really want to help you and improve my skills. I'm ready to work hard and a lot.

brigand:
@timdorr you can use enzyme with snapshot tests, and even configure it so that `expect(wrapper).toMatchSnapshot()` works.

```js
  // package.json
  "jest": {
    "snapshotSerializers": ["enzyme-to-json/serializer"]
  }
```

Btw, I'm not saying enzyme is the better choice here, just that it's still an option.

jimbolla:
I like the idea of redux and react-redux using the same testing tools that ship with create-react-app.

borisyankov:
@timdorr I don't think you are correct. 
Enzyme and Snapshot testing are in no way mutually exclusive.

In fact, Enzyme is still in the Jest documentation, on the same page you linked to.

timdorr:
Ah, looks like that's my mistake then. I wasn't suggesting that we replace all Enzyme DOM testing with snapshots and call it a day, but I was wrong about them being mutually exclusive. 

Nonetheless, throwing in another tool into the toolbox is something I'm hesitant about because it's one more thing to learn. I'd like to reduce fatigue wherever possible, so using the "batteries included" approach to React testing is probably better than layer more knowledge onto our users. 

In any case, snapshot testing is great stuff, so we should start using it rather then worrying about implementation specifics. I do this for all network traffic on my API servers (Rails apps with the excellent [vcr](https://github.com/vcr/vcr)), so this is the DOM equivalent to that. It also makes writing tests a little bit easier, so others can contribute better tests with less effort. Should be a win-win all around.

timdorr:
[...]

PR #2203, opened 2017-01-13
Snapshot tests

Hemko:
#2184 

Added snapshot tests (todomvc and tree-view examples).
https://github.com/reactjs/redux/issues/2184#issuecomment-271313696

### DIFF
--- a/examples/todomvc/src/components/Footer.spec.js
+++ b/examples/todomvc/src/components/Footer.spec.js
@@ -35,10 +35,9 @@ const getTextContent = elem => {
 
 describe('components', () => {
   describe('Footer', () => {
-    it('should render container', () => {
+    it('should render container and filters', () => {
       const { output } = setup()
-      expect(output.type).toBe('footer')
-      expect(output.props.className).toBe('footer')
+      expect(output).toMatchSnapshot()
     })
 
     it('should display active count when 0', () => {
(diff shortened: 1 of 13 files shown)


brigand:
I don't think **replacing** tests with snapshots is good. They should be in addition to manual tests.

Hemko:
Hm... I thought it was means replacing, not adding.

> Try to find some tests that are checking output and change them to snapshot tests.

Ok, i can just add this tests.

Also, i got some troubles. Why this happens?
```
- Snapshot
+ Received

- onChange={[Function mockConstructor]}
+ onChange={[Function anonymous]}
```

brigand:
I think the CI build failures are because of the difference between these in ES5, but them being the same in ES6:

```js
var f = function() {};
var f = function f() {}
```

I'm not sure why babel-jest isn't fixing this.

Hemko:
@brigand Now it's ok?

brigand:
@Hemko looks good to me, but I'm not an authority on this repo. 

For older node versions, what about just disabling snapshots on them? Since there are other tests, it's not a big deal.

Hemko:
@brigand 
Can you suggest the correct way to disabling snapshots for older node version?
I tried this way:
`      if (nodeMajorVersion > 5) {
        expect(output).toMatchSnapshot()
      }
`
But i got the next result:
> 16 obsolete snapshots -> snapshot failure

brigand:
@Hemko apparently this is supposed to be [fixed already](https://github.com/facebook/jest/issues/2618). Try `yarn add jest@18.0.0. I have that version at work and it works correctly. 18.1.0 is currently installed here according to yarn.lock.

For now, we could create a helper that recursively iterates over the output and replace each function with `function f(){}`.

Hemko:
@brigand
What did you mean under helper?
Function which will change the render output?

brigand:
@Hemko on second thought, filtering out the function props would be a pretty crazy solution. By helper I just meant a function that the tests could use to augment the output.

Hemko:
@brigand I don't know how to do that. Can you show some examples?

Hemko:
Perhaps someone knows how to fix this problem?

Additional Information:
https://github.com/facebook/jest/issues/1681

PR #2205, opened 2017-01-15
SSR client: Free memory used for initial state

wmertens:
If the reference is not removed, it cannot be GC'd. Once the store is created, there is no use for the global `window.__PRELOADED_STATE__` variable.


timdorr:
Seems reasonable. Thanks!

PR #2206, opened 2017-01-15
SSR: de-fang JSON embed in example code

wmertens:
It's explained in the doc but it's better to not have any copy/pastable code with security issues.


timdorr:
Thanks!

PR #2208, opened 2017-01-17
Improve combineReducers type safety (TypeScript)

drewnoakes:
This PR flows names and types of state/reducers through the `combineReducers` function.

Using TypeScript's Mapped Types, validate the reducer map object contains all required properties, and that the child reducers return the correct types.

For example:

    interface RootType {
      users: IUser[],
      sessions: ISession[];
    }

    const users = (state: IUser[]) => state; // NOOP
    const sessions = (state: ISession[]) => state; // NOOP

    const reducer = combineReducers<RootType>({ users, sessions });

That worked before, as it does with this PR.

With this change, various errors are caught at compile time:

- Mismatched child reducer name vs. `RootType` property name
- Missing child reducers
- Child reducers that return different types

Current user code that doesn't specify the type on the call to `combineReducers` has the type inferred from the reducer map, so this change doesn't limit them, although the returned reducer's type reflects the reality of the data it will produce.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -47,9 +47,11 @@ export type Reducer<S> = <A extends Action>(state: S, action: A) => S;
 
 /**
  * Object whose values correspond to different reducer functions.
+ *
+ * @template S Combines state object type.
  */
-export interface ReducersMapObject {
-  [key: string]: Reducer<any>;
+export interface ReducersMapObject<S> {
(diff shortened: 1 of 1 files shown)


aikoven:
Already merged into the `next` branch: https://github.com/reactjs/redux/pull/2182

drewnoakes:
Thanks! Looking forward to being able to use it.

PR #2212, opened 2017-01-20
Update applyMiddleware.js

cheerful-li:
## make code more comprehensible ##

when i get into the sourcecode, there get me confused:

middlewareAPI 's dispatch function  mean to use the original store.dispatch, not the composed one later


markerikson:
Hi.  Thanks for the submission.  Although I don't have links right this second, this has been discussed in several previous PRs.  We're not looking to make any further changes to this bit of code right now.  @jimbolla _is_ doing some investigation of alternate approaches for handling this sort of thing.

PR #2218, opened 2017-01-27
Update Middleware.md

lavystord:
change the naïve to naive


onbjerg:
"Naïve" is also valid.

PR #2220, opened 2017-01-28
Added article "a" for grammar consistency

tyreer:
It seems there's a missing article here for the indefinite noun "function", which is referred to as a singular noun with the article "a" in previous sentences.


markerikson:
Thanks!

PR #2222, opened 2017-01-30
Remove .d.ts suffix for imports

BafS:
A little detail but the suffix '.d.ts' is not required for typescript imports (https://www.typescriptlang.org/docs/handbook/module-resolution.html).

The ts linter tell me `[ts] An import path cannot end with a '.d.ts' extension. Consider importing '../../index' instead.`.


timdorr:
Cool. Thanks!

PR #2224, opened 2017-02-02
Remove `Symbol()` from real-world example

### LINKED ISSUE #2217: Remove or polyfill the use of Symbol in the examples
The  [real word example](https://github.com/reactjs/redux/blob/master/examples/real-world/src/middleware/api.js#L76) makes use of JavaScript Symbols. This doesn't work in older browsers¸ also [Create React App](https://github.com/facebookincubator/create-react-app/blob/d1250743adc2abc41d80d566c5f817e1a16da279/packages/react-scripts/config/polyfills.js) doesn't provide any polyfill for Symbols.
I believe that this should be removed from the example. Because It is only used once, and a polyfill needs some [optimizations](https://github.com/facebookincubator/create-react-app/issues/170#issuecomment-235304765) require it. 
What do you think?

yefremov:
**Problem summary**
Real-world example doesn't work on browsers that are not supporting symbol data types.

**Change summary**
Updated `CALL_API` action key to be string primitive.

**Fix verification summary**
Tested on Windows, Internet Explorer 11

Resolves #2217

### DIFF
--- a/examples/real-world/src/middleware/api.js
+++ b/examples/real-world/src/middleware/api.js
@@ -73,7 +73,7 @@ export const Schemas = {
 }
 
 // Action key that carries API call info interpreted by this Redux middleware.
-export const CALL_API = Symbol('Call API')
+export const CALL_API = 'Call API'
 
 // A Redux middleware that interprets actions with CALL_API info specified.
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

PR #2226, opened 2017-02-02
Fixed undefined todo id

rghossi:
When running the basic demo app I was getting undefined for todo id because it was being called as todo.id (but it does not have an id field). Using the array index solved the problem.


timdorr:
The todos do have an id field (check the `ADD_TODO` hander of the reducers): https://github.com/reactjs/redux/blob/master/docs/basics/ExampleTodoList.md#reducerstodosjs

PR #2231, opened 2017-02-05
update combineReducers typings for ts 2.1

Rendalf:
Since TypeScript [corresponds](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html) `keyof` and `Lookup Types` we can check shape of passed reducers map.

Next code is valid in current version, but invalid with updated typings.

```typescript
type A = number
type B = string
type MyState = {
  a: A
  b: B
}

const reducerA: Reducer<A> = (state = 0, action) => {
  ...
}

const reducerB: Reducer<B> = (state = 'hello', action) => {
  ...
}

export default combineReducers<MyState>({
  x: reducerA,
})
```

Updated typings will check shape of `combineReducers` properly.
```typescript
...
export default combineReducers<MyState>({
  a: reducerA,
  b: reducerB,
})
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -48,8 +48,8 @@ export type Reducer<S> = <A extends Action>(state: S, action: A) => S;
 /**
  * Object whose values correspond to different reducer functions.
  */
-export interface ReducersMapObject {
-  [key: string]: Reducer<any>;
+export type ReducersMapObject<S> = {
+  [P in keyof S]: Reducer<S[P]>;
 }
 
 /**
@@ -70,7 +70,7 @@ export interface ReducersMapObject {
  * @returns A reducer function that invokes every reducer inside the passed
  *   object, and builds a state object with the same shape.
  */
-export function combineReducers<S>(reducers: ReducersMapObject): Reducer<S>;
+export function combineReducers<S>(reducers: ReducersMapObject<S>): Reducer<S>;
 
 
 /* store */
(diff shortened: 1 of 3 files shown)


timdorr:
How prevalent is TS 2.1 usage? I don't want to break others' usage of these typings if they haven't upgraded TS. 

(This is exactly what I don't like about having typings in the source repo, btw)

aikoven:
Already merged into the `next` branch: https://github.com/reactjs/redux/pull/2182

@timdorr Switching typings for 2.1 would always be a breaking change. Unfortunately there are little alternatives to bundling typings with Redux. 

To publish typings to `@types` NPM org one has to add them to `DefinitelyTyped` mono-repo, but this makes them quite inconvenient to maintain. 

Another option is to publish them as a separate NPM package, but then TS won't be able to automatically discover it and users would have to explicitly link their compilation to that package like this: https://github.com/reactjs/redux/pull/2182#issuecomment-275904509. This also introduces a risk of fragmentation if no single package is considered *official*.

That said, the current situation already makes us use explicit linking (as in https://github.com/reactjs/redux/pull/2182#issuecomment-275904509) until the next major release, which I don't expect to happen soon.

I'd propose to put typings in a separate GitHub repo under https://github.com/reactjs/, publish as a separate NPM package, and make a link to it in Redux readme.

There's an ongoing discussion that would hopefully let us link that package with `@types` someday: https://github.com/Microsoft/types-publisher/issues/4

Rendalf:
Thanks to @aikoven for explanation

Won't be merged because it already exists into the `next` branch

timdorr:
Thanks for the insight, @aikoven!

PR #2233, opened 2017-02-08
Update MainSection.js

inkinworld:
use arrow function ,the bind() make no sense.


timdorr:
Fixed a small typo. We'll see if that passes tests (it should) and then this can get merged in.

PR #2234, opened 2017-02-09
Destructure and use shorthand object literal syntax

Yahkob:
Seems consistent to destructure and use object literal for this function


timdorr:
I actually prefer the long form in this case, as it's consistent with the docs and shows that you're getting state as an argument to that function.

Yahkob:
Ah good point @timdorr

PR #2236, opened 2017-02-11
Updated makeActionCreator

mahadazad:
Added support for pre-defined values in makeActionCreator


timdorr:
I'd rather keep it simple for this example, given that this is just documentation. Thanks anyways!

PR #2239, opened 2017-02-12
Grammar fix

coryhouse:


markerikson:
Mmm... this one's kinda iffy :)  If anything, I think leaving "modfying" in place and adding "making" in front of "API calls" might be more correct.

MichaelTontchev:
The fix is correct, actually. You have to read

> which means it does not mutate its arguments, perform side effects like API calls or modify values outside of

as
" which means it ((does not (mutate its arguments), (perform side effects like API calls), (modify values outside of ...)"

You distribute the "does not" to all the phrases.

markerikson:
Eh, sure :)

MichaelTontchev:
Clearly not a pressing fix, just wanted to explain it :P

jddxf:
I agree with @markerikson, as here're three "or"s.The first and the third are alike, both in the form of "like A or B".

> which means it does not mutate its arguments, perform side effects like API calls **_or_** modifying values outside of the function, **_or_** call non-pure functions like Date.now() **_or_** Math.random().

MichaelTontchev:
Oh snap, @jddxf , you're right. I was parsing it incorrectly because I didn't get to the last or!

coryhouse:
Perhaps the right move it to reword the long sentence that we're all struggling to properly parse into a bulleted list.

PR #2242, opened 2017-02-14
Update ExampleTodoList.md

EricWuuu:
There is a new 'src/actions/index.js' file and not used. Should update all the import in the container files.


timdorr:
Node and webpack will resolve the `../actions` import to `../actions/index.js`, so it's correct as-is.

PR #2253, opened 2017-02-21
use reducerKey to run reducer

wisdomofgod:
Why not use "reducerkey" to perform only one reducer, instead of executing the loop？

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -15,6 +15,7 @@
  */
 export interface Action {
   type: any;
+  reducerKey?: string;
 }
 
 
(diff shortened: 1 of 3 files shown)


timdorr:
Most reducers should be noops that just return back the same state provided to them. This is both highly optimizable by most JS engines (v8 would make the reducer a hidden class, so access to each key is very very cheap) and not the bulk of the computational load when a dispatch occurs. 

Also, since the plan is to allow reducers running inside combineReducers to consult the overall state, we can't guarantee that a single property of the combined reducers will be affected by a dispatch. That means this optimization will end up producing errors, which could lead to inconsistent state. Predictable and reliable state transition behavior is key to Redux.

PR #2258, opened 2017-02-22
pass current state to subscribers

matanwerbner:
at least in it's vanilla form - it will be usefull if store.subscribe() will have the current state as an argument, instead of needlessly calling store.getState() inside.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -175,7 +175,7 @@ export default function createStore(reducer, preloadedState, enhancer) {
     const listeners = currentListeners = nextListeners
     for (let i = 0; i < listeners.length; i++) {
       const listener = listeners[i]
-      listener()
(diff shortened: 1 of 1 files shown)


markerikson:
This has been proposed numerous times, and is not viable, primarily due to interference with batched dispatches.  Please see #580 and #2214 for discussion.

gaearon:
I don't think it has to do with batching. Batching is problematic when we pass `action`. However passing `state` would be fine. The *problem* with it is that it would burden any enhancers (e.g. DevTools) with reimplementing subscriptions. This is solved by https://github.com/reactjs/redux/pull/1702 but we didn't go ahead with it, and I'm not confident about direction in https://github.com/reactjs/redux/issues/2214, so we likely won't ever fix this.

PR #2265, opened 2017-02-26
Fix issue : 2264

Gregoirevda:
If no static directory is defined, the handleRender middleware function will be called when the browser fetches for `/static/bundle.js` and will give back the index.html

### DIFF
--- a/docs/recipes/ServerRendering.md
+++ b/docs/recipes/ServerRendering.md
@@ -50,6 +50,8 @@ import App from './containers/App'
 const app = Express()
 const port = 3000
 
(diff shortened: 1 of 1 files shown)


docs/recipes/ServerRendering.md
```diff
@@ -50,6 +50,9 @@ import App from './containers/App'
 const app = Express()
 const port = 3000
 
+//Serve static files
+app.use('/static', Express.static('static'));
```
Gregoirevda: Otherwise, the /static/bundle.js file can't be served

timdorr:
Thanks!

PR #2266, opened 2017-02-28
Correct example codes in UsageWithReactRouter.md

tonnamb:
List of changes:

containers/VisibleTodo.js 
- I have added the definition of getVisibleTodos because the case changed from 'SHOW_ALL' (original Todo tutorial) to 'all' (react-router tutorial) and vice versa for 'completed' and 'active'

containers/FilterLink.js
- This is a correction in which it should link to '/' for the 'all' link

add updated definition of index.js since it so different from the original Todo tutorial

### DIFF
--- a/docs/advanced/UsageWithReactRouter.md
+++ b/docs/advanced/UsageWithReactRouter.md
@@ -128,7 +128,7 @@ import { Link } from 'react-router';
 
 const FilterLink = ({ filter, children }) => (
   <Link
-    to={filter === 'all' ? '' : filter}
+    to={filter === 'all' ? '/' : filter}
     activeStyle={{
       textDecoration: 'none',
       color: 'black'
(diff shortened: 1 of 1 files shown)


docs/advanced/UsageWithReactRouter.md
```diff
@@ -174,6 +174,19 @@ Now if you click on `<FilterLink />` you will see that your URL will change from
...
+			return todos
+		case 'completed':
+			return todos.filter(t => t.completed)
+		case 'active':
+			return todos.filter(t => !t.completed)
+		default:
+			return todos
+	}
+}
+
```
timdorr: Rather than duplicating this code, can you simply make it work with `SHOW_ALL`?

docs/advanced/UsageWithReactRouter.md
```diff
@@ -204,6 +217,22 @@ const App = ({ params }) => {
...
+import Root from './components/Root'
+
+let store = createStore(todoApp)
+
+render(
+	<Root store={store} />,
+	document.getElementById('root')
+)
+```
+
```
timdorr: This belongs up below the `components/Root.js` definition. It doesn't have any relation to this section of the doc.

tonnamb: I agree, correcting this.

docs/advanced/UsageWithReactRouter.md
tonnamb: Great idea!!

timdorr:
Thanks!

PR #2275, opened 2017-03-03
simplify compose to include only one special case

flipactual:
Saw an opportunity to remove a special case from the compose method by passing an initial value to `reduce`

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -9,7 +9,5 @@
  * (...args) => f(g(h(...args))).
  */
 export default function compose(...funcs) {
-  return funcs.length === 1
-    ? funcs[0]
-    : funcs.reduce((a, b) => (...args) => a(b(...args)), arg => arg)
+  return funcs.reduce((a, b) => (...args) => a(b(...args)), arg => arg)
 }
--- a/src/compose.js
+++ b/src/compose.js
@@ -8,15 +8,8 @@
  * from right to left. For example, compose(f, g, h) is identical to doing
  * (...args) => f(g(h(...args))).
  */
-
 export default function compose(...funcs) {
-  if (funcs.length === 0) {
-    return arg => arg
-  }
-
-  if (funcs.length === 1) {
-    return funcs[0]
-  }
-
-  return funcs.reduce((a, b) => (...args) => a(b(...args)))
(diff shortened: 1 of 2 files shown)


markerikson:
Hmm.  I appreciate the PR, and generally agree that having fewer special cases is a good thing.  On the other hand, we've had a lot of churn and attempted churn in the various utility functions and stuff like `applyMiddleware`, and I'm hesitant to make changes "just because", even if the tests _do_ pass.

flipactual:
This is what I get for making a late night pull request from a train...

@markerikson I've updated to remove both of the special cases. And it does slightly change the logic, so I entirely understand if you want to hold off.

Would this be an acceptable change down the line?

jimbolla:
The current code is identical to that in [recompose's implementation](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/compose.js) and I think it should always stay that way.

flipactual:
@jimbolla is there a reason, in that case, to not use `compose` from recompose or to break out a shared library?

I'm happy to open this PR upstream, but then this will be out of sync if it's merged.

jimbolla:
It's not worth the effort to make the change, IMO. It's a short simple function that has no real reason to change.

flipactual:
@jimbolla I opened the PR because the function is far from simple with two special cases and somewhat inconsistent returns (everything but the 1 function special case being wrapped)

I understand holding off on making changes, but if it makes the codebase simpler to understand (this function does not currently always do what the jsdoc block says, my change fixes that) and removes logic that was never necessary, I'm hard pressed to understand opposition

gaearon:
Poor `compose`. Everyone who reads Redux source tries to rewrite it. 😄

gaearon:
I appreciate the intent, but I don't think the change is significantly better to warrant another change to a function [that everyone](https://github.com/reactjs/redux/commits/6de14f4881e5fc6a4a6e0ce86e88a967d46802a3/src/utils/compose.js) [keeps changing](https://github.com/reactjs/redux/commits/master/src/compose.js), either for readability, or for performance. I would expect that composing a single function would just return that function, so I’d probably not take this PR. Thanks for sending it though!

PR #2281, opened 2017-03-06

sam-rubin:


markerikson:
Hi.  Appreciate the PR, but I don't see any need to change this.  It's totally fine to define the "presentational" component in the same file that the connection is declared in.  In fact, if anything in the file should be changed, it'd be dropping the re-assignment and just doing `export default connect(null, mapDispatch)(AddTodo)`.

Issue #2245, opened 2017-02-16
Produce main bundle with Rollup


gaearon:
There are unfortunate Babel artifacts in the compiled UMD bundle.
For example:

```js
return e&&e.__esModule?e:{"default":e}}t.__esModule=!0
```

```js
// Oops, we access ["default"] in a hot path!
if(!(0,u["default"])(e))throw Error("Actions must be plain objects. Use custom middleware for async actions.")
```

We should use a better build process that doesn’t produce artifacts like this.

timdorr:
This is due to the oldIE support stuff, right? Shouldn't be happening in next branch. Scope hoisting (aka, rollup) won't fix this by itself.

gaearon:
I meant the Babel `default` indirection which goes away if we compile with Rollup.

TrySound:
@gaearon Do you want a PR?

gaearon:
Sure.

TrySound:
@gaearon Done

Issue #2244, opened 2017-02-16
DEV code is now in production bundle in master


gaearon:
This PR broke Uglify's dead code elimination for DEV: https://github.com/reactjs/redux/pull/2030.

Uglify is not smart enough to understand that if a literal is assigned to a variable, and later that variable gets compared, it's the same as literal. As a result, DEV-only code ends up in the production bundle:

```js
{var c=t[r];"production"!==p&&void 0===e[c]&&(0,l["default"])('No reducer provided for key "'+c+'"'),"function"==typeof e[c
```

I suggest reverting #2030, at least in the current form.

timdorr:
Yeah, not sure why I didn't give that more scrutiny. Luckily, it hasn't been released.

We may want to add a quick check of the gzipped UMD bundle size to the build scripts. We do this on react router and it helps guard against ballooning the bundle size, provided you know what a "good" size looks like. 

I'll revert this once in on desktop (since it can't be done automatically).

gaearon:
>We may want to add a quick check of the gzipped UMD bundle size to the build scripts.

Yes, that would be great!

Andarist:
Do you guys know by any chance how [this](https://github.com/reactjs/redux/commit/0139a9601dae810db533762c2bc4a745a37d4cf6#commitcomment-20910083) can be avoided?

timdorr:
That sounds like a tooling problem. We don't add the polyfill ourselves, so it would seem you need to configure your tooling not to do that. It's similar to checking for a DOM with `typeof window !== 'undefined'`.

Andarist:
yeah, u have reverted your checks to `if (process.env.NODE_ENV !== 'production') {`, but the mentioned commit tried not to only cache this value, but also guarded against lack of `process` like this:
`var NODE_ENV = typeof process !== 'undefined' ? process.env.NODE_ENV : 'development'`
this has triggered (on your previous webpack config too, i see u have switched to rollup now) `process` polyfill to be included

anyhow i'll try to explore possibilities to avoid this, from that I understand Aurelia has its own bundler which does not recognize `process` and packages that use it cannot be properly transpiled by its users

timdorr:
@Andarist It's 50% smaller: https://github.com/reactjs/redux/pull/2283#issuecomment-284913574

PR #2283, opened 2017-03-07
Build umd with rollup

TrySound:
Ref https://github.com/reactjs/redux/issues/2245


TrySound:
40 461 b -> 27 793 b
10 738 b -> 7 503 b

gaearon:
CI seems to fail because of ES3ify check.
But while we're at it, why does the compiled bundle have `.default` anywhere at all?

TrySound:
Nice catch. symbol-observable has only `jsnext:main` which is disabled by default.

timdorr:
@gaearon 
> But while we're at it, why does the compiled bundle have .default anywhere at all?
That is the way Babel 6 handles `export default`. @kentcdodds [wrote about it](https://medium.com/@kentcdodds/misunderstanding-es6-modules-upgrading-babel-tears-and-a-solution-ad2d5ab93ce0#.qrp1xf6p3), but here's why:
```js
// foo.js
const foo = {baz: 42, bar: false}
export default foo
// bar.js
import {baz} from './foo'
```
That is against the spec (though it may have changed since 6.0.0), so they "broke" that behavior to fix the spec. Hence, all defaults are exported in the `default` property of the module.

gaearon:
Right, but if we're bundling with rollup, we need to let it handle ES modules instead of Babel. So we should change the rollup env to *not* do the commonjs transform.

gaearon:
@TrySound Can you post resulting bundles as gists for comparison?

timdorr:
BTW, we're down to:
```
26843 dist/redux.js
 7159 dist/redux.min.js
```
👍 

Here's the latest build artifacts for easy perusal: https://gist.github.com/timdorr/73a8d368c1d0c3d3c83aa73af8f02cbe

timdorr:
LGTM. Most of the cost savings are from the babel import/export transforms and the inter-module imports (`__webpack_require__` stuff). Surprisingly hefty! Here's hoping webpack 2 can add scope hoisting soon...

gaearon:
[This part](https://gist.github.com/timdorr/73a8d368c1d0c3d3c83aa73af8f02cbe#file-redux-js-L512) worries me a little. Does it inline in prod build?

TrySound:
Yeah. This increases bundle. Weird behavior.

gaearon:
Could be https://github.com/reactjs/redux/issues/2244, I'm not sure if it was fixed yet in master.

timdorr:
Now:
```
26843 dist/redux.js
 7023 dist/redux.min.js
```
Keep going!

TrySound:
@gaearon Yeah. Current min release is smaller because of that thing. Inference with code elimination is too hard :) That PR should be reverted.

timdorr:
I'll merge it in here and then merge in this PR. The reason for #2030 are something you can fix with tooling.

TrySound:
Do you guys want to release it as 3.* or 4.*?

timdorr:
@TrySound Can you enable contributor pushing to your branch? Should be a checkbox at the top of the PR.

timdorr:
Once I can push it, reverting that takes us down to:
```
26636 dist/redux.js
 5827 dist/redux.min.js
```
Now we're cooking.

TrySound:
Yay!

@timdorr It's enabled.

timdorr:
Cool, I think we're good. This doesn't change anything outside of the UMD build and does so in entirely a compatible way, so I think it only needs a patch version. Correct me if I'm wrong, otherwise I'll push this as 3.6.1. I don't believe we've had any other non-docs/examples updates since that release.

gaearon:
It's probably safe as a patch. Maybe as a minor for a tiny bit of extra safety.

TrySound:
Okay. I just would like to remove `jsnext:main` which is legacy. Rollup uses `module` field now. But guys still add this field looking at this and similar packages. But this will be breaking change.

timdorr:
Go ahead and do that against the `next` branch, which is our target for 4.0 stuff. 

AFAIK, [webpack is following suit](https://webpack.js.org/configuration/resolve/#resolve-mainfields), so that should be a low-impact breaking change for most folks.

timdorr:
BTW, thanks for the hustle on this!

PR #2286, opened 2017-03-08
Update HOC example in UsingImmutableJS docs

sd2k:
The HOC example in the UsingImmutableJS docs used a named export to import React, but it needs to be a default import.


timdorr:
Thanks!

PR #2287, opened 2017-03-09
Docs: rename variable "middleware" to "middlewares"

CodinCat:
Since it's an array, using `middlewares` seems more clear and also aligns the signature `applyMiddleware(...middlewares)`

### DIFF
--- a/docs/api/applyMiddleware.md
+++ b/docs/api/applyMiddleware.md
@@ -223,17 +223,17 @@ export default connect(
 * If you want to conditionally apply a middleware, make sure to only import it when it's needed:
 
   ```js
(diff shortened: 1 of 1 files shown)


timdorr:
Actually, the "ware" suffix is normally an uncountable word, so does not gain an "s" suffix when pluralized. It's a little more ambiguous in an "invented" word, but if it helps, think of "software" which also doesn't gain a suffix when pluralized.

CodinCat:
Then shouldn't we update the title (signature) to `applyMiddleware(...middleware)`, for consistency

timdorr:
Yep, got it in that file. A PR to get the other references would be appreciated!

PR #2289, opened 2017-03-13
docs(undoRedo): incorrect folder name. change it from 'containers' to 'components'

alonbt:


alonbt:
After reading this section again it turns out there is only one file that contains both presentational component and container component. therefore I'm closing this PR.

PR #2291, opened 2017-03-14
TypeScript typings: reducer state can be undefined

OliverJAsh:


TrySound:
@OliverJAsh It can't be. Redux event doesn't accept undefined in runtime. Use default value to prevent ts error.

OliverJAsh:
@TrySound If it can't be `undefined`, then why does Redux allow a default value?

The `state` in can be `S | undefined`, but the body is guaranteed to receive a `S` if there is a default.

OliverJAsh:
E.g. the type of `(x: number = 1) => x` is `(x?: number) => number` which is equivalent to `(x: number | undefined) => number` if I understand correctly?

TrySound:
@OliverJAsh undefined can be passed to reducer but can't be returned. Then to make it save reducer shouls always return non undefined value. Then there should be a value in passed state. If there's no state then create default one.
`(x: number = 1) => x` and `(x?: number) => number` are different one. In the first example x will be always a number. In the second one you will get NaN some day.

OliverJAsh:
> undefined can be passed to reducer but can't be returned. Then to make it safe reducer should always return non undefined value

Doesn't this sentence match the type definition as per this pull request, i.e. `(state: S | undefined, action: A) => S`?

TrySound:
@OliverJAsh That sentence is only for runtime. Type systems can avoid unsafe type via default parameter.

OliverJAsh:
Ah, got it. So perhaps this is a bug with TypeScript inferring `(x: number = 1) => x` as `(x?: number) => number`—it should be `(x: number) => number`.

TrySound:
In call statement you can avoid parameter and x becomes 1. So practically typescript process them similar. But with `(x?: number) => number` you need also check `typeof x === 'number'`.

PR #2294, opened 2017-03-16
docs(react): fix typo

kud:
`data` is plural, `datum` is the singular.


timdorr:
"the data" refers to the singular collection. It's weird, but it's right as-is. Thanks!

PR #2296, opened 2017-03-19
The better way to complement compose.

quanru:
It's easy to understand than the former in this way.

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -18,5 +18,5 @@ export default function compose(...funcs) {
     return funcs[0]
   }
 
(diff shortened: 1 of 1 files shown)


wuct:
The new approach loses the ability to apply multiple arguments to the first composed function.

gaearon:
I would suggest searching old pull requests for "compose". Everybody who sees that function tries to rewrite it 😅

markerikson:
Yep.  Appreciate the PR, but there's no need for further changes to this function at this point.

gaearon:
For the reference, see my last reply about this: https://github.com/reactjs/redux/pull/2275#issuecomment-284153667.

PR #2297, opened 2017-03-19
Fixed quote types inconsistency in a warning message

Andarist:
While reading the source code I've noticed that this could produce such text:
`Given action an action`
which is really unnatural to say. So I've refactored it a little to be more human-friendly and produce either:
- `Given action "SOME_TYPE"`
- or `Given an action`

Also - if you dont like this `String(variable)` casting approach I may revert this change and amend my PR. But personally I think its more obvious in what it does (casting) + its shorter by 3 chars :P

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -4,10 +4,10 @@ import warning from './utils/warning'
 
 function getUndefinedStateErrorMessage(key, action) {
   const actionType = action && action.type
-  const actionName = (actionType && `"${actionType.toString()}"`) || 'an action'
+  const actionDescription = (actionType && `action "${String(actionType)}"`) || 'an action'
 
   return (
-    `Given action ${actionName}, reducer "${key}" returned undefined. ` +
+    `Given ${actionDescription}, reducer "${key}" returned undefined. ` +
     `To ignore an action, you must explicitly return the previous state. ` +
     `If you want this reducer to hold no value, you can return null instead of undefined.`
   )


timdorr:
[We require that all actions have a `type` field](https://github.com/reactjs/redux/blob/master/src/createStore.js#L157-L162), so I'm actually not sure how that could happen. I would just remove the default text instead.

Andarist:
Yeah, I was kinda wondering why this fallback is there as Ive remember about the check you have mentioned, but thought this could be intended for some reason. Maybe its there because theoretically `combineReducers` outcome could be used outside of `redux`? And the check you have mentioned is in the `dispatch` itself.

At first I thought [this test](https://github.com/reactjs/redux/blob/a56339d0db94e5d5b85a2bfe02ec49ac51af0073/test/combineReducers.spec.js#L81) is testing exactly this, but now I see that 'an action' in the test refers to [this part](https://github.com/Andarist/redux/blob/f81e3aa272085a739dafb6b4d70071719438b636/src/combineReducers.js#L11).

That being said if you decide I should drop this fallback entirely I think [this safe-guard](https://github.com/Andarist/redux/blob/f81e3aa272085a739dafb6b4d70071719438b636/src/combineReducers.js#L11) should be dropped, as we would treat then `.type` to be available always in the reducer.

Let me know what you decide and I'll update my PR accordingly.

Andarist:
@timdorr ping ;)

this is a silly PR really, but would like to have it stop hanging as open in my PR tab, I understand though if you are too busy right now

PR #2299, opened 2017-03-19
Update to TypeScript 2.1

### LINKED ISSUE #2298: Use mapped TypeScript type definition for ReducerMapObject
Feature

The reducer map passed to `combineReducers` is not typechecked against the state in the TypeScript definitions file.

N/A

The reducer map passed to `combineReducers` will be typechecked against the state with the inclusion of type generics and mapped types which were added in TypeScript 2.1.

This would be a breaking change since TypeScript 2.1 is required for mapped types.

AlexanderOtavka:
This allows us to define the reducers map object genetically, so it can be type checked against a state object.  This is a breaking change for projects still using TypeScript versions less than 2.1.

Resolves #2298.

Let me know if there are any issues with this PR that I should take care of.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -47,6 +47,8 @@ export type Reducer<S> = <A extends Action>(state: S, action: A) => S;
 
 /**
  * Object whose values correspond to different reducer functions.
+ *
+ * @template S State object type.
  */
 export type ReducersMapObject<S> = {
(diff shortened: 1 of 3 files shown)


timdorr:
Already done on `next`: https://github.com/reactjs/redux/blob/next/index.d.ts

PR #2304, opened 2017-03-21
applyMiddleware tests are using object property enumeration order

s-coimbra21:
This is really, really minor, but object property enumeration order is not guaranteed despite being fairly consistent in browsers and node, but it doesn't hurt to take that into account and write the tests accordingly.


timdorr:
Thanks!

PR #2310, opened 2017-03-24
Update ThreePrinciples.md

andrey-ponamarev:
```const``` instead of ```let```


andrey-ponamarev:
I can't find any reasons why reducers and store defined with let operator. 
Could you explain if there are any?

timdorr:
Nope, just a random choice.

PR #2315, opened 2017-03-26
Use compose to bind action creator to keep DRY

danieluhl:
Wasn't able to find discussion so apologies if this was discussed. Tradeoffs: this will have a small performance hit due to the checks and reduce call in compose but keeps code DRY.


markerikson:
Hi.  Appreciate the PR, but I don't see a reason to change this.

PR #2316, opened 2017-03-26
docs: Update react-router docs for v4.0.0

### LINKED ISSUE #2300: How to integrate new React-Router (v4.0.0) with React Redux app?
Guys from React-Router have rolled-out new version of their plugin (v4.0.0) and as they say themselves they only have a basic idea of how it should work with Redux. 

And Redux docs is about React Router `^2.7.0` (http://redux.js.org/docs/advanced/UsageWithReactRouter.html)

Will there be an update on how to use Redux with React and new React Router (v4.0.0)?

chakrihacker:
closes #2300


timdorr:
Unfortunately, this isn't complete. The usage of the various components in the example code need to updated as well. In fact, the whole document really should be rewritten, since 4.0 is so different.

chakrihacker:
I would like to take this task

timdorr:
Feel free to submit a PR. I'll close this out for now.

PR #2318, opened 2017-03-26
Fix typings for `compose` function to accept  multiple functions (upto 3) and rest argument

veeramarni:


timdorr:
Isn't this what the previous typings support? I'm not sure what this gives you.

veeramarni:
The change supports the following combinations. Previously released version only supports three functions and Rest arguments, and this change extends to few more combination as I listed below. 

- Two functions and Rest arguments
```
    const store: Store<any> = createStore(
        reducer,
        initialState,
        compose(
            autoRehydrate(),
            ...platformStoreEnhancers,
        ),
    );
```
- One function and Rest arguments
```
    const store: Store<any> = createStore(
        reducer,
        initialState,
        compose(
            ...platformStoreEnhancers,
        ),
    );
```

PR #2323, opened 2017-03-28
upgrade outdated dependencies

jstafford:
Looking through the examples I noticed that the dependencies were getting a little stale. I updated them for my own benefit, and am making a pull request since it's was a bit of a tedious job. In most of the examples I updated the following without making any associated code changes: enzyme from ^2.4.1 to ^2.8.0, humps from ^1.1.0 to ^2.0.0, normalizr from 3.0.0 to ^3.2.2, react-redux from ^4.4.5 to ^5.0.3, react-router from ^2.6.1 to ^3.0.2, redux-undo from ^0.6.1 to ^1.0.0-beta9, todomvc-app-css from ^2.0.6 to ^2.1.0, and react-scripts from  ^0.9.3 to ^0.9.5. In the universal example there were a number of additional dependencies upgraded which are only in that project.

I also upgraded redux-logger from ^2.6.1 to ^3.0.0, which required a minor code change wherever createLogger was imported from redux-logger.

I was unable to upgrade react-router in the real-world example all the way to ^4.0.0. I think I was close to getting it working, but it was taking too long to debug, and seemed like it would need to be a pull request all on it's own. I ran into the same problem on the universal example trying to take webpack all the way to ^2.3.2.


timdorr:
Thanks!

PR #2326, opened 2017-03-29
Add composeReducers util

LucaColonnello:
Add `composeReducers` utility to split complex reducers into multiple chunks handling the same state.

I know this kind of thinks could be a library, but I guess redux should provide a bit more utilities on how to solve some common issues.

This could also help people to avoid write complex reducers and find internal custom solutions to split the code, in favour of an already provided way.

### `reducers/loggedInUser/index.js`

```js
import { composeReducers } from 'redux'
import editInfo from './editInfo';
import uploadPhoto from './uploadPhoto';


const loggedInUser = composeReducers(
  {
    // shared loader for any save or download
    loading: false,

    username: '',
    email: '',
    photos: [],
  },
  editInfo,
  uploadPhoto
)

export default loggedInUser;
```

### `reducers/loggedInUser/editInfo.js`

```js
export default function editInfo(state, action) {
  switch (action.type) {
    // might be an async operation
    case 'EDITING_LOGGED_IN_USER_INFO':
    return {
      ...state,
      loading: true
    }

    case 'SET_LOGGED_IN_USER_USERNAME':
    return {
      ...state,
      loading: false,
      username: action.payload
    }

    case 'SET_LOGGED_IN_USER_EMAIL':
    return {
      ...state,
      loading: false,
      email: action.payload
    }

    default:
    return state
  }
}
```

### `reducers/loggedInUser/uploadPhoto.js`

```js
export default function editInfo(state, action) {
  switch (action.type) {
    case 'UPLOADING_LOGGED_IN_USER_PHOTO':
    case 'DOWNLOADING_LOGGED_IN_USER_PHOTOS':
    return {
      ...state,
      loading: true
    }

    case 'ADD_LOGGED_IN_USER_PHOTO':
    return {
      ...state,
      loading: false,
      photos: [...state.photos, action.payload]
    }

    case 'SET_LOGGED_IN_USER_PHOTOS':
    return {
      ...state,
      loading: false,
      photos: action.payload
    }

    default:
    return state
  }
}
```

### DIFF
--- a/docs/api/composeReducers.md
+++ b/docs/api/composeReducers.md
@@ -1,0 +1,116 @@
+# `composeReducers(initialState, ...reducers)`
+
+Creates a new reducer composing passed in reducers from right to left.
+
+You might want to use this utility to split a complex reducer in small chunks handling the same state.
+Is always better to divide in more chunks reducers that are performing too much calculation
+for a better testing and readability.
+
+Sometimes you can simply flat your state dividing it into more parent keys and use `combineReducers`, which is fine and neat. But sometimes you might need to handle the same state across multiple reducers as they were a big one, which is where a solution like this helps you.
+
+#### Arguments
+
+1. `initialState` (*any*): the initial state to be used by default if no state is passed by redux to the created reducer.
+2. `reducers` (*Array*): The reducers to compose. Each reducer is expected to accept the state and the dispatched action. Its return value will be the new state, if the action matches, to the reducer standing to the left, and so on.
+
+#### Returns
+
+(*Function*): The final reducer obtained by composing the given ones from right to left.
+
+#### Example
+
+This example demonstrates how to use `composeReducers` to split a complex reducer in small easy to test and read ones.
+
+
+#### `reducers/loggedInUser/index.js`
+
+```js
+import { composeReducers } from 'redux'
+import editInfo from './editInfo';
+import uploadPhoto from './uploadPhoto';
+
+
+const loggedInUser = composeReducers(
+  {
+    // shared loader for any save or download
+    loading: false,
+
+    username: '',
+    email: '',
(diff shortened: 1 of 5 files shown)


markerikson:
Hi.  First, thanks for the PR.  Looks like you definitely spent some time putting this together, including docs, typings, and tests, so this is certainly a step above the usual "change a couple lines" PR we often get.  So, I appreciate the effort and the interest.

That said, my _first_ reaction is that we'd be unlikely to add any more built-in reducer utilities as part of the core library.  Technically, `combineReducers` itself doesn't even _need_ to be built-in to Redux, but since "slice reducers" are the suggested standard approach, it made sense to include a utility that caters to that use case.

Functionality-wise, this looks almost equivalent to https://github.com/acdlite/reduce-reducers.  The main difference, I think, is that `composeReducers` takes an `initialState` arg, while `reduceReducers` just takes the functions.  

If nothing else, this is at least worth publishing as a utility library.  I have a list of many existing reducer utilities at https://github.com/markerikson/redux-ecosystem-links/blob/master/reducers.md (a list that _sadly_ needs some cleanup and organization), so there's well-established precedent for people publishing reducer utilities as libraries.

I'll leave this open for discussion for now.

gaearon:
Thanks for the PR but I'd rather not add this to the core. The difference between `compose`, `composeReducers` and `combineReducers` will be a major understanding hurdle for a lot of people. I don't think it is useful enough to offset this problem.

LucaColonnello:
I'll see if I can discuss with @acdlite to address this in his library, maybe make more sense. Thanks for the time spent to review this!

markerikson:
Sure.  Again, this is definitely a good PR overall, just not something we feel is needed in the core.  Thanks!

Issue #2334, opened 2017-04-03
Remove References to Sanity in Combine Reducers


seanmhanson:
The file `combineReducers.js` contains language like `assertReducerSanity` and `sanityError` that is generally viewed as ableist and potentially offensive to people living with mental illness. In addition, it doesn't accurately describe what we're trying to do in this check, and using an idom like this also makes it less approachable for developers whose first language is not English.

I'll throw a PR up in a second since this is a super tiny change, but I'm proposing to call it `assertReducerShape` and describe the error as a `shapeAssertionError` so that we're more precise and more inclusive.

timdorr:
Closing in favor of the PR you opened up. (Thanks for that!)

PR #2335, opened 2017-04-03
Replace `sanity` language in combineReducers

seanmhanson:
This addresses issue https://github.com/reactjs/redux/issues/2334

This replaces `sanityError` and `assertReducerSanity` with more descriptive, less ableist language that describes what the check is actually doing (checking and asserting the shape). I'm proposing to call it `assertReducerShape` and describe the error as a `shapeAssertionError` so that we're more precise and more inclusive at the same time!

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -54,7 +54,7 @@ function getUnexpectedStateShapeWarningMessage(inputState, reducers, action, une
   }
 }
 
-function assertReducerSanity(reducers) {
+function assertReducerShape(reducers) {
   Object.keys(reducers).forEach(key => {
     const reducer = reducers[key]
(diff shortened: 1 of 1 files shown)


timdorr:
Yep, that's a more accurate description of the check being done and the intended scope.

However, I very much disagree with this being ableist language. "Sanity" in this case is the literal definition of the word: "reasonable and rational behavior." That is rather ambiguous in the case of a code assertion (hence, it should be changed), but has nothing to do with inclusion issues or stigmatizing mental disease.

gaearon:
Thanks for the change! I was using the commonly known definition of [sanity check](https://en.wikipedia.org/wiki/Sanity_check) and didn't really think about other meanings. I agree the change is for the better. 👍

PR #2336, opened 2017-04-03
added an import for TodoList to first VisibleTodoList example

perryao:
When the connect function usage is shown with VisibleTodoList, the initial example doesn't import the TodoList component. This is properly shown further down in the containers/VisibleTodoList section, but it should be shown in both places so it's known where the reference to TodoList comes from.


timdorr:
It's defined higher up in the file. These snippets are only showing newly-added code for the examples. Adding this is redundant and makes the code just slightly more complex to read.

PR #2337, opened 2017-04-04
return a curried bindActionCreators function if only actionCreators are present

nadimtuhin:
I have been using `bindActionCreators` for some time and I found that it can be curried. Once curried `bindActionCreators` becomes easier to use moreover it does not break previous codes.

```js
// previous
export default connect(null,
    (dispatch, props) => bindActionCreators({
        addTodo,
        deleteTodo
    }, dispatch)
)(AppComponent);

// after curried
export default connect(null,
    bindActionCreators({
        addTodo,
        deleteTodo
    })
)(AppComponent);
```

I have also added a spec to test this new functionality.

### DIFF
--- a/src/bindActionCreators.js
+++ b/src/bindActionCreators.js
@@ -26,6 +26,11 @@ function bindActionCreator(actionCreator, dispatch) {
  * function.
  */
 export default function bindActionCreators(actionCreators, dispatch) {
+  // return a curried function if only actionCreators are present
+  if(arguments.length === 1) {
+    return (dispatch) => bindActionCreators(actionCreators, dispatch);
+  }
+
   if (typeof actionCreators === 'function') {
     return bindActionCreator(actionCreators, dispatch)
   }
--- a/src/bindActionCreators.js
+++ b/src/bindActionCreators.js
@@ -27,8 +27,8 @@ function bindActionCreator(actionCreator, dispatch) {
  */
 export default function bindActionCreators(actionCreators, dispatch) {
   // return a curried function if only actionCreators are present
-  if(arguments.length === 1) {
-    return (dispatch) => bindActionCreators(actionCreators, dispatch);
+  if (arguments.length === 1) {
+    return dispatch => bindActionCreators(actionCreators, dispatch)
   }
 
   if (typeof actionCreators === 'function') {
--- a/test/bindActionCreators.spec.js
+++ b/test/bindActionCreators.spec.js
@@ -118,6 +118,6 @@ describe('bindActionCreators', () => {
     ])
     expect(console.error).toHaveBeenCalled()
     global.console = _console
-  });
+  })
 
 })
--- a/test/bindActionCreators.spec.js
+++ b/test/bindActionCreators.spec.js
@@ -95,4 +95,29 @@ describe('bindActionCreators', () => {
       'Did you write "import ActionCreators from" instead of "import * as ActionCreators from"?'
     )
   })
+
+  it('returns a curried bindActionCreators function', () => {
+    const _console = console
+    global.console = { error: jest.fn() }
+
+    const curriedBindActionCreators = bindActionCreators(actionCreators)
+    const boundActionCreators = curriedBindActionCreators(store.dispatch)
+
+    expect(
+      Object.keys(boundActionCreators)
+    ).toEqual(
+      Object.keys(actionCreatorFunctions)
+    )
+
+    const action = boundActionCreators.addTodo('Hello')
+    expect(action).toEqual(
+      actionCreators.addTodo('Hello')
+    )
+    expect(store.getState()).toEqual([
+      { id: 1, text: 'Hello' }
+    ])
+    expect(console.error).toHaveBeenCalled()
+    global.console = _console
+  });
+
 })


src/bindActionCreators.js
```diff
@@ -26,6 +26,11 @@ function bindActionCreator(actionCreator, dispatch) {
  * function.
  */
 export default function bindActionCreators(actionCreators, dispatch) {
+  // return a curried function if only actionCreators are present
+  if(arguments.length === 1) {
+    return (dispatch) => bindActionCreators(actionCreators, dispatch);
```
timdorr: Nit: With a single argument arrow function, the parens are optional. I normally leave them off. Also, we don't use semi colors, so you can drop that.

src/bindActionCreators.js
```diff
@@ -26,6 +26,11 @@ function bindActionCreator(actionCreator, dispatch) {
  * function.
  */
 export default function bindActionCreators(actionCreators, dispatch) {
+  // return a curried function if only actionCreators are present
+  if(arguments.length === 1) {
```
timdorr: We use spaces after `if`.

test/bindActionCreators.spec.js
```diff
@@ -95,4 +95,29 @@ describe('bindActionCreators', () => {
...
+    const action = boundActionCreators.addTodo('Hello')
+    expect(action).toEqual(
+      actionCreators.addTodo('Hello')
+    )
+    expect(store.getState()).toEqual([
+      { id: 1, text: 'Hello' }
+    ])
+    expect(console.error).toHaveBeenCalled()
+    global.console = _console
+  });
```
timdorr: Another semi here.

nadimtuhin:
Complied with the project code style :)

timdorr:
Thanks!

Any thoughts from others on this? It could easily be released as another minor (new API, not a change to existing behavior). I have yet to release the Rollup-ed build, so I need to do another release soon anyways.

markerikson:
What's the main use case for this?  I don't see much difference between passing a curried `bindActionCreators` result in that example, vs just passing the object full of action creators explicitly.

I suppose there may be some value in having the currying aspect outside of the React-Redux use case.

nadimtuhin:
@markerikson I did not quite understand your comment here. In any case I am providing a full example why currying can be helpful.

> What's the main use case for this? I don't see much difference between passing a curried bindActionCreators result in that example, vs just passing the object full of action creators explicitly.

```js
const TodoComponent = ({ todos, addTodo, deleteTodo }) => (
  <div>
    <input onEnter={e => addTodo(e.target.value )} />
    <ul>
      { todos.map(todo => 
        <li key={todo.id}>
          {todo} 
          <button onClick={e => deleteTodo(todo.id)}>x</button>
        </li>
      ) }
    </ul>
  </div>
)
```

### 1: naive approach to disaptch actions
```js
export default connect(null,
  dispatch => {
    return {
      addTodo: (todo) => dispatch(addTodo(todo)),
      deleteTodo: (todo) => dispatch(deleteTodo(todo))
    }
   }
)(TodoComponent)
```

### 2: a better approach to dispatch
```js
export default connect(null,
  dispatch => bindActionCreators({
    addTodo,
    deleteTodo
  }, dispatch)
)(TodoComponent)
```

### 3: there is a better way a bit hacky but gets rid of all those dispatch
```js
export default connect(null,
  bindActionCreators.bind(null, {
    addTodo,
    deleteTodo
  })
)(TodoComponent)
```

### 4: this pr will make this code more elegant
```
export default connect(null,
  bindActionCreators({
    addTodo,
    deleteTodo
  })
)(TodoComponent)
```

Currying is a very important concept of functional programming. Now that react, redux ecosystem has been preaching functional paradigm for a while I think people will expect this kind of behavior as I did :)

markerikson:
The `connect` function currently supports passing an object full of action creators as the second argument, instead of an actual `mapDispatch` function:

```js
const actionCreators = {
    addTodo,
    deleteTodo
};

export default connect(null, actionCreators)(TodoComponent);
```

Under the hood, that just automatically runs the object through `bindActionCreators`, per your "Example 2: a better approach".  So, from a visual perspective, that's actually simpler than using a curried `bindActionCreators`, because the application code never even has to use `bindActionCreators`.

Because of that, I don't see any actual benefit for this in relation to React-Redux, unless there's something I'm missing.  Because the change is in Redux itself, it's possible it could have benefits in relation to other Redux/UI bindings.

Basically, I'm not specifically against this, but I am saying that the original stated use case is kind of pointless given `connect`'s support of the object shorthand argument.

nadimtuhin:
@markerikson sorry I did not see this API :D now I too think this pr is not needed anymore

nadimtuhin:
I would rather write some docs so others can use this feature :)

markerikson:
The React-Redux docs do describe the behavior, per https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options : 

> [mapDispatchToProps(dispatch, [ownProps]): dispatchProps] (Object or Function): If an object is passed, each function inside it is assumed to be a Redux action creator. An object with the same function names, but with every action creator wrapped into a dispatch call so they may be invoked directly, will be merged into the component’s props.

I talked about the use of the object shorthand syntax in my post at http://blog.isquaredsoftware.com/2016/10/idiomatic-redux-why-use-action-creators/ , and there's a number of other mentions of it online.  It's just not as widely known overall.

timdorr:
OK, could be a cool use case, but you can curry externally if you *really* need this. Thanks for the work and responses, @nadimtuhin!

PR #2339, opened 2017-04-04
change 'terse' to 'concise'

danidewitt:


timdorr:
Sure, why not. The definitions are interchangeable here, but this fits a tiny bit better.

PR #2340, opened 2017-04-05
docs(observable): es-observable GitHub repo moved to tc39

Haroenv:
The repository still redirects, but it makes sense to use the name that's now in use


timdorr:
Thanks!

PR #2341, opened 2017-04-05
Updating docs for default values w/ preloadedState

### LINKED ISSUE #2338: Reducers need to provide default value, even when using preloadedState
Bug/unexpected behavior

When the store is created, and an initial `preloadedState` is provided, reducers are called with the `@@redux/INIT` action and an `undefined` state. This causes startup to fail, unless the reducers provide default values for state. [Here's a codepen](https://codepen.io/anon/pen/yMrLzg?editors=0010) which replicates the issue.

When providing a `preloadedState`, I would expect to not need to provide a default value in my reducer, since I've already provided one in the form of `preloadedState`. I understand that this is to ensure that reducers are properly-written, in that they don't return `undefined` even when passed `undefined`. However, if a field in the store is populated using `preloadedState` and never changed, it seems silly to have to do either:

```javascript
const propReducer = (state = null, action) => state
```
or

```javascript
const preloadedState = window['__PRELOADED_STATE__']
const propReducer = (state = preloadedState.prop, action) => state
```

At the very least, I think the documentation for using preloaded state should be updated to include this, because it wasn't entirely obvious (without reading the redux source code) why I needed to provide this default when, in my mind, I already had.

The browser/OS/version aren't really relevant, since the unexpected behavior will occur in any scenario.

kengorab:
Fixes #2338, including a note about default values in reducers, even when using preloadedState

### DIFF
--- a/docs/recipes/reducers/InitializingState.md
+++ b/docs/recipes/reducers/InitializingState.md
@@ -1,6 +1,6 @@
 # Initializing State
 
-There are two main ways to initialize state for your application.  The `createStore` method can accept an optional `preloadedState` value as its second argument.  Reducers can also specify an initial value by looking for an incoming state argument that is `undefined`, and returning the value they'd like to use as a default.  This can either be done with an explicit check inside the reducer, or by using the ES6 default argument value syntax: `function myReducer(state = someDefaultValue, action)`.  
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks for writing this up!

PR #2350, opened 2017-04-10
Remove Markdown syntax from code block

justinbach:
I think the `**` prior to `VALUE` in the Immutable sample HOC is a typo.


timdorr:
Thanks!

PR #2357, opened 2017-04-13
Rollup bundles

TrySound:
I'd like to propose going further with flat bundles and prepare redux even for bundlers.
This PR contains merge with master.
/cc @timdorr


timdorr:
The next branch has breaking changes (for 4.0 eventually). It shouldn't be a part of this PR. Isolate just the changes to the rollup config and resubmit.

PR #2363, opened 2017-04-20
Update docs with React v15.5.0

kwnccc:
Changing the import of `PropTypes`, from `react` to `prop-types` package, as of React v15.5.0 update.


gaearon:
Thanks!

PR #2367, opened 2017-04-21
Use the state accessors in the tests, instead of testing the store st…

kevinrutherford:
…ate directly

### DIFF
--- a/examples/shopping-cart/src/reducers/products.spec.js
+++ b/examples/shopping-cart/src/reducers/products.spec.js
@@ -2,7 +2,7 @@ import reducer, * as products from './products'
 
 describe('reducers', () => {
   describe('products', () => {
-    var state
+    let state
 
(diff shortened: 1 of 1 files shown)


examples/shopping-cart/src/reducers/products.spec.js
```diff
@@ -1,58 +1,82 @@
...
-            id: 1,
-            title: 'Product 1'
-          },
-          {
-            id: 2,
-            title: 'Product 2'
-          }
-        ]
-      }
+    var state
```
urfolomeus: Should this perhaps be a `let` in keeping with the rest of the file?

kevinrutherford: +1

timdorr:
Yeah, I think this helps quite a bit. The test descriptions are a good improvement too. Thanks!

PR #2369, opened 2017-04-21
Update examples with React v15.5.0

kwnccc:
I've updated all the redux examples with react dependencies to v15.5.0. 😀


gaearon:
Looks great, thanks!

PR #2371, opened 2017-04-24
Update WritingTests.md

mccambridge:
I wasn't able to query the DOM with `.find()` until I called Enzyme with `mount()` or `render()` instead of `shallow()`


timdorr:
Ah, good call. Thanks!

PR #2372, opened 2017-04-25
Update todomvc specs with React v15.5.0 adding react-test-renderer

kwnccc:
Just a follow up update for some tests on todomvc example


timdorr:
Thank you! Any others that need updating?

kwnccc:
@timdorr for examples I can find no other reference to either `TestUtils` or `createRenderer`, so there must not be any other test needs updating.

PR #2374, opened 2017-04-26
Remove redundant import

skube:
Since you talk about `browserHistory` a little later down anyway, it may confuse readers why it's included early on.


timdorr:
Thanks!

PR #2375, opened 2017-04-27
add a hook to redux-devtools-extension

gfx:
I think `examples/todos` is the most important example for React Redux, so it is worth considering to add the [redux-devtools-extension](https://github.com/zalmoxisus/redux-devtools-extension) middleware if users have the extension in Chrome.

It looks like:

screen shot 2017-04-27 at 13 45 38

### DIFF
--- a/examples/todos/src/index.js
+++ b/examples/todos/src/index.js
@@ -5,7 +5,10 @@ import { Provider } from 'react-redux'
 import App from './components/App'
 import reducer from './reducers'
 
-const store = createStore(reducer)
+const store = createStore(
+  reducer,
+   window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()
+)
 
 render(
   <Provider store={store}>


examples/todos/src/index.js
```diff
@@ -5,7 +5,10 @@ import { Provider } from 'react-redux'
 import App from './components/App'
 import reducer from './reducers'
 
-const store = createStore(reducer)
+const store = createStore(
+  reducer,
+   window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()
+)
```
Florian-R: Maybe a comment stating it's not mandatory? I'm afraid it will confuse newcomers in the current state.

timdorr:
I appreciate the PR, but the todos example is intended to be bare bones and not get into more intermediate stuff, such as Dev Tools.

examples/todos/src/index.js
gfx: commented! fdaae83

PR #2376, opened 2017-04-28
Simplify createStore(), remove nextListeners & ensureCanMutateNextListeners()

robertpenner:
- If we treat `currentListeners` as a read-only array, `createStore()`'s implementation can be simplified. 
- `nextListeners` and `ensureCanMutateNextListeners()` are unnecessary when we assign new arrays to `currentListeners` but never mutate it.
- The code is easier to follow when we don't have to keep track of current vs. next listeners.
- `dispatch()` remains free of the cost of creating new arrays.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -57,14 +57,7 @@ export default function createStore(reducer, preloadedState, enhancer) {
   let currentReducer = reducer
   let currentState = preloadedState
   let currentListeners = []
-  let nextListeners = currentListeners
   let isDispatching = false
-
-  function ensureCanMutateNextListeners() {
-    if (nextListeners === currentListeners) {
-      nextListeners = currentListeners.slice()
-    }
-  }
 
   /**
    * Reads the state tree managed by the store.
@@ -105,8 +98,7 @@ export default function createStore(reducer, preloadedState, enhancer) {
 
     let isSubscribed = true
 
-    ensureCanMutateNextListeners()
-    nextListeners.push(listener)
+    currentListeners = currentListeners.concat(listener)
 
     return function unsubscribe() {
       if (!isSubscribed) {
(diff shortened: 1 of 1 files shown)


markerikson:
I'm not clear on the exact history of why the current listeners logic is the way it is, but I'm pretty sure it has to do with the subtleties regarding more dispatches happening in subscription callbacks, and things like that.  

My default response to  PRs like this that mostly make changes for the sake of style is "No, unless it really is fixing an actual bug".  I wouldn't want to consider merging something like this until I fully understand exactly why the current code is written that way.

I would request that you look back through the history of `createStore` and review why the current code structure exists, then recap the reasons in this thread.  If this change violates any of the existing assumptions about behavior and timing, then it's not an option.

timdorr:
This would be a breaking change. We have very specific behaviors for how subscriptions are handled, particularly if you dispatch or subscribe/unsubscribe from within a listener. 

I'd rather do this via an external library, so we can avoid dealing with the subtleties ourselves. Check out #1729 for an example.

robertpenner:
@timdorr How do you know this is a breaking change when it passes all the unit tests?

robertpenner:
@markerikson Thanks for the feedback. I read the history of `createStore()` carefully before making this change, and I confirmed that this new code satisfies the following required behaviors of Redux:

- [x] supports multiple subscriptions
- [x] only removes listener once when unsubscribe is called
- [x] only removes relevant listener when unsubscribe is called
- [x] supports removing a subscription within a subscription
- [x] delays unsubscribe until the end of current dispatch
- [x] delays subscribe until the end of current dispatch
- [x] uses the last snapshot of subscribers during nested dispatch
- [x] provides an up-to-date state when a subscriber is notified
- [x] does not leak private listeners array
- [x] only accepts plain object actions
- [x] handles nested dispatches gracefully
- [x] does not allow dispatch() from within a reducer
- [x] recovers from an error within a reducer
- [x] throws if listener is not a function
- [x] should be subscribable
- [x] should throw a TypeError if an observer object is not supplied to subscribe
- [x] should return a subscription object when subscribed
- [x] should pass an integration test with no unsubscribe
- [x] should pass an integration test with an unsubscribe
- [x] should pass an integration test with a common library (RxJS)

PR #2382, opened 2017-05-01

nitin2195:


timdorr:
Thanks for this, but we don't need to add more complexity to this example. More files == more someone has to think about when reading through the code.

PR #2385, opened 2017-05-02
Apply prettier to docs

CodinCat:
(This PR is related to #2381)

Hi, I applied [Prettier](https://github.com/prettier/prettier) to the code blocks of documentation. Besides, it caught many syntax errors (missing commas, parentheses, fullwidth characters..)  and I have fixed them manually.

Also the [Store type declaration](https://github.com/reactjs/redux/blob/master/docs/Glossary.md#store) in Glossary.md cannot be parsed by prettier. Sorry I'm not familiar with flow, should I add commas to it? Like below:

```js
type Store = {
  dispatch: Dispatch,
  getState: () => State,
  subscribe: (listener: () => void) => () => void,
  replaceReducer: (reducer: Reducer) => void
}
```

PR #2387, opened 2017-05-03
Prettier to introduction folder

CodinCat:
#2385

### DIFF
--- a/docs/introduction/CoreConcepts.md
+++ b/docs/introduction/CoreConcepts.md
@@ -34,24 +34,25 @@ It would be hard to write such a function for a big app, so we write smaller fun
 ```js
 function visibilityFilter(state = 'SHOW_ALL', action) {
   if (action.type === 'SET_VISIBILITY_FILTER') {
-    return action.filter;
+    return action.filter
   } else {
-    return state;
+    return state
   }
 }
 
 function todos(state = [], action) {
   switch (action.type) {
-  case 'ADD_TODO':
-    return state.concat([{ text: action.text, completed: false }]);
-  case 'TOGGLE_TODO':
-    return state.map((todo, index) =>
-      action.index === index ?
-        { text: todo.text, completed: !todo.completed } :
-        todo
-   )
-  default:
-    return state;
+    case 'ADD_TODO':
+      return state.concat([{ text: action.text, completed: false }])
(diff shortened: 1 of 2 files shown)


timdorr:
This one is good as-is. Thanks!

gaearon:
Hmm. If we adopt Prettier maybe we can also bring semicolons back 😈 ? I only adopted no-semi style because that's what rackt styleguide said so two years ago.

gaearon:
In fact it's been a little annoying to defend the use of no-semi in Redux considered I never liked no-semi in the first place.

timdorr:
I'd just say it is what it is at this point. The pain in converting to no-semi was bad enough that I don't really want to live through that again. It basically breaks every open PR. And other than preference, it doesn't really net us any tangible benefits (at least not enough to warrant the effort required).

markerikson:
I'm a four-space/semicolons person myself, but I also hate formatting PRs clogging up file history.  So, yes, I'd rather not make changes just for the sake of changes (and once again, considering the relative stability of the lib, don't think it's worth the hassle at this point).

CodinCat:
So.. do we have any conclusion on this? It won't be too difficult to bring semis back at this point since I haven't made any changes to the reviews of PRs, just need to re-run Prettier. But after that it may take more works to add them back.

gaearon:
I don't care—let's leave it as is then.

>It basically breaks every open PR.

Easy to fix by running `prettier` in those PRs but I see your point.
(FWIW we did this with React and it was fine for us.)

gaearon:
Also, since we're running Prettier anyway, we're already breaking every PR, are we not?

timdorr:
This is only on the code blocks for the docs.

CodinCat:
Personally I prefer no-semi. *But* this is documentation, maybe with semicolons is less confusing for beginners?

What we need here is just consistency. Some chapters are already in semi style (mostly the *StructuringReducers* section) so whatever with or without semis, we need to make a lot of changes anyway. But I truly believe it’s worth it since it improves the quality of docs (and fixes some errors).

Hope I can have a confirmation from you guys. 😸

CodinCat:
I'm continuing the work. If we really need to bring semis back I can open another PR for that. For now let's just cleanup the docs.

PR #2390, opened 2017-05-03
Prettier to recipes folder

CodinCat:
#2385

### DIFF
--- a/docs/recipes/ComputingDerivedData.md
+++ b/docs/recipes/ComputingDerivedData.md
@@ -24,24 +24,21 @@ const getVisibleTodos = (todos, filter) => {
   }
 }
 
-const mapStateToProps = (state) => {
+const mapStateToProps = state => {
   return {
     todos: getVisibleTodos(state.todos, state.visibilityFilter)
   }
 }
 
-const mapDispatchToProps = (dispatch) => {
+const mapDispatchToProps = dispatch => {
   return {
-    onTodoClick: (id) => {
+    onTodoClick: id => {
       dispatch(toggleTodo(id))
     }
   }
 }
 
-const VisibleTodoList = connect(
-  mapStateToProps,
-  mapDispatchToProps
-)(TodoList)
+const VisibleTodoList = connect(mapStateToProps, mapDispatchToProps)(TodoList)
 
 export default VisibleTodoList
 ```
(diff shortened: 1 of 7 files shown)


docs/recipes/UsingImmutableJS.md
```diff
@@ -168,26 +168,26 @@ Never let a plain JavaScript object contain Immutable.JS properties. Equally, ne
...
 
-```
+```js
 // avoid
-const newObj = { key: value};
-const newState = state.setIn(['prop1’], newObj); // <-- newObj has been added as a plain
-    // JavaScript object - NOT as an Immutable.JS Map
+const newObj = { key: value }
+const newState = state.setIn(['prop1'], newObj) // <-- newObj has been added as a plain
+// JavaScript object - NOT as an Immutable.JS Map
```
CodinCat: This comment may need some fixing. Also at line 190

docs/recipes/ReducingBoilerplate.md
```diff
@@ -347,12 +348,7 @@ The middleware that interprets such actions could look like this:
 ```js
 function callAPIMiddleware({ dispatch, getState }) {
   return next => action => {
-    const {
-      types,
-      callAPI,
-      shouldCallAPI = () => true,
-      payload = {}
-    } = action
+    const { types, callAPI, shouldCallAPI = () => true, payload = {} } = action
```
timdorr: Another one where we should leave it alone.

docs/recipes/ComputingDerivedData.md
```diff
@@ -108,24 +104,21 @@ import { toggleTodo } from '../actions'
...
       dispatch(toggleTodo(id))
     }
   }
 }
 
-const VisibleTodoList = connect(
-  mapStateToProps,
-  mapDispatchToProps
-)(TodoList)
+const VisibleTodoList = connect(mapStateToProps, mapDispatchToProps)(TodoList)
```
timdorr: More connect calls that should be left alone.

docs/recipes/UsingImmutableJS.md
timdorr: Yeah, that's weird formatting to begin with. I'd do this:
```js
// avoid
const newObj = { key: value }
const newState = state.setIn(['prop1'], newObj) 
// newObj has been added as a plain JavaScript object, NOT as an Immutable.JS Map

// recommended
const newObj = { key: value }
const newState = state.setIn(['prop1'], fromJS(newObj)) 
// newObj is now an Immutable.JS Map
```

docs/recipes/WritingTests.md
```diff
@@ -371,7 +368,9 @@ In order to be able to test the App component itself without having to deal with
 import { connect } from 'react-redux'
 
 // Use named export for unconnected component (for tests)
-export class App extends Component { /* ... */ }
+export class App extends Component {
+  /* ... */
+}
```
timdorr: I think this one you can revert too.

timdorr: And the one above.

CodinCat:
Some weird GitHub bug always shows the comment on ComputingDerivedData.md Line 121 which I've fixed..

PR #2393, opened 2017-05-03
Give a clearer warning that react-router guide is outdated

donpinkus:
If you are skimming this guide, or already have react-router installed you will not read the section on "installing react router" - which will result in some very confused users.

Moving the warning to the top of the doc until I (or someone else) has a chance to update it to react-router 4.0


docs/advanced/UsageWithReactRouter.md
```diff
@@ -1,5 +1,7 @@
 # Usage with React Router
 
+Warning: This guide assumes you are using `react-router@^2.7.0`. As of May 2017 the current version of React Router is 4+. If you have just run `npm install --save react-router`, you will be using a version that does not match this guide.
```
timdorr: This applies to both 2.x and 3.x. They are the same, just with different deprecations.

PR #2398, opened 2017-05-06
Changed order of function call in code

furkanpoyraz:
The function needs to be declared before the actual call.


markerikson:
Since this is Javascript, the function declaration should actually get hoisted to the top of the scope, so the existing code should work as-is.

That said, I guess it doesn't hurt to swap them.

Issue #186, opened 2015-06-28
Anti-pattern: reacting to @@INIT action


gaearon:
In the new docs (#140) we should clarify that any actions prefixed with `@@` are *not* meant to be handled. For example, you should never try to handle `@@INIT`. We might enforce that by slightly randomizing names (e.g. `@@INIT_2hj3jh34`).

Handling `@@INIT` manually will break hot reloading. It is invoked at every hot reload, so if you do your initial data transformation there, it won't work the second time.

I've seen people using `@@INIT` to hydrate the plain objects received from the server into the Immutable state. Instead of handling `@@INIT`, you need to consider one of two options:

* convert your whole plain object state to Immutable right away before passing it to Redux;
* check if the `state` is a plain object and use `fromJS` before handling the action.

```js
function myReducer(state = {}, action) {
  if (!isImmutable(state)) { // some kind of typeof check here?
    state = Immutable.fromJS(state);
  }

  ...
}
```

It is easy to abstract this away with a `createImmutableReducer` function like this:

```js
export default function createImmutableReducer(initialState, handlers) {
  return (state = initialState, action) => {
    if (!isImmutable(state)) {
      // Hydrate server plain object state
      state = Immutable.fromJS(state);
    }

    const handler = handlers[action.type];
    if (handler) {
      state = handler(state, action);
    }

    if (!isImmutable(state)) {
      throw new Error('Reducers must return Immutable objects.');
    }

    return state;
  };
}
```

It can then be used for creating reducers from `action constant -> handler` map.

hummlas:
@gaearon What is  "@@"? I'm assuming it's ES6, perhaps related to symbols, but google doesn't like special characters so it ungoogleable :(

dariocravero:
@hummlas it's random :)

hummlas:
Ah, I see, like old-school underscore prefixes, e.g. `var _privateValue = "Private";`? :)

It seem to be used in official spec-stuff as well, like iterables
>"In order to be iterable, an object must implement the @@iterator method, meaning that the object (or one of the objects up its prototype chain) must have a property with a Symbol.iterator key"

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Iteration_protocols

This confuses me a bit as you're supposed to implement an `@@iterator` method, which, to me, seems to be at odds with @@ meaning "internal"?

gaearon:
[...]

PR #2408, opened 2017-05-16
fix assertReducerShape

jeno5980515:
Hi, I remove the part that you check the private namespace is used or not.

I think the code is no use for the condition.
If the namespace is used, the reducer should return a initial state instead of undefined. 
Otherwise if it really return undefined it should block at the previous error.
Only check is undefined or not is not working for check the private namespace is used or not.

I think it should check the initial state and the state return from the reducer which give private namespace action are deep different or not.
However it will reduce lot's performance I think.

Thanks!

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -69,17 +69,6 @@ function assertReducerShape(reducers) {
       )
     }
 
-    const type = '@@redux/PROBE_UNKNOWN_ACTION_' + Math.random().toString(36).substring(7).split('').join('.')
-    if (typeof reducer(undefined, { type }) === 'undefined') {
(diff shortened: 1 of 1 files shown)


timdorr:
This was added in #259 to fix #186. This is correct as-is and should not be removed.

jeno5980515:
I had read the reference you give.
However, I think the check code still not throw any error if I give a reducer with handle the action with '@@redux/INIT' or '@@INIT'.
By the way if the reducer return undefined, it should throw error at the
 initialState === 'undefined'
not at 
typeof reducer(undefined, { type }) === 'undefined'

Thanks!

PR #2411, opened 2017-05-16
Replaced the for loops with forEach

muralikrishna8:
In all the `for` loops used we need to define the `const key = key[i]`, which is not necessary. We can replace that with the `forEach`

### DIFF
--- a/src/bindActionCreators.js
+++ b/src/bindActionCreators.js
@@ -39,14 +39,13 @@ export default function bindActionCreators(actionCreators, dispatch) {
 
   const keys = Object.keys(actionCreators)
   const boundActionCreators = {}
-  for (let i = 0; i < keys.length; i++) {
-    const key = keys[i]
+  keys.forEach(key => {
     const actionCreator = actionCreators[key]
     if (typeof actionCreator === 'function') {
       boundActionCreators[key] = bindActionCreator(actionCreator, dispatch)
     } else {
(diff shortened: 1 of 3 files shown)


markerikson:
Hi.  Appreciate the PR, but I believe this was implemented as a `for` loop as a bit of a performance optimization.  The current code works as-is, and there's not sufficient reason to make changes to it.  Thanks anyway!

muralikrishna8:
But I've observed that the `forEach` has used in a couple of places in combineReducers [line 58](https://github.com/reactjs/redux/blob/master/src/combineReducers.js#L58), [line 43](https://github.com/reactjs/redux/blob/master/src/combineReducers.js#L43), if you believe we are using `for` loop as a bit of performance, even these two should be `for` loop instead of `forEach`.

And when we talk about the **pure functions** and functional way of doing stuff, isn't the `for` loop is mutating the `i` value there??

markerikson:
In those particular cases, the use of `forEach` is not part of the main logic for `combineReducers`, and will only be executed in development mode.

Also, you may have a bit of a misunderstanding for what a "pure function" is.  A "pure function" is simply any function that only bases its output on its inputs, and doesn't have any "side effects" that affect anything else (like mutating its inputs, `console.log()` statements, AJAX calls, etc).  It's entirely fine for a pure function to use mutation in the process of doing its work.  For example, you can immutably update an object by making a copy and directly mutating the copy.  So, there's nothing wrong with declaring a loop counter and updating it.

PR #2413, opened 2017-05-17
Add 100% test coverage for todos-flow example

lifeiscontent:
I wanted to learn more about Jest testing and flow, so I took it upon myself to extend this example and build it so it had 100% test coverage.

please let me know if I should update anything and I'll do as accordingly.

I struggled to find a solid example project on how to test all the things within react/redux so I think this might be a great example for others to learn from on how to test with jest + learn flow

sidenote: I'm also okay with refactoring to not use enzyme, but it would be appreciated if someone could share some guidance on how to test without it as I'm currently unsure of how to do that.


timdorr:
Woah, lots of changes here. 

I think to clear up a bunch of it, we should remove the flow typings files. They should be automatically downloaded via the npm scripts, but should be .gitignore'd from the repo. If they change in any way, that's going to create some useless fluff commits. 

Let's start with that.

timdorr:
Also, we don't currently use semicolons, so you can also remove those to remove a good chunk of changes.

timdorr:
Lastly, the breaking up of the visibility filters stuff into separate files and addition of of reselect may be overkill. I'd try to keep this to the same structure as the plain todos example. Basically, this should be the same thing with Flow typings and tests added.

brigand:
@timdorr fwiw [Why do I need to commit the libdefs that `flow-typed` installs for my project?](https://github.com/flowtype/flow-typed#why-do-i-need-to-commit-the-libdefs-that-flow-typed-installs-for-my-project)

lifeiscontent:
@timdorr the reason I commited the libdefs is exactly the reason @brigand commented about. But yes, I can remove it my main concern is just to give people a better intro into flow/testing

lifeiscontent:
@timdorr as for reselect, I wanted to take an approach that leaned more on the side of real-world than simplistic. But if it's an issue, I guess I can change it back.

lifeiscontent:
anyone know why tests are failing? they're all passing locally.

timdorr:
Your imports are not using the same capitalization as your files. You probably have a Mac, which has a case-insensitive filesystem. Travis runs on Linux, which is case-sensitive. Simple as that.

examples/todos-flow/src/components/todo-list.js
```diff
@@ -0,0 +1,22 @@
...
+
+const TodoList = ({ todos, onTodoClick }: Props) => (
+  <ul>
+    {todos.map(todo => (
+      <Todo key={todo.id} {...todo} onClick={() => onTodoClick(todo.id)} />
+    ))}
+  </ul>
+);
+
+export default TodoList;
```
timdorr: BTW, this file should be named `TodoList.js` to match the component name that it exports.

examples/todos-flow/src/containers/add-todo.js
```diff
@@ -0,0 +1,51 @@
...
+          </button>
+        </form>
+      </div>
+    );
+  }
+}
+
+const connector: Connector<{}, Props> = connect();
+
+export default connector(AddTodo);
```
timdorr: This file should be named `AddTodo.js`

examples/todos-flow/src/containers/filter-link.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+  };
+};
+
+const connector: Connector<OwnProps, Props> = connect(
+  mapStateToProps,
+  mapDispatchToProps
+);
+
+export default connector(Link);
```
timdorr: This should be named `FilterLink.js` and I would make sure the displayName on the component is set to `FilterLink` as well.

examples/todos-flow/src/containers/visible-todo-list.js
```diff
@@ -0,0 +1,32 @@
...
+    }
+  };
+};
+
+const connector: Connector<{}, Props> = connect(
+  mapStateToProps,
+  mapDispatchToProps
+);
+
+export default connector(TodoList);
```
timdorr: This should be named `VisibleTodoList.js` and the displayName on the component should be`VisibleTodoList` as well.

lifeiscontent:
@timdorr all fixed. 👍 What else should I clean up?

lifeiscontent:
@timdorr I just fixed conflicts with master, ready to merge again.

timdorr:
I think we're all good here. Thanks for the hard work!

PR #2414, opened 2017-05-17
change var to const

muralikrishna8:
Changing the `var` to `const` in the [examples/testAll.js](https://github.com/reactjs/redux/blob/master/examples/testAll.js) file to make everything in ES6


timdorr:
Thanks!

PR #2418, opened 2017-05-20
update react-scripts to 1.0.1

ohmyjersh:


gaearon:
Thanks for PR! We’ll probably go with #2419 instead since it’s more comprehensive and includes changes to lockfiles too.

PR #2425, opened 2017-05-27
Remove unnecessary render call

marcelodeveloper:
Hi guys, you don't need it. The subscribe method will handle it to you.


TrySound:
@marcelodeveloper It won't. `subscribe` only adds listeners.
https://github.com/reactjs/redux/blob/master/src/createStore.js#L101

markerikson:
Correct - the example needs to do an initial render, otherwise nothing will get displayed.

PR #2430, opened 2017-05-29
Fix redux-mock-store libdef for todos-flow

lifeiscontent:
Just updating a libdef to keep todos-flow at 100%


timdorr:
Hmm, this is going to get annoying. Are we sure we can't just dynamically pull the typedefs at build/test time?

lifeiscontent:
@timdorr should I just add a post install script for the package.json?

timdorr:
Yeah, that's probably going to be best. Manually maintaining these files is going to create a lot of noise in the commits over time.

lifeiscontent:
@timdorr all updated 👍

timdorr:
Awesome! Thanks!

PR #2433, opened 2017-05-31
give redux-subscriber some attention in StoreSetup.md

wbern:
It has both technical scalability and a good level of documentation. It took me a while to find it, so I hope this can change things for future readers.


timdorr:
Thanks!

markerikson:
For what it's worth, there's actually a wide variety of additional store subscription utility libs out there.  See https://github.com/markerikson/redux-ecosystem-links/blob/master/store.md#store-change-subscriptions for a list.

PR #2442, opened 2017-06-07
Fix: the initial state of todoApp

jiangpengnju:
When combine the splitted reducers, shouldn't the initial state of ```todoApp``` the composition of initial states of splitted reducers?
 
```js
function todos(state = [], action) {
  switch (action.type) {
    case ADD_TODO:
      return [
        ...state,
        {
          text: action.text,
          completed: false
        }
      ]
    case TOGGLE_TODO:
      return state.map((todo, index) => {
        if (index === action.index) {
          return Object.assign({}, todo, {
            completed: !todo.completed
          })
        }
        return todo
      })
    default:
      return state
  }
}

function visibilityFilter(state = SHOW_ALL, action) {
  switch (action.type) {
    case SET_VISIBILITY_FILTER:
      return action.filter
    default:
      return state
  }
}

function todoApp(state = initialState, action) {
  return {
    visibilityFilter: visibilityFilter(state.visibilityFilter, action),
    todos: todos(state.todos, action)
  }
}
```

### DIFF
--- a/docs/basics/Reducers.md
+++ b/docs/basics/Reducers.md
@@ -303,7 +303,7 @@ function visibilityFilter(state = SHOW_ALL, action) {
   }
 }
 
-function todoApp(state = {}, action) {
+function todoApp(state = initialState, action) {
   return {
     visibilityFilter: visibilityFilter(state.visibilityFilter, action),
     todos: todos(state.todos, action)


timdorr:
Actually, in this case, no. The default value for the reducer argument only applies if the input value is undefined. If you pass in that initialState value, the "sub-reducers" will not get an undefined state and will behalf differently if they were moved up to be the top level reducer. But if you pass an empty object, they will get undefined states and perform correctly. Less surprising behaviors that way (especially in tests).

jiangpengnju:
Thank you very much for replying. Got it after re-reading [ES6 default arguments syntax](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/default_parameters).

PR #2443, opened 2017-06-07
indexOf -> includes

denysdovhan:
`includes` is now available in most of the modern browsers. Let's use it instead of ugly `indexOf`


gaearon:
Thanks for PR but it is an explicit goal to support older browsers. We generally don't take changes that make only *us* feel better :-). We could take this with a fallback if there was a real benefit to using the new API, but since they're equivalent, we'll keep using indexOf. Cheers!

gaearon:
(If we changed that then the next person would send a PR to change it back in a few months. I'm open to changing examples to use ES2015 runtime APIs in a couple of years.)

Issue #2445, opened 2017-06-09
Feature request for a notify middleware: Prevent notifying all listeners in certain conditions


zeachco:
Hi, this is a feature request

**What is the current behavior?**
Every dispatch cause all listeners to be called without the possibility to prevent it.

The "hackish" to do it would be with the `enhancer` function from `createStore`. I don't find this clean for the following reason:
It require to rewrite `createStore` completely which make me wonder why do we require Redux in the first place if it's to pass a function that replace it completely.

**What is the expected behavior?**
I think allowing to pass a middleware would be helpfull for cases such as Redux used with Immutable.js. A middleware could verify that the state has muted and not call uselessly all the listeners if not. Other usecase could also apply such as "locking" dispatch in certain conditions for other performances reasons.

I think the goal of Redux is to notify the state changes, not to notify that a dispatch happen otherwise his role wouldn't be to hold a store a well but more a simple event bus system. Maybe I didn't get correctly the misson of Redux here so let me know if I'm in the wrong here.

---
The current Redux version is 3.6.0
this applies for all browsers / node environments and all OS

---
I currently have a working version of this which need polishing so I would provide the PR if this feature is accepted.

Thank you

timdorr:
This isn't really the job of redux, as it's supposed to be small, simple, and somewhat dumb in it's implementation. It shouldn't distinguish who to notify because those listeners can use the state in a variety of ways. Basically, it can't and shouldn't know how state is used. 

It sounds like what you want is reselect: https://github.com/reactjs/reselect

zeachco:
@timdorr thanks for the fast answer, I think I didn't correctly described the issue here.

I do not implies that Redux should handle when or when not to notify listeners, but being able to receive a function that does that check for him which knows about the data structure etc.

The point here is that in a complex application with multiple subscribes everywhere we have no ways to prevent a loop of dispatcher to being called and this sometimes kills the application. Additionally, some nice tools such as [reduxDevTool](https://github.com/gaearon/redux-devtools) can spam a lots of dispatch information which just crash the browser or generate so much noise that it becomes useless.

I will give you an example of a usage and let me know if that makes sense for you
```javascript
// User can specify when he wants the notify to happens
// In this example (so it make sense) I use Immutable.js
// so I can test if there is a change in any node of the store with one super fast check
const preventUselessNotifies = (oldState, newState) => !(oldState && oldState.equals && oldState.equals(newState));

store = createStore(reducers, initState);
[...]

PR #2446, opened 2017-06-09
Proposition (WIP) added notify middleware to createStore API + fixed tests

zeachco:
This is a WIP for https://github.com/reactjs/redux/issues/2445
- [ ] add tests for `store.setNotifyMiddleware`


markerikson:
Appreciate the suggestion, but this is not something we plan to include as it can be handled in "userland" as a store enhancer.

PR #2455, opened 2017-06-18
Remove `bindActionsCreators()` warning

WRONGWAY4YOU:
This PR makes so that there is no warning printed out when calling `bindActionCreators()` function with the first argument containing key "default" (which is not a function).

This is done in order to allow library users to store their actions in the module in which actions _also_ exported as a `default` object. So, commit just skips that object.

### DIFF
--- a/src/bindActionCreators.js
+++ b/src/bindActionCreators.js
@@ -45,6 +45,29 @@ export default function bindActionCreators(actionCreators, dispatch) {
     if (typeof actionCreator === 'function') {
       boundActionCreators[key] = bindActionCreator(actionCreator, dispatch)
     } else {
+      // If the key is `default`, and it is not a function <=>
+      // programmer wanted to do something like this:
+      // ```js
+      // import * as actions from 'actions'
+      //
+      // ...
+      //
+      // const actions = bindActionCreators(actions, dispatch)
+      // ```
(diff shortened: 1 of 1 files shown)


zardak:
As per https://github.com/erikras/ducks-modular-redux#usage
You can get every export while omitting the default one using `import * as actions from './module'`.

WRONGWAY4YOU:
Is it good that the user, when reading [documentation](http://redux.js.org/docs/api/bindActionCreators.html#somecomponentjs) (where is nothing said about the library that you proposing), should spend something like a 5-10 minutes trying to figure out where and why this warning came from?

It is very **usual** that the user wants to export the action creators _also_ as default, without using _any other_ library.

zardak:
My bad, that syntax actually does include default export. (Also, I linked to a writeup on one way of structuring action creators and reducers, it's not a library).

Seems totally sane to me now.

timdorr:
The `default` key is [a side effect of Babel transpilation](https://babeljs.io/docs/plugins/transform-es2015-modules-commonjs/). It is not part of the ES Modules spec. We shouldn't be excepting an implementation detail of the tooling, despite it's popularity. Once module are usable without Babel in the future or if you're using a module-aware bundler (webpack 2 or rollup), this will end up causing more problems that it solves.

PR #2464, opened 2017-06-21
Fixed 'is undefined' error and prop checking in Counter example

Krizzu:
Added method binding in constructor + prop checking.

### DIFF
--- a/examples/counter/src/components/Counter.js
+++ b/examples/counter/src/components/Counter.js
@@ -2,19 +2,19 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 
 class Counter extends Component {
-  static propTypes = {
-    value: PropTypes.number.isRequired,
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

heracek:
@Krizzu Just out of curiosity, why is binding in constructor better?

Krizzu:
It's not like it's better, but changed it because `eslint` was throwing 'undefined` errors.

timdorr:
This is actually caused by the upgrade to eslint 4.0 at the top level and CRA using the wrong version: https://github.com/facebookincubator/create-react-app/issues/2604#issuecomment-311168618. 

I'll likely revert this once https://github.com/facebookincubator/create-react-app/issues/2644 lands.

heracek:
Ok, thanks!

PR #2467, opened 2017-06-21
Fixed the definition of 'Reducer<S>' for TypeScript 2.4.

DanielRosenwasser:
This is going to be a bit roundabout, but bear with me. 😄 

Stricter checks on generics in TypeScript 2.4 correctly errored on the fact that reducers passed to `combineReducers` were expected to be more general than they actually needed to be.

Since the `A` type parameter was never used as part of the return type -  it was only used as a constraint - the appropriate next step was to replace the parameter type of `A` with the `Action` type directly:

```ts
export type Reducer<S> = (state: S, action: Action) => S;
```

However, because of excess property checks, this introduced errors when object literals were passed directly to a reducer. For example:

```ts
const rootReducer: Reducer<RootState> = combineReducers({
  todos: todosReducer,
  counter: counterReducer,
})

const rootState: RootState = rootReducer(undefined, {
  type: 'ADD_TODO',
  text: 'test',
//~~~~~~~~~~~~ excess property
})
```

Thus, I created a new `AnyAction` which is just `Action` with an index signature. This type is separate from `Action` to avoid polluting subtypes of `Action` with an index signature.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -17,6 +17,17 @@ export interface Action {
   type: any;
 }
 
+/**
+ * An Action type which accepts any other properties.
+ * This is mainly for the use of the `Reducer` type.
+ * This is not part of `Action` itself to prevent users who are extending `Action.
+ * @private
+ */
+export interface AnyAction extends Action {
+  // Allows any extra properties to be defined in an action.
+  [extraProps: string]: any;
+}
+
 
 /* reducers */
 
@@ -43,7 +54,7 @@ export interface Action {
  *
  * @template S State object type.
  */
-export type Reducer<S> = <A extends Action>(state: S, action: A) => S;
+export type Reducer<S> = (state: S, action: AnyAction) => S;
 
 /**
  * Object whose values correspond to different reducer functions.


jimsugg:
+1 on this. Redux is unusable with Typescript 2.4 without this (or something like it.) See [https://github.com/Microsoft/TypeScript/issues/16795](url)

morlay:
Could we use another way to fix this. 
With the Generics Default (typescript@2.3+)

```ts
export type Reducer<S, A extends Action = Action> = (state: S, action: A) => S;
```
```ts
// Dispatch need to update too.
export interface Dispatch<A extends Action = Action> {
    (action: A): A;
}
```

timdorr:
Thanks for looking into this. Can someone also make a PR against the `next` branch so things are in sync for the next major release? That's using TS 2.1 (it can be upgraded to 2.4, if you want).

jimsugg:
I tested morlay's suggestion inside my project. The alternative suggestion for the Reducer definition seems to work fine:
`export type Reducer<S, A extends Action = Action> = (state: S, action: A) => S;`

However, the suggestion for the Dispatch definition causes many problems. If you just do that replacement for the Dispatch definition, it immediately produces error TS2428: All declarations of Dispatch must have identical parameters. There are a number of other references to Dispatch in the declaration file that would need to be consistent with this. In my project, I have no problem with the previous Dispatch definition, and I am not sure if this redefinition is really needed.

jimsugg:
If you did change the Dispatch definition, it would probably need to be this:
```ts
export interface Dispatch<S, A extends Action = Action> {
    (action: A): A;
}
```

The 'S' parameter was missing in morley's post. I'm still having trouble with even that substitution, however.

jimsugg:
Thanks for merging Daniel's fix, Tim. I think that's the right way to go for now. Hopefully this will get published soon.

DanielRosenwasser:
Thanks!

morlay:
@jimsugg 'S' parameter in Dispatch is not necessary.  we can keep `S` for now, but i think we can drop it in future.

lukescott:
This actually broke my code. I have custom types like this:
```ts
export interface MyAction extends Action {
// ...
}
```

And now `MyAction` is no longer correct because it isn't `AnyAction`.

lukescott:
Why was `Reducer` changed? If you have an excess property a new type should be defined. That's the point of strict typing. With the example of the first post you can do something like this:

```ts
const rootState: RootState = rootReducer(undefined, {
  type: 'ADD_TODO',
  text: 'test',
//~~~~~~~~~~~~ excess property
} as AnyAction) // cast to new type in PR `AnyAction`
```

With `Reducer` changed custom types are failing.

jimsugg:
@lukescott - Just as an experiment, I suggest you try suggestion from @morlay to see if that works for you:
```ts
export type Reducer<S, A extends Action = Action> = (state: S, action: A) => S;
```
In my test, this also fixes the issue with TS 2.4. So this might be a better solution than the one than introduces AnyAction.

lukescott:
Yep that works. This works as well:
```ts
export type Reducer<S> = (state: S, action: Action) => S;
```

jimsugg:
@lukescott - To reiterate why this is happening (you asked why ```Reducer``` was changed) - combineReducers causes errors in Typescript 2.4 with the current definition of ```Reducer```.

lukescott:
@jimsugg Looks like all my issues were TypeScript 2.4, and some fumbling on my part. `AnyAction` does work. Although as you said, the above solution you mentioned may work better. I also found it works without the `, A extends Action = Action>` part as well.

blocka:
Is there anyway to take advantage of this now, without waiting for it to be shipped? Or do I have to downgrade TS?

aikoven:
@blocka You can use this: https://github.com/reactjs/redux/pull/2182#issuecomment-275904509

ahstro:
Could this not warrant a patch release? It's basically the only change since 3.7.1, so wouldn't have to worry about other changes.

bluesh:
mark

jimsugg:
I add my request for a quick rollout of this fix. This is a blocking issue for Redux users wanting to move to TS 2.4.

aikoven:
I guess this is a non-breaking change and can be released as a patch.

toranb:
@aikoven any idea when this new type def will be in a released? Is it safe to assume it will show up in v3.7.2?

aikoven:
cc. @timdorr

timdorr:
OK, 3.7.2 is out now.

sbuzonas:
I'm not incredibly familiar with how to cope with this change. Would someone that is familiar with what has been done here be kind enough to take a look at this [broken build](https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/253877657)?

Error:
```
ERROR: 23:57  expect  TypeScript compile error: Argument of type '{ router: (state?: RouterState | undefined, action?: RouterAction | undefined) => RouterState; }' is not assignable to parameter of type 'ReducersMapObject'.
  Property 'router' is incompatible with index signature.
    Type '(state?: RouterState | undefined, action?: RouterAction | undefined) => RouterState' is not assignable to type 'Reducer<any>'.
      Types of parameters 'action' and 'action' are incompatible.
        Type 'AnyAction' is not assignable to type 'RouterAction | undefined'.
          Type 'AnyAction' is not assignable to type 'RouterAction'.
            Property 'payload' is missing in type 'AnyAction'.
The following packages had errors: react-router-redux
Error: There was a test failure.
    at Object.<anonymous> (/Users/sbuzonas/Projects/typescript/DefinitelyTyped/node_modules/types-publisher/src/tester/test-runner.ts:93:9)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/sbuzonas/Projects/typescript/DefinitelyTyped/node_modules/types-publisher/bin/tester/test-runner.js:4:58)
```

The `RouterAction` interface is defined [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-router-redux/index.d.ts#L57-L60).

The test that is failing is located [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-router-redux/react-router-redux-tests.tsx).

lukescott:
@sbuzonas edit index.d.ts of redux in node modules and try the alternate Reducer definition that doesn't use AnyAction and see if that solves your issue. If it does it may need to be patched again. 

The AnyAction worked for me, but it may not work in all cases.

huy-nguyen:
@sbuzonas There's a pending [PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/18071) for `react-router-redux` in `DefinitelyTyped` that tries to fix the problem. Can you look at it and comment on whether that will fix your problem?

sbuzonas:
@lukescott The alternative does work, the problem is on master the typescript version is 1.8 and fails to pass the test.

@huy-nguyen That PR would work, but it sacrifices the safety of type checking. It would allow a circumstance where you send an incorrect action to the wrong reducer.

Issue #2469, opened 2017-06-23
Real world example does not work on master


headquarters:
1. Clone the repo on the master branch
2. `cd redux/examples/real-world`
3. `npm install`
4. `npm start`

The real world example fails with the following errors:

```
./src/containers/App.js
  Line 9:   'propTypes' is not defined           no-undef
  Line 18:  'handleDismissClick' is not defined  no-undef
  Line 23:  'handleChange' is not defined        no-undef
```

If I pin all the dependencies to the exact version in examples/real-world/package.json, do another `npm install`, and try again I get similar errors but only reported for UserPage.js. 

So this appears to be an issue with the [class properties transform](https://babeljs.io/docs/plugins/transform-class-properties/) not working properly, but I couldn't pinpoint the exact cause. If I figure out a solution I'll put up a PR.

markerikson:
Those do not appear to be "real" errors, but rather ESLint warnings that are being marked as errors.  I think CRA may have somewhat recently decided to enforce some ESLint rules as errors.

Not sure on the right way to take care of this.

headquarters:
Ok, that makes a lot more sense considering everything is actually defined that is reported. When linter errors take over the screen and don't let you actually see the app, though, I still consider them "real". 😄 

How about a PR to disable those with /* eslint */ rules in those files?

headquarters:
Some extra information: the example works if you do NOT do `yarn` or `npm install` at the top level of the redux repo. 

So, if you only install the modules in examples/real-world, it works. If you go up to the root redux folder, install the modules, then go back and reinstall in real-world and re-run the example app then these errors show up.

markerikson:
Huh.  That's... interesting.  Some side effect of ESLint config resolution, I guess?

bbwstar:
Hi Mark

How are you doing?

I am Sasha and beginner for learning react-redux

I would like to learn about react and react-redux

Please give me good advice

Looking forward to hearing with you

Thanks, Sasha

markerikson:
@bbwstar : This is an issue tracker, not a help forum.

That said, I keep a big list of links to high-quality tutorials and articles on React, Redux, and related topics, at https://github.com/markerikson/react-redux-links . Specifically intended to be a great starting point for anyone trying to learn the ecosystem, as well as a solid source of good info on more advanced topics.  You should also check out the Reactiflux chat channels on Discord, at https://www.reactiflux.com .

PR #2470, opened 2017-06-23
Ignoring eslint rule for defined class properties

headquarters:
Fixes https://github.com/reactjs/redux/issues/2469

This is the _simplest_ solution so the example continues to run, whether or not users have installed the top-level redux node modules. However, it may not be the approach the maintainers want to take, so feedback welcome!


timdorr:
Eh, let's do this for now to reduce end user headaches while eslint and CRA figure it out.

PR #2473, opened 2017-06-26
Revert #2279

artgillespie:
See discussion in #2279


timdorr:
Ah, beat me to it. Thanks!

Issue #2459, opened 2017-06-19
'Is not defined' errors in counter example


dimitri-a:
When I run npm start in the redux\examples\counter I get :

Failed to compile.

./src/components/Counter.js
  Line 5:   'propTypes' is not defined       no-undef
  Line 11:  'incrementIfOdd' is not defined  no-undef
  Line 17:  'incrementAsync' is not defined  no-undef

Any ideas?

yaempioy:
1. Add constructor code follow.
```
constructor (props) {
    super(props)
    this.incrementIfOdd = this.incrementIfOdd.bind(this)
    this.incrementAsync = this.incrementAsync.bind(this)
  }
```
2. Change code function follow.
```
incrementIfOdd () {
    if (this.props.value % 2 !== 0) {
      this.props.onIncrement()
    }
  }

  incrementAsync () {
    setTimeout(this.props.onIncrement, 1000)
  }
```

timdorr:
A PR would be best to get this fixed.

Krizzu:
Hey,

Also, to get rid of `propTypes is not defined`, we could add prop checking to a class, like this:

```
Counter.propTypes = {
  value: PropTypes.number.isRequired,
  onIncrement: PropTypes.func.isRequired,
  onDecrement: PropTypes.func.isRequired
}
```

I can send a PR for this fix

PR #2474, opened 2017-06-26
Fixed 'is undefined' error and prop checking in Async example

leiming:
Fix the same error as #2464 #2459  for Async example.

If it is acceptable, I will fix all examples which includes #2469

### DIFF
--- a/examples/async/src/containers/App.js
+++ b/examples/async/src/containers/App.js
@@ -6,12 +6,10 @@ import Picker from '../components/Picker'
 import Posts from '../components/Posts'
 
 class App extends Component {
-  static propTypes = {
-    selectedReddit: PropTypes.string.isRequired,
-    posts: PropTypes.array.isRequired,
-    isFetching: PropTypes.bool.isRequired,
-    lastUpdated: PropTypes.number,
-    dispatch: PropTypes.func.isRequired
+  constructor(props) {
+    super(props);
+    this.handleChange = this.handleChange.bind(this);
+    this.handleRefreshClick = this.handleRefreshClick.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


timdorr:
Does our/CRA's eslint config not support class properties and statics?

leiming:
Thanks for replying, @timdorr.

I think CRA's ESLint has supported [babel-plugin-transform-class-properties](https://github.com/facebookincubator/create-react-app/blob/master/packages/babel-preset-react-app/index.js#L13). However it doesn't take effort in example cases. 

I try to add a `.eslintrc` file in `redux/examples/real-world/src`, and even install both `babel-plugin-transform-class-properties` and `babel-preset-react-app`:

```json
{
  "extends": "react-app",
  "plugins": ["transform-class-properties"]
}
```

then:

```bash
cd redux/examples/real-world/src
yarn add -D babel-preset-react-app babel-plugin-transform-class-properties
yarn start
```
It still throw error, seems ignore all ESLint config:

```
Failed to compile.

./src/containers/App.js
  Line 9:   'propTypes' is not defined           no-undef
  Line 18:  'handleDismissClick' is not defined  no-undef
  Line 23:  'handleChange' is not defined        no-undef

Search for the keywords to learn more about each error.
```

timdorr:
Oh, this looks like it's coming from the eslint install at the root of the repo: https://github.com/facebookincubator/create-react-app/issues/2604#issuecomment-311168618

Downgrading back to eslint 3.x should fix it, I believe.

leiming:
OK, I will keep track of https://github.com/facebookincubator/create-react-app/issues/2644

timdorr:
Since it's something that will be fixed upstream, I'll wait on that fix before making code changes here. Thanks for bringing this to our attention!

PR #2476, opened 2017-06-27
Building: redux-auto

codemeasandwich:
redux-auto - generate actions & reducers from your file system


timdorr:
Sorry, this isn't a binding. They are anything that plugs into another library that runs alongside the Redux store.

PR #2478, opened 2017-06-28
Updated the definition of 'Reducer' and 'Dispatch' for TypeScript 2.3+

### LINKED ISSUE #2424: [Typescript] Reducer definition causes Typescript to fall with error
Redux: 3.6.0
Typescript: 2.3.3
`compilerOptions.noImplicitAny` is set to `true`.

Redux has next definitions.
```ts
export interface Action {
  type: any;
}

export type Reducer<S> = <A extends Action>(state: S, action: A) => S;
```

Reducer:
```ts
interface AuthReducerState {
  user?: {
    id: number;
    email: string;
  };
}

export const authReducer: Reducer<AuthReducerState> = (state, action) => {
  switch (action.type) {
    default:
      return state;
  }
};
```

Errors:
```
TS7006: Parameter 'state' implicitly has 'any' type.
TS7006: Parameter 'action' implicitly has 'any' type.
```

If definitons would have been changed it would solve problem.
```ts
export interface Action {
  type: string;
  [key: any]: any;
}

export type Reducer<S> = (state: S, action: Action) => S;
```

Please explain what I do wrong.

morlay:
Another way to fix #2424 

And update Dispatch to make it more simple. ( confuse by the `S` of `Dispatch<S>`)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -43,7 +43,7 @@ export interface Action {
  *
  * @template S State object type.
  */
-export type Reducer<S> = <A extends Action>(state: S, action: A) => S;
+export type Reducer<S, A extends Action = Action> = (state: S, action: A) => S;
 
 /**
  * Object whose values correspond to different reducer functions.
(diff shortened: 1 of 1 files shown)


timdorr:
Can you make these against the `next` branch? We'll be upgrading TS there since that's the target for the next major release.

morlay:
Sure

PR #2479, opened 2017-06-29
Updated the definition of 'Reducer' and 'Dispatch' for TypeScript 2.3+

morlay:


index.d.ts
```diff
@@ -43,7 +43,7 @@ export interface Action {
  *
  * @template S State object type.
  */
-export type Reducer<S> = <A extends Action>(state: S | undefined, action: A) => S;
+export type Reducer<S, A extends Action = Action> = (state: S, action: A) => S;
```
aikoven: Why is `S | undefined` removed? Every reducer must accept `undefined` as the first arg to initialize the state.

index.d.ts
```diff
@@ -93,8 +93,8 @@ export function combineReducers<S>(reducers: ReducersMapObject<S>): Reducer<S>;
  * transform, delay, ignore, or otherwise interpret actions or async actions
  * before passing them to the next middleware.
  */
-export interface Dispatch<S> {
-    <A extends Action>(action: A): A;
+export interface Dispatch<A extends Action = Action> {
```
aikoven: I'm not sure this is correct. Redux store is able to dispatch any possible action, so we can't constrain it to a specific subtype.

Also, here's the rationale for making state type a type parameter for `Dispatch`: https://github.com/reactjs/redux/pull/1537
However, for the whole year, I've only seen it being used in `redux-thunk`. Moreover, it would give type-checking benefits only when `dispatch` is called directly via `store.dispatch(...)`, but if we e.g. use `connect` from `react-redux`, we'd still have to declare state type manually.

In light of the above, I tend to agree that we should consider removing the `S` type parameter from `Dispatch`.

index.d.ts
morlay: I don't think so.
For each `Reducer`, they may need to have the same structure of `State`, includes the initial state.

And we used to  define `Reducer` with defaults like:
```ts
const someReducer: Reducer<State> = (state = {} as State  /* this is may not undefined */, action) => state
```

index.d.ts
morlay: Yes, `redux-thunk` and `redux-promise` have different type for Dispatch. 
But I think they should be re-defined in that packages.
Or just use  `any` here :-)

index.d.ts
aikoven: Redux itself passes `undefined` to the reducer on Store init. Typings must reflect this.

morlay: Sure, I have reverted this change.

index.d.ts
morlay: How about declare `Dispatch` like this:
```ts
export interface Dispatch<TInput = Action, TReturn = Action> {
    (input: TInput): TReturn;
}
```
for `redux-thunk` user, can use like `store.dispatch<() => Action>(() => ({ type: "action" }))`
for `redux-promise` user, can use like `store.dispatch<Action, Promise<Action>>(someAction)`

aikoven: The correct typing for `redux-promise` is something like
```ts
type Dispatch = <T>(action: PromiseAction<T>) => Promise<T>;
```

Unfortunately there's no way to express this via type parameters. Currently we augment Dispatch signature in third-party middleware typings via module augmentation, and it works fine. There's a chance that with TS 2.4 we could extend Dispatch during store enhancement. I'm going to make some experiments when I have time.

index.d.ts
```diff
@@ -43,7 +43,7 @@ export interface Action {
  *
  * @template S State object type.
  */
-export type Reducer<S> = <A extends Action>(state: S | undefined, action: A) => S;
+export type Reducer<S, A extends Action = Action> = (state: S | undefined, action: A) => S;
```
aikoven: I think this should conform with https://github.com/reactjs/redux/pull/2467

aikoven:
I am sorry for nagging, in my experience, it is very hard to fix typings without introducing a breaking change, so I just want to make sure we come up with the best typings possible for Redux v4. Otherwise, we would have to wait forever for the next major.

PR #2485, opened 2017-07-01
Feature/typescript todo example

jurosh:
Typescript Todos Example - ticket: https://github.com/reactjs/redux/issues/2489

---

Hi guys,

not sure if there are some fans of TypeScript ✌️ - but definitely it's good alternative to Flow. For Redux there is already Flow todos example - but **TypeScript example project** is **missing**, and you need to search it in alternative sources.

So I thought, might be nice idea to create todos also in TypeScript! 😏

Sub-project is based on Flow example, rewritten to typescript, added required typings and changed [React create app scripts to support typescript](https://github.com/wmonk/create-react-app-typescript).

Hope this might be good start for new guys adopting redux in TypeScript. And for experienced TypeScript devs to get always fresh ideas from this example (maintained by redux community).

Soo.. what do you think?


jurosh:
Hi guys, you did great work on Redux typescript typings, can you take a look at my PR please @DanielRosenwasser @aikoven @raineorshine @use-strict

timdorr:
@jurosh I'd rather not involve a fork of react-scripts if we can avoid it. If there are divergences in functionality, examples may behave differently between the examples. I'd like to keep things as normalized as possible. Thank you, though!

PR #2490, opened 2017-07-04
Add size-limit

ai:
[size-limit](https://github.com/ai/size-limit) is tool to prevent JS libraries bloat. If somebody will add dependency with unexpected big size, Size Limit will return error and Travis CI will reject pull request.

Size Limit creates webpack project in memory and adds your lib as dependency to get real cost with all dependencies and polyfills for `process`.

Last but not least, it shows also why your lib has this size. Thanks to Webpack Bundle Analyze.

Right now MobX, PostCSS, Autoprefixer and JSS uses it. `size-limit` really help me during Logux development (I reduce size 4x, because I in Bundle Analyzer found some big dependencies and unnecessary `process` polyfill). And it is good guard for PRs.

@timdorr what do you think about this small linter for project?

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   "scripts": {
     "clean": "rimraf lib dist es coverage",
     "lint": "eslint src test build",
+    "size": "size-limit 3KB lib/index.js",
     "test": "cross-env BABEL_ENV=commonjs jest",
     "test:watch": "yarn test -- --watch",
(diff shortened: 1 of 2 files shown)


markerikson:
Appreciate the offer, but I don't think that's going to be particularly relevant here.  Redux is already minimal, and we recently started using Rollup for our builds, so they're even smaller.

PR #2496, opened 2017-07-09
writing tests for middleware example changed

brigand:
I think thunk is a better example than 'singleDispatch'. The example is also more copy/paste-able.


brigand:
Fixed some issues, and tested it against the actual redux-thunk. Ready for review.

timdorr:
LGTM. Anyone else want to give it a sanity check?

markerikson:
LGTM too.

Issue #1179, opened 2015-12-24
Precompile tests and ship them with NPM package


gaearon:
Just had an interesting discussion on Twitter about how [not shipping tests breaks `npm explore`](https://twitter.com/ljharb/status/680090399122145284). Maybe we should precompile tests and make sure they don't depend on Babel by the time `npm test` executes inside a published package. I wonder if any lib does that, and whether it's worth doing something similar.

gaearon:
The current concern is that running `npm install` and `npm test` inside the NPM package folder (not the GitHub clone!) won't work because it doesn't contain the Webpack config, babelrc, etc. But we don't want to include babelrc because of issues like #1032.

ljharb:
Basically, this is only problematic because the presence of configs inside node_modules breaks some *broken* tools - in other words, anything that delves inside `node_modules` is broken, because it's supposed to be a black box. The need to work around this exists, but it's still very frustrating that it impacts the entire ecosystem just because some people didn't want to transpile before publishing :-/

gaearon:
@ide What's the official React Native stance on this? Do you plan to encourage people to publish ES6 RN modules? Or is the plan to go back to treating node_modules as ES5 black boxes?

ide:
@gaeron it's been really good to have people publishing ES6/7 code. It's probably a little late to ship Babel with Node but if it were a possibility I think I would want to try it because it's mostly been a success for RN -- it's really good that people don't need to learn how to set up Babel for each project and source maps just work. So for React Native (and soon generic React components once RN is refactored to depend on the React package) I think we'll encourage, or at least support, publishing of ES6/7 code to npm.

gaearon:
Haha I'm so ambivalent about this. Probably going to be good long term, but painful for non-RN-exclusive maintainers today.

ljharb:
I don't think it will be good long term. Until the community settles on a convention where "main" is ES5, and "something else" is ES6, it's user-hostile to publish non-ES5.

ide:
Some thoughts: 
- Babel should provide an option that RN can use to tell Babel to ignore .babelrc files under node_modules. We still want Babel to pick up .babelrc files in the user's project though, so we don't want to ignore all .babelrc files. This is the only scalable solution that currently comes to mind. We don't want every package author to have to omit .babelrc.
- RN can and should continue to work with ES5 packages.
[...]

PR #2505, opened 2017-07-13
Remove `src` directory from npm package

KELiON:
`./src` directory is not used in npm package, only `./es` (as source for ES-module), `./lib` (as a source for old js syntax module) and `./dist` (as UMD module), so it can be removed. It reduces size of installed package by 21.4%

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,6 @@
     "dist",
     "lib",
     "es",
-    "src",
     "index.d.ts"
   ],
   "scripts": {


TrySound:
@KELiON Percents are not significant here. This library is too small. Look at leaflet package :)

KELiON:
@TrySound it is small, but it *is* significant. Imagine you have 100 node projects on your machine, each projects contains 100 npm packages. Even if each package is as small as redux (~224kb), size of all installed packages is `100*100*224KB=2.136GB`. So, by reducing each package by 21.4% we can save almost 0.5GB, that is not so small, especially when you have small SSD drive.

timdorr:
Sorry, but this is intentional. We want folks to be able to browse the source from inside the npm package. There is some npm tooling that relies on this.

TrySound:
@KELiON For 100 projects it's really small price. But if you have 100 projects with 22mb of leaflet and remove all crazy stuff in their package you can save 2GB which is really significant, because it's only ONE dependency. Take a look at highly fragmented projects (babel, jest, webpack...) or just big projects.

KELiON:
@timdorr just curious – do you have an example of such tool?

timdorr:
We had some folks wanting to look at things through `npm explore` https://docs.npmjs.com/cli/explore

timdorr:
See #1179

Issue #2508, opened 2017-07-13
Can't export combineReducers with TypeScript as AnyAction is private


alexmngn:
**Do you want to request a *feature* or report a *bug*?**  
Bug

**What is the current behavior?**

When I try to export `combineReducers`, TypeScript throw the following error:
> error TS4082: Default export of the module has or is using private name 'AnyAction'.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar.**

I simply export `combineReducers`

```
export default combineReducers({
  myReducerOne: oneReducer,
  myReducerTwo: twoReducer,
});
```

**What is the expected behavior?**

No error from typescript, we should be allowed to export `combineReducers`

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**

Typescript 2.4.1
Redux 3.7.2

I noticed the `@private` flag has been added yesterday as a comment in the definition code

```
/**
 * An Action type which accepts any other properties.
 * This is mainly for the use of the `Reducer` type.
 * This is not part of `Action` itself to prevent users who are extending `Action.
 * @private
 */
export interface AnyAction extends Action {
  // Allows any extra properties to be defined in an action.
  [extraProps: string]: any;
}
```

Typescript doesn't like this at all.
Any reason why this is flagged as private?

trusktr:
Regarding the all-too-common "has or is using private name" error, I've opened a request to fix issues like these in TypeScript by bringing declaration files to parity with language features. https://github.com/microsoft/TypeScript/issues/35822

PR #2510, opened 2017-07-14
Remove @private flag on AnyAction type definition

alexmngn:
This fixes the issue https://github.com/reactjs/redux/issues/2508


aikoven:
👍

I wonder if we can make a test for this. Am I right that you get this error when trying to generate type declarations from your code? We could possibly make a TS file that exports everything from redux and then in tests we compile it with `--noEmit`

timdorr:
Thanks!

PR #2514, opened 2017-07-14
Add library size check in CI

siddharthkp:
This will help in **keeping** redux tiny.

It's basically,

this in the terminal: 

```
 PASS  ./dist/redux.min.js: 2.18kB < maxSize 2.5kB gzip
```

and something like this in the pull request: 

1. Setting the file and threshold:

    The config sits inside `package.json`, the maxSize is set to 2.5kB right now.
You can add multiple files here. ([docs](https://github.com/siddharthkp/bundlesize/blob/master/README.md#configuration))

2. Authorize bundlesize:

    `bundlesize` needs access to github API with scope `repo:status`, You can follow the [steps here](https://github.com/siddharthkp/bundlesize/blob/master/README.md#2-build-status)

--------

Changes made:

- [x] Add bundlesize as dev dependency
- [x] Add config for bundlesize
- [x] Move `build` before `test` in `prepare` script
- [x] Split `test` script to `test:unit` and `test:size`
- [ ] Add github token to travis environment variables (you have to do this)

Happy to hear your feedback 😄

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -910,9 +910,9 @@
       "integrity": "sha1-Jw8HbFpywC9bZaR9+Uxf46J4iS8="
     },
     "bundlesize": {
-      "version": "0.11.1",
-      "resolved": "https://registry.npmjs.org/bundlesize/-/bundlesize-0.11.1.tgz",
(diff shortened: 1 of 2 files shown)


timdorr:
Hmm, I like this. Would be interesting for our efforts in #2358.

@gaearon Thoughts? Is this something you care about at all?

siddharthkp:
@timdorr Anything I can do to help? 😅

PR #2515, opened 2017-07-15
Make reducer generics more generic

sbuzonas:
This enforces that the action parameter extends `Action` for strongly typed implementations, and defaults to `AnyAction` for the case where an action may not be explicitly typed.

Enhancement to #2467 
Addresses an issue discovered in a failed build for DefinitelyTyped/DefinitelyTyped#17777

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -54,13 +54,13 @@ export interface AnyAction extends Action {
  *
  * @template S State object type.
  */
-export type Reducer<S> = (state: S, action: AnyAction) => S;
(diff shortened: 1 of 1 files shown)


sbuzonas:
This doesn't work because of the typescript version. It would need 2.3.

Should it be against `next` instead? The current latest version breaks a lot of types that extend `Action` rather than `AnyAction`

sbuzonas:
Closing because I have no idea how to correctly fix this. Opened #2516 for discussion.

Issue #2516, opened 2017-07-15
TypeScript: Can't use strict custom Action types any more


sbuzonas:
**Do you want to request a *feature* or report a *bug*?**  
Bug

**What is the current behavior?**
TypeScript declarations extending `Action` are broken because the `Reducer` type requires action to be `AnyAction`.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar.**

```ts
const CALL_HISTORY_METHOD: string = "something";

interface RouterAction extends Action {
  type: typeof CALL_HISTORY_METHOD;
  payload: any;
}

interface RouterState { }

interface State {
  router: RouterState;
}

function routerReducer(state?: RouterState, action?: RouterAction): RouterState { }

const reducers: Reducer<State> = combineReducers({router: routerReducer});
```

**What is the expected behavior?**
I should be able to strictly type the action my reducer expects.

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**
This has worked since 3.4.0. (Some vague indication that 3.0.8, but no idea where the types came from prior to 3.4)

sbuzonas:
The issue isn't in the definition here. It's in the other definition, `AnyAction` is more accurate to the actual behavior.

Linking DefinitelyTyped/DefinitelyTyped#17652 and DefinitelyTyped/DefinitelyTyped#18071

PR #2517, opened 2017-07-15
Fix the definition of 'AnyAction'

yenshih:
This PR fixes Issue [#2516](https://github.com/reactjs/redux/issues/2516)

Reference: TypeScript Repo Issue [#17223](https://github.com/Microsoft/TypeScript/issues/17223)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -23,10 +23,7 @@ export interface Action {
  * This is not part of `Action` itself to prevent users who are extending `Action.
  * @private
  */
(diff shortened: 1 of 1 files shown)


sbuzonas:
This removes the safety allowing for non objects to be passed to a reducer. It also removes the requirement of the `type` property... which isn't technically an implementation requirement, but more of a standard practice.

timdorr:
Superceded by #2563

PR #2523, opened 2017-07-18
Make ReducersMapObject more typesafe

osi-oswald:
This will catch misspelling of properties inside combineReducers and complain if reducers don't reduce the correct child state type.

("S = any" for backwards compatibility)


osi-oswald:
Oh just found out this is already in the 'next' branch

Issue #2481, opened 2017-06-29
Error in types for middleware (typescript 2.4.1)


artem-malko:
**It is a bug**

I have this simple middleware:
```
import { Middleware, Store, Dispatch, Action } from 'redux';
import { AppState } from 'types/appState';

export default function createSimpleMiddleware(): Middleware {
  return (store: Store<AppState>) => (next: Dispatch<AppState>) => (action: Action): Action => {
    return next(action);
  };
}
```
It is useless, but there is a problem) This code  works with typescript version 2.3.x and below. But in typescript 2.4 there is an Error.

**What is the current behavior?**
In typescript I have this Error:
```
Types of parameters 'store' and 'api' are incompatible.
    Type 'MiddlewareAPI<S>' is not assignable to type 'Store<AppState>'.
```

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar.**

You can find demo code at the top of current issue.

**What is the expected behavior?**

I expect, that there won't be any errors.

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**

Typescript: 2.4.1
Redux: 3.7.1
NodeJS: 6.9.1
OS: MacOS

timdorr:
Just merged this in yesterday: 2467

artem-malko:
@timdorr but my bug is not about reducers. It is about middleware)

artem-malko:
Fix from #2467 fixes many problems, yes, but nit for middleware.

timdorr:
Sorry, this is all sort of a mess and I'm not well-versed in TS. Is #2483 related?

morlay:
@artem-malko could you test this pr #2479 with ts@2.4.1, it works well in my project.

artem-malko:
@timdorr I understand. No #2483 is about reducers too, not about middleware)
@morlay no, my example is not working( 
```
Types of parameters 'store' and 'api' are incompatible.
    Type 'MiddlewareAPI<S>' is not assignable to type 'Store<AppState>'.
```

And if I change code to: 
```
import { Middleware, MiddlewareAPI, Dispatch, Action } from 'redux';
import { AppState } from 'types/appState';

export default function createSimpleMiddleware(): Middleware {
  return (store: MiddlewareAPI<AppState>) => (next: Dispatch<AppState>) => (action: Action): Action => {
    return next(action);
  };
}
```

It has errors too:
```
[...]

PR #2530, opened 2017-07-26
Improve typings for Middleware

davidka:
This PR fixes Issue #2481

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -278,6 +278,10 @@ export interface MiddlewareAPI<S> {
  * logging actions, performing side effects like routing, or turning an
  * asynchronous API call into a series of synchronous actions.
  */
+export interface SpecificMiddleware<S> {
+  (api: MiddlewareAPI<S>): (next: Dispatch<S>) => Dispatch<S>;
+}
+
 export interface Middleware {
   <S>(api: MiddlewareAPI<S>): (next: Dispatch<S>) => Dispatch<S>;
 }
--- a/test/typescript/middleware.ts
+++ b/test/typescript/middleware.ts
@@ -1,5 +1,5 @@
 import {
-  Middleware, MiddlewareAPI,
+  Middleware, SpecificMiddleware, MiddlewareAPI,
   applyMiddleware, createStore, Dispatch, Reducer, Action
 } from "../../index";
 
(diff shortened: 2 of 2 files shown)


timdorr:
Can I get some eyes on this to confirm this fixes the issue? Thanks!

tbo:
This doesn't fix the issue for me. I get additional error messages:

```
    ERROR in [...]/node_modules/redux-thunk/index.d.ts
    (14,22): error TS2314: Generic type 'Middleware<S>' requires 1 type argument(s).
    ERROR in [...]/node_modules/redux-thunk/index.d.ts
    (15,42): error TS2314: Generic type 'Middleware<S>' requires 1 type argument(s).
    ERROR in [...]/shared/createStore.ts
    (27,3): error TS2345: Argument of type '(store: MiddlewareAPI<IState>) => (next: any) => (action: any) => any' is not assignable to parameter
 of type 'GenericMiddleware'.
      Types of parameters 'store' and 'api' are incompatible.
        Type 'MiddlewareAPI<S>' is not assignable to type 'MiddlewareAPI<IState>'.
          Type 'S' is not assignable to type 'IState'.
```

aikoven:
This will break any existing usage of `Middleware` type since it now became a generic type.

We should rather add a new type without touching existing ones.

davidka:
I could do it the other way and introduce a new type SpecificMiddleware: 

```
export interface SpecificMiddleware<S> {
  (api: MiddlewareAPI<S>): (next: Dispatch<S>) => Dispatch<S>;
}

export interface Middleware {
  <S>(api: MiddlewareAPI<S>): (next: Dispatch<S>) => Dispatch<S>;
}
```

But if we would update the typescript dependency from 1.8.0 to 2.4.2, we could make it even more elegant with only one type: 

```
export interface Middleware<T=any> {
    <S>(api: MiddlewareAPI<S>): (next: Dispatch<S>) => Dispatch<S>;
    (api: MiddlewareAPI<T>): (next: Dispatch<T>) => Dispatch<T>;
  }
```

@timdorr Does anything stand against?

tbo:
@davidka I don't think we should force everybody to use the latest typescript version. Some people might run into problems with unresolved conflicts in other dependencies or have other restrictions in place.

davidka:
@tbo yeah that make sense. 
I have now added an extra typing `SpecificMiddleware` which accepts a specific type.
But I do not really like that.

davidka:
@timdorr Could we merge that?

tbo:
I retested this and it still doesn't fix the issue for me. Maybe it is limited to my specific setup, so I would suggest, that somebody else should test this as well.

timdorr:
Superceded by #2563

PR #2531, opened 2017-07-27
Minor copy update.

haleyhuiliu:


timdorr:
Actually this is correct as is: https://english.stackexchange.com/questions/105116/is-it-a-user-or-an-user

Thanks anyways!

PR #2534, opened 2017-07-27
fix shopping-cart example to show inventory

ahuff44:
I noticed at https://codesandbox.io/s/github/reactjs/redux/tree/master/examples/shopping-cart that the number of items left in stock isn't shown; I dug around and figured out that the `inventory` prop wasn't being passed down


timdorr:
Neat! Thanks!

PR #2535, opened 2017-07-28
Update FAQ: Why should I use action creators?

maxhallinan:
This adds a "Why should I use action creators?" section to the Actions FAQ. The answer is based on sources suggested in #1785 . cc: @markerikson


docs/faq/CodeStructure.md
```diff
@@ -72,3 +73,27 @@ Find the balance between these two extremes, and you will master Redux.
 - [#1165: Where to put business logic / validation?](https://github.com/reactjs/redux/issues/1165)
 - [#1171: Recommendations for best practices regarding action-creators, reducers, and selectors](https://github.com/reactjs/redux/issues/1171)
 - [Stack Overflow: Accessing Redux state in an action creator?](http://stackoverflow.com/questions/35667249/accessing-redux-state-in-an-action-creator/35674575)
+
+
+<a id="structure-action-creators"></a>
+### Why should I use action creators?
+
+Action creators are superior to inline actions for several reasons.
```
markerikson: Let's rephrase this, and add a bit of extra info in here as well.

First, I'd say that action creators are not a required part of Redux, and you can dispatch actions directly from a component if you want, but they are a good practice for several reasons.  (I'd also remove the phrase "superior to inline actions".)

Would also be nice if we can squeeze in a mention that they aren't "just" an idea with Redux, but that Redux inherited the idea from the original Flux architecture, per the Flux docs at https://facebook.github.io/react/blog/2014/07/30/flux-actions-and-the-dispatcher.html#actions-and-actioncreators .

maxhallinan: Thanks for the review! I agree - 'superior to inline actions' has an imperious tone I didn't intend.

maxhallinan:
@markerikson, I've expanded the first paragraph with your suggestions. Let me know what you think.

markerikson:
I like it.  Stamped!

maxhallinan:
Thanks for the merge!

PR #2537, opened 2017-07-28
Update FAQ: "Dispatch multiple actions" links

maxhallinan:
This updates links for "Should I dispatch multiple actions in a row from one action creator?" in the Actions FAQ per #1785 . cc: @markerikson


docs/faq/Actions.md
```diff
@@ -108,9 +108,14 @@ Try to avoid dispatching several times synchronously in a row in the places wher
...
+
 **Discussions**
 - [#597: Valid to dispatch multiple actions from an event handler?](https://github.com/reactjs/redux/issues/597)
 - [#959: Multiple actions one dispatch?](https://github.com/reactjs/redux/issues/959)
 - [Stack Overflow: Should I use one or several action types to represent this async action?](http://stackoverflow.com/questions/33637740/should-i-use-one-or-several-action-types-to-represent-this-async-action/33816695)
 - [Stack Overflow: Do events and actions have a 1:1 relationship in Redux?](http://stackoverflow.com/questions/35406707/do-events-and-actions-have-a-11-relationship-in-redux/35410524)
 - [Stack Overflow: Should actions be handled by reducers to related actions or generated by action creators themselves?](http://stackoverflow.com/questions/33220776/should-actions-like-showing-hiding-loading-screens-be-handled-by-reducers-to-rel/33226443#33226443)
+- [Twitter: "redux-thunk is too powerful and is a foot-gun..."](https://twitter.com/intelligibabble/status/800103510624727040)
```
maxhallinan: @markerikson do you think this is worth including? The most interesting aspect of this thread is Dan Abramov's comment that redux-thunk wasn't intended to be the canonical approach for asynchronous dispatching. But the thread doesn't illuminate problems with this pattern or clarify what "convention-based middleware" means. My concern is that this link would undermine confidence in a common pattern without helping to improve on it. It seems more like interesting ephemera for someone who likes to think about these things and less like a step up for someone in need of answers.

markerikson: Hmm.  Yeah, we can maybe leave it out, especially since my "Thoughts on Thunks" article quotes large portions of the thread and addresses the question in greater detail.  (Also, for reference, the "convention-based middleware" phrase refers to libs like Leland's `redux-pack` middleware, which provides a more specific API for dispatching actions when a promise is resolved and rejected.)

maxhallinan:
@markerikson Ok, I removed the "thunk is foot gun" link. Are there any other housekeeping tasks that could be added to this PR? The other links listed for addition on #1785 (re: Code Structure and Performance) have already been added.

markerikson:
Eh, let's go ahead and get these in.  I need to step back and review what's been added, and what still needs to be added out of that list of stuff in #1785.

maxhallinan:
Sounds good. Thanks for the merge.

PR #2538, opened 2017-08-01
Add missing SET_VISIBILITY_FILTER import

ariabov:
The reducer example uses `SET_VISIBILITY_FILTER` constant,  but the import statement prior to the example is not explicitly importing it. This change explicitly imports the constant.


markerikson:
Thanks!

PR #2539, opened 2017-08-01
Make easy for compose

shfshanyue:
It should be behave like the first function if only give one for compose.
so it could be easier.


markerikson:
This function has been tweaked numerous times.  The current implementation is sufficient, and we have no reason to change it further.

PR #2540, opened 2017-08-01
Line 404 -414 Agreement

KevinEverywhere:
TodoList on line 414 should be replaced with VisibleTodoList to match the import statement in line 404. Either TodoList or VisibleTodoList works functionally, but VisibleTodoList demonstrates implementation of container component.


timdorr:
Thanks!

PR #2546, opened 2017-08-06
Use a callback ref in the Real World Example

seethemhigh:
Hello. Here's a minor PR for you that replaces a string ref with a callback.


markerikson:
Thanks, appreciate it!

PR #2548, opened 2017-08-07
Corrected the name of the link for immutable.js library.

Tetsuya3850:
It was previously confused with the article's title above.
Was a bit puzzled when reading the docs. So I made a simple replace.
Hope you could merge it!


markerikson:
Thanks!

PR #2549, opened 2017-08-07
added missing invalidateSubreddit action

ahmetcetin:
invalidateSubreddit function and its constant were missing in the actions.js after async functions changes introduced.


markerikson:
Thanks!

PR #2550, opened 2017-08-09
Update to Babel 7

hzoo:
If y'all don't want to upgrade now that's ok, just a test PR then.

Could also change the explicit plugins in favor of https://github.com/babel/babel-preset-env with the [exclude](https://github.com/babel/babel-preset-env#exclude) option?

```
    "check-es2015-constants",
    "transform-es2015-arrow-functions",
    "transform-es2015-block-scoped-functions",
    "transform-es2015-block-scoping",
    ["transform-es2015-classes", { "loose": true }],
    ["transform-es2015-computed-properties", { "loose": true }],
    ["transform-es2015-destructuring", { "loose": true }],
    ["transform-es2015-for-of", { "loose": true }],
    "transform-es2015-function-name",
    "transform-es2015-literals",
    "transform-es2015-object-super",
    "transform-es2015-parameters",
    "transform-es2015-shorthand-properties",
    ["transform-es2015-spread", { "loose": true }],
    "transform-es2015-sticky-regex",
    ["transform-es2015-template-literals", { "loose": true }],
    "transform-es2015-unicode-regex",

    // exclude
    "transform-es2015-duplicate-keys"
    "transform-es2015-typeof-symbol"
    "transform-regenerator"
    "transform-exponentiation-operator"
    "transform-async-to-generator"
    "syntax-trailing-function-commas"

    "transform-object-rest-spread",
    "transform-es3-member-expression-literals",
    "transform-es3-property-literals"
```


hzoo:
Hmm failure is due to using an old version of traverse (some dep isn't updated, maybe cache issue)

timdorr:
I assume we'd want to wait for at least the beta. Any rough schedule for release?

timdorr:
Also see the last time we updated: #1347

I don't think the explicit transforms are necessary anymore. Although, if we want to target a specific browser/node version, we should only update that on the `next` branch.

hzoo:
@timdorr just fyi alpha shouldn't be any different in the beta for most projects unless you use something custom - https://github.com/babel/babel/milestone/9 need to review 2 prs. We can still run all the transforms.

Oh didn't know about the `next` branch

timdorr:
No worries. I'm happy to merge this in on the master branch, as long as the same targets (IE8, I think?) are supported. We can move up to IE11 in the next major.

hzoo:
Yeah was my intention to keep it the same, just whether we wanted to use the same plugins or do it via preset-env

timdorr:
@hzoo I'm going to close this out for now, since it's just version bumps. Once 7.0 goes rc or final, I can look into re-opening or making a new PR. We can also finally get on preset-env! 👍  Thanks for looking into this and good luck on the path to 7.0.0!

PR #2551, opened 2017-08-11
Allow updating currentState when calling replaceReducer

### LINKED ISSUE #2533: Feature: Allow updating currentState when calling replaceReducer
I would like to suggest adding an optional param to `replaceReducer` that would allow updating the `currentState`. This is needed when using code splitting with SSR (see use case below).

Proposed code change (in bold)
<pre>
function replaceReducer(nextReducer<strong>, nextState</strong>) {
  if (typeof nextReducer !== 'function') {
  }

  currentReducer = nextReducer
  <strong>currentState = nextState || currentState</strong>
  dispatch({ type: ActionTypes.INIT })
}
</pre>
This change will allow hydrating the store with state that was inlined in the page when doing SSR but belongs to a reducer that is being loaded after the store has been initialized because of code splitting.

Use case: https://gist.github.com/cvbuelow/ea7bd4deb7824fdb65d81cd2b8c60f5e
This is based off @gaearon's [gist](https://gist.github.com/gaearon/0a2213881b5d53973514) and addresses the SSR concerns mentioned there. Look at the [latest revision](https://gist.github.com/cvbuelow/ea7bd4deb7824fdb65d81cd2b8c60f5e/revisions#diff-8b3d2f22fc7e4f74eeb25762947f7c36) in my fork to see the diff.

Let me know If this feature looks ok and I can submit a PR.

cvbuelow:
Closes #2533

### DIFF
--- a/docs/Glossary.md
+++ b/docs/Glossary.md
@@ -97,7 +97,7 @@ type Store = {
   dispatch: Dispatch
   getState: () => State
   subscribe: (listener: () => void) => () => void
-  replaceReducer: (reducer: Reducer) => void
+  replaceReducer: (reducer: Reducer, nextState?: State) => void
 }
 ```
 
@@ -107,7 +107,7 @@ There should only be a single store in a Redux app, as the composition happens o
 - [`dispatch(action)`](api/Store.md#dispatch) is the base dispatch function described above.
 - [`getState()`](api/Store.md#getState) returns the current state of the store.
 - [`subscribe(listener)`](api/Store.md#subscribe) registers a function to be called on state changes.
-- [`replaceReducer(nextReducer)`](api/Store.md#replaceReducer) can be used to implement hot reloading and code splitting. Most likely you won't use it.
+- [`replaceReducer(nextReducer, [nextState])`](api/Store.md#replaceReducer) can be used to implement hot reloading and code splitting. Most likely you won't use it.
 
 See the complete [store API reference](api/Store.md#dispatch) for more details.
 
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -18,7 +18,7 @@ This section documents the complete Redux API. Keep in mind that Redux is only c
   * [getState()](Store.md#getState)
   * [dispatch(action)](Store.md#dispatch)
   * [subscribe(listener)](Store.md#subscribe)
-  * [replaceReducer(nextReducer)](Store.md#replaceReducer)
+  * [replaceReducer(nextReducer, [nextState])](Store.md#replaceReducer)
 
 ### Importing
 
--- a/docs/api/Store.md
+++ b/docs/api/Store.md
@@ -15,7 +15,7 @@ To create it, pass your root [reducing function](../Glossary.md#reducer) to [`cr
 - [`getState()`](#getState)
 - [`dispatch(action)`](#dispatch)
 - [`subscribe(listener)`](#subscribe)
-- [`replaceReducer(nextReducer)`](#replaceReducer)
+- [`replaceReducer(nextReducer, [nextState])`](#replaceReducer)
 
 ## Store Methods
 
@@ -125,7 +125,7 @@ unsubscribe()
 
 <hr>
 
-### <a id='replaceReducer'></a>[`replaceReducer(nextReducer)`](#replaceReducer)
+### <a id='replaceReducer'></a>[`replaceReducer(nextReducer, [nextState])`](#replaceReducer)
 
 Replaces the reducer currently used by the store to calculate the state.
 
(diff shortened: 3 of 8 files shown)


faceyspacey:
I found this PR because I wanted to see if anyone had tried to create a PR to allow you to provide an `action` as a second argument. It would be dispatched instead of the`INIT` action. 

The thinking is that if you have a new reducer, it likely happened as a result of code splitting which likely happened as a result of a route change action. That means the route change action already happened, and the the new reducer won't properly respond to it.

It seems, the approach you're going for is creating the state the reducers would have and then preloading it. For the use case I'm thinking of, that's quite a bit of extra work. 

My use case can be handled by an additional action, but why have your reducers do 2x the work when it could be just one additional action. So I wonder if this would also address your use case. If you dispatch the correct action, you shouldn't have to preloaded the state--the state should be *arrived at* as a result of the action.

In short, the implementation would be this:


```js
function replaceReducer(nextReducer, nextAction) {
  if (typeof nextReducer !== 'function') {
  }

  currentReducer = nextReducer
  const action = nextAction || { type: ActionTypes.INIT }
  dispatch(action)
}
```

cvbuelow:
@faceyspacey my use case is specifically for doing ssr with code splitting and react-router <= v3
as you can see here: https://gist.github.com/cvbuelow/ea7bd4deb7824fdb65d81cd2b8c60f5e
I don't need to manually build the state because I already have it from the ssr. Just need to do something like
```js
const nextState = {
  ...store.getState(),
  [name]: store.preloadedState[name]
};
```
I guess I *could* do this by firing off an action like you are suggesting but that would require adding more code to all my reducers to populate it:
```js
/* store.js */
export function injectAsyncReducer(store, name, asyncReducer) {
  const nextAction = {
    type: `INIT_STATE_${name.toUpperCase()}`,
    state: store.preloadedState[name]
  };
  store.asyncReducers[name] = asyncReducer;
  store.replaceReducer(createReducer(store.asyncReducers), nextAction);
}

/* ./reducers/comments.js */
export default function comments(state = initialState, action) {
  switch (action.type) {
    case "INIT_STATE_COMMENTS":
      return action.state;
    ...
  }
}
```
And then for each nested reducer I would need to do the same. So yes, I think it would be possible, just not as clean.

faceyspacey:
i see. but your preloaded state is always the same as it was in the beginning. It seems your also sending down all the state you might need for your reducers on first load. Typically you won't just have all potential state sitting around, but you'll also get it along with the reducer yourself--likely through the strategy of firing an action (whether in `replaceReducer`) or as an action immediately after it like we do now.

It seems you're building your app based around "component level" data fetching (and subsequent state fetching) rather than route level (based on some action triggering it). So you don't have a predictable action that correlates with this component mounting. Or do you? If you do, that action should also trigger this reducer to be in the correct state, rather than having to do `"INIT_STATE_COMMENTS"`.

I have a feeling the state as the 2nd arg really isn't the general purpose approach here. The action is more flexible. Using `state` means you have to create an ad hoc way to get the upcoming state. What you implemented is just one possible case, but I think many people would want the state to be not just the `preLoadedState`, but state that might have evolved by the point code splitting occurs. That's the thing. 

An action would better solve that case. Essentially it allows the same reducer to respond to multiple actions that could trigger the code splitting, or even just a single one with a differing payload--rather than just one state preloaded state received at the beginning for all reducers.

Just some thoughts..

cvbuelow:
Again, my use case is for the first page load with ssr, code splitting, and react-router <= v3. So the actions were already performed on the server resulting in the initial state. Yes, I am building my app around route level data fetching but I don't want to have to call the action again on the client because it would trigger the api calls again. So the only way to set the state is like I'm proposing.

Also I see `replaceReducer` as following the same signature as `createStore`. I mean really `replaceReducer` could probably be refactored to something called `updateStore` and share code with `createStore`. Just a thought.

As for making the second arg the action, it doesn't seem to me like this adds a lot of value aside from saving the time it takes to execute the INIT action since you can just do
```js
store.dispatch(nextAction());
```
If you feel that the savings in execution time is truly important then maybe we need
```js
replaceReducer(nextReducer, nextState, nextAction)
```

faceyspacey:
Well they obviously solve all. I honestly just don't see the Redux people even bothering either either of the 3 options. But it would be nice if they did as the game is finally changing to heavily code split plus SSR universal apps.

Question: why do u need replaceReducer if u do this on first load? Can't u just use create store?

cvbuelow:
Good question. Because of code splitting then I would have to call `createStore` in the `require.ensure` in the `getComponent` of the route.
https://gist.github.com/cvbuelow/ea7bd4deb7824fdb65d81cd2b8c60f5e#file-routes-js-L17

cvbuelow:
@timdorr have you had a chance to look at this?

cvbuelow:
After speaking with @gaearon about this he suggested adding a layer to inject the state before calling `combineReducers`. Tried it out and it works great: https://gist.github.com/cvbuelow/ea7bd4deb7824fdb65d81cd2b8c60f5e

PR #2555, opened 2017-08-14
shopping-cart example: add the ability to remove items from your cart

stormasm:
I thought it would be nice to add in the functionality to remove items from your
cart prior to checkout.  Sometimes you make a mistake and didn't mean to add
in some items, this way we can now delete items from the cart prior to clicking
on the checkout button.


examples/shopping-cart/src/actions/index.js
```diff
@@ -17,12 +17,23 @@ const addToCartUnsafe = productId => ({
   productId
 })
 
+const removeFromCartUnsafe = (productId,quantityById) => ({
```
timdorr: Why is this called "unsafe"? Sounds dangerous... ⚠️

stormasm: Hi Tim, Thanks for reviewing my code. I was following the naming convention of the **addToCartUnsafe**, but happy to rename the method, what is a better name for that method ? We already have **removeFromCart** so let me know, thanks... Also, the tests are failing, so I am working on fixing the tests as well.  I will do another commit once I fix the failing tests.  Stay tuned...

stormasm:
Tim, I fixed the failing test, added some new tests related to removing items from the Cart and renamed the method you pointed out from "Unsafe" to "Complete"...  Hopefully, you have the ability to rebase my commits down to one commit if you accept my pull request.  Thanks, ---Michael

stormasm:
Tim, I added lots more functionality to the shopping cart since my last message to you on August 16.  

Interestingly enough you have a testimonial on the reactabular website https://reactabular.js.org/#/ 

So you might find my new additions to the shopping cart interesting as I use reactabular exclusively for the view.

Here is my new repo https://github.com/stormasm/receipt 
along with the code running here: https://stormasm.github.io/receipt/  

My plan is to blow away my repo https://github.com/stormasm/redux where the pull request originated since it doesn't look like my pull request is going to land in redux.  I assume you can delete my pull request on your end as well. 

We might want to put a link to my new repo in the Readme of the shopping-cart repo ?  

Let me know if you have any other ideas along these lines.  

Thanks --- Michael

timdorr:
@stormasm Thanks for the PR, but I'd rather not increase the complexity of this example. This one is already pretty bloated, so it gets harder for others to read through and keep everything in their mind as they learn it. So, thank you, but I'm going to close this out.

stormasm:
Tim,

I would agree with you ...  Keeping it simple makes sense..
Thanks for taking a look...
I will keep you posted on other stuff I find that might be pertinent.
Thanks again for the review,
Michael

Issue #2501, opened 2017-07-11
"dispatch" passed to middlewares only passes through first argument


Asvarox:
**Do you want to request a *feature* or report a *bug*?** Bug

**What is the current behavior?**
Right now a "dispatch" function that's passed to middlewares passes through only the first argument.

From `applyMiddleware.js`:
```javascript
    var middlewareAPI = {
      getState: store.getState,
      dispatch: (action) => dispatch(action)
    }
```
I have a custom middleware that accepts a dispatch call with 2 arguments. This works fine in, say, `mapDispatchToProps` (from `react-redux`) as it uses `store.dispatch`, but in `dispatch` calls in redux-thunk, the second argument disappears.

**What is the expected behavior?**
All the arguments are passed to the dispatch:
```javascript
    var middlewareAPI = {
      getState: store.getState,
      dispatch: (...args) => dispatch(...args)
    }
```

There are ways to work around it (you can put all the arguments in an array or object), additional arguments can also be stripped by any middleware in middlewares chain, but still it would make things more consistent.

If it makes sense I can submit a PR.

timdorr:
Yeah, go ahead and make a PR for this. I'd be interested to see if there are any objections.

markerikson:
Erm... potentially.  This is _very_ similar to the issue discussed in #1813 .  As far as I know, all existing middlewares and enhancers assume that there's only one argument being passed to `dispatch`.  So, this would only work if `applyMiddleware` is the first enhancer, _and_ your middleware is the first one in the chain.  

Out of curiosity, @Asvarox , what is your current use case?  What does your custom middleware do?

Asvarox:
@markerikson sure. Given a dispatch is called with function as first argument and an array as second, it will `dispatch` the result of calling the function passing contents of the array as arguments:
```javascript
// Written "by hand",this is the "core" of actual code
const middleware = ({ dispatch }) => next => (actionCreator, args) => dispatch(actionCreator(...args));
```
The reasoning is to make testing thunks easier. Ihave quite a few of these that dispatch other actions - by dispatching the result of calling another action creator. That forced me to assert what these action creators return - if one of them started to return another thunk instead of plain object, it broke the tests. More so it was then pretty hard to fix the assertion.

[...]

PR #2560, opened 2017-08-15
Make middleware API dispatch pass through all call arguments

Asvarox:
As discussed in #2501, there's a bit of inconsistency as you can call `store.dispatch` with multiple arguments just fine, but in context of middlewares (say, thunk) only the first one is passed through. This PR tries to fix that.

### DIFF
--- a/src/applyMiddleware.js
+++ b/src/applyMiddleware.js
@@ -24,7 +24,7 @@ export default function applyMiddleware(...middlewares) {
 
     const middlewareAPI = {
       getState: store.getState,
-      dispatch: (action) => dispatch(action)
+      dispatch: (...args) => dispatch(...args)
     }
     chain = middlewares.map(middleware => middleware(middlewareAPI))
     dispatch = compose(...chain)(store.dispatch)
--- a/test/applyMiddleware.spec.js
+++ b/test/applyMiddleware.spec.js
@@ -92,6 +92,26 @@ describe('applyMiddleware', () => {
     })
   })
 
+    it('passes through all arguments of dispatch calls from within middleware', () => {
+        const spy = jest.fn()
+        const testCallArgs = ['test']
+        function multiArgMiddleware() {
+            return next => (action, callArgs) => {
+                if (Array.isArray(callArgs)) {
+                    return action(...callArgs)
+                }
+                return next(action)
+            }
+        }
+        function dummyMiddleware({ dispatch }) {
+            return next => action => dispatch(action, testCallArgs)
+        }
+
+        const store = createStore(reducers.todos, applyMiddleware(multiArgMiddleware, dummyMiddleware))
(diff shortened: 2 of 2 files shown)


timdorr:
This should be fine. It definitely isn't breaking. Middleware can still block other args from being passed through to `next()`, but it at least opens up the possibility to do so.

gaearon:
I think the reason we didn't do this initially is because we wanted to make it easier for all middleware to be compatible. Now we need to either force all middlewares to pass it through or live with a situation where adding a seemingly harmless middleware can break the other code in a somewhat non-obvious way.

gaearon:
As for the use case in https://github.com/reactjs/redux/issues/2501, I don't quite see what prevents the user from passing an array: `dispatch([myThunk, arg1, arg2])`.

timdorr:
Middleware changes the definition of what dispatch's args can be anyways. So whether you use an array or allow multivariate functions, it's really no different because the function signature is changed. It's odd for a loosely-designed API to have a restriction thrown in there.

markerikson:
@timdorr : I'd disagree with the phrase "changes the definition of what dispatch's args can be", at least to some extent.  Yes, they allow you to pass in something other than a plain action, but the working assumption thus far has been that there's only _one_ value being passed through, much like a promise pipeline.  

As you said, allowing multiple args here shouldn't break any existing behavior, but given that all existing middlewares assume there's only one arg, I'm not sure about the actual benefit.  (This is very related to the very long discussion in #1813 .)

gaearon:
IMO allowing this will benefit a tiny minority of users (who already have a fine workaround) but will add cost to every middleware author who will eventually be asked to support this. This change, to be successful, will have to trigger a cascade of changes in every library that delegates to dispatch. And it makes Redux harder to type (which is already hard). In my eyes the cost outweighs the benefit.

src/applyMiddleware.js
```diff
@@ -24,7 +24,7 @@ export default function applyMiddleware(...middlewares) {
 
     const middlewareAPI = {
       getState: store.getState,
-      dispatch: (action) => dispatch(action)
+      dispatch: (...args) => dispatch(...args)
```
sergeysova: why not just return `dispatch`?

```js
const middlewareAPI = {
  getState: store.getState,
  dispatch,
}
```

FranciscoMSM: Hey! Can someone have an explanation for this change?

timdorr: @FranciscoMSM This particular change is to allow middleware to change the arguments of dispatch and pass that through. 

The reason for the extra function is to remove direct access to `dispatch` so middleware doesn't try to attach properties or other hidden metadata to it. Each middleware gets its own unique copy of `dispatch` this way, and there is less chance of interference between middlewares.

Issue #1648, opened 2016-04-20
TypeScript definition for StoreEnhancer prevents state type extension


fdecampredon:
With the current type definition of `StoreEnhancer`:

``` typescript
export type StoreEnhancer<S> = (next: StoreEnhancerStoreCreator<S>) => StoreEnhancerStoreCreator<S>;
export type GenericStoreEnhancer = <S>(next: StoreEnhancerStoreCreator<S>) => StoreEnhancerStoreCreator<S>;
export type StoreEnhancerStoreCreator<S> = (reducer: Reducer<S>, initialState: S) => Store<S>;
```

We cannot define an enhancer that change the type of the sate : 

``` typescript

declare function reducerWithFoo<S>(reducer: Reducer<S>): Reducer<S & { foo: string }>;

function test<S>(next: StoreEnhancerStoreCreator<S>): StoreEnhancerStoreCreator<S & { foo: string }> {
  return (reducer: Reducer<S>, initialSate: S & { foo: string }) => next(reducerWithFoo(reducer), initialSate);
}
```

Here we try to create an enhancer that add a `foo` field to the sate, but typescript errors saying (in short) that `S` is not compatible with `S & { foo: string }`.
This is due to the fact that the `next` function passed as arguments is not generic and so cannot capture the type from arguments.

Perhaps the solution would be to add a generic version of  the  `StoreEnhancerStoreCreator` type, and to change the signature of StoreEnhancer to use that type for the `next` argument.

definitions : 

``` typescript
export type StoreEnhancer<S> = (next: GenericStoreEnhancerCreator) => StoreEnhancerStoreCreator<S>;
export type GenericStoreEnhancer = <S>(next: GenericStoreEnhancerCreator) => StoreEnhancerStoreCreator<S>;
export type StoreEnhancerStoreCreator<S> = (reducer: Reducer<S>, initialState: S) => Store<S>;
export type GenericStoreEnhancerCreator = <S>(reducer: Reducer<S>, initialState: S) => Store<S>;
```

example: 

``` typescript
declare function reducerWithFoo<S>(reducer: Reducer<S>): Reducer<S & { foo: string }>;

function test<S>(next: GenericStoreEnhancerCreator): StoreEnhancerStoreCreator<S & { foo: string }> {
  return (reducer: Reducer<S>, initialSate: S & { foo: string }) => next(reducerWithFoo(reducer), initialSate);
} // no error
```

gaearon:
[...]

PR #2563, opened 2017-08-16
Revamp TypeScript typing with more type safety

pelotom:
This fixes many issues with the current TypeScript typings, while remaining mostly* backwards compatible:

- Add a type parameter to `Action` tracking the type of the `type` tag.
- Stop using the irrelevant state type param on `Dispatch` (but keep it around for backwards compatibility).
- Add a type parameter to `Dispatch` which is the type of things which may be dispatched (previously `Dispatch` had a polymorphic apply which allowed anything to be dispatched).
- Add type parameters `A` and `N` to `Store` which respectively track the actions it may dispatch and all other non-action things which may be dispatched. Then its `dispatch` method can take `A | N`.
- Add a type parameter `A` to `Reducer` which tracks the actions it understands, instead of just using `AnyAction`.
- Use [type parameter defaulting](https://blog.mariusschulz.com/2017/06/02/typescript-2-3-generic-parameter-defaults) everywhere to make this as backwards-compatible as possible.

With these changes, type inference for a lot of redux functions improves dramatically, and thus we get more type safety with less effort. I won't pretend these are perfect; in particular I think the middleware/enhancer functions could use some work to make the `N` variables accumulate as appropriate, but it's hopefully a step in the right direction.

As part of this, upgraded the TypeScript dependency to `^2.4.2`.

\* As you can see from some of the modified testcases, in a few rare cases it's necessary to modify type parameter annotations, e.g. when calling `combineReducers`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -46,14 +46,14 @@ export interface Action<T = any> {
  * @template S The type of state consumed and produced by this reducer.
  * @template A The type of actions the reducer can potentially respond to.
  */
-export type Reducer<S = {}, A extends Action = Action> = (state: S | undefined, action: A) => S;
+export type Reducer<S = any, A extends Action = Action> = (state: S | undefined, action: A) => S;
 
 /**
  * Object whose values correspond to different reducer functions.
  *
  * @template A The type of actions the reducers can potentially respond to.
  */
-export type ReducersMapObject<S = {}, A extends Action = Action> = {
+export type ReducersMapObject<S = any, A extends Action = Action> = {
   [K in keyof S]: Reducer<S[K], A>;
 }
 
@@ -121,7 +121,7 @@ export interface Unsubscribe {
  * @template A the type of actions which may be dispatched by this store.
  * @template N The type of non-actions which may be dispatched by this store.
  */
-export interface Store<S = {}, A extends Action = Action, N = never> {
+export interface Store<S = any, A extends Action = Action, N = never> {
   /**
    * Dispatches an action. It is the only way to trigger a state change.
    *
--- a/index.d.ts
+++ b/index.d.ts
@@ -75,7 +75,7 @@ export type ReducersMapObject<S = {}, A extends Action = Action> = {
  * @returns A reducer function that invokes every reducer inside the passed
  *   object, and builds a state object with the same shape.
  */
-export function combineReducers<S, A extends Action>(reducers: ReducersMapObject<S, A>): Reducer<S, A>;
+export function combineReducers<S, A extends Action = Action>(reducers: ReducersMapObject<S, A>): Reducer<S, A>;
 
 
 /* store */
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,4 @@
+
 /**
  * An *action* is a plain object that represents an intention to change the
  * state. Actions are the only way to get data into the store. Any data,
@@ -12,11 +13,12 @@
  * Other than `type`, the structure of an action object is really up to you.
  * If you're interested, check out Flux Standard Action for recommendations on
  * how actions should be constructed.
+ *
+ * @template T the type of the action's `type` tag.
  */
-export interface Action {
-  type: any;
+export interface Action<T = any> {
+  type: T;
 }
-
 
 /* reducers */
 
@@ -41,15 +43,18 @@ export interface Action {
  *
  * *Do not put API calls into reducers.*
  *
- * @template S State object type.
+ * @template S The type of state consumed and produced by this reducer.
+ * @template A The type of actions the reducer can potentially respond to.
  */
-export type Reducer<S> = <A extends Action>(state: S | undefined, action: A) => S;
+export type Reducer<S = {}, A extends Action = Action> = (state: S | undefined, action: A) => S;
 
 /**
  * Object whose values correspond to different reducer functions.
+ *
+ * @template A The type of actions the reducers can potentially respond to.
  */
-export type ReducersMapObject<S> = {
-  [K in keyof S]: Reducer<S[K]>;
+export type ReducersMapObject<S = {}, A extends Action = Action> = {
+  [K in keyof S]: Reducer<S[K], A>;
 }
 
 /**
@@ -70,7 +75,7 @@ export type ReducersMapObject<S> = {
  * @returns A reducer function that invokes every reducer inside the passed
  *   object, and builds a state object with the same shape.
  */
-export function combineReducers<S>(reducers: ReducersMapObject<S>): Reducer<S>;
+export function combineReducers<S, A extends Action>(reducers: ReducersMapObject<S, A>): Reducer<S, A>;
 
 
 /* store */
@@ -92,9 +97,12 @@ export function combineReducers<S>(reducers: ReducersMapObject<S>): Reducer<S>;
  * function to handle async actions in addition to actions. Middleware may
  * transform, delay, ignore, or otherwise interpret actions or async actions
  * before passing them to the next middleware.
+ *
+ * @template S unused, here only for backwards compatibility.
+ * @template D the type of things (actions or otherwise) which may be dispatched.
  */
-export interface Dispatch<S> {
-    <A extends Action>(action: A): A;
+export interface Dispatch<S = any, D = Action> {
+    <A extends D>(action: A): A;
 }
 
 /**
@@ -109,9 +117,11 @@ export interface Unsubscribe {
  * There should only be a single store in a Redux app, as the composition
  * happens on the reducer level.
  *
- * @template S State object type.
+ * @template S The type of state held by this store.
+ * @template A the type of actions which may be dispatched by this store.
+ * @template N The type of non-actions which may be dispatched by this store.
  */
-export interface Store<S> {
+export interface Store<S = {}, A extends Action = Action, N = never> {
   /**
    * Dispatches an action. It is the only way to trigger a state change.
    *
@@ -138,7 +148,7 @@ export interface Store<S> {
    * Note that, if you use a custom middleware, it may wrap `dispatch()` to
    * return something else (for example, a Promise you can await).
    */
-  dispatch: Dispatch<S>;
+  dispatch: Dispatch<any, A | N>;
 
   /**
    * Reads the state tree managed by the store.
@@ -182,7 +192,7 @@ export interface Store<S> {
    *
    * @param nextReducer The reducer for the store to use instead.
    */
-  replaceReducer(nextReducer: Reducer<S>): void;
+  replaceReducer(nextReducer: Reducer<S, A>): void;
 }
 
 /**
@@ -191,11 +201,13 @@ export interface Store<S> {
  * `createStore(reducer, preloadedState)` exported from the Redux package, from
  * store creators that are returned from the store enhancers.
  *
- * @template S State object type.
+ * @template S The type of state to be held by the store.
+ * @template A The type of actions which may be dispatched.
+ * @template D The type of all things which may be dispatched.
  */
 export interface StoreCreator {
-  <S>(reducer: Reducer<S>, enhancer?: StoreEnhancer<S>): Store<S>;
-  <S>(reducer: Reducer<S>, preloadedState: S, enhancer?: StoreEnhancer<S>): Store<S>;
+  <S, A extends Action, N>(reducer: Reducer<S, A>, enhancer?: StoreEnhancer<N>): Store<S, A, N>;
+  <S, A extends Action, N>(reducer: Reducer<S, A>, preloadedState: S, enhancer?: StoreEnhancer<N>): Store<S, A, N>;
 }
 
 /**
@@ -215,10 +227,11 @@ export interface StoreCreator {
  * provided by the developer tools. It is what makes time travel possible
  * without the app being aware it is happening. Amusingly, the Redux
  * middleware implementation is itself a store enhancer.
+ *
  */
-export type StoreEnhancer<S> = (next: StoreEnhancerStoreCreator<S>) => StoreEnhancerStoreCreator<S>;
-export type GenericStoreEnhancer = <S>(next: StoreEnhancerStoreCreator<S>) => StoreEnhancerStoreCreator<S>;
-export type StoreEnhancerStoreCreator<S> = (reducer: Reducer<S>, preloadedState?: S) => Store<S>;
+export type StoreEnhancer<N = never> = (next: StoreEnhancerStoreCreator<N>) => StoreEnhancerStoreCreator<N>;
+export type GenericStoreEnhancer<N = never> = StoreEnhancer<N>;
+export type StoreEnhancerStoreCreator<N = never> = <S = any, A extends Action = Action>(reducer: Reducer<S, A>, preloadedState?: S) => Store<S, A, N>;
 
 /**
  * Creates a Redux store that holds the state tree.
@@ -253,8 +266,8 @@ export const createStore: StoreCreator;
 
 /* middleware */
 
-export interface MiddlewareAPI<S> {
-  dispatch: Dispatch<S>;
+export interface MiddlewareAPI<S = any, D = Action> {
+  dispatch: Dispatch<any, D>;
   getState(): S;
 }
 
@@ -268,7 +281,7 @@ export interface MiddlewareAPI<S> {
  * asynchronous API call into a series of synchronous actions.
  */
 export interface Middleware {
-  <S>(api: MiddlewareAPI<S>): (next: Dispatch<S>) => Dispatch<S>;
+  <S = any, D = Action>(api: MiddlewareAPI<S, D>): (next: Dispatch<any, D>) => Dispatch<any, D>;
 }
 
 /**
@@ -317,8 +330,8 @@ export interface ActionCreator<A> {
 /**
  * Object whose values are action creator functions.
  */
-export interface ActionCreatorsMapObject {
-  [key: string]: ActionCreator<any>;
+export interface ActionCreatorsMapObject<A = any> {
+  [key: string]: ActionCreator<A>;
 }
 
 /**
@@ -340,19 +353,19 @@ export interface ActionCreatorsMapObject {
  *   creator wrapped into the `dispatch` call. If you passed a function as
  *   `actionCreator`, the return value will also be a single function.
  */
-export function bindActionCreators<A extends ActionCreator<any>>(actionCreator: A, dispatch: Dispatch<any>): A;
+export function bindActionCreators<A, C extends ActionCreator<A>>(actionCreator: C, dispatch: Dispatch<any, A>): C;
 
 export function bindActionCreators<
   A extends ActionCreator<any>,
   B extends ActionCreator<any>
-  >(actionCreator: A, dispatch: Dispatch<any>): B;
+  >(actionCreator: A, dispatch: Dispatch<any, any>): B;
 
-export function bindActionCreators<M extends ActionCreatorsMapObject>(actionCreators: M, dispatch: Dispatch<any>): M;
+export function bindActionCreators<A, M extends ActionCreatorsMapObject<A>>(actionCreators: M, dispatch: Dispatch<any, A>): M;
 
 export function bindActionCreators<
-  M extends ActionCreatorsMapObject,
-  N extends ActionCreatorsMapObject
-  >(actionCreators: M, dispatch: Dispatch<any>): N;
+  M extends ActionCreatorsMapObject<any>,
+  N extends ActionCreatorsMapObject<any>
+  >(actionCreators: M, dispatch: Dispatch<any, any>): N;
 
 
 /* compose */
--- a/index.d.ts
+++ b/index.d.ts
@@ -101,7 +101,7 @@ export function combineReducers<S, A extends Action = Action>(reducers: Reducers
  * @template S unused, here only for backwards compatibility.
  * @template D the type of things (actions or otherwise) which may be dispatched.
  */
-export interface Dispatch<S = any, D = Action> {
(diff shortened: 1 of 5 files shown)


test/typescript/compose.ts
```diff
@@ -36,4 +36,4 @@ const t11: number = compose(stringToNumber, numberToString, stringToNumber,
 
 
 const funcs = [stringToNumber, numberToString, stringToNumber];
-const t12 = compose(...funcs)('bar', 42, true);
```
pelotom: This appears to have just been a bug in the old test case, which the newer version of TypeScript caught.

index.d.ts
```diff
@@ -12,23 +13,13 @@
...
-/**
- * An Action type which accepts any other properties.
- * This is mainly for the use of the `Reducer` type.
- * This is not part of `Action` itself to prevent users who are extending `Action.
- * @private
- */
-export interface AnyAction extends Action {
-  // Allows any extra properties to be defined in an action.
-  [extraProps: string]: any;
+export interface Action<T = any> {
```
aikoven: Am I right that the intended use case for the type argument is using string literal types?

index.d.ts
```diff
@@ -52,15 +43,18 @@ export interface AnyAction extends Action {
  *
  * *Do not put API calls into reducers.*
  *
- * @template S State object type.
+ * @template S The type of state consumed and produced by this reducer.
+ * @template A The type of actions the reducer can potentially respond to.
  */
-export type Reducer<S> = (state: S, action: AnyAction) => S;
+export type Reducer<S = {}, A extends Action = Action> = (state: S, action: A) => S;
```
aikoven: Is there a particular reason for `S` to default to `{}`?

Also, I don't think that it is correct to constraint the type of `action`. A reducer must be able to accept every possible action, which is *any object with `type` property*. It should just ignore all actions that don't affect its state.

index.d.ts
```diff
@@ -52,15 +43,18 @@ export interface AnyAction extends Action {
...
 
 /**
  * Object whose values correspond to different reducer functions.
+ *
+ * @template A The type of actions the reducers can potentially respond to.
  */
-export interface ReducersMapObject {
-  [key: string]: Reducer<any>;
+export type ReducersMapObject<S = {}, A extends Action = Action> = {
+  [K in keyof S]: Reducer<S[K], A>;
```
aikoven: Already done on the `next` branch: https://github.com/reactjs/redux/blob/next/index.d.ts#L52

This PR should probably target the `next`.

index.d.ts
pelotom: Yes

index.d.ts
pelotom: > Is there a particular reason for `S` to default to `{}`?

Not really, I was just on a defaulting rampage.

> Also, I don't think that it is correct to constraint the type of action. A reducer must be able to accept every possible action, which is any object with type property. It should just ignore all actions that don't affect its state.

It's perfectly valid to constrain the type of actions that a reducer handles. It must be able to accept any action in your app, but not any arbitrary object with a `type` field. This is critical to making the library nice to use in TypeScript because it allows the type system to automatically refine the types for you as you learn information by inspecting the `type` field. For example:

```ts
// The universe of all actions used in your app
export type AnyAction =
  | SetGreeting
  | IncrementCounter
  | SetCompletionStatus

export interface SetGreeting {
  type: 'SET_GREETING'
  greeting: string
}

export interface IncrementCounter {
  type: 'INCREMENT_COUNTER'
  counterIndex: number
}

export interface SetCompletionStatus {
  type: 'SET_COMPLETION_STATUS'
  completed: boolean
}

const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
  switch (action.type) {
    case: 'SET_GREETING':
      return { ...state, greeting: action.greeting }
    case: 'INCREMENT_COUNTER':
      return { ...state, counters: state.counters[action.counterIndex] + 1 }
    case: 'SET_COMPLETION_STATUS':
      return { ...state, completed: action.completed }
  }
}
```

What's super cool about this is
1. It's completely type safe. Inside each `case`, the type of `action` has been refined such that it knows what fields should be available on it and what types they have. If I tried to access `action.counterIndex` in a branch where I hadn't checked that `action.type === 'INCREMENT_COUNTER'` it would be a type error.
1. We get exhaustiveness checking, so that no `default` case is required because the type system has proven that you've handled all possible action types.
1. You can perform an app-wide refactoring on the string literals being used for the action `type` fields and they will also change these switch cases.

index.d.ts
pelotom: Ok, I'll do that.

index.d.ts
aikoven: There's always an `init` action which is internal to Redux whose `type` can potentially be made random.

pelotom: Yes, but your app can pretend that doesn't exist for all intents and purposes.

aikoven: To me making types that don't agree with the runtime is an abuse of type system.

However, I can see that this approach gives really nice DX, so I tend to agree with you. It should be up to developer to decide the level of strictness they want.

index.d.ts
```diff
@@ -12,23 +13,13 @@
...
-  type: any;
-}
-
-/**
- * An Action type which accepts any other properties.
- * This is mainly for the use of the `Reducer` type.
- * This is not part of `Action` itself to prevent users who are extending `Action.
- * @private
- */
-export interface AnyAction extends Action {
```
aikoven: Are you sure `AnyAction` can be safely removed? Its absence broke stuff on TS 2.4

See https://github.com/reactjs/redux/pull/2467

index.d.ts
pelotom: You could certainly opt to include the init action in your top-level union of action types if you want to capture absolutely all actions that might be passing through your reducer. The point is to constrain the types to some known, closed universe so that you can have type safety and semantic awareness for refactoring.

index.d.ts
pelotom: Yes, since `Reducer`s are parameterized by action type now it should no longer be necessary.

timwangdev:
Looking forward to see any updates on this PR

pelotom:
Rebased this onto `next` as requested, afaik it should be good to merge.

index.d.ts
```diff
@@ -41,15 +43,18 @@ export interface Action {
  *
  * *Do not put API calls into reducers.*
  *
- * @template S State object type.
+ * @template S The type of state consumed and produced by this reducer.
+ * @template A The type of actions the reducer can potentially respond to.
  */
-export type Reducer<S> = <A extends Action>(state: S | undefined, action: A) => S;
+export type Reducer<S = {}, A extends Action = Action> = (state: S | undefined, action: A) => S;
```
aikoven: Still, do we really want to have `S` default to `{}`? Why?

index.d.ts
```diff
@@ -41,15 +43,18 @@ export interface Action {
...
 
 /**
  * Object whose values correspond to different reducer functions.
+ *
+ * @template A The type of actions the reducers can potentially respond to.
  */
-export type ReducersMapObject<S> = {
-  [K in keyof S]: Reducer<S[K]>;
+export type ReducersMapObject<S = {}, A extends Action = Action> = {
```
aikoven: Default for `S` here is unclear for me too.

index.d.ts
```diff
@@ -70,7 +75,7 @@ export type ReducersMapObject<S> = {
  * @returns A reducer function that invokes every reducer inside the passed
  *   object, and builds a state object with the same shape.
  */
-export function combineReducers<S>(reducers: ReducersMapObject<S>): Reducer<S>;
+export function combineReducers<S, A extends Action>(reducers: ReducersMapObject<S, A>): Reducer<S, A>;
```
aikoven: Let's add a default type for `A`

index.d.ts
```diff
@@ -92,9 +97,12 @@ export function combineReducers<S>(reducers: ReducersMapObject<S>): Reducer<S>;
  * function to handle async actions in addition to actions. Middleware may
  * transform, delay, ignore, or otherwise interpret actions or async actions
  * before passing them to the next middleware.
+ *
+ * @template S unused, here only for backwards compatibility.
```
aikoven: I think we should put more thought into this. These changes are already breaking, so why should we bother about compatibility here?

The `S` type parameter brought much more misconception and inconvenience than benefit. It was initially added to allow us to strongly type-check dispatching thunks: as you can see [here](https://github.com/gaearon/redux-thunk/blob/v2.2.0/index.d.ts), we augment the `Dispatch` interface to add a new signature for thunks that uses `S`.

But in real-world it appeared that we rarely use `dispatch` as a method of the Store, where it has a fixed state type, but rather as an injected function (like in `react-redux`), where we have to declare its type manually:
```ts
const mapDispatchToProps = (dispatch: Dispatch<MyState>) => ...
```
For users who don't use `redux-thunk` this led to annoying need to write `Dispatch<any>` instead of just `Dispatch`.

We could just remove `S` altogether to simplify things; `redux-thunk` users would still be able to get partial type safety by declaring their state type in `ThunkAction` type parameters. The one major drawback of this is that it would make it much harder to migrate to the new typings.

Otherwise we should make this docstring point to the use case of `S`.

index.d.ts
```diff
@@ -109,9 +117,11 @@ export interface Unsubscribe {
  * There should only be a single store in a Redux app, as the composition
  * happens on the reducer level.
  *
- * @template S State object type.
+ * @template S The type of state held by this store.
+ * @template A the type of actions which may be dispatched by this store.
+ * @template N The type of non-actions which may be dispatched by this store.
  */
-export interface Store<S> {
+export interface Store<S = {}, A extends Action = Action, N = never> {
```
aikoven: Again, the default for `S` is unclear.

aikoven: Dispatching non-actions is handled by augmenting the `Dispatch` interface. There's a difference between that and adding another type to the union: for example, when dispatching a thunk, `dispatch` returns the return value of the thunk, not the thunk itself.

index.d.ts
pelotom: I was trying to keep this PR as backwards-compatible as possible, but if that's not important I'm more than happy to remove it.

index.d.ts
pelotom: Having defaults means you can just reference `ReducersMapObject` if you don't care about the types involved. That tends to happen if you have a generic type parameter `T extends ReducersMapObject` for example.

index.d.ts
pelotom: 👍

index.d.ts
pelotom: As I said I'm happy to remove the state parameter and I definitely think this should be done at some point... If you think this PR is the right place for it I'll remove it.

index.d.ts
pelotom: See previous comment... if that doesn't sway you, I'll remove the default.

pelotom: So are you advocating removal of the `N` parameter?

KingHenne: Please don't. Augmenting Dispatch for things like redux-thunk is broken, see https://github.com/gaearon/redux-thunk/issues/82. Still no one gave a solution how typing middlewares can work when redux-thunk has augmented Dispatch.

aikoven: I agree that module augmentation is more of a workaround rather than an ideal solution. 

Still, it gives us something that we can't achieve with just extending the type of `dispatch` argument: it allows us to set up a mapping from input type to output type of `dispatch` calls, which is different from one middleware to another.
Having just `Dispatch<Action | ThunkAction>` is incorrect, because it will work as if `dispatch(thunk)` returned the thunk itself.

Also, from what I see in the issue, the error comes from the definition of `Middleware`, and not from the module augmentation.

Considering the ideal solution: adding `dispatch` signatures happens during store enhancement by `applyMiddleware(...)`. We could try to come up with smarter typings for `StoreEnhancer` and `Middleware` to get signature augmentation without module augmentation. Back then we didn't succeed with it, but now TS is much more powerful.

https://github.com/reactjs/redux/pull/1526
https://github.com/reactjs/redux/issues/1648

index.d.ts
aikoven: Sorry, I don't get how it aids backward compatibility. Previous `Reducer` type did have `S` as a required parameter.

index.d.ts
aikoven: This makes some sense. But wouldn't `ReducersMapObject<{}>` be the same as just `{}`? Therefore `T extends {}` only means that `T` is not `null` or `undefined`.

index.d.ts
pelotom: Hm, I appear to have been confused when I made this comment. I was thinking of the state parameter to the `Dispatch` type, which is completely unnecessary but I was keeping around for backwards compatibility.

The argument for defaulting `S` here is the same as for `ReducersMapObject` and `Store` below... so that we can declare type constraints like `R extends Reducer` when we don't care about the type arguments.

aikoven: Thanks, this makes sense. The only thing that bothers me: if the state type of reducer includes `null`, that reducer won't be assignable to `Reducer` (without parameters).

A very simplified example:
```ts
type Reducer<S = {}> = (state: S | undefined) => S;

const r: Reducer = (state: null | undefined = null): null => {
  return state;
}
```
Gives
```
Type '(state?: null | undefined) => null' is not assignable to type 'Reducer<{}>'.
    Type '{} | undefined' is not assignable to type 'null | undefined'.
      Type '{}' is not assignable to type 'null | undefined'.
        Type '{}' is not assignable to type 'null'.
```

Works fine after setting `S = {} | null`.

index.d.ts
pelotom: You raise a good point and I think it means I chose the wrong default... instead it should be `{ [k: string]: any }`. That way `ReducersMapObject` should be equivalent to `{ [k: string]: Reducer<any> }`?

index.d.ts
pelotom: Yeah, I'm coming to think the default type should probably be either  be `S = any` or `S = {} | null | undefined`... a "top" type, since `S` is in a contravariant position.

index.d.ts
aikoven: Yeah, I think that should work.

index.d.ts
aikoven: I think you're right, except that `undefined` for the state is prohibited by Redux. Yet I'm not sure if we gain any type safety by excluding it.

As far as I know, `{} | null | undefined` is the same as `any`.

index.d.ts
pelotom: Maybe the `Dispatch` type should still take 2 type parameters, but one is the input ("action") type, and one is the output type?

index.d.ts
KingHenne: Remember though, `undefined` is prohibited to return from a reducer but not as argument, which is passed with the initial call.

pelotom: It's not quite the same, because for example if

```ts
declare x: {} | null | undefined
x.foo // this is a type error
declare y: any
y.foo // this is fine
```

`any` isn't really a type, it's more like a directive not to perform type checking. I'd say `{} | null` is the most "correct" thing to use here, but a lot of people tend to use `any` in situations like this because it's easy on the eyes. I could go either way.

pelotom: > Remember though, `undefined` is prohibited to return from a reducer but not as argument, which is passed with the initial call.

The definition covers this since the state argument is of type `S | undefined`.

index.d.ts
KingHenne: > Maybe the Dispatch type should still take 2 type parameters, but one is the input ("action") type, and one is the output type?

Sounds good to me.

KingHenne: After thinking a bit more about it, I don't think that works. Using a union type for the Action and a union type for the return value does not really help here (e.g. with redux-thunk). What you need instead is two different signatures. I'm thinking the Action shouldn't be the type param (for Store and Middleware) but instead the whole Dispatch should be.

index.d.ts
aikoven: > It's not quite the same

You are right, I used the wrong word. They are equivalent in terms of assignability.

aikoven:
Guys, let's continue our discussion here instead of code comments because they get hidden after code updates.

> I'm thinking the Action shouldn't be the type param (for Store and Middleware) but instead the whole Dispatch should be.

I think that's the way to go. The type for Middleware would be something like
```ts
interface Middleware<NewDispatch> {
  (api: MiddlewareAPI): (next: Dispatch) => Dispatch | NewDispatch;
}
```
But there's one problem: `NewDispatch` can potentially be generic, and I don't know of a way to have a generic type as a type parameter. E.g. in `redux-thunk` our `NewDispatch` would have the state type as a type parameter, and we need to somehow make it the same as in `MiddlewareAPI`.

KingHenne:
This is tricky. If you think this further, than the `MiddlewareAPI` interface would probably defined like this:

```ts
export interface MiddlewareAPI<S = any, D = Dispatch> {
  dispatch: D;
  getState(): S;
}
```

But then in the `Middleware` interface you couldn't decide whether to set `Dispatch` or `NewDispatch` on `MiddlewareAPI` because this depends on whether the middleware is the left-most in `applyMiddleware`, right?

KingHenne:
@aikoven To answer your question, would this be sufficient?

```ts
export interface MiddlewareAPI<S = any, D = Dispatch> {
  dispatch: D;
  getState(): S;
}

export interface Middleware<NewDispatch = Dispatch, D = Dispatch> {
  <S = any>(api: MiddlewareAPI<S, NewDispatch | D>): (
    next: Dispatch
  ) => NewDispatch | D;
}
```

And in `redux-thunk`:

```ts
import {Middleware, Dispatch as ReduxDispatch} from 'redux';

export type ThunkAction<R, S, E> = (
  dispatch: Dispatch<S>,
  getState: () => S,
  extraArgument: E
) => R;

export interface Dispatch<S> extends ReduxDispatch {
  <R, E>(asyncAction: ThunkAction<R, S, E>): R;
}

declare const thunk: Middleware<Dispatch<any>> & {
  withExtraArgument(extraArgument: any): Middleware;
};

export default thunk;
```

The `any` doesn't hurt here, does it?

KingHenne:
Or even better:

```ts
import {Middleware, Dispatch as ReduxDispatch} from 'redux';

export type ThunkAction<R, S, E> = (
  dispatch: Dispatch,
  getState: () => S,
  extraArgument: E
) => R;

export interface Dispatch extends ReduxDispatch {
  <R, S, E>(asyncAction: ThunkAction<R, S, E>): R;
}

declare const thunk: Middleware<Dispatch> & {
  withExtraArgument(extraArgument: any): Middleware<Dispatch>;
};

export default thunk;
```

KingHenne:
That does not compile though in my current project (using the `Dispatch` interface from `redux-thunk` in thunks). The problem is the action type. We want to define the actions with either a string literal union type or a string enum (which I prefer since this became available) so TS can narrow down the payload type in a switch for example. But if you also handle other actions as for example `LocationChangeAction` and `RouterAction` from `react-router-redux` you can not put them in your `Action` union type because this breaks the type narrowing based on the `type`. This can only be solved by defining `Dispatch` (in your project) like this:

```ts
import {
  LocationChangeAction,
  RouterAction,
  RouterState,
} from 'react-router-redux';
import {ThunkAction as ReduxThunkAction} from 'redux-thunk';

import {Action} from '../src/actions';

export interface AppState {
  routing: RouterState;
  ...
}

export type ThunkAction<Result> = ReduxThunkAction<
  Action,
  Result,
  AppState,
  undefined
>;

export interface Dispatch {
  <Result>(action: ThunkAction<Result>): Result;
  (action: LocationChangeAction | RouterAction):
    | LocationChangeAction
    | RouterAction;
  (action: Action): Action;
}
```

This `Dispatch` interface can then be used in thunks.

pelotom:
It seems like fixing the type of `Dispatch` in an ideal way still has some issues to be worked through. Is there any reason not to merge this PR in the meantime? Let's not let perfect be the enemy of good.

aikoven:
@pelotom The only thing that bothers me is the `N` parameter. I don't see it as a viable solution and it will be very hard to transition from it if it gets to a release.

I am pretty sure there exists a good solution for `Dispatch` extension, although it may be subject for another PR.

Otherwise, the changes look very well done to me.

timdorr:
OK, I'm merging it in!

timdorr:
Because we didn't have `next` on the CI config, this didn't get tests run. Turns out, there's a breakage:

> Via https://travis-ci.org/reactjs/redux/builds/291276494
> FAIL test/typescript.spec.js
> ● TypeScript definitions › should compile against index.d.ts
> AssertionError: Semantic: /home/travis/build/reactjs/redux/index.d.ts (103,18): All declarations of 'Dispatch' must have identical type parameters.

@pelotom @aikoven @KingHenne or anyone else want to provide a fix? I'd like to push forward on Redux 4.0 and this is a blocker.

pelotom:
@timdorr hmm, what am I doing wrong?

```
> yarn test
yarn run v1.2.1
$ cross-env BABEL_ENV=commonjs jest
 PASS  test/typescript.spec.js
 PASS  test/createStore.spec.js
 PASS  test/bindActionCreators.spec.js
 PASS  test/compose.spec.js
 PASS  test/combineReducers.spec.js
 PASS  test/applyMiddleware.spec.js
 PASS  test/utils/warning.spec.js
```

pelotom:
Never mind, I reproduced it now.

pelotom:
@timdorr opened a PR: #2674

domachine:
So glad that this got merged. Is there anything I can help to get this released to npm?

timdorr:
It's in 4.0.0-beta.1. `npm install redux@next`

domachine:
Ah great! Thank you so much.

PR #2570, opened 2017-08-19
Improve note about WordPress framework

m4sk1n:
The current one looks like it's implementation of Redux library for WordPress (which has no sense…)


gaearon:
This is better. Thanks.

PR #2571, opened 2017-08-20
Annotate action.js headings

stevekinney:
As it stands, there are multiple headings entitled `action.js`. On the official docs page, it offers a anchor to each heading, but this doesn't work because it will just match the `id` of the first `#### action.js` heading. This change adds some context, which will—in turn—make each heading unique.


timdorr:
Thanks!

PR #2573, opened 2017-08-20
Spelling error

PashaSchool:
just simple error where this word 'naїve' should look like this 'native'


timdorr:
This is correct as-is: https://en.wiktionary.org/wiki/naive#English

PR #2575, opened 2017-08-22
TodoMVC example: add items in the correct order

### LINKED ISSUE #2574: TodoMVC example: items are in the wrong order
In `examples/todomvc`, new TODO items are added to the top instead of at the bottom.

New items should be added to the bottom instead, to be consistent with every other TodoMVC example (e.g. [React](http://todomvc.com/examples/react/)) and to allow for easier comparison.

mathiasbynens:
New TODO items must be added to the bottom instead of at the top.

Closes #2574.


timdorr:
Thanks!

PR #2579, opened 2017-08-23
Add redux-data-structures to Ecosystem

adrienjt:
I wonder if you all would consider [redux-data-structures](https://redux-data-structures.js.org/) as a "staff pick". I got good feedback on it (and 85 stars on GitHub so far).


markerikson:
Yeah, I think it's a useful and interesting enough item to add.

PR #2582, opened 2017-08-24
Changed unsubscribe from let to const

omerts:
Using const is more correct since you don't want unsubscribe to be mutated


timdorr:
Sure. Thanks!

PR #2585, opened 2017-08-25
Fix argument inconsistency in Store.md

Elfo404:
The argument in the title and in the arguments list should be consistent between each other


timdorr:
nextThanks!

PR #2586, opened 2017-08-27
Update FAQ: When should I learn Redux?

maxhallinan:
This adds a "When should I learn Redux?" section to the General FAQ. The answer includes links suggested in #1785 . cc: @markerikson


timdorr:
I feel like this should be combined with the "When should I use Redux?" question. They are quite related in my mind.

maxhallinan:
I made the same comment on [1785](https://github.com/reactjs/redux/issues/1785#issuecomment-318618213). While I'm not confidant that I covered everything Mark intended for "When should I learn Redux?", it seems like there are two related but distinct ideas:

- The best time to learn Redux is when you have a problem that Redux solves. ("When should I learn Redux?"). 
- The problems that Redux solves are... ("When should I use Redux?")

Keeping the question about learning separate creates a space for the broader point about avoiding "JavaScript fatigue". But that question does give rise and flow directly into the question of good use cases. So maybe placing "When should I learn Redux?" before "When should I use Redux?" helps to keep the answers simple and the flow coherent.

markerikson:
Yeah, I agree that they're two separate questions.  It's entirely possible to _know_ Redux, but not be sure when you actually ought to use it.

docs/faq/General.md
```diff
@@ -71,3 +72,22 @@ Redux is originally written in ES6 and transpiled for production into ES5 with W
...
+**Articles**
+
+- [Deciding What Not To Learn](http://gedd.ski/post/what-not-to-learn/)
+- [How to learn web frameworks](https://ux.shopify.com/how-to-learn-web-frameworks-9d447cb71e68)
+- [Redux vs MobX vs Flux vs... Do you even need that?](https://goshakkk.name/redux-vs-mobx-vs-flux-etoomanychoices/)
+
+**Discussions**
+
+- [Ask HN: Overwhelmed with learning front-end, how do I proceed?](https://news.ycombinator.com/item?id=12882816)
+- [Twitter: If you want to teach someone to use an abstraction...](https://twitter.com/acemarke/status/901329101088215044)
```
markerikson: Hah! I love that this got added just a day or two after I tweeted it :)

docs/faq/General.md
```diff
@@ -71,3 +72,22 @@ Redux is originally written in ES6 and transpiled for production into ES5 with W
 > The new Counter Vanilla example is aimed to dispel the myth that Redux requires Webpack, React, hot reloading, sagas, action creators, constants, Babel, npm, CSS modules, decorators, fluent Latin, an Egghead subscription, a PhD, or an Exceeds Expectations O.W.L. level.
 
 >Nope, it's just HTML, some artisanal `<script>` tags, and plain old DOM manipulation. Enjoy!
+
+
+<a id="general-when-to-learn"></a>
+### When should I learn Redux?
+
+What to learn can be an overwhelming question for a JavaScript developer. It helps to narrow the range of options by focusing on problems you find in your work. Redux is a pattern for managing application state. If you do not have problems with state management, you might not need to learn Redux. Especially if you are new to JavaScript application development, we encourage you to start with a library (like React) for creating data-driven user interfaces. Such a library might be all you need to build your application. If your application becomes so complex that you are confused about where state is stored or how state changes, then it might be a good time to learn Redux.
```
markerikson: Pretty good start.  Got a couple ideas for rephrasing things.

First, let me quote my "standard advice for learning React":

> On that note, definitely don't over-complicate the learning process by trying to learn many different things at once. Some people will say you should use a "boilerplate" to learn React, and they're wrong - boilerplate projects almost always come with too many pieces configured, and are confusing for beginners.

> Instead, the best advice is to focus on learning React itself first. Once you have a good understanding of how React works, you will better appreciate why a state management library like Redux can be useful, and you can learn about other tools later.

I'd like to see if we can use some of that phrasing.  It's not so much that React is a "data-driven UI library", but that React already has a built-in pattern for managing state.  

I definitely like that last sentence on "becoming so complex you're confused".

Another aspect I'd like to get into this is that we advise learning React first just so there's fewer concepts to understand at once.  It's certainly possible for people to tackle them both at the same time, but for most people it's easier to focus on React first, then Redux after that.

docs/faq/General.md
```diff
@@ -71,3 +72,22 @@ Redux is originally written in ES6 and transpiled for production into ES5 with W
...
+- [Deciding What Not To Learn](http://gedd.ski/post/what-not-to-learn/)
+- [How to learn web frameworks](https://ux.shopify.com/how-to-learn-web-frameworks-9d447cb71e68)
+- [Redux vs MobX vs Flux vs... Do you even need that?](https://goshakkk.name/redux-vs-mobx-vs-flux-etoomanychoices/)
+
+**Discussions**
+
+- [Ask HN: Overwhelmed with learning front-end, how do I proceed?](https://news.ycombinator.com/item?id=12882816)
+- [Twitter: If you want to teach someone to use an abstraction...](https://twitter.com/acemarke/status/901329101088215044)
+- [Twitter: it was never intended to be learned before...](https://twitter.com/dan_abramov/status/739961787295117312)
+- [Twitter: Learning Redux before React?](https://twitter.com/dan_abramov/status/739962098030137344)
```
markerikson: A couple more very relevant quotes from the last few days to add.  I suppose these could easily go under "When should I use Redux?" too.

- https://twitter.com/raquelxmoss/status/901576285020856320
- https://twitter.com/garetmckinley/status/901500556568645634

markerikson:
I can also see a valid point to putting "When should I learn?" in front of "When should I use?".  I'm not 100% sure I want to do it, but it's worth trying it out to see how it looks.

docs/faq/General.md
maxhallinan: 😀

docs/faq/General.md
maxhallinan: I've added these links.

docs/faq/General.md
maxhallinan: I've revised this to be more explicit that a UI library is possibly sufficient when the library includes a state management system. And I've added emphasis that beginners are best off starting without Redux. I'm reluctant to comment on React more specifically because it undermines the important point that Redux is not coupled to React. Saying "focus on React first, then Redux after that" could imply a preference for React more fundamentally. So I tried to incorporate the principles of what you're saying with less emphasis on React.

maxhallinan:
@markerikson thanks for the review! I've updated this branch with a second draft that (I hope) addresses your feedback. I also moved "When should I learn..." before "When should I use..." and removed language about learning from "When should I use...". Let me know what you think.

markerikson:
I'd say this looks good enough for now.  Stamped!

PR #2596, opened 2017-08-31
removed isRequired from todos, otherwise demo was broken

tylik1:
With "isRequired" prop, assigned to todos, the demo will fail, as initially there are no todos, when the app starts


jimbolla:
It should be given an empty array, otherwise todos.map(...) would throw an error.

timdorr:
Jim's correct. The initial state of the Redux store will provide an array.

PR #2597, opened 2017-09-01
[TypeScript] Add Dispatch default any type parameter

timwangdev:
Since` <S>` in generic `Dispatch` is not used in its interface definition. 
And `Dispatch<any>` pattern is commonly used in TypeScript redux projects.


timdorr:
Dupe of #2563, which is more comprehensive.

Issue #992, opened 2015-11-02
Support for typed actions (Typescript)


nickknw:
Related to these two issues:
 - https://github.com/rackt/redux/issues/355
 - https://github.com/rackt/redux/issues/437

Hi Dan,

The company I am with is using Redux as well as Typescript, and we would like to add some additional compile-time guarantees to our actions. Currently this isn't possible because of the restriction that actions must be plain objects. You said that this is because 'record/replay and associated features won't work with an action class' and I get that those are valuable features and don't want to lose them. But I had code using Typescript classes that (when the redux devtools were enabled) could still record, cancel and replay actions just fine (using the code I posted in issue 355). 

When I remove the devtools I just get errors from that check that actions are plain objects and execution stop there. To me it seems like it is too strict, checking more than is really required. I would much rather have a check that requires that the actions are serializable. I know performance is a concern - IMO it shouldn't have to guarantee/prove they are serializable, just check that the action objects themselves say that they are, and then trust them. 
(Not sure why the check doesn't also fail when devtools are enabled - maybe you can shed light on that).

I would really like to be able to leverage the advantages of both Typescript and Redux. Hope you can help me (and others using both of these tools) arrive at a solution.

I also want to reiterate that (as much as I wish it were otherwise!) Typescript does not have discriminated unions, so solutions based on that approach will not work.

Thank you!

gaearon:
Does TypeScript let you type-check shapes of plain objects instead of defining classes?
My only issue is with using instances of classes for actions.

Wouldn't [TypeScript interfaces](http://blogs.msdn.com/b/typescript/archive/2013/01/24/interfaces-walkthrough.aspx) be enough to strongly type plain object actions?

nickknw:
Almost - the problem with those is that they only exist at compile-time. So the reducer will have no way to tell which incoming action is of which interface. ...Unless I keep the 'type' property on each action and then manually cast them after checking that property. And at that point I'm not getting any extra type-safety over just leaving them as plain objects.

subjectix:
[...]

PR #2599, opened 2017-09-05
Optimize dispatch plain object check

### LINKED ISSUE #2598: Using lodash.isPlainObject in dispatch has performance overhead
I just created a performance comparison between Redux and [Repatch](https://github.com/jaystack/repatch): https://github.com/jaystack/redux-repatch-performance-comparison.

This comparison dispatches a big amount of pure actions (simple increment), and I found that Redux is too slow as we expect.

I discovered, that [this assertion](https://github.com/reactjs/redux/blob/master/src/createStore.js#L150) at `dispatch` causes this performance decreasing.

I suggest a simpler checking, like:
```javascript
action && action.constructor === Object
```
if it's not necessary that action objects can made by `Object.create(null)`.

With this enhancing redux dispatching will be 2 magnitude faster.

timdorr:
Closes #2598

This boosts perf by not exhaustively checking every action of it's object-ness. Dispatch is approximately 56x faster. Across a million runs of simple counter reducer:

```
master: 4690.358ms
nodash: 82.821ms
```

I also applied the same thing in combineReducers. I believe the check is sufficient, but I'd love a second set of eyes on it.

We also no longer need to bundle any of lodash, so this saves a good number of bytes too! From the UMD build:
```
26715 master/redux.js
21518 nodash/redux.js
5761 master/redux.min.js
4813 nodash/redux.min.js
```


markerikson:
Looks okay at first glance, although the nuances of JS object detection are something I've never had to deal with.  Do we have some tests around this scenario?

src/combineReducers.js
```diff
@@ -26,7 +25,7 @@ function getUnexpectedStateShapeWarningMessage(inputState, reducers, action, une
     )
   }
 
-  if (!isPlainObject(inputState)) {
+  if (!inputState || inputState.constructor !== Object) {
```
taggon: This seems to have the same problem with an old-fashioned `isArray` implementation. It always fails if the action comes from another window.

markerikson:
I saw an email from a review comment come through, but I'm not seeing that comment as I'm looking at the diff.  The email said:

> This seems to have the same problem with an old-fashioned `isArray` implementation.  It always fails if the action comes from another window.

Anyone more knowledgeable have thoughts on that?

timdorr:
Another alternative to checking `constructor` is to `toString` the object and see if that is `[object Object]`.

milesj:
Constructor checks should be enough IMO.

jdalton:
The `isPlainObject` check is certainly more robust, but if it's not needed, and a simpler check is doable, more power to ya. Does anyone know the history of using the plain object check over say just a traditional object type check?

timdorr:
@jdalton It's a good safety net for someone that makes a mistake when setting up something like redux-thunk or another middleware. Happens a fair amount and is more friendly to beginners.

jdalton:
> Happens a fair amount and is more friendly to beginners.

Could it be something that's available in dev builds and removed in production? Are similar debug helpers handled that way?

timdorr:
We have [some dev-only checks](https://github.com/reactjs/redux/blob/7e745ce543baecb835173220e7bb51f17b19c875/src/combineReducers.js#L108) in combineReducers, but not much else. Unfortunately, it's easy to miss a lot of these issues without 100% test coverage, so having "friendlier" errors in production is pretty helpful to our users. Other than this strict check, they are mostly fairly simple and performant enough to warrant leaving it.

AFAIK, React still uses the dev-expression babel plugin and `__DEV__` checks in their code.

markerikson:
Couple more questions:

1. What scenarios are likely to result in someone dispatching objects from `Object.create(null)`, or from another window?  
2. What checks would be necessary to catch those?

Per 1, I know I've seen several addons that do stuff like serializing actions over `postMessage`, Electron main-vs-render interop, and similar concepts.  I'm not sure if any of those would run into trouble with this change.

timdorr:
> Object.create(null)

Seems unlikely. Most folks use object literals.

>  from another window

This I have zero clue about. Not saying that it's not possible, just that I've never done it and don't know the implications. Can someone wire up a demo so we can test it out?

BTW, my company does this in our Chrome extension. We have a background page with the Redux store that is replicated to the frontend content scripts. We do phone calls, so we need a single place to maintain our dialer's state. I'll ask the engineer who built that out.

src/combineReducers.js
gaearon: I don't think this would work across iframes or multiple realms. Does it? Are we sure that this is a superset of Lodash's `true` answer rather than a subset or intersection?

gaearon:
My two cents:

1. Cross-iframe and cross-realm things can get weird and cause false negatives. I'm really happy with the effort that went into Lodash implementation.

2. I don't think benchmarks are very relevant. If you're dispatching more than one or two actions in a second you already have a performance problem because your subscribers are executing much more often. You might want to reconsider using Redux if dispatching is your bottleneck.

timdorr:
<Retitling this so it doesn't come across as a dig against lodash>

src/combineReducers.js
jdalton: > I don't think this would work across iframes or multiple realms. Does it?

No it wouldn't. Or objects created with `Object.create(null)`

timdorr: Yes, I know about this. Hence the PR instead of just blasting this into master and calling it a day 😄 

It's not a use case I'm familiar with, so I'd love some help understanding it and coming up with some examples or even tests for it.

markerikson:
More research questions:

1. What edge cases do all the individual checks in `lodash.isPlainObject` actually handle?
2. Which of those checks are the most expensive?
3. Which of the edge cases are we actually concerned about in practice?

markerikson:
@gaearon , I've anecdotally seen plenty of cases where people are firing off _way_ more than "one or two actions a second".  I agree that subscribers are generally the bottleneck, but if we can reasonably say that there's some "plain object" edge cases that we aren't concerned with and improve dispatching perf overall, it's worth at least investigating.

src/combineReducers.js
phsantiago: Sry for the noob question but why `inputState.constructor !== Object` instead of `typeof inputState === 'object'`

gaearon:
The problem is that we'll likely get *false negatives* for plain objects rather than false positives. In other words we'll now throw for things that are actually completely okay (and *are* plain objects).

This is hugely disruptive to people who already rely on current behavior in their apps, sometimes without a good alternative.

>I've anecdotally seen plenty of cases where people are firing off way more than "one or two actions a second"

OK. Let's say 60 dispatches per second is the maximum I would consider reasonable. Does this PR make a measurable difference in this case?

src/combineReducers.js
TrySound: @phsantiago Because object type can represent any object `new Number` or `new Array` for example. Instance of `Object` is always plain.

rodrigorm:
@gaearon I have run a sample with 60 dispatches before and apter applying the patch:

```
$ node .
master: 0.164ms
nodash: 0.162ms
```

jayhasyee:
@timdorr @gaearon I can suggest a **lazy assertion**.
1) first we check the `action` by my previous less strict suggestion
2) if it does not meet with the first restriction, we can check with `isPlainObject` or something else

**Benefits: we do not punish the regular users, but in edge cases there may be performance overheads**

IMHO I think we should leave Lodash, because that's a heavy dependency.
My new suggestion:
```javascript
if (!action || (action.constructor !== Object && !action.__proto__ && typeof action !== 'object')) {
  throw new Error('Actions must be plain objects. ' + 'Use custom middleware for async actions.');
}
```

gaearon:
I like this. Any downsides?

gaearon:
Are there objects with `constructor === Object` but that are not considered plain objects by Lodash?

jayhasyee:
I can imagine some frequently used dispatch: logging. I've used Redux for this, and I was glad by its performance. I think the performance improving is a cardinal purpose, because Redux is a very essential library in the most of React projects. If we can make it smaller and faster, that can only be good.

jayhasyee:
@gaearon I tried my snippet with regular and `Object.create(null)` actions. Both of them worked.
With regular actions I measured `70-90 ms`.
With the other that was `200-300 ms`.
I am not sure that covers all of cases that `isPlainObject` performs. But I can agree, that the lazy assertion is our first approximation, and removing Lodash may be an other issue.

Idea: we should use Lodash's test cases.

gaearon:
What I'm trying to say is whatever the cost of `isPlainObject` is, the cost of calling subscribers and reducers will likely dwarf it in real projects. Can we please test this on a real project with 60 dispatches per second before proceeding?

jimbolla:
react-redux also uses lodash's isPlainObject to test the results of mapXToProps. It only does it during dev builds, not production. It seems like the intent of this code is the same; to make sure the developer isn't doing anything wacky. By the time it gets to prod build, the issue should have been resolved.

src/combineReducers.js
aminpaks: I'm very newbie. Would you explain what is the down side if the action is not plain?
Thanks in advance.

TrySound: @aminpaks It means it's not serializable -> may have side effects -> not predictable. And maybe redux devtools won't work too :)

aminpaks: I'm very new to Redux, I actually started a few months back using it. It's an amazing approach and thanks for all your work. We are using heavily Typescript to statically type check in our codebase, and at some point I realized that the boilerplate in defining the actions and their types is killing our team (maybe everyone!).

This is an example of defining [actions in NgRx](https://github.com/ngrx/example-app/blob/master/src/app/actions/collection.ts) project. Literally they are defined in three different place for different purposes. Looking at this recently I took [another approach](https://github.com/aminpaks/angular-redux-router/blob/master/src/app/store/store.utils.ts) to keep track of actions and its typings:
```ts
// Already defined in store definitions
import { ItemX } from '...';

// This action will have a payload as type of `ItemX`
export const AHigherOrderAction = actionCreator<ItemX>('[Project X] A Higher Order Action');


... 
// In reducer
import { AHigherOrderAction } from './actions';
...
export function reducer(state: ATypeForThisPortionOfStore = initialState, action: Action) {
  switch(action.type) {
  case AHigherOrderAction.type:
    return { ...state }
  default:
    return state;
}


...
// dispatch the action
store.dispatch(new AHigherOrderAction(aVariableWithAProperType));
...
```
Currently to be able to make it work I'm converting these actions (functions) to plain objects by a middleware.
I understand there is a performance concern in converting part (I did [this benchmark](https://jsperf.com/convert-higher-order-actions-to-plain-object-actions) to see what's the best approach), but I believe it's worth using this approach in terms the amount of code we save.

Long story short, you already mentioned some reasons that the actions are supposed to be plain. What if redux api can offer something to convert these actions internally. Cause currently we have some issues in which other middlewares are skipping the conversion and this crashes Redux.

Any thoughts?

TrySound: @aminpaks I already passed through this overengineering with actions as classes. Currently my boilerplate is
```ts
// actions.js
export type Action
  = { type: 'DO_ONE_THING' }
  | { type: 'DO_ANOTHER_THING' };

// reducers.js
import { Action } from './actions.js';

export const reducer = (state: State, action: Action) => {
  switch () {
    case 'DO_ONE_THING':
      return state;
    default:
      return state;
  }
};
```

TrySound: There is not any benefits in classes.

nasreddineskandrani: @TrySound in you example in the reducer file => Can IDE with intellisence resolve DO_ONE_THING ?

TrySound: @nasreddineskandrani Nope, but type checker will say what's wrong. For IDE fans there is small boilerplate like this
```js
export const DO_ONE_THING = 'DO_ONE_THING';
```

aminpaks: @TrySound Correct me if I'm wrong, according to you:
> It means it's not serializable -> may have side effects -> not predictable. And maybe redux devtools won't work too :)

If this won't break anything developer should be able to bypass it, no?

nasreddineskandrani: @TrySound 
please read enum string introduced in ts2.4 section here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-4.html
enum string => give you intellisense ability (save time) and type checking.

markerikson: @aminpaks , @nasreddineskandrani , @TrySound : I do appreciate the enthusiasm and interest, but this discussion would probably be better off in a different location.  

The Reactiflux chat channels at https://www.reactiflux.com always have plenty of people around who are happy to discuss Redux-related topics.  There's somewhat less discussion of TypeScript or NgRX, but there might be some people who can discuss those.

nasreddineskandrani: @markerikson I hesitate to comment but since the code was in a sub thread. I gived my 2 cents. Thanks for the link I ll be happy to exchange about redux their.

lukescott:
I think `object.constructor === Object` would only not work with cross-frames if:

1 - The frame is the same origin
2 - You compare a variable from one frame to the other - An `Object` in one frame is different from an `Object` in another.

But if you're using postMessage, the communication is serialized, so it shouldn't be an issue. Typically taking a variable from another frame is a bad idea in general.

timdorr:
I'm going to make this held off for 4.0, so we can get more extensive testing via the prerelease cycle. It's now rebased against the `next` branch.

Just to be clear, I'm on the same side of not seeing this being about a speed optimization, as dispatches should be relatively infrequent in most applications and not a bottleneck. Instead, it's a good chance to drop an external dependency that brings a fair amount of bytes with it. It's not huge, but every little bit helps.

But the key blocker here is equivalency of the plain object check. I'd love a simple function for doing that. What would be most helpful here is a test harness I can use to ensure this is equivalent. Something around message passing via iframes and fun edge case stuff. I think we can do it :+1:

gaearon:
I'd be fine with making the check more *relaxed* but not more constrained.

timdorr:
I'm wondering if we need to check for it being a plain object at all. As long as it has a `type` property, even if that's got some getter magic layered on top, it should be good to go through the dispatch function and on to the reducer. 

Should we just get rid of this check? What's the motivation behind it? Is there any downside to letting folks do `class Action { type = '' }` or other non-object-y things?

markerikson:
Somewhat relevant to the discussion: #304, #306 .

Still digging back through the history to see exactly when that check was first added, but it looks to be there pretty much from the start.

markerikson:
I'd say that the `isPlainObject(action)` check is a combination of a linting check and an enforcement of intent.  On the linting side, it's meant to avoid other values like promises, functions, or random class instances get through, so that the reducers don't get surprised by what shows up.

Meanwhile, the _intended_ use of Redux is to only allow plain objects to reach the reducers, as in "this is how we want you do use it".  Serializability is a large part of that.  See Dan's comments at https://github.com/reactjs/redux/issues/992#issuecomment-153138700 and following for specific reasoning.

markerikson:
Interesting to read [the history of `isPlainObject.js](https://github.com/reactjs/redux/commits/76113474ddabc809ff38349f9489c16a8ed12ee0/src/utils/isPlainObject.js) , and the discussion in PR #131 where we previously swapped Lodash for homegrown utils the first time.  Note the prior discussion about sizes, and also see [Dan's comments about how he wants `isPlainObject` to behave](https://github.com/reactjs/redux/pull/131#r32723289).

timdorr:
OK, I've brought back our home-grown isPlainObject with a much faster implementation and with less bytes. I also brought back our previous tests to ensure it works.

To prove it out in the browser context, here's a test-case of the exact same code: https://gist.github.com/timdorr/326ed608a9cbe67f247c2fbc709cdba3 

I've provided both a true and false case. Performance is still pretty good:
```
next: 4690.358ms
nodash: 807.397ms
```
Byte savings are still there too:
```
28582 next/redux.js
23763 nodash/redux.js
 7135 next/redux.min.js
 6340 nodash/redux.min.js
```
This should satisfy everyone's concerns, I believe. But let me know what obvious stuff I missed 😄

jdalton:
Are `null` prototype objects, `Object.create(null)` considered plain objects for y'all?

timdorr:
I consider that a very rare enough edge case to warrant not covering it. It's certainly acceptable to start rejecting it in a major revision.

jdalton:
Cool! I like while loop walking approach. I'll probably use it for Lodash too 😸

timdorr:
Yeah, it's neat. 

Just so others understand its purpose: It handles cross-realm objects by walking up the prototype chain until it reaches Object and checks if that "Object" has the same prototype as our potential plain object. 

In the case of something coming from an iframe, that Object is different than your own local Object because they come from two different JavaScript contexts. So, we use this technique to get to that definition of Object for the object instance. 

It serves as a good perf boost because you can short circuit a fair number of cases before they would get to the slower toString() calls (that's the part that slows down lodash/isPlainObject and is-plain-object).

jdalton:
@timdorr 
> It serves as a good perf boost because you can short circuit a fair number of cases before they would get to the slower toString() calls (that's the part that slows down lodash/isPlainObject and is-plain-object).

Yep, now imagine if you put that effort into improving Lodash instead of ripping it out.

timdorr:
I still want the byte savings. Lodash would be helpful if we made more extensive use of it, but for this simple case, keeping it in house and only testing the cases we care about is best.

jdalton:
> I still want the byte savings. Lodash would be helpful if we made more extensive use of it, but for this simple case, keeping it in house and only testing the cases we care about is best.

Ya'll haven't even used the webpack plugins that reduce its size.

> keeping it in house and only testing the cases we care about is best.

I donnoo Redux used Lodash, removed it, went back, it's a whole *thing*.

gaearon:
I would feel more comfortable if we attempted to contribute this upstream to Lodash first before forking.

timdorr:
> Ya'll haven't even used the webpack plugins that reduce its size.

We don't use webpack.

timdorr:
lodash/lodash#3483

jdalton:
@timdorr 
>> Ya'll haven't even used the webpack plugins that reduce its size.
> We don't use webpack.

Ah, you all did at one point. No worries, despite the name, [the plugin](https://github.com/lodash/lodash-webpack-plugin) works for rollup too :grinning:

@gaearon
> I would feel more comfortable if we attempted to contribute this upstream to Lodash first before forking.

🤘 Teamwork makes the dream work! This is in time to land for Lodash v5 (Jan) which will be a 95% smaller ES module package *(out of the box small w/o using plugins)*. The focus on Lodash v5 is simplifying things down, leveraging new language features where appropriate, and moving full steam ahead with ESM (since all modern evergreen browser will support it) and we'll use [@std/esm](https://github.com/standard-things/esm) for Node.

timdorr:
Cool, we've got that in there for whenever v5 can land. I'll revisit using it there, when we can get some nice perf *and* size benefits.

For now, I'd like to proceed with our own solution. It gets us these benefits today, meaning I can get a 4.0 beta out in the next day or two. Switching back is very easy (delete some files and change some imports), so it's not a blocker to adoption IMHO. If this proves to be error-prone, I will happily revert the change and leave things untouched. But I'd like to prove that out with betas and RCs in the real world, rather than just the limited set of edge cases my brain can dream up. We have some exceptionally creative users out there, so I'm sure they can find the breaking points 😄

Noitidart:
Thanks for getting the redux folks to upstream this. I depend on lodash a ton. Great to see!

PR #2607, opened 2017-09-12
Improve combineReducers type def with keyof

garbles:
With this change, the following code will _now_ fail the TypeScript static check. At present it does not fail.

```ts
type State = {
  a: A;
  b: B;
  c: C;
};

function aReducer(state: A, action: AnyAction): A;
function bReducer(state: B, action: AnyAction): B;
function cReducer(state: C, action: AnyAction): C;

// should fail: no key "c" using cReducer
combineReducers<State>({
  a: aReducer,
  b: bReducer
});

// should fail: "a" is using wrong reducer
combineReducers<State>({
  a: bReducer,
  b: bReducer,
  c: cReducer
});
```

 Previously, you could have passed _any_ object to `combineReducers` and the checker was OK with it. e.g.

```ts
combineReducers<State>({
  fart: () => { return null };
});
```

The definition for compose was broken for TypeScript @ 2.X so I fixed that as well by removing [this line](https://github.com/reactjs/redux/pull/2607/files#diff-b52768974e6bc0faccb7d4b75b162c99L386).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -59,9 +59,9 @@ export type Reducer<S> = (state: S, action: AnyAction) => S;
 /**
  * Object whose values correspond to different reducer functions.
  */
-export interface ReducersMapObject {
-  [key: string]: Reducer<any>;
-}
+export type ReducersMapObject<S extends { [key: string]: any }> = {
+  [K in keyof S]: Reducer<S[K]>;
+};
 
 /**
  * Turns an object whose values are different reducer functions, into a single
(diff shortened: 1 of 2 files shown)


timdorr:
This is already implemented on the next branch: https://github.com/reactjs/redux/blob/next/index.d.ts#L52

garbles:
Whoops. @timdorr thanks for the FYI.

timdorr:
No worries, it happens a lot because the branch isn't very visible.

nico1510:
@timdorr  are there plans to merge it into master from the next branch ?

PR #2617, opened 2017-09-19
Documentation addition for best practice when using ImmutableJS and Selectors.

hally9k:


timdorr:
Thanks, but this assumes Reselect usage, which isn't introduced in this document.

PR #2618, opened 2017-09-21
Update the article 'Javascript and Immutability' url

holajiawei:
'Javascript and Immutability' The url is redirected to wrong location.


timdorr:
Thanks!

PR #2622, opened 2017-09-24
Add tests and move routes within <Router> for the real-world example

sebko:
My first PR to an OSS project! 🎉 

**Potential issue**
- Writing a render test of the `<Root/>` component in the "real-world" example yields the following error message
```
Invariant Violation: You should not use <Route> or withRouter() outside a <Router>
```
While there currently aren't any tests rendering the Root component, it may be worth fixing the structure of the components to avoid future questions on why a render test of the `<Root/ >` component fails.

**Proposed changes**
- Move Routes within `<Router>` - followed this guide [https://github.com/reactjs/redux/blob/master/docs/advanced/UsageWithReactRouter.md](https://github.com/reactjs/redux/blob/master/docs/advanced/UsageWithReactRouter.md)
- Add tests - I had originally mocked out the store with `redux-mock-store` but then received a console.error in regards to DevTools not being properly configured.

I'm not sure if my difficulties testing the Root component were a symptom of not having the `<Provider>` within `src/index.js` or whether I wasn't mocking something correctly - most examples I've seen have the `<Provider>` within `src/index.js` without any test files.

### DIFF
--- a/examples/real-world/package.json
+++ b/examples/real-world/package.json
@@ -26,6 +26,6 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "eject": "react-scripts eject",
-    "test": "react-scripts test"
+    "test": "react-scripts test --env=jsdom"
   }
 }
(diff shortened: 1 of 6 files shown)


timdorr:
Thanks, but you should be using a MemoryRouter and popping <Root> inside of that in the test environment. In general, I discourage jsdom usage because of weird quirky differences in behavior.

PR #2623, opened 2017-09-24
use mapped type for combineReducers

mmkal:
Enforces using valid property names for reducers, and having the reducers return valid objects. 

Uses TypeScript 2.x feature https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types


mmkal:
Ah. Just seen https://github.com/reactjs/redux/commit/8033325697a4d1aa635a672502828c463d6f3478 and https://github.com/reactjs/redux/pull/2563 - please ignore!

PR #2624, opened 2017-09-26
Update UsageWithReact.md

spditner:
At this point in the tutorial, there is no id on a Todo instance, and the TOGGLE_TODO action is using the array index of todos -- so pass that index down to the Todo component.


timdorr:
Thanks!

PR #2627, opened 2017-09-27
Include mention of combineReducers

bmuenzenmeyer:
Updating troubleshooting.md after everything else checked out, but closer inspection of my app revealed I omitted this step.

> using Github's tooling for a quick PR, I did a quick scan of https://github.com/reactjs/redux/blob/master/CONTRIBUTING.md and didn't see anything else needed.


timdorr:
I don't really understand the suggested solution. It seems ambiguous as to what you're supposed to do.

timdorr:
Also, this is specific to users of combineReducers, which isn't everyone. I think this is too specific to add. Thanks anyways!

Issue #2631, opened 2017-09-28
Update all redux with react examples to use React 16


tigarcia:
Currently, all of the examples use React 15.5.  Updating to use react 16.

tigarcia:
PR is here https://github.com/reactjs/redux/pull/2632

PR #2632, opened 2017-09-28
Update examples to use React 16 Issue #2631

tigarcia:
Updating the package.json files of all the examples to use React 16.


timdorr:
Unfortunately, this won't work by itself. Our example tests use enzyme currently, which will require an upgrade to support React 16.

tigarcia:
Ah, sorry about that. I ran the tests in the root before I sent the PR.  I guess I should have gone into each example and run the tests from there. Could this be done as part of the issue I saw to swap out the testing framework with jest ([#2184](https://github.com/reactjs/redux/issues/2184))?

timdorr:
Since this has aged quite a bit, I'm going to close it out. But if you want to revive it, I'll be happy to re-open. Thanks!

PR #2635, opened 2017-10-01
Fix Broken "toggle-all" in TodoMVC Example

quinnnned:
As of todomvc-app-css v2.1.0, the "toggle-all" chevron decoration is no longer applied to the checkbox, but instead to its associated label.

Prior to this commit, the todomvc example did not use a label, so the v2.1.0 change had the effect of hiding the "toggle-all" control.

This commit adds the label -- fixing this bug.

Tests have been updated to reflect these changes.

### DIFF
--- a/examples/todomvc/src/components/MainSection.js
+++ b/examples/todomvc/src/components/MainSection.js
@@ -30,10 +30,13 @@ export default class MainSection extends Component {
     const { todos, actions } = this.props
     if (todos.length > 0) {
       return (
-        <input className="toggle-all"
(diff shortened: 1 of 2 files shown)


quinnnned:
Additional Context:

[breaking change to todomvc-app.css](https://github.com/tastejs/todomvc-app-css/commit/808217fab49a5e2da2d5e52aee417c62807cec28) 

[similar PR to fix vuex example](https://github.com/vuejs/vuex/pull/761)

timdorr:
Thanks!

Issue #1398, opened 2016-02-14
passing an invalid reducer to combineReducers() silently fails


nabn:
If I pass something to `combineReducers()` that doesn't really return a valid reducer, it silently fails. But if it's the only reducer passed to `combineReducers()`, it throws an error saying valid reducers were not passed. I'm not sure if it's a bug, but sure does look like one to me.

Isn't it better if it warned me if one of the arguments passed to `combineReducers()` wasn't a valid reducer?

I ran into this issue when I forgot to do `export default myReducer` in one of my reducer files.

aakashsigdel:
yes, I think the combineReducer just checks if the object's provided value is a function or not, and doesn't warn if the the function passed is undefined.
```javascript
if (typeof reducers[key] === 'function')  {
  finalReducers[key] = reducers[key];
}
````

When pushing the value to `finalReducer` array, we could make it throw an error or give a warning if the value of the reducer passed is not a function

we could have something like

```javascript
if (typeof reducers[key] === 'function')  {
  finalReducers[key] = reducers[key];
} else {
  throw 'The reducer' + reducers[key] + 'is not a valid type of reducer'
}
```

gaearon:
The reason we have this check is that it used to be a popular pattern to `import * from './reducers'` which will give you an object with some potentially unrelated fields, e.g. `__esModule: true`.

Maybe we should give up on this behavior since we don’t encourage `import *` for reducers anymore but that would be a breaking change.

pke:
Also I might add that a reducer that throws also does so silently. You only see that its supposed to be reduced state is missing in the state.

I had a wrong import for action types and my switch statement then basically said:

```
switch (action.type) {
  case undefined.ACTION_NAME
   ....
}
```

A `sanityError` is saved in combineReducers, though its message does not contain which reducer was insane. That happens because the call to the reducer inside `assertReducerSanity` is not wrapped in `try...catch`

I should be able to make a PR fixing the error reporting.

hoangnm:
Is there any news for this ? I see that it's silently failed because in createStore.js (line 170), we just skip the error. Can we catch the error and log to the console, it helps easier to debug when something wrong...

Huanzhang89:
Agree with @gaearon on this one but maybe in the next major release since this will be a breaking change for a lot of production apps using `import *` for reducers?

timdorr:
[...]

PR #2636, opened 2017-10-02
Warns when type of reducers passed is invalid to avoid silent failure

jekku:
Attempts to solve issue #1398 .

Undefined reducers are still checked for intuitive debugging.
Warnings are now given to the console if a passed reducer is not a function.
Warnings are also given to the console if a whole es module imported via the * syntax is passed to combineReducers.

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -106,8 +106,11 @@ export default function combineReducers(reducers) {
     const key = reducerKeys[i]
 
     if (process.env.NODE_ENV !== 'production') {
-      if (typeof reducers[key] === 'undefined') {
+      const reducerUnderInspection = reducers[key]
+      if (typeof reducerUnderInspection === 'undefined') {
         warning(`No reducer provided for key "${key}"`)
+      } else if (typeof reducerUnderInspection !== 'function') {
+        warning(`Reducer provided for "${key}" is not a function. Received a ${typeof key}.`)
       }
     }
 
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -106,11 +106,12 @@ export default function combineReducers(reducers) {
     const key = reducerKeys[i]
 
     if (process.env.NODE_ENV !== 'production') {
-      const reducerUnderInspection = reducers[key]
-      if (typeof reducerUnderInspection === 'undefined') {
+      if (typeof reducers[key] === 'undefined') {
         warning(`No reducer provided for key "${key}"`)
-      } else if (typeof reducerUnderInspection !== 'function') {
-        warning(`Reducer provided for "${key}" is not a function. Received a ${typeof key}.`)
+      } if (key === '__esModule') {
+        warning(`Passing a whole ES Module to combine reducers is discouraged.`)
+      } else if (typeof reducers[key] !== 'function') {
+        warning(`Reducer provided for "${key}" is not a function. Received type: ${typeof key}.`)
       }
     }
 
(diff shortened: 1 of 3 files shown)


test/combineReducers.spec.js
```diff
@@ -18,14 +18,30 @@ describe('Utils', () => {
...
         another: { nested: 'object' },
         stack: (state = []) => state
       })
 
+      expect(spy.mock.calls[0][0]).toMatch(
+        /Reducer provided for "fake" is not a function/
+      )
+
+      expect(spy.mock.calls[1][0]).toMatch(
+          /Reducer provided for "broken" is not a function/
```
rstacruz: Indentation :)

test/combineReducers.spec.js
```diff
@@ -18,14 +18,30 @@ describe('Utils', () => {
       expect(s2).toEqual({ counter: 1, stack: [ 'a' ] })
     })
 
-    it('ignores all props which are not a function', () => {
+    it('warns on props which are not a function and excludes them', () => {
+      const preSpy = console.error
+      const spy = jest.fn()
+      console.error = spy
```
rstacruz: You've overridden `console.error`, but there's no mechanism here to put it back after the test.

The Jest way to do this is to make a mock of `utils/warning` to capture any calls to `warning()`.

```js
jest.mock('../src/helpers/warning')
```

https://facebook.github.io/jest/docs/en/manual-mocks.html has more info :)

rstacruz:
(Maintainers: sorry for the impromptu review; @jekku is a colleague of mine who's asked his coworkers to review his contribution.)

test/combineReducers.spec.js
```diff
@@ -18,14 +18,30 @@ describe('Utils', () => {
...
+
       const reducer = combineReducers({
         fake: true,
         broken: 'string',
         another: { nested: 'object' },
         stack: (state = []) => state
       })
 
+      expect(spy.mock.calls[0][0]).toMatch(
```
rstacruz: Another (less-precise, but terser) way to do this is

```js
expect(spy).toHaveBeenCalledWith('Reducer provided for "fake" is not a function')
```

See: https://devhints.io/jest

jekku: Thank you! :)

src/combineReducers.js
```diff
@@ -106,8 +106,11 @@ export default function combineReducers(reducers) {
     const key = reducerKeys[i]
 
     if (process.env.NODE_ENV !== 'production') {
-      if (typeof reducers[key] === 'undefined') {
+      const reducerUnderInspection = reducers[key]
```
timdorr: You shouldn't need to extract this out. Just keep using `reducers[key]`. It's probably not as fast, but this is dev-only code.

test/combineReducers.spec.js
timdorr: They can do what we do in other tests. Just add this at the end:
```js
       spy.mockClear()
       console.error = preSpy
```

timdorr:
Whoops, got a little cocky with those test updates. Serves me right for trying to use the web interface to edit 😄  Feel free to revert that!

timdorr:
One thing to note: We tried this on bindActionCreators and had to revert there as well: #2279  

Unfortunately, with most people transpiling modules through Babel, there are side effects when the pattern of `import * as reducers from './reducers'` is used. Unless we can somehow be rock solid in protecting against this without hacks, I'm not sure this is going to be mergeable.

jekku:
@timdorr would throwing out an exception when an `import *` is invoked to combineReducers, help?

timdorr:
@jekku Sorry for not responding. At least currently, there isn't a rock-solid, cross-platform way of doing that. You can check for `__esModule`, but that may not come across depending on how folks have their transpilation set up. 

Unfortunately, fixing for this kind of issue isn't going to be possible in the current ecosystem. I had to revert a similar warning in a previous release because it caused other sorts of headaches like this. 

Unless someone can think up a novel way to resolve it that covers both native and loosely transpiled environments, I'm not sure there's a solution.

PR #2637, opened 2017-10-02
Added store transactions

uditalias:
# Transactions

## Why
When using libraries like [react-redux](https://github.com/reactjs/react-redux), each action dispatched through the store will trigger all store listeners.

### Goals
1. To enable the option to dispatch multiple actions (from a middleware for example) without triggering the store listeners for each action.
2. Increase application performance

### How it works
Let's take [react-redux](https://github.com/reactjs/react-redux) as an example. Each non-falsy `mapStateToProps` will create a store subscription. When you dispatch a redux action, all store subscription get called. What if I want to dispatch multiple actions in "one shot" and only when I finish I want redux store to call my listeners?

This is where transactions comes in. you create a transaction with actions dispatches within the transaction body, and when its finish, the store listeners gets called.

```javascript
store.transaction(() => {
    store.dispatch(addTodo('Hello'));
    store.dispatch(filterTodos('new'));
});

// store listeners will be called after transaction body
```

### PR Includes

1. `store.transaction` method to create a transaction.
2. Enhanced `dispatch` method to enable skip listeners when transaction scope is running.
3. Added tests.
4. Fixed tests to work with this PR.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -8,7 +8,8 @@ import $$observable from 'symbol-observable'
  * Do not reference these action types directly in your code.
  */
 export const ActionTypes = {
-  INIT: '@@redux/INIT'
+  INIT: '@@redux/INIT',
+  TRANSACTION: '@@redux/TRANSACTION'
 }
 
 /**
@@ -59,6 +60,8 @@ export default function createStore(reducer, preloadedState, enhancer) {
   let currentListeners = []
   let nextListeners = currentListeners
   let isDispatching = false
+  let isTransaction = false
+  let dispatchCalledInTransaction = false
 
   function ensureCanMutateNextListeners() {
     if (nextListeners === currentListeners) {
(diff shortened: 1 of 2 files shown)


timdorr:
Interesting, but I think this would be better implemented as a store enhancer. Batched actions have [been solved](https://github.com/tappleby/redux-batched-subscribe) that way [before](https://github.com/tshelburne/redux-batched-actions), on both the pre- and post-dispatch side of things.

I'd rather leave this up to the community to solve, than try to bake in a solution.

markerikson:
Yep, there's [many existing ways to dispatched batched actions in userland](https://github.com/markerikson/redux-ecosystem-links/blob/master/store.md#batching-and-notifications), and if you want to create a new one, then a store enhancer is the right way to do it.

If you do turn this into an enhancer, please submit a PR to add it as an entry in my addons catalog.

uditalias:
Thank you guys
@markerikson I'll create a store enhancer

PR #2641, opened 2017-10-03
Make bindActionCreators transparently pass `this`.

### LINKED ISSUE #2638: Feature Request: pass `this` through bound actionCreator
Is there any chance of changing `bindActionCreator` from:
```js
function bindActionCreator(actionCreator, dispatch) {
  return (...args) => dispatch(actionCreator(...args))
}
```

to:
```js
function bindActionCreator(actionCreator, dispatch) {
  return function() { return dispatch(actionCreator.apply(this, arguments)); };
}
```

The existing behaviour explicitly sets `this` to undefined so this [shouldn't](https://xkcd.com/1172/) be a breaking change.

I have some mixins/decorators that I apply to action creators by composition such that the arguments don't change but I can adjust settings in the enhancer by passing an object through `this`.  If I use some of the convenience functions like `connect` I lose the ability to pass through these objects.

Parakleta:
Resolves #2638

### DIFF
--- a/src/bindActionCreators.js
+++ b/src/bindActionCreators.js
@@ -1,5 +1,5 @@
 function bindActionCreator(actionCreator, dispatch) {
-  return (...args) => dispatch(actionCreator(...args))
+  return function() { return dispatch(actionCreator.apply(this, arguments)) }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


timdorr:
OK, seems minor. Let's do it.

Parakleta:
Thanks.

PR #2647, opened 2017-10-06
Doc fix: updateObjectInArray replaces with new value, not an obj.

dlrice:


timdorr:
Thanks!

markerikson:
Actually, I really did intend the original usage.  The intent was the `action.item` might be a partial object, not the entire thing.  The edited version might be the way you'd want to do it if your action always includes the entire new item, but that's not what I had in mind when I wrote this.

Appreciate the PR, but I'm going to revert this back to the original version.

timdorr:
Ah, thanks. I wasn't 100% on that.

markerikson:
Yeah, I probably didn't make it entirely clear in the writeup, but that's what I was thinking when I wrote it.

PR #2650, opened 2017-10-06
updates the testing section on async actions to use fetch-mock

kellyrmilligan:
nock doesn't work as far as I can tell. 

https://medium.com/@kellyrmilligan/testing-async-actions-in-redux-with-isomorphic-fetch-and-fetch-mock-35f98c6c2ee7

specifically mentioned in that article:

https://github.com/node-nock/nock/issues/409
https://github.com/node-nock/nock/issues/591


timdorr:
Thanks!

Issue #1636, opened 2016-04-19
Redux.replaceReducer throw warning "Unexpected key found"


andreechristian:
This issue was moved from https://github.com/reactjs/react-redux/issues/360

> Hello,
> I tried to do a code splitting on route change.
> 
> When the route changes to '/register', i dynamically add "registration" as one of the root state using replaceReducer and it work just fine.
> Once the registration completed i want to remove the reducer using replaceReducer again, but instead it return the following warning :
> 
> Unexpected key "registration" found in initialState argument passed to createStore. Expected to find one of the known reducer keys instead: ... Unexpected keys will be ignored.
> 
> Thanks in advance

andreechristian:
Basically, i cannot remove a 'sub state' once it is added to the store :(

gaearon:
Technically this is a warning rather than an error so you can. But it is annoying. I think we should fix this.

cpeddecord:
I'm running into this warning after adding serialized state from a server handoff whilst routing into a route with a dynamically added reducer. Only happens on initial load as my routes initialize on the client after the store. Wondering if there's a good workaround here...

andreechristian:
I think after changing `current reducer` into `next reducer`, we should trim unnecessary state from current state based on `next reducer` keys. Or if the store doesn't have multiple reducers, do something like recreating store state based on next reducers default action => and then assign current state matched data. And only after that we should dispatch the INIT.

andreechristian:
Also @gaearon, although this is a warning, it seemed to me that the state doesn't necessarily changed, i still got 'registration' as one of my root state. Or is it expected? But if it is expected, then my 'registration' state will gonna be a dead meat right there, right? The reducer's gone but the state remains.

gaearon:
I would removing that warning and checking which tests fail. They likely show reasonable ideas or why this warning was added. Then we can consider how to make it works well for all use cases.

flucivja:
Hi, 

I came across same issue when I am trying to build modular application. The basic idea is that once module is loaded it will inject its own reducers and once module is unloaded it should remove its own reducers. 
I've accomplished it with usage with react-router and using onEnter and onLeave hooks. Basically onEnter hook inject module's reducers with replaceReducer and dispatch actions 'a1', 'a2', 'a3' on the injected reducers. Then once user navigates outside of the module, module`s reducers are removed in onLeave hook also with usage of replaceReducer. 
[...]

PR #2651, opened 2017-10-07
Do not warn if replaceReducer causes reducer shape to change

Kaivosukeltaja:
Attempts to solve issue #1636.
`replaceReducer` calls the new reducer with empty objects as parameter to get the desired state shape, then filters out the keys in the current state that don't exist in the new one.
Added new test to verify functionality.
My first PR to Redux, please be gentle :)

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -196,6 +196,17 @@ export default function createStore(reducer, preloadedState, enhancer) {
       throw new Error('Expected the nextReducer to be a function.')
     }
 
+    const reducerShape = nextReducer({}, {})
+
+    if (typeof currentState === 'object' && Array.isArray(currentState) === false) {
+      currentState = Object.keys(currentState).reduce((targetShape, key) => {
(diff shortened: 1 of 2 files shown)


quinnnned:
I don't believe this will work for unusually deep or unusually shallow state trees.

1. **Deep**: This will only fix the warning for a top-layer combineReducers.  A deeper state tree with multiple layers of combineReducers would still produce the warning.

2. **Shallow**: Consider the following reducer:
````js
export default (state = {}, action) => {
    switch (action.type) {
        case "ADD_KEY_VALUE_PAIR":
            return {
                ...state,
                [action.key]: action.value
            }
        default:
            return state
    }
}
````

This state tree is a simple collection of key-value pairs.  The proposed solution would wipe the state back to an empty object

timdorr:
I appreciate the PR, but I think #2673 is the way to go here. Thanks anyways!

PR #2652, opened 2017-10-08
Use acition.id instead of action.index

nuronialBlock:
In the documentation, the todo object has no field named as index. But there is a field named as id.


nuronialBlock:
Sorry, Found my mistake. I was following both example todoApp and naratted todoApp.

markerikson:
Hi.  Appreciate the PR, but the current use of `index` is correct.  The second argument to a `map()` callback will always be the index of that item in the array, and the assumption is that the code creating the action is also including the appropriate index value for the reducer to check against.

nuronialBlock:
Thanks @markerikson and sorry for wasting your time.

PR #2660, opened 2017-10-13
Fix real-world example use of react-router-dom

auswalk:
Corrected use of React Router in real-world example so app runs properly in production build


timdorr:
Oh, there's still a yarn lockfile in there. I'll remove that. You should be using npm to install things.

PR #2662, opened 2017-10-13
Be more specific on line 32, last sentence.

cstoddart:
I would be happy with any change that replaces "it is wiser to generate a unique ID every time SOMETHING NEW is created" with something more specific. As someone trying to learn Redux, I would like to know exactly what you mean here regarding action index, and I think "something new" is too vague.

### DIFF
--- a/docs/basics/Actions.md
+++ b/docs/basics/Actions.md
@@ -29,7 +29,7 @@ import { ADD_TODO, REMOVE_TODO } from '../actionTypes'
 
 Other than `type`, the structure of an action object is really up to you. If you're interested, check out [Flux Standard Action](https://github.com/acdlite/flux-standard-action) for recommendations on how actions could be constructed.
 
-We'll add one more action type to describe a user ticking off a todo as completed. We refer to a particular todo by `index` because we store them in an array. In a real app, it is wiser to generate a unique ID every time something new is created.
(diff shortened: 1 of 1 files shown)


markerikson:
The existing sentence does make sense.  A "real app" isn't going to be creating todos.  A blogging app might be creating Users, Posts, Comments, and so on.

The todos example is simply storing todos in an array, like this:

```js
const todos = [
    {text : "Buy Milk"},
    {text : "Mow Lawn"}
];
```

Those todo entries don't have any kind of an ID value.  So, the only way to identify a given todo is to use its current index in the array.

cstoddart:
What does "every time something new is created" refer to in that sentence? Does it mean every time a new action type is created? Surely it doesn't mean literally any time anything new is added. At the moment I am unsure what this sentence tells me regarding when I should be generating a unique ID for actions.

timdorr:
Every time a new bit of state (normally an object) is added to the store that you want to access frequently or quickly, creating a unique identifier to that data is best. You will not normally store things in arrays, but instead plain objects as essentially hash tables.

cstoddart:
Thank you for the clarification Tim! That definitely makes sense.

I hope they update the sentence to say something like "every time a new object is added to the store".

PR #2664, opened 2017-10-15
Updated Action Creators second paragraph

fuchodeveloper:
Added punctuation to the second paragraph of the Action Creators section to make text easier to understand.
Link to updated section: http://redux.js.org/docs/basics/Actions.html#action-creators


markerikson:
Thanks!

PR #2666, opened 2017-10-16
Remove warning about router 2/3

adimit:
The recent PR by @timdorr for #2300 updated this guide to be compatible with Router v4, so the warning is now very confusing.


timdorr:
Thanks!

PR #2672, opened 2017-10-22
Recursively Prune State on store.replaceReducer

### LINKED ISSUE #1636: Redux.replaceReducer throw warning "Unexpected key found"

quinnnned:
Resolves #1636 -- Prevents combineReducers from warning about unexpected keys after calling replaceReducer.

Adds recursive `pruneState` function to createStore, which uses the differences between the current and next state shapes to determine which state branches have been removed (are not acknowledged by the next reducer).  The removed state branches are then removed from the current state.

The state shape is defined here as the union of every path which begins at the state tree root and terminates at a non-plain-object leaf (boolean, number, string, array, etc).

Properties of a plain object which do not appear in the current or next shape are not pruned.  This covers the use case where a plain object is being used as a dynamic collection of key-value pairs.

If any plain object is unchanged by the pruning process, the original object is returned instead of a new one.

The state shape for a given reducer is determined by invoking it with an undefined state and the INIT action.

The combineReducers warning does not appear in production, so this pruning process also does not take place in production.

Added unit test which fails (generates warning) when the pruning process does not take place.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -66,6 +66,27 @@ export default function createStore(reducer, preloadedState, enhancer) {
     }
   }
 
+  function pruneState(state, currentShape, nextShape) {
+    if (![state, currentShape, nextShape].every(isPlainObject)) {
+      return state
+    }
+    let hasChanged = false
+    const stateKeys = Object.keys(state)
+    const nextState = {...state}
+    for (let i = 0; i < stateKeys.length; i++) {
+      const key = stateKeys[i]
+      if (key in currentShape) {
+        if (key in nextShape) {
+          nextState[key] = pruneState(state[key], currentShape[key], nextShape[key])
+        } else {
+          delete nextState[key]
+        }
+      }
+      hasChanged = hasChanged || nextState[key] !== state[key]
+    }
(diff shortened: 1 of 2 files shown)


timdorr:
I appreciate the PR, but I think #2673 is the way to go here. If I'm convinced otherwise, we can reopen this.

quinnnned:
Okay. I was looking for a way to supress the error without having to change redux's production behavior.

timdorr:
I think we can do that in 4.0, which would seem to be a better solution. It certainly opens up a few other special case handlers as well (particularly for HMR cases).

Issue #1736, opened 2016-05-16
Action never seen when dispatched on Store.


ramakay:
I have a non React project with the following code
.

*index.js*

```
import configureStore from 'js/store/configureStore'

/* Redux Imports */
import { createPage } from 'js/actions/page'

// Boot the App
document.addEventListener('DOMContentLoaded', function () {
  const initialState = window.__INITIAL_STATE__
  const pageId = window.__INITIAL_STATE__.page
  const store = configureStore(initialState)
  store.dispatch(createPage(pageId))
})

*actions/page.js*
function initPage (pageId) {
  return {
    type: 'CREATE_PAGE',
    page: pageId
  }
}

export function createPage (pageId) {
  return dispatch => {
    dispatch(initPage(pageId))
  }
}
```

*reducers/page.js*

```
/**
 * Page Mediators get imported here
 * Page Mediators are brokers that hold page initialization.
 * The bundle includes all page mediators but they will not be activated
 * unless thre reducer calls them.
 */
import * as cart from 'js/templates/cart'

/**
 * This is a edge case, index.js boots the app and we are subscribing to 
 * a Redux @@INIT action instead of adding one more layer of actions
 * which leads to unintended scenarios.
 */

const page = (state = {page: []} , action) => {

  console.log(state)
  switch (action.type) {
    case '@@INIT':
      console.log('initializing page:' + state.id)
      console.log(state)
      dispatch({type: 'CREATE_PAGE',page: state.id})
      cart.init()
      return state
    case 'CREATE_PAGE':
      console.log('initializing page:' + state.id)
      cart.init()
      return state
    default:
      return state
  }
}
export default page
```

----
The problem I have is that `actions/page.js` never sees the `CREATE_PAGE` action. Any thoughs?

gaearon:
Unrelated but

>/**
> * This is a edge case, index.js boots the app and we are subscribing to 
> * a Redux @@INIT action instead of adding one more layer of actions
> * which leads to unintended scenarios.
> */

can break your app any day.

We [specifically say](https://github.com/reactjs/redux/blob/99118402ca3334d68de4822dfa4f6461a80da31d/src/createStore.js#L4-L9):

>/**
> * These are **private** action types reserved by Redux.
> * For any unknown actions, you must return the current state.
> * If the current state is undefined, you must return the initial state.
> * **Do not reference these action types directly in your code.**
> */

ramakay:
[...]

PR #2673, opened 2017-10-22
REPLACE action for replaceReducers

### LINKED ISSUE #1636: Redux.replaceReducer throw warning "Unexpected key found"

timdorr:
Fixes #1636 

Provides a separate private action for when we replace reducers. This allows us to specifically handle this case, which resolves a false dev-only warning for combineReducers currently.

### DIFF
--- a/src/utils/actionTypes.js
+++ b/src/utils/actionTypes.js
@@ -5,7 +5,8 @@
  * Do not reference these action types directly in your code.
  */
 var ActionTypes = {
-  INIT: '@@redux/INIT'
+  INIT: '@@redux/INIT',
+  REPLACE: '@@redux/REPLACE'
 }
 
 export default ActionTypes
--- a/src/utils/actionTypes.js
+++ b/src/utils/actionTypes.js
@@ -5,8 +5,8 @@
  * Do not reference these action types directly in your code.
  */
 const ActionTypes = {
-  INIT: '@@redux/INIT',
-  REPLACE: '@@redux/REPLACE'
+  INIT: '@@redux/INIT' + Math.random().toString(36).substring(7).split('').join('.'),
+  REPLACE: '@@redux/REPLACE' + Math.random().toString(36).substring(7).split('').join('.')
 }
 
 export default ActionTypes
--- a/src/utils/actionTypes.js
+++ b/src/utils/actionTypes.js
@@ -4,7 +4,7 @@
  * If the current state is undefined, you must return the initial state.
  * Do not reference these action types directly in your code.
  */
-var ActionTypes = {
+const ActionTypes = {
   INIT: '@@redux/INIT',
   REPLACE: '@@redux/REPLACE'
 }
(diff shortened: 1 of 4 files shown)


quinnnned:
Hey that test looks familiar. 😉 Glad you liked it.

quinnnned:
@timdorr what if the nextReducer depends on the INIT action for proper behavior?  Or is it just understood that a reducer must not depend on the INIT action?

timdorr:
> Hey that test looks familiar. 😉 Glad you liked it. 

Yes, thank you! I'll amend the commit so you get credit.

quinnnned:
Gotcha.  I also just saw Dan's advice in #186.  Should have done a quick google first.

timdorr:
As requested in https://github.com/reactjs/redux/issues/1736#issuecomment-219484915, this also adds a randomized name to the private actions, making them harder to look for.

src/utils/actionTypes.js
```diff
@@ -4,8 +4,9 @@
  * If the current state is undefined, you must return the initial state.
  * Do not reference these action types directly in your code.
  */
-var ActionTypes = {
-  INIT: '@@redux/INIT'
+const ActionTypes = {
+  INIT: '@@redux/INIT' + Math.random().toString(36).substring(7).split('').join('.'),
+  REPLACE: '@@redux/REPLACE' + Math.random().toString(36).substring(7).split('').join('.')
```
asantos00: Does it makes sense to have this *random token generator* in a function ? (since it being used at least twice here)

timdorr: Probably, if we start having more than just two. I'm find with the duplication for now.

lvivski: Why do you need two separate random strings generated? Seems that you can generate it once and use in both action types.

timdorr: Yes, but that seems like splitting hairs. It doesn't really matter.

lvivski: Ok, fair enough. There's also https://github.com/reactjs/redux/blob/master/src/combineReducers.js#L72. Does it make any sense to put all of them in the same place?

quinnnned: I'm with @timdorr here.  When applying the D.R.Y. principle, it is critical to distinguish between essential duplication and coincidental duplication, because trying to "DRY-up" the latter kind can increase code complexity.  In cases where it is not clear what kind of duplication it is, one best practice is to permit the duplication until it causes actual pain.  This seems to be what @timdorr is doing.

markerikson: @quinned : highly relevant here: https://twitter.com/acemarke/status/901329101088215044 :)

timdorr:
Without further objections, I think this is OK to merge in. If anyone has any issues with it, feel free to revert or submit a PR with improvements.

PR #2671, opened 2017-10-22
Fix TypeScript tests to pass for new TypeScript definitions

evgenyt1:
Fixed typescript tests for new d.ts
Use prepare instead of prepublish in package.json (same as in master branch).
Update typescript in yarn.lock to comply to what's in package.json

I'm using the `next` branch to have new d.ts and strongly typed actions.

However wondering if the `next` branch is going to be synced with the current master?

PR #2674, opened 2017-10-22
Fix TS definitions test for new Dispatch typing

pelotom:


timdorr:
BTW, there is another PR (that I busted with `next` updates just now), that has a similar, but different set of fixes: https://github.com/reactjs/redux/pull/2671/commits/b9278d1e9daef210208a4faae20266c6a96f3a05

Any thoughts on the comparison between the two?

pelotom:
By and large it's equivalent, but these definitions aren't quite right:

```ts
// dispatch.ts
declare module "../../" {
  export interface Dispatch<D = Action> {
    <R>(asyncAction: (dispatch: Dispatch<D>, getState: () => D) => R): R;
    // ^ `getState()` should not return the action type `D`
  }
}
```

```ts
// middleware.ts
declare module "../../" {
    export interface Dispatch<D> {
        <R,S>(asyncAction: (dispatch: Dispatch<D>, getState: () => S) => R): R;
        // ^ `getState()` should not be parameterized by the state type `S`; the type
        // of the state is not determined by the caller
    }
}
```

timdorr:
Thanks. I'm starting to get my bearings with TS stuff, but our typings are *really* dense and I need some hand-holding :)

I think this looks good. Since we're now passing, this should get my other PRs unblocked. Thank you!

pelotom:
No problem, looking forward to seeing these typings on `master`!

PR #2675, opened 2017-10-23
Upgrade flow version to 0.57.3 for flow-redux example

villesau:
Wanted to see if the typings for react-redux works with latest version of Flow. Most of the changes are from flow-upgrade, but change from event.target to event.currentTarget and error if root element is not found needed to be added manually.

### DIFF
--- a/examples/todos-flow/src/__tests__/components/TodoList.test.js
+++ b/examples/todos-flow/src/__tests__/components/TodoList.test.js
@@ -1,6 +1,6 @@
 // @flow
 
-import React from 'react';
+import * as React from 'react';
 import { shallow } from 'enzyme';
 
 import TodoList from '../../components/TodoList';
--- a/examples/todos-flow/src/__tests__/components/TodoList.test.js
+++ b/examples/todos-flow/src/__tests__/components/TodoList.test.js
@@ -1,6 +1,6 @@
 // @flow
 
-import * as React from 'react';
+import React from 'react';
 import { shallow } from 'enzyme';
 
 import TodoList from '../../components/TodoList';
--- a/examples/todos-flow/src/components/Link.js
+++ b/examples/todos-flow/src/components/Link.js
@@ -1,10 +1,10 @@
 // @flow
 
-import * as React from 'react';
+import React, { type Node } from 'react';
 
 export type Props = {
   active: boolean,
-  children?: React.Node,
+  children?: Node,
   onClick: () => void
 };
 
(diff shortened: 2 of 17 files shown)


examples/todos-flow/src/__tests__/components/TodoList.test.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-import React from 'react';
+import * as React from 'react';
```
timdorr: These lines should be left alone. This changes behavior in a subtle way and isn't a best practice.

examples/todos-flow/src/components/Link.js
```diff
@@ -1,10 +1,10 @@
 // @flow
 
-import React from 'react';
+import * as React from 'react';
 
 export type Props = {
   active: boolean,
-  children?: React$Element<any>,
+  children?: React.Node,
```
milesj: `React$Node` avoids the wildcard import.

examples/todos-flow/src/__tests__/components/TodoList.test.js
villesau: Interesting! How it changes the behaviour? This result was purely from flow-upgrade and can be changed ofc.

Flow docs encourage this way of importing React too:

>Note: We import React as a namespace here with `import * as React from 'react'` instead of as a default with `import React from 'react'`. When importing React as an ES module you may use either style, but importing as a namespace gives you access to React’s utility types.

see: https://flow.org/en/docs/react/components/

examples/todos-flow/src/components/Link.js
villesau: I can also import `import React, { type Node } from 'react';`

examples/todos-flow/src/__tests__/components/TodoList.test.js
timdorr: The primary thing is it changes the import from "just the default please" to "everything including the default." This makes it harder for tree-shaking bundlers to work out what is and is not used from a module. It's subtle, but important in its implications.

timdorr:
Thanks for this!

examples/todos-flow/src/__tests__/components/TodoList.test.js
villesau: Right, thanks for the clarification :) Indeed it makes tree shaking problem more difficult to solve.

PR #2677, opened 2017-10-24
Counter

taras:


timdorr:
Thanks for the PR, but the point of this example is to keep it simple and not use extraneous libraries. It's an intro to Redux.

PR #2678, opened 2017-10-24
MigratingToRedux

trevdor:
**What does this PR do?**
Clarifies which migration process is being described in the guide.

**Why?**
First, thanks for the docs! This is a big help as we convert our few bits of Flux to Redux.
However, when referencing this guide, a coworker and I were both confused at first glance, thinking the "Your process will look like this" line was referring to the previous line that describes migrating from Redux back to a related lib. 
Suggested edit is to move that line to the end so it's clear it's a note rather than the subject of the steps.


timdorr:
I'd just remove that line entirely. I'm not sure what that's doing in a doc about migrating *to* Redux.

trevdor:
Seems reasonable to me. Removed it.

timdorr:
Thanks!

PR #2681, opened 2017-10-25
Update markdown syntax on DataFlow.md

kjad:
Minor markdown fix for documentation. I noticed that there was text being rendered in a code block (due to 4 space indent instead of 2 spaces), when it most likely should just be rendered as normal text.


markerikson:
Hmm.  The current deployed version of the docs looks okay to me, but the tweak shouldn't hurt anything.  Thanks!

PR #2686, opened 2017-10-31
Update WritingTests.md

GaryPWhite:
https://github.com/wheresrhys/fetch-mock/issues/104

I wrangled with why the testing for `'isomorphic-fetch'` and `'fetch-mock'` didn't work before finding the above issue, and realizing that you must `import 'isomorphic-fetch'` as a global for `'fetch-mock'` to override it. Wanted to update the docs so the next person doesn't have the same issue :)

### DIFF
--- a/docs/recipes/WritingTests.md
+++ b/docs/recipes/WritingTests.md
@@ -80,7 +80,7 @@ For async action creators using [Redux Thunk](https://github.com/gaearon/redux-t
 #### Example
 
 ```js
-import fetch from 'isomorphic-fetch'
+import 'isomorphic-fetch'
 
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

GaryPWhite:
May also be worth changing, but i'm not sure:

in the response promise:
```
fetchTodosSuccess(json.body)
```
should be
```
fetchTodosSuccess(json)
```
based on the setup, it might be more appropriate to change the whole line from:
```
.then(json => dispatch(fetchTodosSuccess(json.body)))
```
to
```
.then(body => dispatch(fetchTodosSuccess(body)))
```

PR #2687, opened 2017-10-31
Update WritingTests.md

GaryPWhite:
Another change, same day! :)

found another little hiccup earlier, and I mentioned it in the [PR](https://github.com/reactjs/redux/pull/2686)

in the response promise:
```Javascript
fetchTodosSuccess(json.body)
```
should be
```Javascript
fetchTodosSuccess(json)
```
based on the setup, it might be more appropriate to change the whole line from:

```Javascript
.then(json => dispatch(fetchTodosSuccess(json.body)))
```
to

```Javascript
.then(body => dispatch(fetchTodosSuccess(body)))
```


timdorr:
Thanks!

PR #2689, opened 2017-11-02
no need to require `rootReducer` again

stevemao:
[`babel-preset-react-app` does not transform modules to CJS](https://github.com/facebookincubator/create-react-app/blob/master/packages/babel-preset-react-app/index.js#L107-L108) so I don't think we need to require `rootReducer` again

### DIFF
--- a/examples/real-world/src/store/configureStore.dev.js
+++ b/examples/real-world/src/store/configureStore.dev.js
@@ -18,8 +18,7 @@ const configureStore = preloadedState => {
   if (module.hot) {
     // Enable Webpack hot module replacement for reducers
     module.hot.accept('../reducers', () => {
-      const nextRootReducer = require('../reducers').default
-      store.replaceReducer(nextRootReducer)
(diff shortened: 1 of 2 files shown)


timdorr:
Yes, we do so that the module will be reloaded. We have to use a require statement to get it to reload properly.

stevemao:
Thanks @timdorr . 
https://github.com/gaearon/react-hot-loader#getting-started
In step 4, 

```js
if (module.hot) {
  module.hot.accept('./containers/App', () => { render(App) })
}
```

`App` component is not re`require`ed, why this is different here? 
Also is there a way to get rid of commonjs and just use ES modules?

gaearon:
If you use webpack 2 then an extra require is unnecessary.

gaearon:
Since we do use it, I think PR is probably correct. Needs somebody to verify it :-)

timdorr:
I guess you're right. Didn't know about that functionality.

The example is on webpack 2.6.1, so this is good to go.

PR #2692, opened 2017-11-02
remove unnecessary path to file from docs

djstojadin:
When require is given the path of a folder, node.js will look for an index.js file in that folder so no need for /index


timdorr:
Thanks!

PR #2696, opened 2017-11-04
Updates to build for babel-preset-env

hmillison:
Per #1342, this updates the redux build to use babel-preset-env to simplify things.

Updated the rollup config to use babel's `external-helpers` plugin per the projects recommendation for babel configuration (https://github.com/rollup/rollup-plugin-babel#configuring-babel)

I did a quick test in IE11 with a trivial create-react-app npm linked with this branch of redux and did not see anything issues 💯


timdorr:
Found some improvements in bundle sizes in rollup/rollup-plugin-babel#100. While we have to wait for an upstream fix, it's at least got a workaround for now.

The build failure appears to be something else going on (not 100% sure of the cause, but it finally reproduced itself locally, so I can fix it). That's not your fault, so this looks good to go. Thanks for the help!

PR #2697, opened 2017-11-05
(feature) add delete todos functionality

Zadvornyi:


timdorr:
Thanks, but this has been suggested a few times. We're looking to keep the examples in sync with the documentation, which doesn't include anything about deletion. So, for now, we're going to keep this functionality out of the example.

PR #2699, opened 2017-11-06
Use Object Literal Shorthand Syntax

prathamesh7pute:
Looks like it's a miss as its already done at other instances like productId


markerikson:
Thanks!

PR #2700, opened 2017-11-07
Updated typescript to 2.1v. Improved type for combineReducers function.

in19farkt:
Hi everybody. Here is my proposal: let's update Typescript version from 1.8 to 2.1 and use the strength of awesome `keyof` feature, that let us restrict fields of our redux state. 

The current problem is: we can pass almost everything to `combineReducers` and Typescript will not complain about anything. It will lead to poor user experience and make us to check the correctness of the state ourselves. Although there is a way to tell Typescript how check all fields that we pass to  `combineReducers` — `keyof`. It is supported only from Typescript 2.1, so I have updated its version, rewrote the typings and updated the tests for `compose`. The examples below demonstrate the idea quite well:

```typescript
// We have few fields, let it be just string for example:
type IFirstField = 'first';
type ISecondField = 'second';
type IThirdField = 'third';

// Our redux state contains all those fields:
interface IData {
  first: IFirstField;
  second: ISecondField;
  third: IThirdField;
}

// Current typing should be:
type Reducer<S> = (state: S, action: AnyAction) => S;
interface ReducersMapObject {
  [key: string]: Reducer<any>;
}
type combineReducers = <S>(reducers: ReducersMapObject) => Reducer<S>;

// The problem is shown below
// typescript just swallow error and tell us nothing:
const invalidReducer2 = combineReducers<IData>({
  first: (state: ISecondField, action: AnyAction) => state,  // notice we have used incorrect fields
  second: (state: IFirstField, action: AnyAction) => state,  // here should be ISecondField
});
const invalidReducer1 = combineReducers<IData>({}); // we could even pass nothing!

// Proposed typings with "keyof"
// Now all state fields are bound to exactly fields
type Reducer<S> = (state: S, action: AnyAction) => S;
type ReducersMapObject<S> = {
  [key in keyof S]: Reducer<S[key]>;
}
type combineReducers = <S>(reducers: ReducersMapObject<S>) => Reducer<S>;

// Now Typescript complains about miscorrect types:
const invalidReducer2 = combineReducers<IData>({
  first: (state: ISecondField, action: AnyAction) => state,  // notice, here is ISecondField
  second: (state: IFirstField, action: AnyAction) => state,
  third: (state: IThirdField, action: AnyAction) => state,  
});  // Yeah, we'll see error here
const invalidReducer1 = combineReducers<IData>({});  //  Eventually Typescript will cry!

// But with correct fields Typescript is happy :)
const validReducer = combineReducers<IData>({
  first: (state: IFirstField, action: AnyAction) => state,
  second: (state: ISecondField, action: AnyAction) => state,
  third: (state: IThirdField, action: AnyAction) => state,
});
```

Hope it will be usefull pull request for community and looking forward for your feedback. Thank you!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
+}
--- a/index.d.ts
+++ b/index.d.ts
@@ -58,8 +58,8 @@ export type Reducer<S> = (state: S, action: AnyAction) => S;
 /**
  * Object whose values correspond to different reducer functions.
  */
-export interface ReducersMapObject {
-  [key: string]: Reducer<any>;
+export type ReducersMapObject<S> = {
+  [key in keyof S]: Reducer<S[key]>;
 }
 
 /**
@@ -80,7 +80,7 @@ export interface ReducersMapObject {
  * @returns A reducer function that invokes every reducer inside the passed
  *   object, and builds a state object with the same shape.
  */
-export function combineReducers<S>(reducers: ReducersMapObject): Reducer<S>;
+export function combineReducers<S>(reducers: ReducersMapObject<S>): Reducer<S>;
 
 
 /* store */
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,6 +50,11 @@
         }
       }
     },
+    "agentkeepalive": {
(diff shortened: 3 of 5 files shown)


in19farkt:
Ah. Just seen 8033325 and #2563 - please ignore!

PR #2705, opened 2017-11-08
Simplify the definition of reducers

vkatsikaros:
Instead of defining reducers by what actions are not, define reducers by what they are. A reminder of what actions are is kept in the text, for education purposes.

I am not a react/redux expert so feel free to correct any errors in the new text.

However, wording is subjective and a matter of taste. So, I'll understand if you won't consider merging the change.

### DIFF
--- a/docs/basics/Reducers.md
+++ b/docs/basics/Reducers.md
@@ -1,6 +1,6 @@
 # Reducers
 
-[Actions](./Actions.md) describe the fact that *something happened*, but don't specify how the application's state changes in response. This is the job of reducers.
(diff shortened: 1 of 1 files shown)


timdorr:
LGTM. Since this is relatively important text, I'd rather get a second 👍 before merging.

naw:
I like the direction and intent of the change. I like the first sentence. However, removing the words after "how" from the last sentence changes the meaning of the sentence. It begs the question "how what?", and the mind will fill in "how something happened" rather than "how the state changes".

vkatsikaros:
Thanks @timdorr and @naw!

I pushed a change to address https://github.com/reactjs/redux/pull/2705#issuecomment-342996961

timdorr:
@markerikson Any thoughts?

timdorr:
LGTM, so I'm going to merge it in. Thanks!

PR #2708, opened 2017-11-10
update TS definition - made ReducersMapObject generic

Dante-101:
More accurate TS definitions. Helps in highlighting issues with the child reducers during compile time.


timdorr:
This is already done on the next branch: https://github.com/reactjs/redux/tree/next

Dante-101:
Do we know a release date for the next branch?

PR #2711, opened 2017-11-12
Get rid of console warnings in TodoApp of docs

ACollectionOfAtoms:
Add default case for `getVisibleTodos` and remove `#` from href in `Link.js`


timdorr:
Thanks!

PR #2716, opened 2017-11-17
todos list

KairatKempirbaev:
state.todos should be used since it is an array of  todos not just state itself


KairatKempirbaev:
Pardon, haven't noticed that you pass .
todos(state.todos, action)
Calling every state just state confused. Nothing good happens at 2 am. Sorry

PR #2719, opened 2017-11-21
adds shared route module

PeteSchuster:
Move routes to a shared module to prevent redundancy.


timdorr:
Thanks, but this adds another extra <div> in there. The duplication is fine as far as I'm concerned.

PeteSchuster:
yea, my other thought was to have one Root.js file, and return null for `<DevTools />` if node.env is prod?

If a application has 50+ routes, you'd be hard pressed to keep maintaining 100 Routes between two files vs 50 routes in one file.

PR #2722, opened 2017-11-22
Rollup update

Andarist:
`rollup@0.50` doesn't require an `externalHelpersWhitelist`. It's capable of removing `asyncGenerator` babel helper on its own.


timdorr:
Thanks

PR #2723, opened 2017-11-22
Call makeMapStateToProps

mohamed-seada-1994:
call makeMapStateToProps to get reference to mapStateToProps


timdorr:
This is correct as-is. It's the factory pattern.

PR #2726, opened 2017-11-24
enhance replaceReducer with optional nextState parameter

budarin:
It helps to recover app state which is rendered on server as sync app, but client is built as code-splitted app.

Each dynamic module injects its reducers when it has been loaded.
But at the moment clients store is initialized with only root reducers from root module and store state lost most of the data from the server.

**_nextState_** parameters helps to recover server state for those code-splitted app at the first render


timdorr:
This was last tried with #2551. You can do this on your own: https://gist.github.com/cvbuelow/ea7bd4deb7824fdb65d81cd2b8c60f5e

This doesn't belong in the core library, because it leads to abuse and buggy behavior.

budarin:
@timdorr thanks for the link.

But why it leads to abuse and buggy behavior?
It's well know design pattern - code split for routes.
In case of a large application with a very big state there's no logic to keep all reducers in one place.

PR #2728, opened 2017-11-25
Simplify composer

ivan-nikitovic:


TrySound:
@ivan-nikitovic It's less readable. Also it makes the first case without functions slower. And arrow function directly exported with default will be less debuggable.

ivan-nikitovic:
@TrySound you are right, function is probably better choice at least from standpoint of consistency with other exports. 

Can you please explain me how this makes first case without functions slower, I don't get that?

TrySound:
@ivan-nikitovic In that case your code allocates two functions: first and iteration.

ivan-nikitovic:
@johanneslumpe I thought that @TrySound talks about result of the compose function, not on the compose function itself, because result of both function is completely the same in each case.

In current(_master_) version of `compose` function check `funcs.length === 1` is redundant with default behavior of `reduce` function which will return first element as a result if gets called with single element array - directly without calling iterator function. That actually inspired this change.

Of course, with the check, single function case works better, as you mentioned, because there is no `reduce` function call or function allocation for `iterator` (_function exits earlier_), but it's completely unnecessary for more realistic use case when you are composing two or more functions. In that case additional `funcs.length === 1` check is waste from performance standpoint. 

BTW, @TrySound , @johanneslumpe thank you for the answer. 😃

PR #2731, opened 2017-11-27
Make clear where reducers act on a state slice instead of on whole state

hjvanranden:
Make clear where reducers act on a state slice instead of on the whole state, by renaming their parameter 'state' to 'stateSlice'.

Motivation: My colleague and I were independently confused on the content of parameter 'state' in the reducers that act on a state slice.

### DIFF
--- a/docs/basics/Reducers.md
+++ b/docs/basics/Reducers.md
@@ -210,18 +210,18 @@ function todoApp(state = initialState, action) {
 Is there a way to make it easier to comprehend? It seems like `todos` and `visibilityFilter` are updated completely independently. Sometimes state fields depend on one another and more consideration is required, but in our case we can easily split updating `todos` into a separate function:
 
 ```js
-function todos(state = [], action) {
+function todos(stateSlice = [], action) {
   switch (action.type) {
     case ADD_TODO:
       return [
-        ...state,
+        ...stateSlice,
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

markerikson:
I'm not entirely sure I agree with the naming change here.

Part of the idea of reducers like this is that as far as the function is considered, this _is_ the entire state.  The fact that it's just one slice is a concern of the parent reducer function, not this one.  In fact, you could reuse the function in another context, or even as a root reducer.

docs/basics/Reducers.md
```diff
@@ -396,27 +396,27 @@ import {
 } from './actions'
 const { SHOW_ALL } = VisibilityFilters
 
-function visibilityFilter(state = SHOW_ALL, action) {
+function visibilityFilter(stateSlice = SHOW_ALL, action) {
   switch (action.type) {
     case SET_VISIBILITY_FILTER:
-      return action.filter
+      return stateSlice.filter
```
naw: This change here is incorrect. cc @timdorr

naw:
Perhaps call it `visibilityState` in the visibility reducer and `todosState` in the todos reducer? This might address both of your concerns, @markerikson and @hjvanranden

Issue #2735, opened 2017-11-28
update docs to use babel-preset-env


hozefaj:
**What is the current behavior?**
The current docs have example of using `es2015` as the babel preset. 

**What is the expected behavior?**
Better form would be to use [`babel-preset-env`](https://github.com/babel/babel/tree/master/packages/babel-preset-env) rather. Since this can confuse new folks to on whether to use `es2015`...`es2016`....or what?
Rather have the docs updated to use `env` instead.

Happy to send a PR, if this change is okayed.

markerikson:
Yes, that would be a good idea - we'd definitely take a PR to update that.

hozefaj:
Cool. Will send one later today.

hozefaj:
The Contributing guide says to [publish](https://github.com/reactjs/redux/blob/master/CONTRIBUTING.md#publishing-the-docs) & [cleaning](https://github.com/reactjs/redux/blob/master/CONTRIBUTING.md#cleaning-the-docs) the doc, however I would assume that you would that to happen only after the PR's are merged.

@markerikson Can you please let me if I need execute those 2 commands or they are meant to be done at a time when bunch of updates are released.

PR #2736, opened 2017-11-28
update docs to use babel-preset-env instead of es2015

hozefaj:
With reference to issue #2735 updating docs to use babel-preset-env instead of es2015.


timdorr:
Fixed a couple small issues. Thanks!

Issue #2742, opened 2017-11-29
examples/async code has reddit as variable instead of subreddit, which the variables refer to


yasuf:
Has this been discussed before?

timdorr:
Can you point to the line of code that's wrong? I don't see any issues.

yasuf:
The file would be https://github.com/reactjs/redux/tree/master/examples/async/src/actions/index.js.

Let me elaborate about this @timdorr:

We're fetching posts for an specific "Subreddit" not just reddit posts and the actions refer to posts as belonging to a "reddit" which is confusing to read. 

For example:
www.reddit.com/r/askreddit has posts every day and the _subreddit_ is **"AskReddit"**.
We're not fetching posts for "reddits" but for "subreddits"

For example in https://github.com/reactjs/redux/blob/master/examples/async/src/actions/index.js#L7 we the action generator for an action called `SELECT_REDDIT` but we're not selecting a reddit but a SUBREDDIT.

https://en.oxforddictionaries.com/definition/subreddit

And this erroneous concept of "reddits" is in the entire example not just in that particular actions file

yasuf:
I can make a PR, I already have the code ready, I just didn't want to open it if there was a better reasoning for calling them "reddits". Sorry for not elaborating better the first time

PR #2743, opened 2017-11-29
Rename variable names to subreddit instead of reddit

yasuf:
See https://github.com/reactjs/redux/issues/2742.


timdorr:
Oh, I see. It's nomenclature, not a mis-used variable. Interestingly, this was already done in the docs: https://github.com/reactjs/redux/blob/master/docs/advanced/ExampleRedditAPI.md

Thanks!

Issue #2628, opened 2017-09-27
Standards-compliant ES module build


magnonellie:
**Do you want to request a *feature* or report a *bug*?** 

Feature

**What is the current behavior?**

Redux doesn't have a standards compliant es-module build that allows it to run in the browser as a native module without compilation from the user.

**What is the expected behavior?**

The build in the `es` directory should probably be made standards compliant by either rewriting the imports in the source or using rollup (which doesn't fix node_modules imports AFAIK) to produce a build of Redux that runs directly in the browser with a `<script type="module">` tag.

**What's missing?**

The problem right now is that browsers supporting es-modules doesn't support importing with loosely defined paths such as `./compose` or `lodash/isPlainObject`. Instead, the full actual path has to be defined with extensions and everything.

timdorr:
Unfortunately, this stuff is so new, no one else in the ecosystem is doing this. Even if we were to update with correct URLs, our dependent packages wouldn't and would break things.

For the moment, you can use unpkg's module feature to accomplish this: https://unpkg.com/redux@3.7.2/es/index.js?module

magnonellie:
That's a shame since I just noticed that the normal dist version doesn't work either when importing from a module. Fails with this:
```
Uncaught TypeError: Cannot read property 'Redux' of undefined
    at freeGlobal (redux.js:4)
    at redux.js:5
```

PR #2744, opened 2017-11-30
feat(es-modules): native es-module support

pixelass:
adds changes required to run redux via native es-modules in supported browsers

As requested in #2628 

`symbol-ovservable@1.0.0` added support. 

This PR simply ensures correct paths and file extensions, therefore allows to be imported via native es-modules in the browser.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -1,7 +1,7 @@
-import $$observable from 'symbol-observable'
+import $$observable from '../node_modules/symbol-observable/es/index.js'
 
-import ActionTypes from './utils/actionTypes'
-import isPlainObject from './utils/isPlainObject'
+import ActionTypes from './utils/actionTypes.js'
+import isPlainObject from './utils/isPlainObject.js'
 
 /**
  * Creates a Redux store that holds the state tree.
--- a/package-lock.json
+++ b/package-lock.json
@@ -11133,7 +11133,7 @@
       "integrity": "sha512-vmvP5y/oJIJmXKHY36PIjVeI/46Sny6BMBa7/ou2zsNz1PiqU/Gtcz1GujnHz5Qlxncv+J9VlWmttnshqFj3Kg==",
       "dev": true,
       "requires": {
-        "symbol-observable": "1.0.4"
+        "symbol-observable": "1.1.0"
       }
     },
     "safe-buffer": {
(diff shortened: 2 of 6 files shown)


src/createStore.js
```diff
@@ -1,7 +1,7 @@
-import $$observable from 'symbol-observable'
+import $$observable from '../node_modules/symbol-observable/es/index.js'
```
timdorr: This doesn't work on the web.

pixelass: why not?

symbol-observable 1.1.0 is compatible

timdorr: There isn't a `node_modules` folder on the web.

pixelass: Yes there is. This is how this feature is supposed to work.
You are supposed to upload node_modules that are imported.

This is the entire idea and reason this works in the first place.

Take a look at polymer, lodash etc. They follow the same patttern. 

I think you might have missed this feature in the latest browser releases

* https://hackernoon.com/native-es-modules-ready-for-prime-time-87c64d294d3c
* https://jakearchibald.com/2017/es-modules-in-browsers/
  * https://jakearchibald.com/2017/es-modules-in-browsers/#comment-3334824431
* more needed?

timdorr: No, that makes the assumption you're using npm or nodejs, both of which don't exist in the browser context. If I were to manually download the code here, independent of npm, it wouldn't work with this form. 

You're also dependant on the exact filesystem layout of the symbol-observable module. They can change that at any time (we're doing the same thing in redux 4.0), so you're relying on a private API essentially.

This is what I was getting at in the last issue where this stuff isn't ready for prime time.

pixelass:
I am testing this in a project and it works as expected. I might have to make some minor changes but it generally works in the browser.

It's really just a matter of ensuring the right path to node_modules

redux-browser

pixelass:
Why did you close this issue? Can you please explain that in other words than "This won't work".

src/createStore.js
pixelass: Thank you for the explanation. At least I understand your concerns now.

If users download the code here they'd probably use the dist version, right?  
How would it work without npm if they download it? 

The PR simply adds file extensions and the full path to node_modules. It doesn't even cause a breaking change IMHO.

Changes in symbol-observable should not matter since they use semver and the version is fixed in the package.json. I don't understand that concern at all, sorry. If you deliver the wrong version then it's not an issue with the package/module but a mistake the developer made. 

I see that I made an error with the relative path to `node_modules`. It cannot work like this  but I think if the path were absolute `/node_modules` which is the same in every npm project it should work.

I've tested it in a project using polymer, polymer-redux and redux, The setup is straightforward and everything is vanilla, no issues whatsoever. I'll gladly test it in one of my react projects built with webpack and babel too.

I also am willing to provide several test cases or forks of well known products built on redux to prove the compatibility. 

I'd just really love to see this feature in the near future.

Thank you for understanding.

timdorr:
Because it makes assumptions about how people will build and use this project. We cannot assume all users have npm and node in their toolchains. 

It also won't work with our bundle format changes in 4.0. We're packing everything as [a single bundle](https://unpkg.com/redux@4.0.0-beta.1/es/redux.js), so the files won't exist to be loaded.

timdorr:
To be clear, the issue I have isn't with your implementation or our support of ESM, it's with the ecosystem in general. We've only *just* gotten modules in browsers. The toolchain ecosystem needs time to catch up. It's not there yet, but it will be soon. Once it is, we can work on it at this level. Right now, it's just too premature.

pixelass:
@timdorr full respect for your precaution.

Polymer and lodash (..etc) already work like this. Neither project has dependencies so therefore it might not be the same, I see that. The only dependencies redux has already made this step which is the only reason I thought it would make sense to try pushing this feature.

I guess for the time being I'll use my fork and try to mirror all new features back into my fork, to make it easier to switch back when redux adds this feature.

Thank you

PR #2754, opened 2017-12-12
Initialize default state instead of using bad style

ailadson:
I was reading through docs with my friend. We both found the example visually confusing, which is important because this is the documentation.


timdorr:
I disagree. Having to create and maintain another variable, however minor, is arguably more complex.

markerikson:
Eh, fwiw, I personally write my initial state variables separately.

PR #2757, opened 2017-12-15
Update TodoMVC example to be more in line with other examples

tomipaul:
- Store visibilityFilter in the state
- Update presentational components
- Add container components
- Update reducer code to use composition
- Add Reselect


timdorr:
Sorry I haven't had a chance to look at this. It's fairly extensive, so it may take some time. If others want to make some review comments, go right ahead. I'm not the sole gate keeper for contributions to this project 😄

examples/todomvc/src/components/MainSection.js
```diff
@@ -1,82 +1,40 @@
...
-    if (todos.length > 0) {
-      return (
+import VisibleTodoList from '../containers/VisibleTodoList'
+
+const MainSection = (props) =>  {
+  const { todosCount, completedCount, actions } = props
+  return (
+    <section className="main">
+      {
+        (todosCount > 0) ?
```
raunofreiberg: Ditto for the ternary described above.

examples/todomvc/src/components/MainSection.js
```diff
@@ -1,82 +1,40 @@
...
-  }
-
-  renderToggleAll(completedCount) {
-    const { todos, actions } = this.props
-    if (todos.length > 0) {
-      return (
+import VisibleTodoList from '../containers/VisibleTodoList'
+
+const MainSection = (props) =>  {
+  const { todosCount, completedCount, actions } = props
```
raunofreiberg: I would just destructure these inside the parameters of the function and have an implicit return.

examples/todomvc/src/components/MainSection.js
```diff
@@ -1,82 +1,40 @@
...
+            className="toggle-all"
+            type="checkbox"
+            checked={completedCount === todosCount}
+          />
+          <label onClick={actions.completeAllTodos}/>
+        </span> : null
+      }
+      <VisibleTodoList />
+      {
+        (todosCount > 0) ?
```
raunofreiberg: Ditto.

examples/todomvc/src/components/Footer.js
```diff
@@ -9,64 +9,40 @@ const FILTER_TITLES = {
...
+        {Object.keys(FILTER_TITLES).map(filter =>
+          <li key={filter}>
+            <FilterLink filter={filter}>
+              {FILTER_TITLES[filter]}
+            </FilterLink>
+          </li>
+        )}
+      </ul>
+      {
+        (completedCount > 0) ?
```
raunofreiberg: How about doing `!!completedCount && <Component ... />` instead of the ternary? Would also get rid of the `null` being returned if the condition is not met.

examples/todomvc/src/actions/index.js
```diff
@@ -4,5 +4,6 @@ export const addTodo = text => ({ type: types.ADD_TODO, text })
 export const deleteTodo = id => ({ type: types.DELETE_TODO, id })
 export const editTodo = (id, text) => ({ type: types.EDIT_TODO, id, text })
 export const completeTodo = id => ({ type: types.COMPLETE_TODO, id })
-export const completeAll = () => ({ type: types.COMPLETE_ALL })
+export const completeAllTodos = () => ({ type: types.COMPLETE_ALL_TODOS })
```
raunofreiberg: Why the rename? Seems like the context is obvious here regardless.

examples/todomvc/src/components/Link.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import classnames from 'classnames'
+
+const Link = ({ active, children, setFilter }) => {
+  return (
```
raunofreiberg: Use an implicit return. No need for the block.

examples/todomvc/src/containers/FilterLink.js
```diff
@@ -0,0 +1,20 @@
...
+    dispatch(setVisibilityFilter(ownProps.filter))
+  }
+})
+
+const FilterLink = connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(Link)
+
+export default FilterLink
```
raunofreiberg: I would prefer doing

```
export default connect(
   mapStateToProps,
   mapDispatchToProps
)(Link)
```

But I'm not sure if this is a convention or not, I just don't see the point in the intermediary variable that is disposable anyway.

examples/todomvc/src/containers/Header.js
```diff
@@ -0,0 +1,26 @@
...
+      placeholder="What needs to be done?"
+    />
+  </header>
+)
+
+Header.propTypes = {
+  addTodo: PropTypes.func.isRequired
+}
+
+export default connect(null, { addTodo })(Header)
```
raunofreiberg: Even more reasons to use the convention described above then if this is already in use here.

examples/todomvc/src/containers/MainSection.js
```diff
@@ -0,0 +1,25 @@
+import { connect } from 'react-redux'
+import * as TodoActions from '../actions'
+import { bindActionCreators } from 'redux'
+import MainSection from '../components/MainSection'
+import { getCompletedTodoCount } from '../selectors'
+
+
+const mapStateToProps = (state) => {
```
raunofreiberg: Seems like in some cases you're using an implicit return for an object and in some cases an explicit one. I would recommend keeping it consistent and using an implicit return in all cases then, unless you require a block for extra statements.

examples/todomvc/src/actions/index.js
tomipaul: Yes, it is obvious. I just feel having `Todos` attached to the `completeAll` will make it much more explicit and consistent with the others.

examples/todomvc/src/components/MainSection.js
tomipaul: Sure, thanks

timdorr:
@tomipaul Any chance you can address those comments and also resolve the merge conflicts?

tomipaul:
_Oops_ Somehow I missed your comment @timdorr 

Back when the comments were made, I started addressing them before some other work took me off.
I will just complete it and fix the conflict
In the next 24 hours, the PR should look good.

timdorr:
That should be fixed in master now too. Waiting on a build to confirm that and then I'll merge this in. Thanks for keeping up with this!

PR #2759, opened 2017-12-16
Clarify references to property values

gitname:
I rephrased the statements in order to emphasize that each of the "different reducing functions" is a *property* value, not an *object* value.

Example of an *object* value. I felt confused when I read the plural version (i.e. "an object whose value**s**...").

```js
var object = objectValue;
```

Examples of *property* values:

```js
var object = objectValue;
object.property1 = propertyValue1;
object.property2 = propertyValue2;
```

### DIFF
--- a/docs/api/combineReducers.md
+++ b/docs/api/combineReducers.md
@@ -2,7 +2,7 @@
 
 As your app grows more complex, you'll want to split your [reducing function](../Glossary.md#reducer) into separate functions, each managing independent parts of the [state](../Glossary.md#state).
 
-The `combineReducers` helper function turns an object whose values are different reducing functions into a single reducing function you can pass to [`createStore`](createStore.md).
+The `combineReducers` helper function turns an object whose property values are different reducing functions into a single reducing function you can pass to [`createStore`](createStore.md).
 
 The resulting reducer calls every child reducer, and gathers their results into a single state object.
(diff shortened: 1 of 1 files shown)


markerikson:
I'm not sure I understand the distinction you're making here.

Terminology-wise, for a JS object like `{a : 42, b : "blah"}`, `a` and `b` are "keys", and `42` and `"blah"` are "values".  I don't think I've heard the phrase "property values" before, and I don't see how that's any clearer.

gitname:
When I think of an object in JavaScript, I think of it as a collection of properties, where each **property** has a key and a value. I think of each value as belonging to a property. That's why I referred to the value as a "property value."

But, since each **property** belongs to an object anyway, I can see how saying "property value" is unnecessarily specific (at least in colloquial usage, as the [ECMAScript 2015 spec](https://www.ecma-international.org/ecma-262/6.0/#sec-object-type "ECMAScript 2016 spec - Section 6.1.7 - The Object Type") refers to the thing you and I called "keys" as "key *values*" -- but I promise I didn't know that\* when I submitted the PR).

Anyway, with your explanation in mind, the original phrasing makes more sense to me than it first did. Thanks for explaining your rationale behind closing the PR.

---

\* Meaning, I wasn't trying to be pedantic.

PR #2763, opened 2017-12-18
Better middleware order in TS test

krawaller:
Not that it matters much, but the `logger` and `thunk` middlewares in the TS test code are in the "wrong" order. This PR switches them around to follow real-world usage.


timdorr:
Thanks!

PR #2766, opened 2017-12-18
A11y Fixes

knittingcodemonkey:
- Update color scheme to darker versions with more contrast
- Adjusted markdown to comply with header tag hierarchy
- Add alt tag and aria-label to logo


timdorr:
LGTM. Thanks!

Issue #2769, opened 2017-12-20
Product quantity are not displayed into cart on shopping-cart example


linconnu:
**Do you want to request a *feature* or report a *bug*?**  
Report a *bug* in shopping-cart example.

**What is the current behavior?**
Quantity of product added into cart are not displayed, only quantity in available products are.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar.**
Just click on "Add to cart" button. In cart, no quantity are provided.

**What is the expected behavior?**
According to the code, we should see the quantity of products added into the cart.

PR #2770, opened 2017-12-20
ProductItem and Cart gives different prop for product quantity

linconnu:
https://github.com/reactjs/redux/issues/2769


timdorr:
Nice catch. Thanks!

Issue #2740, opened 2017-11-29
Incorrect typescript typings for middlewares


nenadalm:
Hi. It looks like typings don't work nicely for middlewares because I can't use type of state in it, because it's type is `S` (see the comment in code):

```typescript
import * as redux from 'redux';

interface State {
    app: {
        title: string;
    }
}

const initialState: State = {
    app: {
        title: 'awesome app',
    }
}

function createMiddleware(): redux.Middleware {
    return function<S>(api: redux.MiddlewareAPI<S>) {
        return function (next: redux.Dispatch<S>): redux.Dispatch<S> {
            return function (action) {
                const state = api.getState();
                // error is shown here because `app` does not exist on type `S`
                console.log('app title: ', state.app.title);

                next(action);
            };
        };
    }
}

function appReducer(state: State, action) {
    return state;
}

redux.createStore(
    appReducer,
    initialState,
    redux.applyMiddleware(createMiddleware())
)
```

My colleague advised me to extend `redux.Middleware` but then typings don't work either because following code should fail to compile, but it compiles just fine (see the comment in code):
```typescript
import * as redux from 'redux';

interface State {
    app: {
        title: string;
    }
}

export interface Middleware extends redux.Middleware {
    (api: redux.MiddlewareAPI<State>): (next: redux.Dispatch<State>) => redux.Dispatch<State>;
}

function createMiddleware(): Middleware {
    return function(api: redux.MiddlewareAPI<State>) {
        return function (next: redux.Dispatch<State>): redux.Dispatch<State> {
            return function (action) {
                const state = api.getState();
                // `state` is of type `State2` which doesn't have `app` property, but typescript happily compiles this
                console.log('app title: ', state.app.title);

                next(action);
            };
        };
    }
}

interface State2 {
    wtf: string;
}

const initialState2: State2 = {
    wtf: 'ttf'
}

function appReducer(state: State2, action) {
    return state;
}

redux.createStore(
    appReducer,
    initialState2,
    redux.applyMiddleware(createMiddleware())
)
```

environment:
- redux: 3.7.2
- typescript: 2.6.2

so `applyMiddleware` should probably take generic parameter for state and pass it to the middlewares?
```typescript
// current definition
export function applyMiddleware(...middlewares: Middleware[]): GenericStoreEnhancer;

// should be probably defined something like:
export function applyMiddleware<S>(...middlewares: Middleware<S>[]): GenericStoreEnhancer;
```

timdorr:
[...]

PR #2773, opened 2017-12-21
Update TypeScript typings

aikoven:
This PR continues work from https://github.com/reactjs/redux/pull/2563 and fixes https://github.com/reactjs/redux/issues/1648, https://github.com/reactjs/redux/issues/2602 and https://github.com/reactjs/redux/issues/2740.

Tests are rewritten to cover different supported approaches. All TS strict checks are enabled in tests.

## Reducers

* Reducers support simple definition where action shape is not type-checked to aid the transition from JS to TS. 
* Discriminated union for actions is still supported as well.
* A test is added for writing reducers with type guards.
* Typings for `combineReducers` are updated to allow strict combination using discriminated unions, as well as combining custom and third-party reducers.

See [tests](https://github.com/aikoven/redux/blob/typescript-typings/test/typescript/reducers.ts).

## Dispatch

* Default `Dispatch` type only accepts plain actions. It also supports restricting action types using discriminated unions.
* Extra signatures are added using intersection type e.g. `Dispatch & PromiseDispatch`.

See [tests](https://github.com/aikoven/redux/blob/typescript-typings/test/typescript/dispatch.ts) and [tests for injected dispatch](https://github.com/aikoven/redux/blob/typescript-typings/test/typescript/injectedDispatch.ts) e.g. `connect` from `react-redux`.

## Middleware

* Middleware type supports declaring extra Dispatch signatures which are added to the Store type by `applyMiddleware` enhancer.
* Middleware type supports restricting the state type.

See [tests](https://github.com/aikoven/redux/blob/typescript-typings/test/typescript/middleware.ts).

## Store Enhancers

* Store enhancer type supports extending Store type. Particularly, Dispatch signature is extending using this mechanism. It also supports for adding custom methods.
* Support for enhancers extending the state type.

See [tests](https://github.com/aikoven/redux/blob/typescript-typings/test/typescript/enhancers.ts).

cc. @pelotom @KingHenne


index.d.ts
```diff
@@ -46,7 +56,7 @@ export interface Action<T = any> {
  * @template S The type of state consumed and produced by this reducer.
  * @template A The type of actions the reducer can potentially respond to.
  */
-export type Reducer<S = any, A extends Action = Action> = (state: S | undefined, action: A) => S;
+export type Reducer<S = any, A extends Action = AnyAction> = (state: S | undefined, action: A) => S;
```
ikokostya: Actually, this is not needed, because any interface with additional properties extends `Action` interface.

aikoven: The default type for `A` is needed so that any reducer acting on the state `State` would be assignable to `Reducer<State>` (with omitted action type parameter).

You may want to change it back to `Action` and check out the tests that break.

ikokostya: Ah, I see now. Thanks.

timdorr:
Since no one has come in and yelled "Holy balls, this is terrible!", I'm going to assume this is at least somewhat of an improvement and merge it in. 

I swear, I'll learn TS at some point...

aikoven:
@jayphelps Old `MiddlewareAPI` of Redux `v3` didn't have `Dispatch` parameter:

```ts
export interface MiddlewareAPI<S> {
  dispatch: Dispatch<S>;
  getState(): S;
}
```

The new one has a default for it:

```ts
export interface MiddlewareAPI<D extends Dispatch = Dispatch, S = any> {
  dispatch: D;
  getState(): S;
}
```

If you don't need a custom dispatch signature, you can just use the built-in `Dispatch` type: `MiddlewareAPI<Dispatch, MyState>`, or even have it inferred for you, see [the test](https://github.com/reduxjs/redux/blob/v4.0.0/test/typescript/middleware.ts#L87-L101) for middleware that expects exact state type.

> It seems like it would have been a breaking change for the average TS person upgrading.

That's why we have bumped a major version 🙂

PR #2774, opened 2017-12-25
Restructure ecosystem page

### LINKED ISSUE #2588: Restructure "Ecosystem" docs page
The "Ecosystem" page is currently a bit of a scattershot list ([current version for reference](https://github.com/reactjs/redux/blob/ab5c9593bc304880c3089fc0bfcc2c3a28fd21d3/docs/introduction/Ecosystem.md)).  There's a mixture of random tutorial links, translations, bindings, middleware, and other addons.  

The page describes itself as "staff picks", and also that "the Redux staff has vetted them personally".  Really, at this point that's not exactly true.  We've had lots of people submit their own libs over time, and while someone did presumably approve each merged PR, I'm not sure we really "vetted" things per se.

On a similar note, there's a great quote from Andrew and Dan at https://github.com/reactjs/redux/issues/215#issuecomment-118639342:

> As @gaearon said once, (I can't remember where... probably Slack) we're aiming to be like the Koa of Flux libraries. Eventually, once the community is more mature, the plan is to maintain a collection of "blessed" plugins and extensions, possibly under a reduxjs GitHub organization.

I don't think we're going to actually maintain any additional libs at this point, but we can certainly highlight ones that are widely used or specifically interesting in some better ways.

I talked with Dan at ReactRally about revamping this page.  I don't have an absolutely specific plan in mind yet, but some rough thoughts:

- Move any tutorials to a separate page.  In fact, we really ought to have a separate "Learning Resources" docs page, and put stuff in there instead.
- Review all existing links for relevance
- Stop advertising "Awesome-Redux" at the top of the page.  It's okay to leave it in a list of "Useful Resources", but in my (very biased) opinion, it's a link dump that's insufficiently organized.  (FWIW, I think Dan actually suggested that change first, and Sindre Sorhus felt it was poor enough that he didn't accept it into his "awesome-awesome" list.)
- Rebuild the list of "blessed" addons from scratch:
  - Come up with a list of use cases or categories
  - Specifically pick out 3-5 entries for each use case that we want to highlight
  - For each one, give a description of what it is, why you might want to use it (or pros/cons), and hopefully even give a 10-line snippet of what using it looks like

markerikson:
Splits the "Ecosystem" docs page into "Learning Resources" and "Ecosystem", and updates both pages with content from my articles list and addons list.

Fixes #2588 .

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -6,6 +6,7 @@
   * [Core Concepts](/docs/introduction/CoreConcepts.md)
   * [Three Principles](/docs/introduction/ThreePrinciples.md)
   * [Prior Art](/docs/introduction/PriorArt.md)
+  * [Learning Resources](/docs/introduction/LearningResources.md)
   * [Ecosystem](/docs/introduction/Ecosystem.md)
   * [Examples](/docs/introduction/Examples.md)
 * [Basics](/docs/basics/README.md)
--- a/docs/introduction/Ecosystem.md
+++ b/docs/introduction/Ecosystem.md
@@ -1,101 +1,807 @@
 # Ecosystem
 
-Redux is a tiny library, but its contracts and APIs are carefully chosen to spawn an ecosystem of tools and extensions.
+Redux is a tiny library, but its contracts and APIs are carefully chosen to spawn an ecosystem of tools and extensions, and the community has created a wide variety of helpful addons, libraries, and tools.  You don't need to use any of these addons to use Redux, but they can help make it easier to implement features and solve problems in your application.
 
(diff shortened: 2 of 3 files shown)


ishankbahl:
Hey @markerikson , should there be a section for recommended boilerplates too?

markerikson:
@ishankbahl : maybe, maybe not.  The general stance of the React and Redux teams is that we don't recommend "boilerplates", especially for learners, because they're generally too complex and full of things that most people don't need.

ishankbahl:
React docs have a starter kits [section](https://reactjs.org/community/starter-kits.html) under tools.

markerikson:
Hmm.  Don't think I realized that page was there.

Really, though, that page completely works as a list of "Redux starters" too :)

alexkrolick:
On the topic of this PR:
Separating libraries and learning resources is a good idea 👍 

Separately:
The community starter kits (now "Other Starter Kits") section is pretty loosely maintained in the React docs. Dan just did a sweep to clear out some of the ones that don't fit the criteria well, but there isn't a "notability" requirement or anything like that and people add their own all the time. A page like this would almost be better as a wiki or community-upvoted list, since maintainers don't don't vet every addition and the community should feel free to add to it and rearrange it. A task for another day...

markerikson:
Since there's been no further comments, I'll merge this in and will hopefully get the docs republished soon.

PR #2775, opened 2017-12-25
Fix rm unexpectedKeyCache prototype property

stkevintan:
some protoytpe property(eg, `toString` `valueOf` ) may affect the unexpectedKeyCache, using `object.create(null)` instead of `{}` will be more robust


markerikson:
Hi.  Appreciate the PR, but I'm afraid I don't see the point of this change.  We don't do any assertions or checking on the type of `unexpectedKeyCache` - we're just using it as a lookup table.  Declaring it as a plain object literal should be fine.

stkevintan:
OK，thx for your reply.

PR #2776, opened 2017-12-27
Fixes breaking toggle example

vik-y:
Todo toggle was not working because `toggleTodo` action was never dispatched. Took a while to figure out as a react beginner. Has been fixed in the example given here: https://redux.js.org/docs/basics/ExampleTodoList.html


timdorr:
Actually, a feature of `connect`'s `mapDispatchToProps` in object form is that it will automatically bind `dispatch` to each action creator. So, this is already done for you (hence the test suite fails here).

PR #2777, opened 2017-12-30
Update setup instructions for testing with Enzyme

mpontus:
Since version 3, Enzyme now requires an adapter to be set up for the specific version of React.

This PR updates [Writing Tests](https://redux.js.org/docs/recipes/WritingTests.html) section of the docs to include the necessary setup instructions.


timdorr:
Thanks!

PR #2779, opened 2017-12-31
Spelling, capitalization, and whitespace changes

tbroadley:
This PR contains four commits with the following purposes:

1. Replace an instance of `presentional` with `presentational`
2. Remove trailing whitespace from two lines
3. Replace an instance of `Devtools` with `DevTools`
4. Replace one instance each of `emphasising`, `realising`, and `optimise` with their spellings in US English, in line with how similar words are spelled elsewhere in the documentation

I think that 1 is the only definite correction. If you don't think that 2, 3, or 4 are changes worth making, I'm happy to remove those commits from the PR.

### DIFF
--- a/docs/basics/UsageWithReact.md
+++ b/docs/basics/UsageWithReact.md
@@ -354,7 +354,7 @@ export default VisibleTodoList
 
 #### `containers/AddTodo.js`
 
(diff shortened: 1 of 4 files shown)


timdorr:
@markerikson Did you want to take a look here?

timdorr:
Not sure why I had to do it twice, but I resolved the conflicts with master.

markerikson:
Looks good.

PR #2784, opened 2018-01-07
Fix punctuation error in Reducers.md

georgemathieson:
Was reading the docs and noticed this probably should be a full stop, not a semi-colon.


markerikson:
Thanks!

PR #2786, opened 2018-01-08
extract the method of generate redux default action

bangwu:


timdorr:
I've decided against doing that for now. Thanks anyways!

PR #2788, opened 2018-01-09
Fix example: todos-with-undo

shuaibird:
* Add action filter to reducer undoableTodos

Details: In the example `todos-with-undo`, currently the reducer `undoableTodos` would respond to every single action, including the action(s) related to the reducer `visibilityFilter`, which should not be what to be expected.


timdorr:
Thanks!

PR #2789, opened 2018-01-10
Update doc: advanced/Middleware

shuaibird:
* Naïve implementation of the applyMiddleware api

Details: the `applyMiddleware` function returns a new `store` object, leaving the original one unmodified, which may cause confusions due to its inconsistence with the previous attempts.


timdorr:
Thanks, but this is interim code and not meant to actually be used. It's fine if it's not perfect; it's not supposed to be 😄

PR #2798, opened 2018-01-19
DOCS: Fix grammatical error in UsageWithReactRouter.html#reading-from-the-url

billfienberg:
Before: 
>This is because we are filtering from `<VisibleTodoList />`'s `mapStateToProps()` is still bound to the `state` and not to the URL.

After:
>This is because we are filtering from `<VisibleTodoList />`'s `mapStateToProps()`, which is still bound to the `state` and not to the URL.


timdorr:
Which thanks you!

Issue #2803, opened 2018-01-22
When run real-world example got error message


dews:
**What is the current behavior?**
When running real-world example got error message 'Unexpected token < in JSON at position 0'

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar.**
From offical example [codesandbox](https://codesandbox.io/s/github/reactjs/redux/tree/master/examples/real-world)
1. type calvin
2. click go
3. click loadmore, wait for loading complete.
4. click loadmore again. Should see an error message on the top.

HTTP Response Headers
`Link:<https://api.github.com/user/5280/starred?page=1>; rel="prev",
 <https://api.github.com/user/5280/starred?page=3>; rel="next", 
 <https://api.github.com/user/5280/starred?page=3>; rel="last", 
 <https://api.github.com/user/5280/starred?page=1>; rel="first"`

Because there is space in form of '<'.

**What is the expected behavior?**
Should not have an error message.

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**
"version": "4.0.0-beta.1"

PR #2804, opened 2018-01-22
Fix real-world example got error message

dews:
I open an issue [#2803](https://github.com/reactjs/redux/issues/2803)


timdorr:
Thanks!

PR #2806, opened 2018-01-23
update fetch error handling documentation

hwigmore:
errors are not separated into their own callback in the first `then`, all responses (success or error) must go through response.json()


timdorr:
The error response from the Reddit API may not be JSON if it's a serious error (usually 500). So, handling the error before processing the content is actually best in this case.

PR #2807, opened 2018-01-24
Add more expect examples to `it` block

iliastsangaris:


timdorr:
Thanks, but we're not testing varying inputs, just that it accepts multiple arguments. One example is sufficient.

Issue #2808, opened 2018-01-24
TypeScript: usage of `DeepPartial` for `preloadedState` is too deep


OliverJAsh:
**Do you want to request a *feature* or report a *bug*?**  

Bug

**What is the current behavior?**

Since https://github.com/reactjs/redux/pull/2679, we use `DeepPartial` for the type of `preloadedState`.

This means that, given a state tree, all properties at any level in the tree are optional.

**What is the expected behavior?**

Given a state tree, only properties whose state is computed by a reducer should be optional. This is because reducers have initial state, but the state within them may be required (unless those properties are explicitly marked as optional or they represent reducers).

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**

N/A

---

/cc @aikoven @dawnmist

I'm not quite sure how we can write the type definition to match the expected behaviour. Any ideas?

Related https://github.com/reactjs/redux/issues/2552

aikoven:
I'm not sure it's possible without some kind of black magic. 
But it might be possible to solve with the upcoming TS 2.8 feature — [Conditional Types](https://github.com/Microsoft/TypeScript/pull/21316).

The crucial part here is `combineReducers`. The `preloadedState` type should be the same as the type of the first argument of the reducer passed to `createStore`. It should also be a subtype of `DeepPartial<S>`. 

Given some type `T` where `DeepPartial<S> <= T <= S` we should be able to produce `Reducer` type that accepts `T` as its first argument, and returns `S`. `combineReducers` should preserve it when mapping over `ReducersMapObject`.

ChrisAckerman:
There’s a comment in #2679 that it needs to be a deep partial, and not just partial. But hunting through the issues and PRs I’m not seeing why? Partial would have fixed the issue without adding new issues... even in the case of nested combined reducers. What am I missing? What’s the scenario that required deep partial?

dawnmist:
`Partial<State>` would have only fixed one level of the issues - if you have a nested tree of reducers the inner ones would again require being fully defined or the entire tree left out entirely.
e.g.
```
{
   reportA
        => options
             => optionA
             => optionB
             => optionC
   reportB,
   ...
}
```
[...]

PR #2809, opened 2018-01-24
TypeScript: `combineReducers` returns reducer that accepts partial state

OliverJAsh:
`combineReducers` returns a reducer that accepts partial state, not the whole state.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -86,7 +86,7 @@ export type ReducersMapObject<S = any, A extends Action = Action> = {
  *   object, and builds a state object with the same shape.
  */
 export function combineReducers<S>(reducers: ReducersMapObject<S, any>): Reducer<S>;
-export function combineReducers<S, A extends Action = AnyAction>(reducers: ReducersMapObject<S, A>): Reducer<S, A>;
+export function combineReducers<S, A extends Action = AnyAction>(reducers: ReducersMapObject<S, A>): (state: Partial<S> | undefined, action: A) => S;
 
 
 /* store */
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -86,7 +86,7 @@ export type ReducersMapObject<S = any, A extends Action = Action> = {
  *   object, and builds a state object with the same shape.
  */
 export function combineReducers<S>(reducers: ReducersMapObject<S, any>): Reducer<S>;
-export function combineReducers<S, A extends Action = AnyAction>(reducers: ReducersMapObject<S, A>): Reducer<S, A>;
```
OliverJAsh: It seems a shame to break apart from the `Reducer` type here, but without adding an extra generic to `Reducer`, I don't see a way around it?

aikoven:
We could consider adding a type
```ts
type CombinedReducer<S, A extends Action = AnyAction> = 
    (state: Partial<S> | undefined, action: A) => S
```
It would be assignable to `Reducer<S, A>`, so you'd be able to use it anywhere like a regular Reducer.

Although when you pass it to a generic function with `S` as a type parameter, it is inferred to `Reducer<Partial<S>>`. Particularly, when nesting `combineReducers`:
```ts
declare const reducer: Reducer<number>;

const combined = combineReducers({
  parent: combineReducers({
    child: reducer,
  }),
});

typeof combined === (
  state: Partial<{parent: Partial<{ child: number }>}> | undefined,
  action: AnyAction
) => {parent: Partial<{ child: number }>}

combined(undefined, {type: ''}).parent.child  // number | undefined, but expected number
```

OliverJAsh:
Thinking about this some more, I think the `state` parameter type needs to use something like `DeepPartial`, to allow partial state to be passed in when using nested `combineReducer`s.

Although `DeepPartial` is problematic in itself: https://github.com/reactjs/redux/issues/2808

timdorr:
Since this stalled, I'm going to close it out.

Issue #2810, opened 2018-01-25
A small step to get your error messages names better


VladBlow:
**Do you want to request a *feature* or report a *bug*?**  

A feature or in other words a proposal.

**What is the current behavior?**

In https://github.com/reactjs/redux/blob/master/src/createStore.js I found a little mistake in error messages: there is a shortage of article in [one](https://github.com/reactjs/redux/blob/master/src/createStore.js#L103) of them. 

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar.**

Not a bug.

**What is the expected behavior?**

I think we should add the article so we can use a function which generates error message like this: `Expected ${something} to be a ${something}.`. It will improve the code for a little bit.

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**

There are no affected versions.

timdorr:
I'm not sure what you mean, but I don't believe this adds anything to the code. A PR would probably be a better form of discussion, as we can talk over specifics.

PR #2811, opened 2018-01-26
Fixed article

VladBlow:
Fixes for https://github.com/reactjs/redux/issues/2810

And what do you say about function for errors like this:
```
function errorWithExpectedType(actual, expected) {
  throw new TypeError(`Expected the ${actual} to be a ${expected}.`)
}
```
Instead of simple TypeError.


timdorr:
Thanks!

I wouldn't worry about the function addition. We don't need the extra complexity.

PR #2814, opened 2018-01-28
enhancement: store.subscribe(listener, filter)

warren-bank:
new (optional) input parameter: `filter`

* its purpose is to conditionally filter when `listener` is called
* when its value is:
  * undefined:
    * `listener` is called as normal
  * `function`:
    * input: (oldState, newState)
    * output: boolean
      * when truthy, `listener` is not informed of this state change
  * `boolean` && true:
    * `listener` is not informed when `newState`
      is equal (by reference) to `oldState`

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -104,7 +104,7 @@ export default function createStore(reducer, preloadedState, enhancer) {
     }
 
     if (typeof filter !== 'function') {
-      filter = (filter === true) ? pure_filter : false
+      filter = filter === true ? pure_filter : false
     }
 
     const nextListener = {listener, filter}
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -98,10 +98,16 @@ export default function createStore(reducer, preloadedState, enhancer) {
    * @param {Function} listener A callback to be invoked on every dispatch.
    * @returns {Function} A function to remove this change listener.
    */
-  function subscribe(listener) {
+  function subscribe(listener, filter) {
     if (typeof listener !== 'function') {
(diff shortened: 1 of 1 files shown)


warren-bank:
Hi.

* I see that one of the tests failed, but I couldn't detect from its output whether the reason had anything to do with my (minor edit to the) code or not. It appears to be something unrelated. I'd be happy to make any necessary changes.

* maybe the name `filter` wasn't the absolute best choice?
  * normally, a filter function returns true to include and false to exclude..
    * whereas this is the other way around
  * would `ignore` or `skip` be preferable?
  * of course, this is strictly internal..
    * users passing a function into `subscribe` can call it anything they like..
    * or make it anonymous (for that matter)

warren-bank:
Unrelated to the error in Travis CI..
I was thinking..

* maybe the name `filter` wasn't the absolute best choice?
  * normally, a filter function returns true to include and false to exclude..
    * whereas this is the other way around
  * would `ignore` or `skip` be preferable?
  * of course, this is strictly internal..
    * users passing a function into `subscribe` can call it anything they like..
    * or make it anonymous (for that matter)

markerikson:
This has been frequently requested, and we have no plans to implement this.  If you'd like to do it yourself, you can implement it as a store enhancer.

For more info, see the FAQ entry on [why state and action are not passed to subscribers](https://redux.js.org/docs/faq/DesignDecisions.html#does-not-pass-state-action-to-subscribers), and the many prior discussions listed in issue #580 .

PR #2818, opened 2018-02-01
Fix example inserting item in immutable array

pgherveou:


markerikson:
Hi.  Appreciate the PR, but the current example is correct as-is.  The second argument to `slice()` is exclusive, so we need to start at that index value to get the rest of the items.  Try running this small example:

```js
const items = ["a", "b", "c", "d", "e"];
const index = 2;

const newItems = [
  ...items.slice(0, index),
  "f",
  ...items.slice(index)
];

console.log(newItems);
// should be: ["a", "b", "f", "c", "d", "e"]
// if we used index + 1, then "c" would be left out

```.

pgherveou:
Oh sorry my bad I misread it, i thought it was  about replacing the value at index

PR #2822, opened 2018-02-03
Typo: Added 'the' before 'case of..'

tyreer:
Minor tweak—but in this sentence, _case_ requires _the_ because the specificity of the prepositional phrase _of this tiny component_.


markerikson:
Thanks!

PR #2825, opened 2018-02-07
Include `replaceReducer` in middleware store API.

afenton90:
At the moment the store that gets passed to middleware does not include the `replaceReducer` function.
It would be really useful to have this for creating middleware that dynamically imports code split reducers.

We also believe this would create a more consistent API for the store object in middleware.

### DIFF
--- a/src/applyMiddleware.js
+++ b/src/applyMiddleware.js
@@ -29,6 +29,7 @@ export default function applyMiddleware(...middlewares) {
 
     const middlewareAPI = {
       getState: store.getState,
+      replaceReducer: store.replaceReducer,
       dispatch: (...args) => dispatch(...args)
     }
     chain = middlewares.map(middleware => middleware(middlewareAPI))
--- a/test/applyMiddleware.spec.js
+++ b/test/applyMiddleware.spec.js
@@ -48,16 +48,18 @@ describe('applyMiddleware', () => {
       }
     }
 
-    const spy = jest.fn();
-    const store = applyMiddleware(test(spy), thunk)(createStore)(reducers.todos);
+    const spy = jest.fn()
+    const store = applyMiddleware(test(spy), thunk)(createStore)(reducers.todos)
 
-    store.dispatch(addTodo('Use Redux'));
-    store.dispatch(addTodo('Flux FTW!'));
+    store.dispatch(addTodo('Use Redux'))
+    store.dispatch(addTodo('Flux FTW!'))
 
-    expect(spy).toHaveBeenCalledWith(expect.objectContaining({
-      replaceReducer: expect.any(Function)
-    }));
-  });
+    expect(spy).toHaveBeenCalledWith(
+      expect.objectContaining({
+        replaceReducer: expect.any(Function)
+      })
+    )
+  })
 
   it('passes recursive dispatches through the middleware chain', () => {
     function test(spyOnMethods) {
--- a/test/applyMiddleware.spec.js
+++ b/test/applyMiddleware.spec.js
@@ -40,6 +40,25 @@ describe('applyMiddleware', () => {
     ])
   })
 
+  it('should pass store.replaceReducer to middleware', () => {
+    function test(spyOnMethods) {
+      return methods => {
+        spyOnMethods(methods)
+        return next => action => next(action)
+      }
+    }
+
+    const spy = jest.fn();
+    const store = applyMiddleware(test(spy), thunk)(createStore)(reducers.todos);
+
+    store.dispatch(addTodo('Use Redux'));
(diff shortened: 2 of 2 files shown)


timdorr:
We don't need to expand the API surface to do this: https://medium.com/front-end-hacking/code-splitting-redux-reducers-4073db30c72e 

You actually want to implement a store enhancer for this. A middleware is intentionally limited to just the dispatch phase.

afenton90:
Thanks for taking a look at the PR @timdorr 
I have read that article previously, and whilst that approach does work somewhat with code splitting, I believe we can do better and not create an abstraction on top of redux.

For me it would be better to open up the Redux middleware API to include `replaceReducer` so the following code splitting pattern could be followed.
```js
// index.js
const middleware = applyMiddleware(dynamicReducers(reducer));
const store = createStore(reducer, {}, middleware);
```

```js
// dynamicReducer.js
const actionToReducerMapping = {
  ACTION_TYPE: { todo: reducer }
};

export default reducer => {
  currentReducers = reducer;
  return store => next => async action => {
    if (actionToReducerMapping[action.type]) {
      // Load reducer asynchronously
      store.replaceReducer(combineReducers({
        ...currentReducers,
        ...actionToReducerMapping[action.type]
      }))
    }
    next(action);
  };
};

```

Having a mapping between actions and their dynamically imported reducers makes as a more scalable solution when combined with the middleware pattern I am proposing.

I don't want to create an abstraction on top of redux to facilitate code splitting. To me it makes sense to work with the functionality of redux to allow this behaviour, with a simple extension of the current API.

ematipico:
In this way you are going to break the separation of concerns and design of the API. At this point a store enhancer is better, isn't it?

afenton90:
I see the enhancer approach as an unnecessary abstraction for the purposes of loading reducers that have been split into separate bundles.
With the middleware approach I feel it is working with the functionality that redux provides, and a more logical place to put this functionality.

I may be missing something with the current middleware API https://github.com/reactjs/redux/blob/master/src/applyMiddleware.js#L30
but there is no explanation for why the store is restricted to these methods alone.

If there is a reason this has been restricted it would be good to document somewhere.
However, at the moment there seems to be no solid reason/examples not to put `replaceReducer` in the middleware API?
I am happy to be corrected though.

timdorr:
> I may be missing something with the current middleware API but there is no explanation for why the store is restricted to these methods alone.

Middleware itself is a store enhancer. It's designed to redefine `dispatch` so you have a consistent place to put async actions and other things like that. Putting them elsewhere leads to weird side effects.

afenton90:
With that being the case would this change not make sense? As this is further enhancing the store?
> You actually want to implement a store enhancer for this.

Actions are being dispatched and as a side effect of those actions being dispatched, reducers are being brought in to fulfil those actions through dynamic imports.

Based on your description above https://github.com/reactjs/redux/pull/2825#issuecomment-364938137 makes sense to put in a middleware.

timdorr:
I mean it's implemented as a store enhancer, not that it is providing store enhancement. What it's doing is different from how it's doing it.

coderas:
I think this is a massively missed opportunity if the PR isn't merged since it simplifies the pattern for dynamically adding reducers to a store by action type.

I'm trying to imagine what weird (as in undesired) side effects could be introduced that couldn't be introduced using any other pattern. Could you elaborate?

afenton90:
Could really do with a deeper explanation of why this addition is undesirable?
At the moment, I'm still not seeing a concrete reason.

timdorr:
`applyMiddleware` is only for affecting the call phase of the `dispatch` lifecycle. So, `dispatch(action)` -> middleware goes here -> `reducer(state, action)`. It's to allow for handling side effects (async, global context stuff, other libs, etc) in just that part of the process. If you need it to do more (such as changing listeners or replacing reducers), then that falls out of the scope of middleware.

afenton90:
Completely agree middleware should be used for introducing side effects.
But what I don't agree with is where the line is being drawn on which side effects are being allowed and which ones aren't.

Other libraries that have the concept of middleware [koa](http://koajs.com/), [express](http://expressjs.com/) allow an open API to introduce behaviour as the consumer sees fit.
The redux API at the moment seems needlessly constrained in its current implementation.

If this is a hard and fast rule of the library lets document it somewhere.
If however, we can be more open (which would be my preference) lets do that.

SathishGovindaraju:
I do agree with @afenton90 . If middlewares are for introducing side effects, then why restrict the normal `store` API? By just passing the actual store (with all of its original API), it allows for much wider use of `applyMiddleware` rather than it just being a `dispatchInterceptor`

PR #2827, opened 2018-02-07
Change destination's link : learn React Router API

TroubadourFabiolas:
Change destination's link to learn React Router API V4 instead of V3


timdorr:
Thanks!

PR #2831, opened 2018-02-09
docs/middleware: minor improvement

spedy:
I think this better summons up Middlewares in redux


timdorr:
Thanks, but it reads better as "between".

PR #2840, opened 2018-02-17
Update Footer to use to use VisibilityFilters from actions

matiasleidemer:
Import and use the VisibilityFilters const from the actions file in the
documentation instead of using a hardcoded string.

I believe this will make things easier for the reader (as it did for me). Also, it doesn't seem to be a good practice to keep repeating constants throughout the codebase.

Some other places in the docs could benefit from this change. I didn't update them yet because I would like some feedback first.


timdorr:
Thanks!

PR #2842, opened 2018-02-17
Fixed broken links in README

shade:


timdorr:
I'm going to hold off on this for a bit. We just switched doc systems, so there are some growing pains. We shouldn't have to update external links.

timdorr:
This is fixed now!

PR #2844, opened 2018-02-18
Update Footer to use to use VisibilityFilters from actions

matiasleidemer:
Import and use the VisibilityFilters const from the actions file in the
documentation instead of using a hardcoded string.

To complement #2840.


timdorr:
Thanks!

PR #2843, opened 2018-02-18
Fix redirects in docs for previous urls

SamyPesse:
This PR fixes the issue about urls mentioned in #2837 and #2842.

You can test it on the test project (https://redux.gitbook.io/redux-test/) which is synced with my fork `SamyPesse/redux`:

- https://redux.gitbook.io/redux-test/docs/introduction/index.html
- https://redux.gitbook.io/redux-test/docs/basics/index.html
- https://redux.gitbook.io/redux-test/docs/advanced/index.html
- https://redux.gitbook.io/redux-test/docs/recipes/index.html
- https://redux.gitbook.io/redux-test/docs/Glossary.html

PR #2837, opened 2018-02-14
New documentation on the upcoming GitBook

SamyPesse:
Next week, we'll introduced a new version of GitBook (https://beta.gitbook.com/) 🎉 . This version brings a better reading experience (mobile and desktop), awesome search (powered by Algolia) and some cool new features.

I've run an initial import of the Redux docs in the new GitBook, you can checkout the output at: 
 https://redux.gitbook.io/docs/

It'd be awesome to get your feedback on this new version, and see if you're interested in switching the Redux documentation to this!

[screen shot 2018-02-14 at 18 10 06](https://redux.gitbook.io/docs/)

This version also includes some advanced features to analyse the quality of your documentation: a search dashboard and insights about the quality of your docs:

screen shot 2018-02-14 at 18 02 03
screen shot 2018-02-14 at 18 02 30

PR #2845, opened 2018-02-18
fix links on home page

clevelandbuckeye:
The links are broken. I'm brand new to this project and the link to 'introduction' shows page not found. Yes these links are also still broken. Just trying to point out that there are broken links on the home page which affects the learning experience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,14 +162,14 @@ If you enjoyed my course, consider supporting Egghead by [buying a subscription]
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


markerikson:
@clevelandbuckeye : yes, we just updated Gitbook (the tool we use to build the docs), and the update cause some changes that we're trying to fix.  For more details, see #2837, #2842 , and #2843.

I'll close this because we should hopefully have things resolved soon, but appreciate the PR!

PR #2848, opened 2018-02-19
Fix examples and todo app with VisibilityFilters export

matiasleidemer:
Kudos to @antoniwan for spotting the issue.


timdorr:
Thanks!

PR #2849, opened 2018-02-19
Add redux actiontyper

alnorris:


markerikson:
Hi.  Thanks for the PR, but I don't think it's something we specifically want to highlight at this time.  However, I'll add it to my [Redux addons catalog](https://github.com/markerikson/redux-ecosystem-links) next time I update it.

PR #2852, opened 2018-02-21
[todos] [refactor] improve the general style of the approach to React

raunofreiberg:
Some code clean-up for easier understanding and better conventions (i.e no temporary variable for connecting a component to the store).

### DIFF
--- a/examples/todos/src/actions/index.js
+++ b/examples/todos/src/actions/index.js
@@ -1,16 +1,16 @@
 let nextTodoId = 0
-export const addTodo = (text) => ({
+export const addTodo = text => ({
   type: 'ADD_TODO',
   id: nextTodoId++,
   text
 })
 
-export const setVisibilityFilter = (filter) => ({
+export const setVisibilityFilter = filter => ({
   type: 'SET_VISIBILITY_FILTER',
   filter
 })
 
-export const toggleTodo = (id) => ({
+export const toggleTodo = id => ({
   type: 'TOGGLE_TODO',
   id
 })
--- a/examples/todos/src/components/Footer.js
+++ b/examples/todos/src/components/Footer.js
@@ -3,21 +3,18 @@ import FilterLink from '../containers/FilterLink'
 import { VisibilityFilters } from '../actions'
 
 const Footer = () => (
-  <p>
-    Show:
-    {" "}
-    <FilterLink filter={ VisibilityFilters.SHOW_ALL }>
+  <div>
+    <span>Show: </span>
+    <FilterLink filter={VisibilityFilters.SHOW_ALL}>
       All
     </FilterLink>
-    {", "}
-    <FilterLink filter={ VisibilityFilters.SHOW_ACTIVE }>
+    <FilterLink filter={VisibilityFilters.SHOW_ACTIVE}>
       Active
     </FilterLink>
-    {", "}
-    <FilterLink filter={ VisibilityFilters.SHOW_COMPLETED }>
+    <FilterLink filter={VisibilityFilters.SHOW_COMPLETED}>
       Completed
     </FilterLink>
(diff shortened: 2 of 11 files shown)


examples/todos/src/components/Footer.js
```diff
@@ -3,21 +3,14 @@ import FilterLink from '../containers/FilterLink'
...
-    </FilterLink>
-  </p>
+  <div>
+      <span>Show: </span>
+      {VisibilityFilters.map((filter, idx) => (
+          <FilterLink key={filter.type} filter={filter.type}>
+              {filter.label}
+          </FilterLink>
+      ))}
+  </div>
```
timdorr: I know this is pedantic, but can we leave this alone? I'd rather make the UI and code 1:1, than essentially be doing meta-programming of the UI.

timdorr:
BTW, any code changes here need to be reflected here: https://github.com/reactjs/redux/blob/master/docs/basics/ExampleTodoList.md

examples/todos/src/components/Footer.js
raunofreiberg: Was wondering about this as well. Amended my last commit, should be restored now.

raunofreiberg:
Seems like the `ExampleTodoList.md` hasn't been updated for a while. Went ahead and unified the code and the docs since the docs were outdated a bit and still used the old style as opposed to the actual code.

examples/todos/src/actions/index.js
```diff
@@ -1,22 +1,22 @@
...
   id
 })
 
 export const VisibilityFilters = {
-  SHOW_ALL: 'SHOW_ALL',
-  SHOW_COMPLETED: 'SHOW_COMPLETED',
-  SHOW_ACTIVE: 'SHOW_ACTIVE'
+ SHOW_ALL: 'SHOW_ALL',
+ SHOW_COMPLETED: 'SHOW_COMPLETED',
+ SHOW_ACTIVE: 'SHOW_ACTIVE'
```
timdorr: This should still be two spaces. Are you running Prettier locally? We've added it to the library code, but not the examples yet. It may be a good idea to introduce it now, since this is already a pretty notable refactor.

raunofreiberg: Hmm. Not sure how that happened, I'm not using Prettier locally either. Maybe consider creating a issue about it separately and it might get taken up in another pull request. Fixed it for now.

raunofreiberg:
@timdorr This should be fine for merging now.

timdorr:
It's got the 👍 from me, but since this is rather large and very visible code, I'd like a second set of eyes on it before merging.

examples/todos/src/containers/VisibleTodoList.js
```diff
@@ -15,17 +15,15 @@ const getVisibleTodos = (todos, filter) => {
...
-const mapStateToProps = (state) => ({
+const mapStateToProps = state => ({
   todos: getVisibleTodos(state.todos, state.visibilityFilter)
 })
 
-const mapDispatchToProps = {
-  onTodoClick: toggleTodo
-}
+const mapDispatchToProps = dispatch => ({
+  toggleTodo: id => dispatch(toggleTodo(id))
```
fportela-ns: i don't understand the purpose if doing this. Yes, it's working, but that's the benefit of doing this over just sending the parameters when calling the action _directly_?

CodeKommissar:
Hi @timdorr! About the coding style, I think the changes made by @raunofreiberg are for the better. It uses const declarations instead of let, has better naming conventions and the arrow functions implicitly return objects. The refactoring (in my opinion) does improves readability.

timdorr:
OK, I think we're good on this. Thanks!

PR #2855, opened 2018-02-25
Update Store.md

reyou:
Hotlinks fixed under Store Methods section.


AllenFang:
@reyou I think your solution can fix the official website anchor but will break the anchor for github README.md, you can check https://github.com/reyou/redux/blob/patch-1/docs/api/Store.md

markerikson:
Yeah, in-page anchors seem to be broken as a result of the upgrade to the new Gitbook version (#2837 ).  We're waiting for an update to Gitbook that will fix those.

PR #2857, opened 2018-02-28
Add return state to visibilityFilter when filter is undefined

VictorWinberg:
Given action "SET_VISIBILITY_FILTER", reducer "visibilityFilter" returned undefined. To ignore an action, you must explicitly return the previous state. If you want this reducer to hold no value, you can return null instead of undefined.


timdorr:
This wouldn't be ignoring the action. The code is handling the `SET_VISIBILITY_FILTER` action explicitly. So, if we put in an undefined filter, then that would be expected output.

VictorWinberg:
Aight, I guess that could be true as well. But then your [redux-devtools](https://github.com/gaearon/redux-devtools) from @gaearon should not return that error-message, since it came straight from that.

PR #2860, opened 2018-03-01
remove two intermediate variables to enhance readability

cyan33:
Hi Redux maintainers! I was reading the source code of the code base when I felt like these two variables, `reducerKeys ` and `finalReducerKeys ` are rather redundant. Besides, readability-wise, `forEach` is much better than the traditional for loop. So I replaced these two variables with `Object.keys` and `forEach`. And AFAIK, the code seems to me clearer.

Point out if I missed anything or this refactor has some potential errors. Thanks! 😄

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -116,11 +116,8 @@ function assertReducerShape(reducers) {
  * passed object, and builds a state object with the same shape.
  */
 export default function combineReducers(reducers) {
-  const reducerKeys = Object.keys(reducers)
   const finalReducers = {}
-  for (let i = 0; i < reducerKeys.length; i++) {
-    const key = reducerKeys[i]
-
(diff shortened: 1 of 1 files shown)


markerikson:
Hi.  Appreciate the PR, but in this case the goal is to optimize perf a bit more rather than make the code "declarative" via a `forEach` loop.  Pretty sure we've people suggest similar changes in the past, and opted not to use them (don't have specific links atm).  So, this isn't something we'd merge at this point.  Thanks for looking at this, though!

brodybits:
> Hi. Appreciate the PR, but in this case the goal is to optimize perf a bit more rather than make the code "declarative" via a forEach loop.

Comments in the source and maybe CONTRIBUTING.md would be nice.

cyan33:
@markerikson Got it, thanks!

PR #2865, opened 2018-03-02
Add a sideEffects: false flag for Webpack 4

timdorr:
See https://github.com/webpack/webpack/tree/master/examples/side-effects

Webpack 4 can more effectively tree shake when a module defines itself as having no side effects (aka, doesn't talk to things external to the module at the top level). I believe we qualify, but I'm putting this in as a PR just to be sure. Looks like it can easily net some byte savings for end users 👍

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,5 +105,6 @@
   },
   "jest": {
     "testRegex": "(/test/.*\\.spec.js)$"
-  }
+  },
+  "sideEffects": false
 }


nickmccurdy:
What's involved with testing this, would it be enough to run CRA on this build of Redux and ensure that it doesn't break? Or would the module have to be analyzed?

timdorr:
Build it with webpack 4 and see if it breaks. I think there's a low likelihood of problems, so I'll throw it into the next release (probably an RC) and see if everyone throws tomatoes back.

PR #2872, opened 2018-03-06
[docs] [examples] remove the open localhost cmd from examples

raunofreiberg:
IMHO, there is no need to specify this command in the documentation since all the examples are based off `create-react-app` and executing the `npm start` script calls the `openBrowser` method that does this by default. Would love to remove this `open ...` command to avoid confusion.


timdorr:
Thanks!

Issue #2854, opened 2018-02-24
FAQ Updates, Part 2: Electric Boogaloo


markerikson:
We were able to add a lot of FAQ updates over in #1785 , but there's still some more updates I'd like added.  Here's the current list of items I'd like in there:

#### General

- When should I learn Redux?
  - https://twitter.com/dan_abramov/status/739961787295117312
  - https://twitter.com/dan_abramov/status/739962098030137344
  - https://github.com/markerikson/react-redux-links/blob/master/basic-concepts.md
- What are pros and cons of using Redux?
  - https://github.com/reactjs/redux/issues/1342#issuecomment-183401969
- How does Redux compare to [Backbone, Angular, MobX]?
  - Redux vs a global variable
    - https://news.ycombinator.com/item?id=13334978 
- How is Redux different from React's context API?  Does context replace Redux?
  - https://daveceddia.com/context-api-vs-redux/
  - https://blog.isquaredsoftware.com/2018/03/redux-not-dead-yet/

#### Code Structure

- Reusable / encapsulated logic/modules?
  - https://github.com/markerikson/redux-ecosystem-links/blob/master/component-state.md
  - https://github.com/markerikson/react-redux-links/blob/master/redux-techniques.md#encapsulation-and-reusability
  - http://randycoulman.com/blog/
  - https://github.com/slorber/scalable-frontend-with-elm-or-redux
  - http://www.webpackbin.com/VkHniHP6b
- Where should websockets and other persistent connections live?
    - In middleware.  See existing examples at https://github.com/markerikson/redux-ecosystem-links/blob/master/middleware-sockets-adapters.md

#### Actions

- Should I use thunks, sagas, or something else?
  - https://blog.isquaredsoftware.com/presentations/workshops/redux-fundamentals/store-api.html#/6
  - https://decembersoft.com/posts/what-is-the-right-way-to-do-asynchronous-operations-in-redux/
  - https://www.reddit.com/r/reactjs/comments/8vglo0/react_developer_map_by_adamgolab/e1nr597/

#### Performance

- Pagination/caching?
  - http://stackoverflow.com/questions/33940015/how-to-choose-the-redux-state-shape-for-an-app-with-list-detail-views-and-pagina
  - https://github.com/facebook/relay/issues/720#issuecomment-174050321
  - https://twitter.com/acemarke/status/804071531844423683
  - https://medium.com/@dcousineau/advanced-redux-entity-normalization-f5f1fe2aefc5
[...]

PR #2879, opened 2018-03-08
docs(FAQ: React-Redux): Why use React-Redux  #2854

stevenandersonz:
Base on the requested issue I update the React-Redux point answering the question Why use React-Redux.

I'll be glad to hear any feedback.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -55,6 +55,7 @@
   - [Why doesn't `mapDispatchToProps` allow use of return values from `getState()` or `mapStateToProps()`?](/docs/faq/DesignDecisions.md#no-asynch-in-mapDispatchToProps)
 - **React Redux**
   - [Why isn't my component re-rendering, or my mapStateToProps running?](/docs/faq/ReactRedux.md#react-not-rerendering)
+  - [Why use React-Redux?](/docs/faq/ReactRedux.md#why-use-react-redux)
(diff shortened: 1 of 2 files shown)


timdorr:
There are a fair number of grammatical issues to clean up as a first step.

I think we should start with explaining that it's possible to use Redux and React without this library, but you will have to end up writing a lot of extra and complex code. Then get into the performance benefits and optimizations that are baked in. 

And that last paragraph is incorrect. The performance optimizations are specifically there to avoid the re-render on every dispatch.

stevenandersonz:
Thank you for your review, yes I agree with you, and I want to do this. 
What I'm going to do then: 
* Is to start explaining that we can use React and Redux just using the functions provided by the store. I think that an App with two components that cares of different pieces of state would be perfect to demonstrate this.
* Then I reference the same app to illustrate how `<Provider/>` save the work of passing the store down as a props
* The last step is to explain how `connect()` performs several equalities checks to avoid unnecessary calls to
 mapStateToProps, mapDispatchToProps, etc.

markerikson:
Hi.  Sorry I never got around to working with this.  I ultimately wound up writing a page for the React-Redux docs at https://react-redux.js.org/introduction/why-use-react-redux , and used a version of that for the FAQ.  Thanks for the PR!

PR #2880, opened 2018-03-09
Remove validator.js suggestion

chrisself:
According to the [validator.js documentation](https://github.com/chriso/validator.js), XSS sanitization was removed from the library in late 2013. Because of this, I believe it makes sense to remove it as a suggestion from the SSR recipe. At the time, the maintainer suggested checking out [Google Caja](https://code.google.com/archive/p/google-caja/source/default/source). I also did some searching on NPM and found the following two packages, both of which seem full featured and popular:

https://www.npmjs.com/package/xss
https://www.npmjs.com/package/xss-filters

Should we replace the suggestion with one of the alternatives above? Definitely open to suggestions — thanks!


timdorr:
Yeah, let's switch out the link instead. xss-filters looks good.

timdorr:
<script>document.write('Thanks!')</script>

PR #2884, opened 2018-03-10
Update AsyncActions.md

drootnar:
typo: `us => use`


markerikson:
Appreciate the PR, but the current phrasing is correct.  The expanded wording would be "this lets us dispatch functions".

Issue #2846, opened 2018-02-18
New Gitbook version for the Redux docs - issues, tweaks, and improvements


BTMPL:
**What is the current behavior?**

Browser scroll position is not reset when navigating around the new Redux website.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar.**

1. go to https://redux.js.org/advanced/middleware or any other "long" page
2. navigate to https://redux.js.org/advanced/async-flow or any other "short" page
3. URL and content is updated, scroll position is not reset and what appears to be "empty" page is shown

**What is the expected behavior?**

Page is scrolled to top on navigation

CC @SamyPesse as this is most probably a Gitbook issue

SamyPesse:
Yes, this is a GitBook issue, a fix is already coming (probably tomorrow).

If you catch other issues like this one, feel free to open a ticket at https://gitbook.canny.io/feature-requests

Thank you 🍻

markerikson:
@SamyPesse : hijacking the thread because you're looking at it.  Since we're doing all this redirect stuff anyway - if I wanted to start shuffling locations of some of the docs pages, is there a way to set up redirects?

For example, let's say I wanted to move /advanced/middleware to /intermediate/middleware .  Could we get that redirected?

SamyPesse:
@markerikson GitBook automatically manages redirects. basically if you move a pageA from `hello/a` to `world/b`, then a redirect is setup.

The main issue come from the GitHub integration. When changes are created from pushes to GitHub, it's much harder to recognises moves and not remove & add. GitBook has its own unique identifier for each pages, but it's not currently synced with GitHub.

If you move a page and modify its content at the same time, it'll not be recognized as a move, but as the deletion of the first page and creation of a new page. In that case, no redirect is setup.

My point is that auto-redirects on GitBook works very well for content without GitHub sync, but not really for projects like Redux on GitHub.

You can't currently enter your own redirects, but that may be a nice feature. For example for projects synced with GitHub, we could have an option in `.gitbook.yaml`:

```yaml
redirects:
  /advanced/middleware: /intermediate/middleware
```

But in a perfect world, I'd prefer to have it work automatically for all projects, this is not something docs maintainers should care about; when reordering content, GitBook should be good at avoiding broken urls.

Do you see another usage for redirects ?

alexkrolick:
[...]

PR #2885, opened 2018-03-11
Dead anchor in Performance.md

nicolas-besnard:


markerikson:
Yeah, unfortunately the new version of Gitbook isn't handling the anchor tags right.  We're tracking this over in #2846  - we're waiting for them to release a new version that handles this correctly.  I'll close this, but thanks for the PR.

PR #2887, opened 2018-03-12
Async Actions Tutorial: Fix broken Promise in thunk action creator

dosentmatter:
The current `onRejected` function resolves with `undefined` because it returns the `console.log()` expression. This means that if the `fetch()` errors, the `onRejected` in the first `then()` will resolve with `undefined`, causing `json` in the second `then()` to be `undefined`. Calling `receivePosts()` with `json === undefined` will error on `json.data`.

There's no reason to use two `then()` calls. But if two `then()` calls are desired, the second `then()` must happen before the `onRejected` function, because the `onRejected` function should resolve with something so it doesn't have an unhandled Promise rejection. In the fix, the `onRejected` function resolves with the `error`. To have the second `then()`, you could nest the `then()` inside the `onResolved` function like so:
```
.then(
  response => Promise.resolve(response.json)
    .then(
      json => dispatch(receivePosts(subreddit, json))
    )
)
```
But there's no point in doing that.

### DIFF
--- a/docs/advanced/AsyncActions.md
+++ b/docs/advanced/AsyncActions.md
@@ -363,18 +363,19 @@ export function fetchPosts(subreddit) {
 
     return fetch(`https://www.reddit.com/r/${subreddit}.json`)
       .then(
-        response => response.json(),
-        // Do not use catch, because that will also catch
-        // any errors in the dispatch and resulting render,
-        // causing a loop of 'Unexpected batch number' errors.
-        // https://github.com/facebook/react/issues/6895
-        error => console.log('An error occurred.', error)
-      )
-      .then(json =>
-        // We can dispatch many times!
-        // Here, we update the app state with the results of the API call.
-
-        dispatch(receivePosts(subreddit, json))
+        response => {
+          // We can dispatch many times!
+          // Here, we update the app state with the results of the API call.
+          return dispatch(receivePosts(subreddit, response.json())
+        },
+        error => {
+          // Do not use catch, because that will also catch
+          // any errors in the dispatch and resulting render,
+          // causing a loop of 'Unexpected batch number' errors.
+          // https://github.com/facebook/react/issues/6895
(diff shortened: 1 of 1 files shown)


timdorr:
The 2nd `then` is needed because [`response.json()` returns a Promise](https://developer.mozilla.org/en-US/docs/Web/API/Body/json) which you can't pass into the dispatch as-is (not without some middleware). So, the first `then` will resolve the Promise before passing to the next `then`. 

While I appreciate the attention to detail in the error path, I'm not particularly worried about the lessons being taught here. The focus is on how you interact with Redux inside of Promises, not the particulars of writing exhaustive Promise handling.

dosentmatter:
Thanks for the response. The docs has a "**Note on `fetch`**" section, so it doesn't exactly assume you are familiar with the `fetch` API. In that case, I think running a `then()` on the `response` makes it clearer that `.json()` is a `Promise`.

The error handling was added in #2331. It kinda sidetracks you from the redux tutorial and I think the original was clearer. But if you are going to add error handling, why not do it correctly? Also, I think the existing code is confusing, because a `catch()` can be done, it just shouldn't be done after the dispatch. The comment telling you to "Do not use catch, because that will also catch any errors in the dispatch" is referencing the dispatch below it so that can be confusing because you haven't seen it yet. These are just my thoughts when reading the tutorial for the first time. I am okay with not changing it. Thanks again!
```
    return fetch(`https://www.reddit.com/r/${subreddit}.json`)
      .then(
        response => {
          // We can dispatch many times!
          // Here, we update the app state with the results of the API call.
          return response.json().then(json => dispatch(receivePosts(subreddit, json)))
        },
        error => {
          // Do not use catch, because that will also catch
          // any errors in the dispatch and resulting render,
          // causing a loop of 'Unexpected batch number' errors.
          // https://github.com/facebook/react/issues/6895
          console.log('An error occurred.', error)
          return error
        }
      )
```

PR #2889, opened 2018-03-13
Update todos flow to follow up-to-date libdefs

villesau:
React-redux flow libdefs has updated significantly since last update and does not need `Connector` type anymore: https://github.com/flowtype/flow-typed/pull/1731 .

I would have updated other packages too, but due to react-scripts issue https://github.com/facebook/create-react-app/issues/871 I couldn't run the scripts.


timdorr:
Thanks!

PR #2891, opened 2018-03-14
Remove dead link from Immutability reading list

esbena:
http://www.t4d.io/javascript-and-immutability/ results in 404, so I removed it.
I did a few searches to see if the page had moved, but I came up empty.


timdorr:
Thanks!

markerikson:
Looks like it's in the Web Archive at http://web.archive.org/web/20161101072408/http://www.t4d.io/javascript-and-immutability/ .  Maybe see if one of the other articles listed at https://github.com/markerikson/react-redux-links/blob/master/immutable-data.md is a suitable replacement ?

PR #2892, opened 2018-03-15
 Put the private action type into actionType.js

zazaliu:


timdorr:
Thanks, but this is intentionally not stored. The probe action is intended to be thrown away and not re-used.

PR #2897, opened 2018-03-16
Fixing typo, asychrony vs asynchronicity

mungojam:


timdorr:
This is right as-is: https://en.wikipedia.org/wiki/Asynchrony_(computer_programming)

mungojam:
@timdorr earlier in the paragraph it says asynchronicity, so is that one wrong maybe instead? It was the inconsistency that stood out

PR #2899, opened 2018-03-22
fix code example for `generating reducers`

seriousManual:


timdorr:
Thanks!

markerikson:
FWIW, the existing example _was_ valid.  It's an ES6 object literal, with a computed key as the function name, and a function declaration directly as part of the object.

Admittedly not the easiest thing to read, but it was valid.

timdorr:
Ah, I didn't know you could combine those two, but that makes sense.

seriousManual:
wow, didn't know this! 
thanks for the clarification, with the added colon and fat arrow the readability might a bit better though :)

PR #2901, opened 2018-03-24
Fixes the broken async example

lawwantsin:


timdorr:
Other than some style changes, I'm not seeing what this fixes. One thing in particular: removing the class field form of the handler functions means you have to manually bind the functions, which is non-ideal.

Issue #2896, opened 2018-03-16
Useful Documentation Explicitly Omitted


Swivelgames:
### Documentation Bug

While perusing the section on `Reselect` and using selectors to create "views" of your reducer state, I noticed a somewhat alarming disclaimer that stopped me in my tracks.

Naturally, I'm utilizing Redux with React. Additionally, I typically have high hopes that the documentation will provide enough information for me to explore the depths of a library or utility and use it optimally.

Unfortunately, when reading about selectors, I was told quite literally that what was about to be shown "requires changes to the [code base] ... and have been omitted for brevity."

Brevity is important in writing documentation, but it's incredibly frustrating when you run upon a disclaimer that essentially says, "The things we're doing in this section? You can't do them without X and Y, but we're not going to tell you how to do X and Y, so good luck." This is less brevity, and instead excluding useful or important information from the documentation. A better solution might be to move that information to another article and reference it from the original.

#### Expected Result
- An additional section, or a link or reference to additional materials for those who would like to do what your documentation is demonstrating.

#### Actual Result
- Disclaimer does not include a link or reference to any explanation of the items in the examples below it.
- OR, disclaimer does not include a link to the example's source code, which could then be used to better understand the additional documentation for those who want to dive deeper.

The most unfortunate part is that I was specifically searching the documentation for this very section in order to understand exactly what I quickly found out was intentionally omitted.

> **Accessing React Props in Selectors**
>
> _This section introduces a hypothetical extension to our app that allows it to support multiple Todo Lists. Please note that a full implementation of this extension requires changes to the reducers, components, actions etc. that aren't directly relevant to the topics discussed and have been omitted for brevity._

(Source: [https://github.com/reactjs/redux/blob/master/docs/recipes/ComputingDerivedData.md](https://github.com/reactjs/redux/blob/master/docs/recipes/ComputingDerivedData.md))

P.s., Keep up the great work!

markerikson:
I'd appreciate it if you could file a PR that expands on this section.  You may want to refer to my post [Idiomatic Redux: Using Reselect Selectors for Performance and Encapsulation](http://blog.isquaredsoftware.com/2017/12/idiomatic-redux-using-reselect-selectors/), as well as some of the articles in the [Redux Reducers and Selectors](https://github.com/markerikson/react-redux-links/blob/master/redux-reducers-selectors.md#selectors) section of my React/Redux links list.

hedgerh:
[...]

PR #2903, opened 2018-03-25
Add more code samples to example in Computing Derived Data recipe

hedgerh:
Added code that shows the reducers/actions needed for the "Accessing props in selectors" example in Computing Derived Data.

Closes https://github.com/reactjs/redux/issues/2896


markerikson:
Thanks!

PR #2906, opened 2018-03-27
Optimize plain object check function

aakashapoorv:
Optimized isPlainObject(obj) function, skipping while loop.
Working well with test `test/utils/isPlainObject.spec.js`

Update not passing test for `expect(isPlainObject(sandbox.fromAnotherRealm)).toBe(true)`


timdorr:
This doesn't work for cross-realm objects (think iframes).

PR #2908, opened 2018-03-28
docs: Change lets to const when creating store

nikolasleblanc:
Noticed that code examples use:

`let store = createStore(...)`

Since store is never reassigned, we should use `const`.


timdorr:
`let`'s update a few more... Thanks!

nikolasleblanc:
Sweet. Thanks @timdorr!

PR #2910, opened 2018-03-28
Check self.process before reading for browser compatibility

### LINKED ISSUE #2907: Remove usage of `process.env.NODE_ENV` in ES module build

Feature

keanulee:
With this patch, browsers can directly run the ES module build of Redux from https://unpkg.com/redux@4.0.0-beta.2/es/redux.js?module and other tools that converts package names to paths.

Fixes #2907


timdorr:
Thanks for this. I think it looks good, but anyone else want to chime in?

jimbolla:
@timdorr  The thing to verify is that those blocks still get removed in the prod build. I suppose comparing the prod build size before and after this change would work.

timdorr:
Yep, I'll want to verify locally to be sure. I'll try to run that through tonight or tomorrow at some point.

kaushik94:
@jimbolla @timdorr the [way react does this check](https://github.com/facebook/react/blob/master/scripts/jest/setupEnvironment.js#L7) is by defining a `__DEV__` variable. Chances are you already knew this. Although I don't know how to define this globally, we can still define something like this in index.js file and export from there to check `if(__DEV__)` , I think this would be a cleaner solution? I would be glad to implement this, and also would like to work on this library for self-improvement. I would be happy if I can contribute, even if it's a small change :)

timdorr:
@kaushik94 That's transformed into a NODE_ENV check during the build process. They use it as a shorthand while coding, but it's not how the build comes out compiled.

PR #2915, opened 2018-04-01
changed const to let in applyMiddleware

isahgaga:
changed const to let in applyMiddleware function in Attempt #6


markerikson:
I initially assumed you were talking about editing the actual implementation, but yeah, that example does seem to have been a bit off :)

PR #2919, opened 2018-04-07
Fix missed case in "observe" type check

zerobias:
Redux briefly checks that observer received in observe function is a valid object and throw TypeError with useful user message if is not. 
However, it missed null case, as `typeof null === 'object'` (well-known js issue) and then failed with the obscure error message `Cannot read property 'next' of null`.  I'm sure that there is no reason to not to show the correct error message in all relevant cases
https://github.com/reactjs/redux/blob/master/src/createStore.js#L237

This PR fixes that inconsistency by throwing TypeError with the same useful message in case of null observer 

Note that this PR is not a breaking change in any way

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -234,7 +234,7 @@ export default function createStore(reducer, preloadedState, enhancer) {
        * emission of values from the observable.
        */
       subscribe(observer) {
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

Issue #2921, opened 2018-04-07
All the tests fails due to missing "testEnvironment" property in Jest config


burnmaniac:
**Do you want to request a *feature* or report a *bug*?**  

Bug

**What is the current behavior?**

All tests fails. More info in the next section of the issue.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar.**

On a fresh install of Ubuntu, I have forked the repo, run `npm install`, run `npm run build` and then 
`npm run test`. All tests failed with the error message:

`SyntaxError: Unexpected token { at _load_jsdom (node_modules/jest-environment-jsdom/build/index.js:18:19)`

**What is the expected behavior?**

All tests should pass without error messages.

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**

Node version: v.4.2.6
NPM version: 3.5.2
OS version: Ubuntu 16.04.4 LTS

I don't know if it worked on previous versions of Redux.

timdorr:
You need a newer version of npm (possibly node as well). You should be using npm@5.

PR #2922, opened 2018-04-07
Add "testEnvironment" to jest config in package.json.

burnmaniac:
This is the fix for #2921, also my first official OSS contribution. I hope I did everything all right.

Thank you!

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     ]
   },
   "jest": {
+    "testEnvironment": "node",
     "testRegex": "(/test/.*\\.spec.js)$"
   },
   "sideEffects": false


markerikson:
Hmm.  To be fair, it looks as if that's directly related to using an older version of Node (per https://github.com/facebook/jest/issues/5276 and https://github.com/standard-things/esm/issues/210 ).

Still, I don't see any reason to _not_ add that flag to `package.json`.  @timdorr  ?

burnmaniac:
Just in case the flag didn't break the tests on newer versions of Node, I know that Travis will do that also, but I've installed newer version of Node (8.11.1 which is the LTS at this moment) and run the tests with the `testEnvironment` flag. All the tests passed.

timdorr:
Just FYI: The suite is failing because of some flow stuff that's out of date, not this change.

I'd actually *want* the suite to fail on that old of a version. It serves as a warning to upgrade (even though it's somewhat opaque). If you're developing a code contribution to Redux, you need to be using the latest npm and should be on an updated node version. That's what I use for publishing and what we use for testing.

Perhaps we should add an `engines` section to the package.json instead?

burnmaniac:
I didn't think of it that way. I didn't want to use that Node version, I just installed it using `apt`. Didn't even check which version was it.

I guess I should first find out on which Node version everything works, then add it like this?

```
{
    engines: {
        "node": ">=6.0.0" // Or which ever version I find minimal
    }
}
```

timdorr:
Actually, I realized that's incorrect. That would apply to the published package, not just the dev environment. We still support node 4 and earlier, since there isn't anything wacky like async or generators in there. 

Not sure what the best option is. I lean towards doing nothing and erroing out for older node versions. Not as nice, but less messy for users.

burnmaniac:
I agree. Feel free to close pull request. 👍

PR #2928, opened 2018-04-12
Remove unneeded chain declaration

BenBrostoff:
Hello - I've been reviewing the Redux source and was trying to figure out what the purpose of assigning`chain` to an empty array was in `applyMiddleware`. There doesn't seem to be a scenario where `chain` is not immediately updated to `middlewares.map(middleware => middleware(middlewareAPI))`. Please let me know if I'm missing something here and thanks for a great library.

### DIFF
--- a/src/applyMiddleware.js
+++ b/src/applyMiddleware.js
@@ -25,13 +25,12 @@ export default function applyMiddleware(...middlewares) {
           `Other middleware would not be applied to this dispatch.`
       )
     }
-    let chain = []
 
     const middlewareAPI = {
       getState: store.getState,
       dispatch: (...args) => dispatch(...args)
     }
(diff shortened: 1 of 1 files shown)


markerikson:
Hmm.  I appreciate the PR, and I'm inclined to agree, but I'm also _very_ hesitant to merge any changes to this file at this point.  We've had numerous PRs over the last few years trying to "tweak" or "fix" things (usually related to the declaration of `dispatch` in here), and they've almost all been wrong.  The current code is there, it works, and I don't see a particularly pressing need to change it further.

I'll ping @timdorr and @jimbolla just to get their thoughts.

jimbolla:
LGTM

timdorr:
OK, at least we're all equally culpable.

PR #2941, opened 2018-04-18
Add redux-saga-requests to ecosystem

klis87:


klis87:
Hi, I am really surprised, as you have `svrcekmichal/redux-axios-middleware` listed, and my library does the same things with many more features (also due to power of `redux-saga`). And it is not really a competition to mentioned library - according to https://github.com/redux-saga/redux-saga/issues/1117 mixing another async middleware with `redux-saga` is often problematic, and this is exactly the reason I created [redux-saga-requests](https://github.com/klis87/redux-saga-requests) , it is just a better choice for someone using `redux-saga`, while `redux-axios-middleware` is a great library for someone not using `redux-saga`.

Btw, I couldnt find my library listed on your Redux catalogue either, maybe u confused it with a different library?

PR #2950, opened 2018-04-19
Remove "Redux itself is very simple"

gaearon:
Reflecting a few years later this was a bit of a silly thing to write in the docs.
Of course it's not simple to people learning it.


timdorr:
I agree!

PR #2953, opened 2018-04-21
Keep state even if there is no reducer

FL0RIANMEYER:
Handy for Electron Main/Render process to have different reducer for shared actions


timdorr:
I'm not understanding why you need this. The reducer(s) should be pure functions, so they should not need existing state in place to construct the next state. If they do, then you are likely doing something wrong with your reducers (most likely, they are not pure functions).

PR #2957, opened 2018-04-24
getState is OnlyRead

lihuanji:
example:

```
function createStore(preloadedState) {
    let currentState = preloadedState
    function getState() {
        return currentState
    }
    return {
        getState,
    }
}

const store = createStore({test: 1});
store.getState().test = 2;
console.log(store.getState()); // The test value here is 2.
```

The state has changed

I think I should return a new object in getState().

thank you.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -72,7 +72,7 @@ export default function createStore(reducer, preloadedState, enhancer) {
       )
     }
 
-    return currentState
+    return {...currentState}
   }
 
   /**


Cryrivers:
I believe this is working as intended. You shouldn’t modify states directly.

markerikson:
Yes - `getState()` just returns whatever the current state value is.  It's up to you to avoid accidentally mutating it.

Issue #2960, opened 2018-04-24
Parentheses in arrow functions


tomasztomys:
**Do you want to request a *feature* or report a *bug*?**  
bug

**What is the current behavior?**
```
arg => {
  aaa;
  bbb;
  return xxx;
}
```

**What is the expected behavior?**
```
(arg) => {
  aaa;
  bbb;
  return xxx;
}
```

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**
latest (since 2 years)

I think that we should add parentheses if we have multi-line body of arrow function (especially with return statement). Like in airbnb proposes (which allowed to skip parentheses only if we have sth like a single line arrow function with 1 arg) or like in Google styleguide (always use parentheses).

timdorr:
We use [Prettier](https://github.com/prettier/prettier) on the code base, so it decides for us.

tomasztomys:
@timdorr Prettier's added a new option in the end of 2017, because issue about parenthesis was the most commented issue in their repos. I think that we should change conf of Prettier in our repo.

PR #2961, opened 2018-04-24
Change arrowParens option of Prettier to "always"

tomasztomys:
#2960

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,4 +1,5 @@
 {
   "semi": false,
-  "singleQuote": true
+  "singleQuote": true,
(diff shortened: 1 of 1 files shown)


timdorr:
I disagree with this. Parens are optional on arrow functions for a reason. It keeps things cleaner without them.

Plus, I'm scarred from years of Clojure usage... :)

tomasztomys:
Why is it cleaner? It's even not looks like function ;)

@gaearon

timdorr:
The arrow (`=>`) of the arrow function makes it look like a function, IMHO. Not the parens.

Issue #2958, opened 2018-04-24
docs: tip on conditionally applying middleware uses `const` erroneously


andreaswachowski:
It seems that [applyMiddleware.md](https://github.com/reactjs/redux/blob/827b627c4011f14ff541db6d1ec162f8e12484fa/docs/api/applyMiddleware.md) contains an incorrect use of `const` in one place. The tip on conditionally applying middleware contains the following code:

```javascript
const middleware = [a, b]
if (process.env.NODE_ENV !== 'production') {
  const c = require('some-debug-middleware')
  const d = require('another-debug-middleware')
  middleware = [...middleware, c, d]
}
```

There's an assignment to `middleware` inside the if-statement, hence it should not be `const`-declared.

timdorr:
If you want to make a change to the docs, a PR is the best way to do this.

PR #2962, opened 2018-04-24
Fix erroneous use of const in docs

andreaswachowski:
`middleware` is reassigned in the if-statement and hence must not be const.

This pull request is in regard to #2958.


timdorr:
Thanks! I think I modified that erroneously when switching some let's to const's.

PR #2964, opened 2018-04-25
Todos example: Always use VisibilityFilters instead of strings

Birkenstab:
Two months ago the VisibilityFilters export was created. This commit removes the missed places where strings were still used and replaces it with VisibilityFilters


timdorr:
Thanks!

PR #2965, opened 2018-04-25
fix: update ExampleTodoList.md with regards to #2964

raunofreiberg:
Noticed this was not updated. Probably should consider syncing the actual code with the markdown files (Gatsby 🤔 )?


timdorr:
Thanks!

PR #2970, opened 2018-04-28
Update Reducers.md to be a bit more succinct

MattAurich:
"describe the fact that *something happened*" can be nothing more than a BEEP that is irregular and inconsistent in it's response to an input, while an action (that is of any use beyond a '*' (all)) is going to describe (as defined by the author) *what happened*, through a series of one or more descriptors/data points.


timdorr:
Thanks!

PR #2972, opened 2018-04-29
Use the latest npm version

retyui:
update `.travis.yml`


timdorr:
Thanks, but we're waiting on some Jest fixes for Node 10 before updating.

PR #2976, opened 2018-05-04
Move PROBE_UNKNOWN_ACTION to utils/actionTypes.js

Nbsaw:
Make the code style more uniform

### DIFF
--- a/src/utils/actionTypes.js
+++ b/src/utils/actionTypes.js
@@ -4,21 +4,17 @@
  * If the current state is undefined, you must return the initial state.
  * Do not reference these action types directly in your code.
  */
+
+const randomString = () => Math.random()
+                               .toString(36)
+                               .substring(7)
+                               .split('')
+                               .join('.')
+
 const ActionTypes = {
-  INIT:
-    '@@redux/INIT' +
-    Math.random()
-      .toString(36)
-      .substring(7)
-      .split('')
-      .join('.'),
-  REPLACE:
-    '@@redux/REPLACE' +
-    Math.random()
-      .toString(36)
-      .substring(7)
-      .split('')
-      .join('.')
+  INIT: `@@redux/INIT${randomString()}`,
+  REPLACE: `@@redux/REPLACE${randomString()}`,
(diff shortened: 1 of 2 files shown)


src/utils/actionTypes.js
```diff
@@ -4,21 +4,18 @@
...
-  REPLACE:
-    '@@redux/REPLACE' +
-    Math.random()
-      .toString(36)
-      .substring(7)
-      .split('')
-      .join('.')
+  INIT: `@@redux/INIT${randomString()}`,
+  REPLACE: `@@redux/REPLACE${randomString()}`,
+  PROBE_UNKNOWN_ACTION: () => `@@redux/PROBE_UNKNOWN_ACTION_${randomString()}`
```
theJian: This type has a different format.
It contains an extra `_` before `randomString`.

Nbsaw: @theJian `PROBE_UNKNOWN_ACTION ` already contains '_' before `randomString`.

theJian: I see. Maybe it's a typo.
But I think we can remove it as those types are grouped together now.

Nbsaw: Ok, I understand what you mean.  But I think we can't delete this `_` . 
Because it may have been intentionally added by previous [versions](https://github.com/reactjs/redux/blob/master/src/combineReducers.js#L81). 

PS: I tried to delete it, test can still pass , So I'm not sure if I can delete it . 🤔

src/utils/actionTypes.js
```diff
@@ -15,7 +15,7 @@ const randomString = () =>
 const ActionTypes = {
   INIT: `@@redux/INIT${randomString()}`,
   REPLACE: `@@redux/REPLACE${randomString()}`,
-  PROBE_UNKNOWN_ACTION: () => `@@redux/PROBE_UNKNOWN_ACTION_${randomString()}`
+  PROBE_UNKNOWN_ACTION: () => `@@redux/PROBE_UNKNOWN_ACTION${randomString()}`
```
Nbsaw: @theJian Thank you for your suggestion I have already deleted

timdorr:
Enough folks have already complained about the code duplication, so why not. I don't really think it's a problem, but it's not going to hurt anyone.

PR #2978, opened 2018-05-05
Removes extra containers/App.js file in todomvc example

JoeDuncko:
I was trying to learn from the todomvc example when I realized that there were two `App.js` files. Looks like the one in `/containers/` isn't used, so I am making this PR to delete it.

Tests
- [ ] Navigate to `/examples/todomvc` in your terminal, `npm install`, `npm start`, make sure the todomvc example compiles and runs


JoeDuncko:
Seems like this stuff was last modified in https://github.com/reactjs/redux/pull/2757 by @tomipaul

timdorr:
Thanks!

tomipaul:
That was an oversight. Thanks @JoeDuncko

PR #2985, opened 2018-05-11
Use same return type for both StoreCreator signatures in TS typings

reklawnos:
When performing a refactor, I noticed that removing the `preloadedState` param from a `createStore` call resulted in a different type than calling `createStore` with `preloadedState`. This is due to the return type of `StoreCreator` being slightly different for each signature. Specifically, the one without `preloadedState` does not include a type param for the action type.

This PR adds the action as a type param for the return type of the `createStore` function without a `preloadedState` param, as well as a couple of additional tests in `tests/typescript/store.ts` that make sure the action type param is included in the return type for both signatures.


timdorr:
Thanks!

PR #3002, opened 2018-05-22
[docs] Returning a newState

BorodinDK:
I think that writing `newState` in this case is more correct.
I propose to make it similar to this text https://github.com/reduxjs/redux/blame/master/docs/faq/Actions.md#L61


timdorr:
Thanks!

PR #3003, opened 2018-05-22
Add unpkg field to package.json

andrewbridge:
This allows unpkg.com to point to the UMD build of redux via https://unpkg.com/redux. Currently, this points at lib/redux.js (as it falls back to using the `main` field) which will not run in browsers.


timdorr:
Thanks!

PR #3004, opened 2018-05-22
Fix MiddleWare.md typo

jhaenchen:
grok isn't a word :)


timdorr:
It is: https://en.wikipedia.org/wiki/Grok

jhaenchen:
@timdorr  That's cute and all but maybe we should use words that people actually understand instead of le cute easter eggs in technical docs?

timdorr:
It's not an easter egg. It's a word: https://www.merriam-webster.com/dictionary/grok https://en.oxforddictionaries.com/definition/grok

jhaenchen:
It’s a cutesy developer word that requires someone who learned normal English to look up. It has no place in a technical document.

Jackson "Jake" Haenchen
Jacksonhaenchen@gmail.com
832-423-3342
________________________________

PR #3008, opened 2018-05-27
console.groupEnd doesn't take parameters

plumpNation:
https://developer.mozilla.org/en-US/docs/Web/API/Console/groupEnd


markerikson:
Thanks!

PR #3012, opened 2018-06-01
Change "there's" to "there are" in Learning Resources documentation

talum:
Because "many resources" is plural, I think the sentence should begin with "There are" rather than "There's" or "There is". 

Grammarly reference: https://www.grammarly.com/blog/there-is-there-are/


timdorr:
Thank's!

PR #3016, opened 2018-06-01
Update UsageWithReact.md

pranayaryal:
I thought naming it as 'FilteredTodoList' is closer to what it means than 'VisibleTodoList'. Thanks.


timdorr:
This is meant to match the example in the repository, so it should be left alone (unless you want to update everything!)

PR #3018, opened 2018-06-04
Fix typescript Store interface definition

testerez:
A dispached action does not actually have to be of type `Action`. For example a thunk function does not have a `type` property.


aikoven:
This is done by extending `Dispatch` type with custom signatures, see [tests](https://github.com/reduxjs/redux/blob/v4.0.0/test/typescript/middleware.ts).

PR #3026, opened 2018-06-15
Warn when reducer is not a function

baibhavx:
``` js
  todos: todosReducer, 
  steps: {} // no warning, this is simply ignored 
})
```
- If the object supplied to combineReducers has a key whose value is not a function, it simply ignores it. With this fix, it gives a warning, letting the user know that each key in the object must point to a function.

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -121,6 +121,8 @@ export default function combineReducers(reducers) {
     if (process.env.NODE_ENV !== 'production') {
       if (typeof reducers[key] === 'undefined') {
         warning(`No reducer provided for key "${key}"`)
+      } else if (typeof reducers[key] !== 'function') {
+        warning(`The reducer for "${key}" is not a function`)
       }
     }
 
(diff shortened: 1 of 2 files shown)


baibhavx:
npm_ci_fail
@timdorr I added the tests to my commit but the build fails at `npm ci` with `zlib: unexpected end of file` and I have no idea why that is happening. If possible, please suggest how to make this pass. Thanks!

src/combineReducers.js
```diff
@@ -121,6 +121,8 @@ export default function combineReducers(reducers) {
     if (process.env.NODE_ENV !== 'production') {
       if (typeof reducers[key] === 'undefined') {
         warning(`No reducer provided for key "${key}"`)
+      } else if (typeof reducers[key] !== 'function') {
+        warning(`"${key}" is not a function`)
```
timdorr: This will probably be confusing by itself. I'd add more context to the message: `The reducer for "${key}" is not a function`

timdorr:
Yeah, something is up with Travis. Not your fault, though! I'll see if I can figure it out...

src/combineReducers.js
baibhavx: Good idea! I've updated the error message to ` The reducer for "${key}" is not a function` format.

PR #3032, opened 2018-06-17
Syntax improvement

partha0103:
Modified the default function to arrow function


markerikson:
Appreciate the PR, but it's not an "improvement" exactly.  They both work, and there's no immediate need to switch this.

PR #3033, opened 2018-06-17
From mutable to immmutable reducer

AlvaroBernalG:
Hello good people of `redux`,

This change makes the reducer use computed properties as the rest of the code base. 

Thanks for maintaining this awesome library!.


markerikson:
Hi.  Appreciate the PR, but the immutability isn't actually needed in this example, because the call to `reduce()` takes a new object as its starting point, so it's okay as-is.  Either approach will work okay, so there's no need to make a change here.

PR #3037, opened 2018-06-21
Add the way of using observable

allenleeok:


timdorr:
Thanks, but this probably belongs in the documentation or a separate example, not randomly added to an existing one.

Issue #2587, opened 2017-08-27
Add documentation for setting up a redux store with create-react-app


SeanPlusPlus:
per this thread:

https://twitter.com/acemarke/status/901929091879841792

markerikson:
Great, thanks for offering to help out.

To recap: skimming through the docs, we have the specific API section for `createStore`, and samples of using `createStore` in our actual example projects, but I don't immediately see good examples of how to go about _using_ `createStore` in a typical application.  I'd like to see a recipe page that covers how to write your own `configureStore` function to encapsulate store creation, and lays out a good sequence of steps that users could adapt to fit their own needs.

In particular, I'd like to use the code and approach from my own [Practical Redux, Part 3: Project Setup](http://blog.isquaredsoftware.com/2016/11/practical-redux-part-3-project-planning-and-setup/) post as a starting point and example.

The recipe _should_ be widely applicable to just about any Redux app, not just CRA, since there CRA itself doesn't care about Redux - it just provides a skeleton React app as a starting point.  However, it may be useful to use CRA as an example of where to call `configureStore()` and use the store in a React app.

markerikson:
It would also be good to first review the docs to see what examples of `createStore` we _do_ have, just in case I'm not thinking of something :)

AuroreM:
Hi @markerikson, 
I'm interesting to work on this, do you know if this issue is still "available" ?

timdorr:
Until they're closed, they're always available to be worked on.

markerikson:
Fixed via #3039 .  Thanks!

PR #3039, opened 2018-06-21
Docs(recipes): add configureStore recipe

jtmcgrath:
Recipe doc for creating a `configureStore` function, based on the discussion in issue #2587.

First draft, so probably doesn't flow perfectly and will likely contains some errors - always find it better to revisit after a couple of days to clean things up. If you spot anything which could be improved please point it out, no matter how small (or large) a detail!

Couple of thoughts/questions:

- Not sure whether this is being added to the right place. In a sense it builds on from the basics section, but on it's not really exploring a basic part of redux.
- Could do with closing statement and links to further reading.
- Should this new code be added to the example? Could be a useful addition, but then it would break the flow of the doc (since it links through to the code at the start) - maybe it should be a separate example?
- There's a bit of redundancy in there (info explained in basics section briefly revisited); is it preferred to cater for people who have come to this page directly by including this, or to keep the doc short and include links to the other pages?

### DIFF
--- a/docs/recipes/ConfiguringYourStore.md
+++ b/docs/recipes/ConfiguringYourStore.md
@@ -91,6 +91,8 @@ Let's add these to our  existing `index.js`.
 - First, we need to import `redux-thunk` plus our `loggerMiddleware` and `monitorReducerEnhancer`, plus two extra functions provided by Redux: `applyMiddleware` and `compose`.
 - We then use `applyMiddleware` to create a store enhancer which will apply our `loggerMiddleware` and the `thunkMiddleware` to the store's dispatch function.
 - Next, we use `compose` to compose our new `middlewareEnhancer` and our `monitorReducerEnhancer` into one function.
+
+    This is needed because you can only pass one enhancer into `createStore`. To use multiple enhancers, you must first compose them into a single larger enhancer, as shown in this example.
 - Finally, we pass this new `composedEnhancers` function into `createStore` as its third argument. _Note: the second argument, which we will ignore, lets you preloaded state into the store._
 
 ```js
--- a/docs/recipes/ConfiguringYourStore.md
+++ b/docs/recipes/ConfiguringYourStore.md
@@ -1,0 +1,180 @@
+# Configuring Your Store
+
+In the [basics section](../basics/README.md), we introduced the fundamental Redux concepts by building an example Todo list app.
+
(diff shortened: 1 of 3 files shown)


markerikson:
Great, thanks!  I'll try to review this in the next day or so.

jtmcgrath:
Good idea - added that section!

markerikson:
While we're at it, let's go ahead and show adding `redux-thunk` as a middleware all the way through, since that's the most common.

jtmcgrath:
Sure - as part of the same example, showing how you'd add multiple middlewares?

markerikson:
Might want to add a little bit of clarification there in the "add to existing index.js" section about why `compose` is needed in the first place (you can only pass one enhancer to `createStore`, so multiples need to be composed together into a single larger enhancer).  Not sure whether it's worth showing it with just the `applyMiddleware` enhancer first to show migrating towards `compose` when the second is needed, or just do them both in one example and describe why it's necessary.

jtmcgrath:
Feels like that makes sense as part of the existing flow, no need to add an extra block of code for it. 🙂

markerikson:
One other item we might want to consider:  HMR setup.  Want to throw that in there too?

markerikson:
@jtmcgrath : quick question, do you have a Twitter handle?

jtmcgrath:
Awesome - just ping me if you'd like any changes, happy to help out! 🙂

Nope, don't have any social media at the moment ...keep meaning to fix that but can never find the time. 😛

markerikson:
Hah, sure.  Just wanted to give you credit when I announce the new page.

Really appreciate the contribution and the fast turnaround on suggested changes.  Would love to have you work on some of the other "missing page" issues as well, if you're up for it!

docs/recipes/ConfiguringYourStore.md
```diff
@@ -0,0 +1,311 @@
...
+    )
+  }
+
+  return store
+}
+```
+
+The new code is wrapped in an `if` statement, so it only runs when our app is not in production mode, and only if the `module.hot` feature is available.
+
+We use Webpack's `module.hot.accept` method to specify which module should be hot reloaded, and what should happen when the module changes. In this case, we're watching the `./reducers` module, and passing the updated `rootReducer` to the `store.replaceReducer` method when it changes.
```
eps1lon: Since parcel is becoming more popular and also [supports the same API](https://parceljs.org/hmr.html) this should not only mention webpack.

markerikson: Sure, I'll go ahead and edit that real fast.

markerikson: Done.  Reworded to:

jtmcgrath:
Awesome - I'll take a look when I've got some time! 😃

PR #3045, opened 2018-07-02
Fix typo in redux-saga example

muhajirframe:


muhajirframe:
I am not sure wether we should change `response` -> `result` or `result` -> `response`

timdorr:
Thanks!

PR #3048, opened 2018-07-03
Shouldn't we pass isFetching from redux?

muhajirframe:
Well, I am not sure from where should we get isFetching as no structure provided here.


timdorr:
Thanks!

PR #3052, opened 2018-07-05
Remove a dead link and an archived example

gc:
The link which says Twitter Mobile uses Redux is a dead link (the tweet was deleted), I looked around and couldn't find a reliable source saying Twitter Mobile uses Redux, and other one links to an archived repository which isn't the best example to show. Here are 2 alternate websites which say they are using redux, if they are a suitable replacement for these examples, they are respectably complex applications: https://www.artsy.net / https://www.vistaprint.com

### DIFF
--- a/docs/faq/Miscellaneous.md
+++ b/docs/faq/Miscellaneous.md
@@ -14,10 +14,8 @@
 
 Yes, lots of them!  To name just a few:
 
-- [Twitter's mobile site](https://twitter.com/necolas/status/727538799966715904)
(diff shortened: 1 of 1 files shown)


markerikson:
Weird that the tweet is gone.  I was able to find it archived here:  http://web.archive.org/web/20180519171815/https://twitter.com/necolas/status/727538799966715904

Perhaps a better example would be this article: https://medium.com/statuscode/dissecting-twitters-redux-store-d7280b62c6b1

PR #3053, opened 2018-07-07
Attempt to fulfill the request of sagas-thunks use cases

giacomocerquone:
Fulfilling @markerikson twitter request https://twitter.com/acemarke/status/1013801105162108928
It may be no sense or even garbage, but at least I wanted to give it a try.
If you think it's something acceptable, you can let me know what to improve.


markerikson:
Thanks, appreciate the PR!

Let's see.  My first observation is that I'd prefer to put this in an existing FAQ page.  The "Actions" page makes the most sense, since that's where we already discuss using middleware for async behavior.

Second, I'd suggest condensing this a bit into one larger question+answer:  "What async middleware should I use? How do you decide between thunks, sagas, observables, or something else?".

Go ahead and update the PR to move this content over to the "Actions" page instead, and condense it into one larger question.  We'll see what else needs to be done from there.

giacomocerquone:
Yeah, I agree that placing it into the "Actions" page make more sense! I think I just misinterpreted your tweet :)

Anyway wouldn't be this title more precise?
"What side effect middleware should I use? How do you decide between thunks, sagas, observables, or something else?"
Since middleware are not used just to perform async logic.

PS. pr is coming :)

giacomocerquone:
any news? @markerikson :)

markerikson:
Sorry about that, forgot this was here.  Sure, let's get it in.

PR #3062, opened 2018-07-16
docs: fix `composeWithDevTools` capitalization

czamp:
Should be imported `composeWithDevTools`,  instead of `composeWithDevtools`.

Misspelling appears to compile fine in the console, but the browser returns an error: `TypeError: Cannot read property 'apply' of undefined`


timdorr:
Thanks!

PR #3065, opened 2018-07-18
 Remove unneeded middlewareAPI.dispatch declaration

angular-moon:
`(...args) => dispatch(...args)` equals `dispatch`


angular-moon:
sorry, I made a mistake!

PR #3067, opened 2018-07-20
Add types for current Observable interpolation interface

### LINKED ISSUE #1723: typings for store.observable

pinyin:
closes #1723

According to #1725, the issue was closed because TypeScript wasn't supporting symbol-named properties at that time, which in my understanding is no longer the case.

Is it the time to add the types now?


timdorr:
Yep, it's a good time. If anyone else wants to chime in with feedback, feel free. But I think this looks good as-is.

timdorr:
Sorry this took so long, but I think it's OK. Thanks!

Issue #2591, opened 2017-08-28
Improve docs related to UI layers like React-Redux


markerikson:
We've got a section on use with React, but it feels a bit weak to me at the moment.  There's several aspects of `connect()` that aren't covered in there, like the object shorthand form of `mapDispatch`, using factory functions, and more.

The React-Redux API docs are also kept separate, and live over in the React-Redux repo itself, at https://github.com/reactjs/react-redux/blob/master/docs/api.md .  Now, I do fully agree that Redux is UI-layer agnostic, but given that React-Redux is the one "official" set of bindings that we maintain, I'd like to add more detail on how to actually use those correctly into our main docs.

In addition, it would be great if we could add some kind of "Use with other UI libraries" section that gives some details on how to use Redux with Angular, Ember, Vue, Backbone, or jQuery.

bgates:
What's the appropriate level to pitch this section (and the rest) of the documentation? I saw your comment in #2295 that _Docs are written in a deliberately verbose style for clarity and learning_; I spent the last few years at The Iron Yard, sometimes introducing Redux to people who were in their 12th week of learning JavaScript, so I'm biased towards even more verbosity. 

What would you think about an expansion of this section, to include
* motivation for using a state management library with React (what React specifically gains from Redux)
* what a higher-order component is and why it's the best choice here
* description of `connect` as a function returning a function
* preceding the todo list example with a smaller example of a single container wrapping a presentational component
* more discussion around `mapStateToProps` and `mapDispatchToProps` (more verbose explanations of what the names mean; how they're just names, useful bc they're descriptive, but either could be replaced by `pickles` if you were consistent and things would still work)
* some discussion of the "shortcuts and power options", though possibly in a different document?

markerikson:
@bgates : those are some very interesting questions and ideas.  (Also, given your background, I would actually like to pick your brain when you have some time free, and discuss your experience teaching Redux to early learners.)

When I wrote this issue, I was really thinking more in terms of adding additional sections, like a trimmed-down version of the React-Redux API docs, and a section on using Redux with other libraries like Angular or Ember.  However, if you have ideas for improving the current [Basics: Usage with React](http://redux.js.org/docs/basics/UsageWithReact.html) page, I'm definitely interested.

bgates:
I'd love to talk about early stage education @markerikson - what's a good channel/time?

I'll spend some time this weekend sketching some thoughts out and getting them up as PRs, if that's ok.

markerikson:
[...]

PR #3070, opened 2018-07-24
The Medium article should be optional reading.

travishohl:
It's great to provide links to more in-depth material, but what if we made it seem more optional? It might be better to allow the reader to keep their momentum here.

This is loosely related to #2591.


timdorr:
Seems like a good change to me.

PR #3077, opened 2018-07-29
fix error when middleware is not a function

DmitryAlekseev:


src/applyMiddleware.js
```diff
@@ -30,7 +30,11 @@ export default function applyMiddleware(...middlewares) {
       getState: store.getState,
       dispatch: (...args) => dispatch(...args)
     }
-    const chain = middlewares.map(middleware => middleware(middlewareAPI))
+    const chain = middlewares.map(middleware => {
+      if (typeof middleware === 'function') {
+        return middleware(middlewareAPI)
+      }
```
theJian: If middleware is not a function then it'll return `undefined`. The result is a chain contains `undefined`. It throws error when you're trying to compose such a chain.

DmitryAlekseev: If I use th construction like this:
const middleware = [__DEV__ && createLogger()];

I spent two days to find why my react-native app crashed on start in the production mode...

timdorr: Yes, but this change won't fix your problem. You also didn't add a test, so you wouldn't see this fail.

timdorr:
Sorry, but I'd rather keep this function simpler. 

If you need to programmatically add middleware, do so without the && shorthand:
```js
const middleware = [
  middleware1,
  middleware2
]

if (__DEV__) middleware.push(devMiddleware)
```

markerikson:
We've had a bunch of PR churn around handling undefined values in our various utility methods.  I'm still not sure whether we ought to be throwing errors in those cases to pre-emptively warn people, or silently swallowing `undefined` values.

PR #3078, opened 2018-07-30
Fix combineReducers example?

rmorrise:
I am totally new to redux, but I was scratching my head over this code example. Please verify whether this edit is correct.


eps1lon:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Property_definitions
See section about shorthand property names (ES2015)

theJian:
Welcome to JavaScript world!

markerikson:
Yep, that's ES6 object shorthand syntax, and it's valid.

PR #3080, opened 2018-07-31
Mark StoreCreator's preloadedState argument as optional

srittau:
When wrapping `createStore()` using typescript, the typescript definition prevents an optional `preloadedState` argument:

```typescript
function myCreateStore<S>(reducer: Reducer<S>, preloadedState?: Redux.DeepPartial<S>) {
    return createStore(reducer, preloadedState);
}
```

This will complain that type `undefined` is not assignable to `DeepPartial<S>`. Marking the `preloadedState` argument as optional fixes this problem.


timdorr:
Interesting that was overlooked, since `StoreEnhancerStoreCreator` already has an optional preloadedState. Funky.

Thanks for the one-character fix!

PR #3082, opened 2018-08-01
update typescript definition for Action interface

cyberhck:
set default type to string
I think we should set default to what's recommended, `any` typing in TypeScript is a bad practice.


cyberhck:
had a look at tests, two failing, fixed one, and one I'm stuck at. middleware.ts @ custom dispatch.

timdorr:
I was able to make this work by allowing MiddlewareAPI to override the Action used for the Dispatch constraint. There's probably another way to do this automatically, but I'm not sure. At the very least, it's backwards compatible.

Any other TS experts want to chime in?

Issue #3085, opened 2018-08-04
Cache `finalReducerKeys.length` to optimize performance in `combineReducers`


KJlmfe:
## Problem:

In `src/combineReducers.js`,  the `finalReducerKeys.length` won't change. So to optimize performance, it would be better to cache it.

```js
export default function combineReducers(reducers) {
  const reducerKeys = Object.keys(reducers)
  const finalReducers = {}
  // omit ....
  const finalReducerKeys = Object.keys(finalReducers)

  // omit ....
  return function combination(state = {}, action) {
  // omit ....
    let hasChanged = false
    const nextState = {}
    // `finalReducerKeys.length` could be cached to optimize performance
    for (let i = 0; i < finalReducerKeys.length; i++) { 
      // omit ....
    }
    return hasChanged ? nextState : state
  }
}
```

## Suggestion:

```js

export default function combineReducers(reducers) {
  const reducerKeys = Object.keys(reducers)
  const finalReducers = {}
  // omit ....
  const finalReducerKeys = Object.keys(finalReducers)
  const finalReducerKeysLength = finalReducerKeys.length // cache it

  // omit ....
  return function combination(state = {}, action) {
  // omit ....
    let hasChanged = false
    const nextState = {}
    for (let i = 0; i < finalReducerKeysLength; i++) {  // use cache
      // omit ....
    }
    return hasChanged ? nextState : state
  }
}

```

PR #3086, opened 2018-08-04
Cache finalReducerKeys.length to optimize perf

KJlmfe:
In src/combineReducers.js, the finalReducerKeys.length won't change. So to optimize performance, it would be better to cache it.

This is related to #3085.


timdorr:
Thanks, but in v8 at least (probably others) this wont have a performance impact. There is a fast path optimization for `.length` access on Arrays, so it essentially compiles down to this form anyways.

PR #3089, opened 2018-08-08
introduce listeners freeze functionality

jevgenig:
**NB** this PR is a call to discussion, as it's something that might or might not become something usable for wide masses

**Problem**

It is not possible to control notifications to subscribers.

**Motivations**

- In some business cases it should be possible to freeze UI.
- Bunching of actions becomes easy controllable by `dispatch(freeze);dispatch(do1);dispatch(do2);dispatch(unfreeze)`

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -51,6 +51,7 @@ export default function createStore(reducer, preloadedState, enhancer) {
   let currentListeners = []
   let nextListeners = currentListeners
   let isDispatching = false
+  let isFrozen = false
 
   function ensureCanMutateNextListeners() {
     if (nextListeners === currentListeners) {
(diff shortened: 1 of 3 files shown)


jevgenig:
Thanks for comment. This is exactly the problem.
It is not possible to do that using middleware, as actions need to continue being dispatched
It is kinda possible, but in a very clumsy way, to implement it using enhancers - by wrapping `subscribe` functionality, with partial copy of the redux's original `subscribe` logic

But I might be really missing something, and will close/revoke this PR if find a proper way to do that.

timdorr:
Look into a store enhancer (which is what applyMiddleware is), which definitely can do this.

jevgenig:
Many thanks for a hint! Works like a charm. Closing PR.

For future explorers:
```
const freeze = { type: "FREEZE" };
const unfreeze = { type: "UNFREEZE" };

function applyFreezableStore() {
    return createStore =>
        (...args) => {
            const store = createStore(...args);
            let isFrozen = false;
            const subscribe = listener => store.subscribe(() => !isFrozen && listener());
            const dispatch = (...a) => {
                switch (a[0].type) {
                    case freeze.type:
                        isFrozen = true;
                        return;
                    case unfreeze.type:
                        isFrozen = false;
                        return;
                    default:
                        return store.dispatch(...a);
                }
            };

            return {
                ...store,
                subscribe,
                dispatch,
            };
        };
}

```

PR #3091, opened 2018-08-09
Docs - consistent teaching in Reducers

edmulraney:
The Reducers doc begins by stating: 
> Remember that actions only describe *what happened*, but don't describe how the application's state changes.

But then it goes on to create actions which don't describe what happened but instead describe the intended state change. For example: `SET_VISIBILITY_FILTER` vs `VISIBILITY_FILTER_CLICKED`.

For consistency you could remove that initial paragraph, or update the actions to describe what happened.

### DIFF
--- a/docs/basics/Reducers.md
+++ b/docs/basics/Reducers.md
@@ -82,12 +82,12 @@ function todoApp(state = initialState, action) {
 }
 ```
 
-Now let's handle `SET_VISIBILITY_FILTER`. All it needs to do is to change `visibilityFilter` on the state. Easy:
+Now let's handle `VISIBILITY_FILTER_CLICKED`. All it needs to do is to change `visibilityFilter` on the state. Easy:
 
 ```js
 function todoApp(state = initialState, action) {
   switch (action.type) {
-    case SET_VISIBILITY_FILTER:
+    case VISIBILITY_FILTER_CLICKED:
       return Object.assign({}, state, {
         visibilityFilter: action.filter
       })
@@ -115,13 +115,13 @@ Note that:
 
 ## Handling More Actions
 
-We have two more actions to handle! Just like we did with `SET_VISIBILITY_FILTER`, we'll import the `ADD_TODO` and `TOGGLE_TODO` actions and then extend our reducer to handle `ADD_TODO`.
+We have two more actions to handle! Just like we did with `VISIBILITY_FILTER_CLICKED`, we'll import the `TODO_SUBMITTED` and `TODO_CLICKED` actions and then extend our reducer to handle `TODO_SUBMITTED`.
 
 ```js
 import {
-  ADD_TODO,
-  TOGGLE_TODO,
-  SET_VISIBILITY_FILTER,
+  TODO_SUBMITTED,
+  TODO_CLICKED,
+  VISIBILITY_FILTER_CLICKED,
   VisibilityFilters
 } from './actions'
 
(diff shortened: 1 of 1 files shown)


timdorr:
You're thinking in terms of UI. "What happened" is in the context of application state. There may be multiple inputs to an action, which is all opaque to the store. For instance, this may be a collaborative view and the filter was set by a remote client from a WebSocket message. 

Essentially, you've turned it into "How it happened", which is what we're not going for with the action names.

edmulraney:
Thanks @timdorr :)

IMO, that make even less sense from an application state perspective. I can't think of an occasion where I'd prefer opaqueness over transparency for the store. In any system I believe you benefit from transparent actions which describe what's actually happened in the system, rather than what the developer intends the state change to be. Dispatching an action with the word `SET_A_B_C` has no guarantee that `A_B_C` will be affected but that is the hope.

With a `SET_DATA` action type approach you're describing something that *hasn't* happened yet, unless you do some mental gymnastics to think of the dispatching of an action occurring after the state mutation has occurred. i.e. the setting of the visibility filter doesn't happen until the action is handled by the reducer. 

The websocket example is exactly what made me think that this doc should be updated. If I were to receive a websocket visibility-filter update I'd expect a separate action such as `VISIBILITY_FILTER_RECEIVED`. Both events would be handled by the same reducer:

```
case VISIBILITY_FILTER_CLICKED: 
case VISIBILITY_FILTER_RECEIVED:
... do the state update
`

`VISIBILITY_FILTER_CLICKED` isn't describing how the state change has happened either (which is good).

An action like `SET_VISIBILITY_FILTER` seems pretty harmless on its own in a small PoC like this. The problem is if that sets a precedent for `SET_X_DATA`'s throughout a real world app, which leads to pretty much imperative style action orchestration: 

```
function someUiSpecificAction() {
  dispatch('SET_VISIBILITY_FILTER', ...)
  dispatch('SET_ANALYTICS_EVENT', ...)
  dispatch('SET_AUDIT_EVENT', ...)
  dispatch('SET_MODAL_CLOSE', ...)
  dispatch('SET_NOTIFICATION_SUCCESS', ...)
}
```
vs
```
function someUiSpecificAction() {
   dispatch('SOME_UI_SPECIFIC_THING_HAPPENED`)
}
```
and then that action being handled in the relevant reducers/middleware.

I thought this was one of Dan's intentions for Redux :)

markerikson:
@edmulraney : please see my thoughts at https://blog.isquaredsoftware.com/2017/05/idiomatic-redux-tao-of-redux-part-2/#action-semantics , and the related links at the bottom of the post.

edmulraney:
@markerikson awesome article! Would be nice to see the docs link to it. 

It looks like the Reducers doc has taken the first half of the quote from Dan:
>The point is to decouple “what happened” from “how the state changes”. “Setter” action creators defeat the purpose, conflating the two.

and omitted the second half which effectively says don't use setter actions (which this doc does). 

Totally get that this is bikesheddy and not worth a long discussion - just thought I'd highlight it as its a doc which new people learning redux will use and it might be better not to have conflicting information. 

Path of least resistance seems to be to omit the first paragraph rather than rename the actions.

markerikson:
@edmulraney : both Part 1 and Part 2 are linked directly in the Redux README :)

Honestly, both "setter"-named actions and "event"-named actions are viable approaches, and I've used both depending on the situation.  There's enough new concepts already in the tutorial - I don't want to make things more complicated and confuse beginners further with a bikeshed discussion there.

edmulraney:
@markerikson I totally agree with you - I think I'm making my case really badly! Let's pretend I prefer the setter-action approach - I would still raise an issue or PR. The issue is that the doc first says ` "Remember that actions only describe what happened, but don't describe how the application's state changes."` (event approach) but then goes on to use the setter approach.  

For a beginner that's just read that actions describe "what happened" to then be shown setter-actions would be confusing.

I hope it's clear that I'm not pushing for this PR to be merged... the point is to resolve the conflicting information. If it doesn't seem conflicting to you - ignore me! :)

PR #3098, opened 2018-08-14
Fix todos-flow example

### LINKED ISSUE #2717: todos-flow example is broken

Bug

In `examples/todos-flow`, the Flow typings are not working correctly.

Just to be clear, this bug isn't about Flow features which don't exist or aren't possible. They used to work perfectly fine, but are currently broken. **The typings work correctly on Flow 0.54, however they do not work on Flow 0.55 and above. The current version is 0.59.**

`TodoList` has a [required prop](https://github.com/reactjs/redux/blob/master/examples/todos-flow/src/components/TodoList.js#L10) `todos` with type `Todos`, with the [following definition](https://github.com/reactjs/redux/blob/master/examples/todos-flow/src/types/todos.js):

```
export type Todo = {
  +id: Id,
  +text: Text,
  +completed: boolean
};

export type Todos = Array<Todo>;
```

However, if you change [`mapStateToProps`](https://github.com/reactjs/redux/blob/master/examples/todos-flow/src/containers/VisibleTodoList.js#L14) in `src/containers/VisibleTodoList.js` to either not provide the `todos` prop, or provide an invalid type, `yarn flow` will incorrectly report `No errors!`.

This incorrectly reports no errors, even though `TodoList` is not receiving a `todos` prop and will result in a runtime error:
```
const mapStateToProps = (state: State) => {
  return {
    NOTtodos: visibleTodosSelector(state)
  };
};
```

This incorrectly reports no errors, even though `TodoList` is receiving a `todos` prop that does not match the `Todos` type, and will result in a runtime error:
```
const mapStateToProps = (state: State) => {
  return {
    todos: "this is a string, not a an array of Todos, but Flow doesn't care"
  };
};
```

To reiterate, both these examples **correctly report errors** on an older version of Flow, 0.54.

Simply change `mapStateToProps` in `src/containers/VisibleTodoList.js` to either of the two examples above, and run `yarn flow`.

Both examples above should result in Flow reporting an error.

This became broken in Flow 0.55. Both `mapStateToProps` examples above **correctly report errors in Flow 0.54**.

I have reported this issue to both Flow and flow-typed, however I wanted to raise awareness on the main Redux repo that Flow typings are currently broken in what is a canonical example in this repo.

I personally think Flow's strengths really shine on a large Redux driven codebase, and Redux support in Flow would ideally be treated almost as much as a first class citizen as React is itself. It's a shame to see this broken for nearly three months now.

https://github.com/facebook/flow/issues/5343

colesimmons:
Fix #2717 

It appears updating all dependencies to the most recent fixed the problem.

### DIFF
--- a/examples/todos-flow/package-lock.json
+++ b/examples/todos-flow/package-lock.json
@@ -1,13 +1,46 @@
 {
   "name": "todos-flow",
-  "version": "0.1.0",
+  "version": "0.0.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@octokit/rest": {
+      "version": "15.9.5",
+      "resolved": "https://registry.npmjs.org/@octokit/rest/-/rest-15.9.5.tgz",
+      "integrity": "sha512-vJEHSTnI4UAbCDTjVSQljPeX81zsQVNj2ruM5Oj5gxOttHD0TcfWeElcJYoITCMxQTgN6Y+bJFo6/+/0CqoacA==",
+      "dev": true,
+      "requires": {
+        "before-after-hook": "^1.1.0",
+        "btoa-lite": "^1.0.0",
+        "debug": "^3.1.0",
+        "http-proxy-agent": "^2.1.0",
+        "https-proxy-agent": "^2.2.0",
+        "lodash": "^4.17.4",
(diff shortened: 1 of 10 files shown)


timdorr:
Thanks!

PR #3108, opened 2018-08-23
use a variable to save reducers[key] in for loop

MinimalistYing:
Maybe this way is more human-readable.
And gain a little bit benefit for performance.

Oops, this didn't pass the CI because below files.
test\createStore.spec.js 
test\compose.spec.js 

I found this is not my fault, even the master branch didn't pass format:check


markerikson:
Appreciate the PR, but I don't think there's enough actual benefit here to make any changes.

PR #3111, opened 2018-08-25
Update ImmutableUpdatePatterns.md

mohsen158:


markerikson:
This doesn't look right.  In that example, `second` is intended to be an object that uses item IDs as keys, not an array.

PR #3118, opened 2018-09-02
Minor improvement on Basic -> Reducers doc.

TurtleShip:
This is a small PR that changes

> All combineReducers() does is _**generate**_ a function that calls your reducers with the slices of state selected according to their keys, and **combining** their results into a single object again.

to

> All combineReducers() does is _**generating**_ a function that calls your reducers with the slices of state selected according to their keys, and **combining** their results into a single object again.

Sorry for my OCD.

### DIFF
--- a/docs/basics/Reducers.md
+++ b/docs/basics/Reducers.md
@@ -367,7 +367,7 @@ function reducer(state = {}, action) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


markerikson:
Hi.  Appreciate the PR, but if the grammar needs to be tweaked, I think it actually needs to go the other way:  "generate" ... "combines".  If you can tweak the PR that way, I can get it in.

PR #3120, opened 2018-09-04
Optimize isPlainObject

hfuuss:
When I saw the isPlainObject function, I found it could be optimized. And did some related tests.

### DIFF
--- a/src/utils/isPlainObject.js
+++ b/src/utils/isPlainObject.js
@@ -5,10 +5,5 @@
 export default function isPlainObject(obj) {
   if (typeof obj !== 'object' || obj === null) return false
 
-  let proto = obj
-  while (Object.getPrototypeOf(proto) !== null) {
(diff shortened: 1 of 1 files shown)


markerikson:
@timdorr , can we get some clarification on what the specific point of the `while` loop is?  I'm sure you had a good reason for putting it there.

Prior discussion / origination: #2599 .

markerikson:
Looks like the specific comment is https://github.com/reduxjs/redux/pull/2599#issuecomment-342849867 : 

> Just so others understand its purpose: It handles cross-realm objects by walking up the prototype chain until it reaches Object and checks if that "Object" has the same prototype as our potential plain object.

> In the case of something coming from an iframe, that Object is different than your own local Object because they come from two different JavaScript contexts. So, we use this technique to get to that definition of Object for the object instance.

Based on that, I think we want to leave it as-is.  Closing.

PR #3124, opened 2018-09-06
Refactor the for loop with ES6 style

sumleo:
Read the code clearly


ematipico:
`for...in` it's more for objects, not arrays.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

timdorr:
Thanks, but this treats arrays as objects, which isn't ideomatic JavaScript.

PR #3132, opened 2018-09-12
Reword UsageWithReact docs wording

BlueHatbRit:
I stumbled over reading "that tells how" and found it a bit awkward, I also think it might be grammatically incorrect though I'm no expert. I've reworded it to "that describes how" which should be equally accessible wording but flows a bit easier.

I went through a few different variations and thought this was the most accessible wording while making sure it still flowed well.


markerikson:
Sure, that works.  Thanks!

PR #3137, opened 2018-09-15
WIP - Examples todomvc with ImmutableJs

arojunior:
Add example with ImmutableJs.

Actually I was working on it a long time ago and today I remembered that I didn't finish. If someone else would help me out, it will be great.


timdorr:
Thanks, but I don't think we need another example specific to a particular immutability library. In fact, I'd like to remove some examples, as we've got too many already.

PR #3139, opened 2018-09-18
Switched for loop with foreach

catcpp:


markerikson:
Appreciate the PR, but there's no reason to actually change this.

catcpp:
I don't know, maybe less code (1 line, lol) :D 
Also, I think forEach is more cute than for

PR #3142, opened 2018-09-20
move Header component out of Header container file

joshalling:
I was going through the todomvc example and I noticed that the Header container file also included the Header component. I moved it and its test into the components directory to clean it up a little bit.

- [x] Navigate to /examples/todomvc in your terminal, npm install, npm start, make sure the todomvc example compiles and runs
- [x] Navigate to /examples/todomvc in your terminal, npm test, and make sure all tests pass


timdorr:
Thanks!

PR #3143, opened 2018-09-20
Add ES-browser build

### LINKED ISSUE #2907: Remove usage of `process.env.NODE_ENV` in ES module build

TimvdLippe:
This build is intended for users who want to use Redux in a browser with ES modules support. The build removes all references to `process` by assuming it is in a production environment.

Additionally, it does not transpile, as all browsers that have ES modules support also support ES2015+

This approach is similar to the one of Vue (https://github.com/vuejs/vue/pull/7110) which I have also been using with success.

Fixes #2907


timdorr:
Hey Tim, it's Tim 😄 	

I'm wondering if we can keep this in the `es` path, but output redux.browser.js or something like that.

What happens with `symbol-observable` in this bundle?

TimvdLippe:
@timdorr Updated, PTAL :smile:

timdorr:
I'm actually mucking about with the babel and rollup configs right now. I'll have a conflict ready for you in just a second...

TimvdLippe:
@timdorr Conflict should be busted :tada:

TimvdLippe:
Hm, let me actually minify just like the minified umd build.

timdorr:
Cool. I renamed it to just redux.mjs for simplicity. Forgot about that extension.

TimvdLippe:
Thanks for the quick review @timdorr ! Can't wait for the next release :tada:

TimvdLippe:
@timdorr What is the ETA of a release of a new version of Redux which includes this PR?

Andarist:
According to ES modules support taken from here - https://github.com/babel/babel/blob/6b8261bbb7ebb8d2980cb30c3b3369678fd99835/packages/babel-preset-env/data/built-in-modules.json#L1-L11 and testing only desktop browsers quickly it seems that used object rest spread syntax in the es/redux.mjs is not supported in those verstions of Edge & Safari.

Depending on what the audience of this bundle is it might need to get adjusted.

Issue #3090, opened 2018-08-08
TOC links in docs don't work


gaearon:
When I click on any table of contents link on e.g. https://redux.js.org/faq/general, nothing happens. In general it seems like anchors are broken?

markerikson:
Are you talking about the hand-written ones at the top of the FAQ pages, or the auto-generated ones on the side?

The hand-written ones have been busted since the Gitbook upgrade a few months ago.  @SamyPesse said they were going update their parser to account for those, but nothing's happened.  

Technically this is a dupe of #3061 .

The auto-generated TOC is actually kind of nice, but my main reason for doing the hand-written ones in the first place is that it's a lot nicer to paste in https://redux.js.org/faq/actions#actions-side-effects instead of https://redux.js.org/faq/actions#how-can-i-represent-side-effects-such-as-ajax-calls-why-do-we-need-things-like-action-creators-thunks-and-middleware-to-do-async-behavior , so it'd be nice if we had those working again.  (Or, a way to force the auto-generated TOC to use the shorter versions.)

Nicholas-Eng:
Hi, I think this issue has to do with the "id" and lack of "name" attributes used on the <a/>. See this [post ](https://stackoverflow.com/questions/5319754/cross-reference-named-anchor-in-markdown/7335259#7335259) for more information.

Is it alright if I take this one as my first issue?

markerikson:
@Nicholas-Eng : if you can figure out a way to make it work, great!  Unfortunately, we don't have any kind of PR changes preview set up for the docs, and my understanding is that Gitbook does more than "just" a standard Markdown transform, so I'm not sure there's a way to verify that changes actually fix the issue other than merging them in.

Based on prior discussions, it sounds as if this is a weakness in Gitbook's transformation logic, and they haven't gotten around to fixing it.

Nicholas-Eng:
I agree. This looks more like an issue on Gitbook's end. I do see that the ids are missing from the anchor tags in the auto generated HTML, so that's why they aren't navigating to the sections.

gaearon:
Shall we remove all TOCs then? It's super confusing.

gaearon:
Also why does it attempt to load Mixpanel? Are we logging analytics for Redux doc readers? Why?

Nicholas-Eng:
I would recommend leaving the right pane ToC and either commenting out or removing the hand-written ToC for now. Maybe once the issue is resolved with Gitbook then bring back the hand-written ToC.

[...]

PR #3144, opened 2018-09-21
Update General.md

lazaroalvarenga:
Updated index links to do an anchor action. The <a> Tag was empty and would not work. Links (###) also work with a "slugify" version of your destiny titles.

This PR belongs to issue #3090


timdorr:
Thanks!

PR #3146, opened 2018-09-21
Update compose.md

mrassili:
The `createStore` function accepts 3 arguments, the second argument (initial state) being optional was omitted without providing an `undefined`

### DIFF
--- a/docs/api/compose.md
+++ b/docs/api/compose.md
@@ -25,6 +25,7 @@ import reducer from '../reducers'
 
 const store = createStore(
   reducer,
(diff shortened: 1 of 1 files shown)


timdorr:
Actually, we check for this and allow it: https://github.com/reduxjs/redux/blob/master/src/createStore.js#L32

mrassili:
Fair enough, I'm new to Redux so please forgive my ignorance

timdorr:
No worries. We're weirdly accommodating with our API :)

PR #3151, opened 2018-09-27
fix: Throw error if createStore is passed several enhancers

### LINKED ISSUE #3114: Add sanity check for accidentally passing two separate enhancers?
I've occasionally seen people passing two separate store enhancers to `createStore()` (as seen in [this SO question](https://stackoverflow.com/q/52046340/62937) ).  This is always an error, and it seems like we could add an additional sanity check to `createStore()` - probably just checking to see if we have 3 args, and if the second and third are both functions.  (I would think we can safely assume that the `preloadedState` argument shouldn't be a function.)  It's a rarer problem, but it's preventable.

Thoughts?

selbekk:
This commit adds a check for whether the user is passing several
enhancers to the `createStore` function.

Fixes #3114.

### DIFF
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -29,6 +29,17 @@ import isPlainObject from './utils/isPlainObject'
  * and subscribe to changes.
  */
 export default function createStore(reducer, preloadedState, enhancer) {
+  if (
+    (typeof preloadedState === 'function' && typeof enhancer === 'function') ||
+    (typeof enhancer === 'function' && typeof arguments[3] === 'function')
+  ) {
+    throw new Error(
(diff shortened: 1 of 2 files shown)


selbekk:
I added two checks - one if you're passing something like 

```js
createStore(reducer, func, anotherFunc) // throws
```

, and another one if you pass initial state as well:

```js
createStore(reducer, { initial: 'state' }, func, anotherFunc) // throws
```

I'm up for changing whatever - but I'd love some feedback first!

markerikson:
Looks good to me.

@timdorr , any objections?

timdorr:
LGTM. Thanks!

PR #3152, opened 2018-09-27
Pass dispatch without wrapping

kkirby:
I'm not sure if there was a reason we are wrapping dispatch. If there is, then close my PR. I was just looking over the code and saw this could be simplified.

### DIFF
--- a/src/applyMiddleware.js
+++ b/src/applyMiddleware.js
@@ -28,7 +28,7 @@ export default function applyMiddleware(...middlewares) {
 
     const middlewareAPI = {
       getState: store.getState,
-      dispatch: (...args) => dispatch(...args)
+      dispatch
     }
     const chain = middlewares.map(middleware => middleware(middlewareAPI))
     dispatch = compose(...chain)(store.dispatch)


timdorr:
Yes, there is a reason. It's so that during middleware setup, they don't attempt to do something silly like try to overwrite `dispatch` with their own value. Because we pass it in as an object, it's a reference to the original value, so you can do bad things with it.

markerikson:
I think we need to add a comment to `applyMiddleware` that says "SECRET_MIDDLEWARE_IMPLEMENTATION_DO_NOT_MODIFY_OR_YOU_WILL_BE_FIRED".

kkirby:
@markerikson I was going to originally open an issue asking if it was wrapped for a reason, but since the issue wasn't a question or a bug, I decided to make a PR and see what the response would be.

markerikson:
Hah, it's okay, it's not you specifically :)  We've just gotten a _lot_ of PRs over the years to modify `applyMiddleware` and `compose` because people either A) don't understand exactly how they work, or B) feel the need to tweak the implementation in some way.

PR #3155, opened 2018-10-01
Switch examples to Babel 7, Jest 23, and Webpack 4

Timer:
This upgrades all examples using `react-scripts` to v2. All examples were tested by hand and are working correctly.

### DIFF
--- a/examples/async/package-lock.json
+++ b/examples/async/package-lock.json
@@ -4,10 +4,1183 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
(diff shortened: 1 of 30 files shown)


timdorr:
Looks like the `async` example fails because it doesn't have any tests. I guess that produces an error in react-scripts now. We should probably [blacklist that package from the test script](https://github.com/reduxjs/redux/blob/master/examples/testAll.js#L21).

gaearon:
Or we could write a few tests :)

timdorr:
Very nice! I agree we should write some tests, but this at least makes it explicit for now. Thank you!

PR #3156, opened 2018-10-02
Force node environment for Jest

TimvdLippe:
This resolves NPM installations that install Redux via a git branch (e.g. `npm install reduxjs/redux#master`). I know that this generally frowned upon, but I wanted to test the latest version that includes the `es-modules` compatible distribution.

Note that I also tried to set the `globals` option in the `package.json#jest`, however the `NODE_ENV` in `.babelrc.js` would still be `production`. Updating the `test` script in `package.json` does force the correct environment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "format:check": "prettier --list-different \"{src,test}/**/*.{js,ts}\" index.d.ts",
     "lint": "eslint src test",
     "pretest": "npm run build",
-    "test": "jest",
+    "test": "NODE_ENV=test jest",
     "test:watch": "npm test -- --watch",
     "test:cov": "npm test -- --coverage",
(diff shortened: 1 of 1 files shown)


timdorr:
I'm confused why this is necessary. [According to Jest:](https://jestjs.io/docs/en/getting-started)

> Note: If you are using a more complicated Babel configuration, using Babel's env option, keep in mind that Jest will automatically define NODE_ENV as test. It will not use development section like Babel does by default when no NODE_ENV is set.

So, this would seem to be redundant, no?

TimvdLippe:
Yes, that was why I was confused as well. However, if you add a `console.log(NODE_ENV)` in `.babelrc.js`, it prints out `production` during `npm install`. This might actually be a bug in `jest` though?

PR #3157, opened 2018-10-03
docs(faq): fix faq toc general anchor links

xphong:
Fixes the TOC links under general with the correct anchor tags


timdorr:
Thanks!

PR #3158, opened 2018-10-03
fix(docs): fix toc links in faq and recipes

xphong:
Fixes the TOC links under faq and recipes with the correct anchor tags


timdorr:
Thanks!

PR #3159, opened 2018-10-05
fix(docs): fix broken links in FAQ toc

xphong:
Fixes the links under actions and immutable data on https://redux.js.org/faq currently linking to a 404 GitHub page.


timdorr:
Thanks once more!

PR #3166, opened 2018-10-09
Update Store.md

Santi-san:


timdorr:
Thanks, but this is fine as-is.

PR #3168, opened 2018-10-10
Use @babel/runtime helpers instead of bundled ones

Andarist:
Main focus of this PR is to use @babel/runtime helpers instead of including their copy in distributed sources.

Additionally I've refactored rollup.config.js to more readable form (in my opinion). I think it's easier to see how different bundles are different now, when their definitions are way shorter at the bottom of the file. If you don't like it I may revert the change.

**Motivation**
Shipping less code - by using @babel/runtime helpers we are able to deduplicate their code across all dependencies using them.

Quick test on the size of `es/redux.js`. It's not perfect because I have not minified away dev-only warnings, but I didn't do that for both cases (pre-PR & post-PR) so I think that results are still comparable.

Using
```sh
npx terser -cm --toplevel es/redux.js | gzip -c | wc -c
```
I got 2958 pre-PR & 2862 post-PR. This is nearly 100b reduction & this includes a rather lengthy reference to the imported module (`@babel/runtime/helpers/esm/objectSpread`) which costs approximately 25b, so in fact this change saves 100b+ (assuming ESM code can be inlined in user bundles thanks to module concatenation aka scope hoisting). 💰

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -6,23 +6,41 @@ import { terser } from 'rollup-plugin-terser'
 
 import pkg from './package.json'
 
-const noDeclarationFiles = { compilerOptions: { declaration: false } };
+const noDeclarationFiles = { compilerOptions: { declaration: false } }
+
+const babelRuntimeVersion = pkg.dependencies['@babel/runtime'].replace(
+  /^[^0-9]*/,
+  ''
+)
+
+const makeExternalPredicate = externalArr => {
+  if (externalArr.length === 0) {
+    return () => false
+  }
+  const pattern = new RegExp(`^(${externalArr.join('|')})($|/)`)
+  return id => pattern.test(id)
+}
 
 export default [
   // CommonJS
   {
     input: 'src/index.ts',
     output: { file: 'lib/redux.js', format: 'cjs', indent: false },
-    external: [
+    external: makeExternalPredicate([
       ...Object.keys(pkg.dependencies || {}),
       ...Object.keys(pkg.peerDependencies || {})
-    ],
+    ]),
     plugins: [
       nodeResolve({
         extensions: ['.ts']
       }),
       typescript({ useTsconfigDeclarationDir: true }),
-      babel()
+      babel({
+        plugins: [
+          ['@babel/plugin-transform-runtime', { version: babelRuntimeVersion }]
+        ],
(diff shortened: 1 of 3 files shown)


rollup.config.js
```diff
@@ -5,100 +5,82 @@ import { terser } from 'rollup-plugin-terser'
...
-      }),
-      babel({
-        exclude: 'node_modules/**'
-      }),
-      replace({
-        'process.env.NODE_ENV': JSON.stringify('development')
-      })
-    ]
-  },
+const createConfig = ({
```
timdorr: I don't want to use factory functions here. Please keep it as a simple config file. It's harder to read through and figure out what the actual output is this way. Yes, it's more verbose, but it's also much more approachable this way.

Andarist:
@timdorr removed the config factory

timdorr:
OK, the latest changes are around 713b/729b for the CJS and ES modules. Not nothing, but not a huge deal either.

I will say I'm generally against adding new dependencies to the project. `@babel/runtime` may be common for many projects, but that isn't always the case. I took the same sort of policy against lodash. Nothing against that library, I just didn't want to incur an additional dependency. For context, [I align with Mike Perham's thoughts on the subject](https://www.mikeperham.com/2016/02/09/kill-your-dependencies/).

I'll leave this open for more input/opinions, but I'm voting for closing this PR. Sorry, Mateusz! I appreciate your efforts here and in past PRs.

Andarist:
> I'll leave this open for more input/opinions, but I'm voting for closing this PR. Sorry, Mateusz! I appreciate your efforts here and in past PRs.

No worries.

> I will say I'm generally against adding new dependencies to the project. @babel/runtime may be common for many projects, but that isn't always the case. I took the same sort of policy against lodash. 

The situation here is different than in case of lodash. `redux` already depends on those helpers, it's just implicit dependency as they are bundled into the code. So this PR actually doesn't add any dependencies, it just moves where this code/helpers/dependencies are located (from redux source to `@babel/runtime` package)

Andarist:
@timdorr react-redux is using approach used in this PR. Would you maybe want me to revive this PR?

timdorr:
Yeah, I've turned around on it. Can you resolve the merge conflicts?

timdorr:
Thanks, Mateusz!

PR #3174, opened 2018-10-15
Added a tip on adding multiple group

daniekpo:
Added a tip on adding multiple group of actions to a component using bindActionCreators. This was copied from Dan Abramov's answer to https://github.com/reduxjs/redux/issues/363


timdorr:
Thanks, but this isn't really about bAC. It's about the Object Spread proposal and Object.assign. Since it's not specific to bAC, I don't think it adds that much to the doc. Thank you for your efforts, though!

PR #3179, opened 2018-10-18
added further reading

JoeCortopassi:


timdorr:
Thanks, but we have other places for links.

JoeCortopassi:
@timdorr would it be better if I added it here? https://redux.js.org/introduction/learningresources#middleware

PR #3181, opened 2018-10-19
Update async action creators test: Remove redundant method

kevgathuku:
In the [fetch-mock docs](http://www.wheresrhys.co.uk/fetch-mock/#api-lifecyclerestore_reset), restore() is an alias for reset()
Remove reset() since both accomplish the same task


timdorr:
Thanks!

kevgathuku:
Awesome. Thanks @timdorr

PR #3183, opened 2018-10-19
chore: modified to forEach statement

wxnet2013:


eps1lon:
Duplicate of #2411 and #3139.

PR #3188, opened 2018-10-21
Coding standard fix

sharazghouri:
semicolon miss end of statements. maybe cause  of any error any time in future


timdorr:
Thanks, but it's safe not to use semicolons and that is the style of this example.

cup:
https://github.com/reduxjs/redux/issues/3204
https://github.com/reduxjs/redux/pull/3187
https://github.com/reduxjs/redux/pull/3188

Issue #3186, opened 2018-10-20
Add a "Code Splitting" docs page.


markerikson:
I would like a new docs page added that describes how to code-split Redux reducers (and possibly other pieces of Redux-related code as well).  We'd add it under the "Recipes" section for now, but move it under the future "Real-World Apps" section described in the docs revamp in #2590 when we get that section up.

This page should explain the general principles and technique for code-splitting reducers (re-running `combineReducers` with the new slice reducers, etc).  It should also point to 2-3 of the best-looking existing userland libraries for handling this more automatically, such as https://redux-dynamic-modules.js.org and https://github.com/ioof-holdings/redux-dynamic-reducer .

For reference on this topic, I have links to a few existing articles on code-splitting reducers in the [Redux Reducers](https://github.com/markerikson/react-redux-links/blob/master/redux-reducers-selectors.md#reducers) page in my links list, and a bunch of links to existing libraries in the [Reducers#Dynamic Reducer Injection](https://github.com/markerikson/redux-ecosystem-links/blob/master/reducers.md#dynamic-reducer-injection) section of my addons list.

Please comment here if you're interested in writing this page!

abettadapur:
I can help in writing this page. I can get started over this weekend

JustFly1984:
Hello everyone! I think my question could be related:

- How could you split reducers based on immutable.js Map as root?

I have pretty much 4 separate redux stores: 

- static
- account
- manager
- admin

account and manager reducers are growing intensively, so I would like to separate some parts to reduce bundle size

markerikson:
@JustFly1984 : there shouldn't be any difference in how you split a state tree with a plain JS object root and an Immutable.js Map root, other than needing to use a specialized version of `combineReducers()` to generate the root reducer.

abettadapur:
Started work [here](/abettadapur/redux/tree/code-splitting)

markerikson:
Merged in #3190 .

PR #3190, opened 2018-10-21
[DRAFT]: Add a Code Splitting page to the 'Recipes' section

abettadapur:
An initial draft at a code splitting page. @markerikson Let me know your thoughts, if I need to add more sections or clarify anything in more detail


abettadapur:
#3186

docs/recipes/CodeSplitting.md
```diff
@@ -0,0 +1,138 @@
...
+}
+```
+
+To add a new reducer, one can now call `store.reducerManager.add("asyncState", asyncReducer)`.
+
+To remove a reducer, one can now call `store.reducerManager.remove("asyncState")`
+
+## Libraries and Frameworks
+There are a few good libraries out there that can help you add the above functionality automatically:
+ * [`redux-dynamic-reducer`](https://github.com/ioof-holdings/redux-dynamic-reducer)
```
mpeyper: Hey, I'm the author/maintainer of `redux-dynamic-reducer` and while I'm stoked you want to include my library in these docs, it's unfortunately out of active development and [will be deprecated soon](https://github.com/ioof-holdings/redux-dynamic-reducer/issues/16).  

Luckily, we do have a new library called [`redux-dynostore`](https://github.com/ioof-holdings/redux-dynostore) which supersedes it, adding more configurability and dynamic sagas support and you still get the `addReducers` function on the store.

markerikson:
I'll try to take a look at this in the next few days - been busy with ReactConf and work.

docs/recipes/CodeSplitting.md
```diff
@@ -0,0 +1,138 @@
+# Code Splitting
+In large web applications, it is often desirable to split up the app code into multiple JS bundles that can be loaded on-demand. This strategy, called 'code splitting', helps to increase performance of your application by reducing the size of the initial JS payload that must be fetched.
+
+To code split with Redux, we want to be able to dynamically add reducers to the store. The default usage of Redux mandates that a single root reducer should to be passed to the `configureStore` call, which makes dynamically adding new reducers more difficult. Below, we discuss some approaches to solving this problem and reference two libraries that provide this functionality.
```
markerikson: Rephrase this a bit.  It's not that "default usage mandates a single root reducer".  It's that there _is_ only one reducer, and the standard usage is to split that up into smaller functions for maintainability.

docs/recipes/CodeSplitting.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Basic Principle
+To dynamically add Reducers to a Redux store, there are two approaches that can be taken
```
markerikson: I'd say there's only one approach - calling `store.replaceReducer()`.  However, there may be some different ways you can handle re-creating the root reducer.

docs/recipes/CodeSplitting.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Using `replaceReducer`
```
markerikson: Let's have a small section on what `replaceReducer()` is here, then move any discussion of `injectReducers()` or similar to an additional header section.

docs/recipes/CodeSplitting.md
```diff
@@ -0,0 +1,138 @@
...
+To add a new reducer, one can now call `store.reducerManager.add("asyncState", asyncReducer)`.
+
+To remove a reducer, one can now call `store.reducerManager.remove("asyncState")`
+
+## Libraries and Frameworks
+There are a few good libraries out there that can help you add the above functionality automatically:
+ * [`redux-dynamic-reducer`](https://github.com/ioof-holdings/redux-dynamic-reducer)
+    * This library exposes the `addReducer` function on the Redux store to accomplish the behavior we described above. It also has React bindings which make it easier to add reducers within the React component lifecycle.
+* [`redux-dynamic-modules`](https://github.com/Microsoft/redux-dynamic-modules)
+    * This library introduces the concept of a 'Redux Module', which is a bundle of Redux artifacts (reducers, middleware) that should be dynamically loaded. It also exposes a React higher-order component to load 'modules' when areas of the application come online. Additionally, it has integrations with libraries like `redux-thunk` and `redux-saga` which also help dynamically load their artifacts (thunks, sagas).
```
markerikson: Might as well toss in a link to the list of additional libs I've seen: https://github.com/markerikson/redux-ecosystem-links/blob/master/reducers.md#dynamic-reducer-injection

markerikson:
Just made those tweaks and merged this in.  @abettadapur , thanks for writing this!

raRaRa:
It would be nice addition to know how to access the store in order to call `injectReducers`. E.g. if I were to write HOC or use it from some routes, how would I access `injectReducers`?

Some people recommend creating and then exporting the store, which other components or logic can import. What would be the recommended way? Could the docs show a small simple HOC in order to achieve this with a connected component?

mpeyper:
Right now, the options are:

1. `export`/`import` the store directly (I would not recommend this approach myself)
2. Use the `store` out of the `ReactReduxContext` export of `react-redux` (not recommend as it is not a "public" API)
3. Create your own `Provider` and `Consumer` components to access the store/inject function
4. Use one of the libraries listed in the docs instead of rolling your own (they usually are using 2. internally but at least the unsafe access is hidden away from your code)

The problem is none of these are great solutions and I'd struggle to recommend any of them or want to endorse them by making them examples in the docs, except perhaps using a library, which I don't think there should be an example of in the `redux` docs.

See [this issue](https://github.com/reduxjs/react-redux/issues/1150) and [this PR](https://github.com/reduxjs/react-redux/pull/1154) for details and please leave feedback if you have any.

mpeyper:
There's no built in way to do this and I'm not sure how I would a achieve this myself but I believe the [redux-dynamic-modules](https://github.com/Microsoft/redux-dynamic-modules) library supports this.

raRaRa:
@mpeyper Yeah they do, but migrating to `redux-dynamic-modules` is not straightforward for everyone. For example, I would need to create a module for all external reducers that I'm currently using. I've opened an issue with proposal that could simplify it.

Thanks

PR #3191, opened 2018-10-23
Update ConfiguringYourStore.md

charliegdev:
No need to wrap `renderApp` inside a new arrow function every time we hot reload; we can just use `renderApp` itself.


markerikson:
Thanks!

Issue #3172, opened 2018-10-12
Change docs examples to always use two spaces


gaearon:
Noticed that some of the docs in the recipes section appear to be using four spaces. This is inconsistent with both the rest of the docs and with prevalent JS style in general.

timdorr:
We should be using Prettier on the docs. It handles Markdown and code blocks within Markdown. Some folks have done this as they have edited the files, due to their local config. But we don't have anything set up in our tooling to enforce this.

Muhammedalpaslan:
### wqsaaaaaaaaaaaaaae

### SDASD

1. 

> `_**DSAD**_`

D
D

DD

- DASD### ###

PR #3192, opened 2018-10-23
Use Prettier.js for consistent markdown formatting #3172

KoleMyers:
Issue reference [#3172](https://github.com/reduxjs/redux/issues/3172) .   

I added `.md` files to the formatting scripts. I'm open to any changes for formatting you think best, as I just used mainly the default options and tab spacing. There are some cases where mutli-line code blocks were condensed to a single line, which may not be best for readability in documentation.

Thanks


timdorr:
@markerikson Think we should get this in now or do so after some docs updates? I don't want to be disruptive, but this would be nice to get in now.

markerikson:
Fine by me.

timdorr:
🚀

Issue #2955, opened 2018-04-22
Add docs around static typing


markerikson:
We don't currently appear to have any documentation around using static types (TS / Flow) with Redux.  It would be beneficial if we had docs on what the type definitions look like, and how to handle correctly typing different scenarios.

buoyantair:
This would be great!

timdorr:
This may actually force me to learn TypeScript...

BTW, I found Dr. Axel Rauschmeyer's intro to TS *really* helpful for getting me off the ground with typing: http://2ality.com/2018/04/type-notation-typescript.html

ch1ll0ut1:
Please provide examples how to:
Augment dispatch so that you can customize return type for specific action types and have this work in bindActionCreators as well

ch1ll0ut1:
For me it seem´s this issue is critical because, as of right now no async action library has implemented correct type augmentation. This and the complexity of implementing it, make it quite difficult to use redux with typescript if you want typed async actions.

eps1lon:
@ch1ll0ut1  Could you point to some untyped examples and explain what the expected behavior of typescript should be for those?

JannicBeck:
Would appreciate this.
I was looking for an example with TS and was surprised that there was none.
[I ended up having to write it myself](https://github.com/JannicBeck/undox/tree/master/examples/react/counter-typescript)
Might be good as a starting point.

ch1ll0ut1:
From https://redux.js.org/advanced/async-actions
```
import fetch from 'cross-fetch'
​
export const REQUEST_POSTS = 'REQUEST_POSTS'
function requestPosts(subreddit) {
  return {
    type: REQUEST_POSTS,
    subreddit
  }
}
​
export const RECEIVE_POSTS = 'RECEIVE_POSTS'
function receivePosts(subreddit, json) {
  return {
    type: RECEIVE_POSTS,
    subreddit,
    posts: json.data.children.map(child => child.data),
    receivedAt: Date.now()
  }
}
​
export const INVALIDATE_SUBREDDIT = 'INVALIDATE_SUBREDDIT'
export function invalidateSubreddit(subreddit) {
  return {
    type: INVALIDATE_SUBREDDIT,
    subreddit
  }
}
​
// Meet our first thunk action creator!
// Though its insides are different, you would use it just like any other action creator:
// store.dispatch(fetchPosts('reactjs'))
​
export function fetchPosts(subreddit) {
  // Thunk middleware knows how to handle functions.
[...]

PR #3201, opened 2018-10-29
[DRAFT]: Docs for usage with typescript

HershVar:
### Status
[January 9, 2019] - Fourth draft completed
[November 12, 2018] - Third draft completed, next revision will include usage with React-Redux & Thunks
[November 10, 2018] - Second draft completed, looking for feedback
[October 29, 2018] - Initial draft ready, looking for feedback

### Description
From #2955 , this adds documentation around using redux with typescript.

### Notes & Considerations

1. I chose to do the bulk of the documentation here around the example of a basic chat application. This allows for the examples to be more practical and interactive (codesandbox).
1. The docs include a codesandbox example for people to explore. If anyone feels something here could be improved on please let me know. Here's the [link](https://codesandbox.io/s/w02m7jm3q7).
2. The implementation of type checking varies slightly between different blog posts. I chose an approach that I consider reasonable. I'm open to suggestions here.

### Misc.
1.  I placed the new doc under the "Advanced" section. If there is a more appropriate place for this specific doc I'd be happy to move it as needed.

2. I haven't linked this doc to the table of contents (README). If I should do that please let me know.


markerikson:
I'll try to take a look at this in the next few days - been busy with ReactConf and work.

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,219 @@
...
+// src/store/chat/actions.ts
+
+import { IMessage, ISendMessageAction, ChatActions } from './types'
+
+export type sendMessageType = (newMessage: IMessage) => ISendMessageAction
+
+export const sendMessage: sendMessageType = newMessage => ({
+  type: ChatActions.SendMessage,
+  payload: newMessage
+})
```
Kovensky: I'd prefer to just declare this as a hoisted function instead of an arrow function:

```ts
import { IMessage, ISendMessageAction, ChatActions } from './types'

export function sendMessage(newMessage: IMessage): ISendMessageAction {
  return {
    type: ChatActions.SendMessage,
    payload: newMessage
  }
}
```

`sendMessageType` is literally just `typeof sendMessage` so I didn't think it was needed to declare separately.

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,219 @@
...
+import { SystemActions, IUpdateSessionAction, ISystemState } from './types'
+
+export type updateSessionType = (
+  newSession: ISystemState
+) => IUpdateSessionAction
+
+export const updateSession: updateSessionType = newSession => ({
+  type: SystemActions.UpdateSession,
+  payload: newSession
+})
```
Kovensky: Same here, can be a hoisted function, don't need the separate `updateSessionType`.

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,219 @@
...
+
+const initialState: IChatState = {
+  messages: []
+}
+
+export const chatReducer: Reducer<IChatState> = (
+  state = initialState,
+  action
+) => {
+  switch (action.type) {
```
Kovensky: Refinements won't work unless you give the second generic argument to `Reducer` with the list of actions you support.

Cryrivers:
Saw @markerikson 's tweet and made a TypeScript + Redux example based our practice. Check it out: https://codesandbox.io/s/62kw7xvnqw (**Note:** Do download and try it in VSCode. The type inference on CodeSandbox is not working perfectly at the moment.)

To List some advantages: 
 1. No need to describe the shape/type of actions and action creators
 2. Automatically map type of reducers and actions to props using helper `WithRedux`
 3. Type-safe actions (see screenshot).

image

HershVar:
Thank you for the feedback so far. Before I work on the necessary changes, some points of discussion:

@Kovensky 
1. There is an optional TSLint rule that forces interface names to start with an "I" so wanted feedback here. Looks like this convention is not that common so I will change it.
2. Regarding hoisted functions, this is fine and I will change it accordingly.
3. Thanks for catching the lack of type checking for reducers. There is a slightly different method for type checking reducers by @Cryrivers . I would like to see what people think before making the appropriate changes.

@CarloPalinckx 
1. There are multiple popular naming conventions for actions. The current names for the actions might be confusing because of the very basic demonstration. For example, in addition to `@CHAT:SEND/MESSAGE`, there could have also been `@CHAT:SEND/FILE`. I think the assumption here is that people reading this section of the docs would already be familiar with redux, meaning they would have an opinion on which convention to use. I'm still willing to change it if you think it will add clarity.

CarloPalinckx:
@HershVar  It seems like all the pages in the docs stay away from this particular convention so I wouldn't introduce that in docs about typescript. I also don't think people people will be very familiar with redux at this state other than have read the previous pages. If you are integrating redux straight into a typescript environment, you would have to have read these docs in order to get started.

HershVar:
@CarloPalinckx 
Good point, I will make those changes in order to stay consistent with the rest of the documentation as you pointed out.

docs/advanced/UsageWithTypescript.md
danpantry: +1. It's really rare to define function types like this in my experience. It would be more intuitive to do this:

```ts
function updateSession(newSession: SystemState) {
  return { type: SystemActions.UpdateSession, payload: newSession }
}
```

You would not commonly define a type alias here unless you needed it to be re-usable

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,219 @@
...
+## Type Checking State
+
+Type checking each slice of state is simple when you understand that each slice of state is just an object.
+
+Describing the state shape of the chat reducer's slice of state:
+
+```ts
+// src/store/chat/types.ts
+
+export interface IMessage {
```
danpantry: It is discouraged to use the `IName` convention in interfaces since TypeScript is structurally, not nominally, typed.

Consider using `Message`, `ChatState`, instead of `IMessage`, `IChatState`.

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,219 @@
...
+
+We will be using TypeScript's enums to declare our action constants. [Enums](https://www.typescriptlang.org/docs/handbook/enums.html) allow us to define a set of named constants.
+
+Chat Action Constants and Actions:
+
+```ts
+// src/store/chat/types.ts
+
+export enum ChatActions {
+  SendMessage = '@CHAT:SEND/MESSAGE'
```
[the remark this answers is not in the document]
danpantry: I personally prefer using raw strings instead of enums. A discriminant union will give you appropriate type checking. The only downside is that your action creators might emit the wrong kind of action. YMMV

One should also mention that exporting an enum here is not useful - the name `ChatActions` to me implies it is a discriminant union type, and the code would greatly benefit from having this for type safety. So, instead of `export enum ChatActions`, consider

```ts
// src/store/chat/types.ts
export type ChatActions =
  | SendMessageAction
```
and of course you could extend that discriminant union. You can then pass that union to a reducer, which enables exhaustive type checking!

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,219 @@
...
+// src/store/chat/reducers.ts
+
+import { Reducer } from 'redux'
+import { IChatState, ChatActions } from './types'
+
+const initialState: IChatState = {
+  messages: []
+}
+
+export const chatReducer: Reducer<IChatState> = (
```
danpantry: I don't think the explicit `Reducer` type is necessary. TypeScripts strength comes in using as few explicit types as possible - Consider using this instead:

```ts
interface SendMessageAction {
   type: typeof ChatActions.SendMessage,
   payload: Message
}
function chatReducer(state: IChatState, action: SendMessageAction) {
  switch (action.type) {
    return {
      messages: [...state.messages, action.payload]
    }
  }
}
```

one downside of this approach is that while the type checking at compile time *is exhaustive*, your function *may receive* inputs that do not match the appropriate actions at runtime. For example, `{ type: '@@redux/init' }`. I don't think there's a way to work around this, though.

docs/advanced/UsageWithTypescript.md
danpantry: +1 to @Kovensky. Again, I would recommend avoiding the `Reducer` type declaration altogether and rely on TypeScripts structural typing and just define the function as a hoisted function.

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,219 @@
...
+
+import { chatReducer } from './chat/reducers'
+import { IChatState } from './chat/types'
+
+export interface IAppState {
+  system: ISystemState
+  chat: IChatState
+}
+
+const rootReducer: Reducer<IAppState> = combineReducers({
```
danpantry: Once again, you do not need to use `Reducer<IAppState>` here.

Also, defining `IAppState` here is not actually required! you can instead do this:

```ts
const rootReducer = combineReducers({ system, chat })
export type AppState = ReturnType<typeof rootReducer>
```

This can save some hassle in larger state trees

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,219 @@
...
+import { Reducer } from 'redux'
+import { SystemActions, ISystemState } from './types'
+
+const initialState: ISystemState = {
+  loggedIn: false,
+  session: '',
+  userName: ''
+}
+
+export const systemReducer: Reducer<ISystemState> = (
```
danpantry: This is more of a criticism of the code than the TypeScript: IMO it's best practice to have one reducer manage *one thing*. Instead of having a state object with multiple values, you should have each reducer manage one part of each state unless that state is instrinically tied together (in most cases, state that is that tightly coupled can be derived state).

For example, `loggedIn: false` could be determined by whether or not `session` is equal to `null`. `session` should be `null` as should `userName` with absent values.

Just my 2c - this could be changed to be two reducers, which would help keep the state flat - which is important in Redux!

```ts
export function sessionReducer(opaqueSessionToken: string | null, action: Action) {
  ...
}

export function usernameReducer(username: string | null, action: Action) {
  ...
}


// src/store/index.ts
  user: userReducer,
  session: sessionReducer,
  chat: chatReducer
})
```

HershVar:
@danpantry 
Thank you for this feedback, in combination with the other responses here I will make the appropriate changes this weekend. Will cc when the update is pushed.

danpantry:
this looks a lot better!

Cryrivers:
@HershVar thanks for mentioning our solution. Please do consider having type-safe reducers by utilizing discriminated unions as well. Otherwise it might confuse beginners.

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,200 @@
...
+
+import { SystemActions, SystemState, SystemActionTypes } from './types'
+
+const initialState: SystemState = {
+  loggedIn: false,
+  session: '',
+  userName: ''
+}
+
+export function systemReducer(state = initialState, action: SystemActionTypes) {
```
Cryrivers: please explicitly indicate the type of return value is `SystemState`, otherwise the return value is not type-safe.

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,200 @@
...
+```ts
+// src/store/chat/reducers.ts
+
+import { ChatState, ChatActions, ChatActionTypes } from './types'
+
+const initialState: ChatState = {
+  messages: []
+}
+
+export function chatReducer(state = initialState, action: ChatActionTypes) {
```
Cryrivers: please explicitly indicate the type of return value is `ChatState`, otherwise the return value is not type-safe.

docs/advanced/UsageWithTypescript.md
Cryrivers:
```suggestion
export function systemReducer(state = initialState, action: SystemActionTypes): SystemState {
```
it would be better because in practice we may accidentally return a wrong object. this change ensures that you need to always return a `SystemState`.

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,206 @@
...
+
+## A Practical Example
+
+We will be going through a simplistic chat application to demonstrate a possible approach to include static typing. This chat application will have two reducers. The _chat reducer_ will focus on storing the chat history and the _system reducer_ will focus on storing session information.
+
+The full source code is available on [codesandbox here](https://codesandbox.io/s/w02m7jm3q7). Note that by going through this example yourself you will experience some of the benefits of using TypeScript.
+
+## Type Checking State
+
+Type checking each slice of state is simple when you understand that each slice of state is just an object.
```
markerikson: I'd caveat this somehow to emphasize "in this example".  Slices of state aren't necessarily objects, they could be anything - a counter, an array, etc.

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,206 @@
...
+
+Describing the state shape of the chat reducer's slice of state:
```
markerikson: Phrasing on this sentence is awkward.  Perhaps something like:  "Adding types to each slice of state is a good place to start.  In this example, we start by describing the type of the chat reducer's slices of state: "

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,206 @@
...
+  message: string
+  timestamp: number
+}
+
+export interface ChatState {
+  messages: Message[]
+}
+```
+
+Describing the state shape of the system reducer's slice of state:
```
markerikson:
```suggestion
And then do the same for the system reducer's slice of state:
```

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,206 @@
...
+export type ChatActionTypes = SendMessageAction
+```
+
+Note that you can use TypeScript's Union Type to express multiple actions like so:
+
+```ts
+export type ChatActionTypes = SendMessageAction | DeleteMessageAction
+```
+
+With these types declared we can now also type check chat's action creators. In this case we are taking advantage of TypeScript's inference to avoid verbosity:
```
markerikson: As opposed to... what?

It might be worth showing a commented-out example of what the "verbose" approach would be in comparison.  (I assume it would involve hand-declaring the type of the action creator.)

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,206 @@
...
+    payload: newSession
+  }
+}
+```
+
+## Type Checking Reducers
+
+Reducers are just pure functions that take the previous state, an action and then return the next state.
+
+Type checked chat reducer:
```
markerikson: Add some description here explaining what exactly is being done in comparison to a plain JS reducer (ie, importing the state and action types and referencing them).

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,206 @@
...
+Note that we are exporting these interfaces to reuse them later in reducers and action creators.
+
+## Type Checking Actions & Action Creators
+
+We will be using TypeScript's enums to declare our action constants. [Enums](https://www.typescriptlang.org/docs/handbook/enums.html) allow us to define a set of named constants.
+
+Chat Action Constants & Shape:
+
+```ts
+// src/store/chat/types.ts
```
markerikson: Ought to discuss why the types are being declared in a separate file - what's the tradeoff here?

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,206 @@
...
+        ...action.payload
+      }
+    }
+    default:
+      return state
+  }
+}
+```
+
+Putting it all together in combine reducers, note that we do not have to explicitly declare a new interface for AppState. We can use `ReturnType` to infer state shape from the `rootReducer`.
```
markerikson:
```suggestion
We now need to generate the root reducer function, which is normally done using `combineReducers`.  Note that we do not have to explicitly declare a new interface for AppState. We can use `ReturnType` to infer state shape from the `rootReducer`.
```

And hey, that state shape inference is cool!

markerikson:
Looks like a good start so far.  Thanks to @HershVar for your work, and everyone else who has offered review suggestions.

That said, I'd _really_ like this to also cover use with React-Redux.  I keep hearing people say that's one of the biggest pain points in using TS with Redux.

I can also see an argument for having a separate "Using TS with React-Redux" page in the React-Redux docs, but I think it's worth having that content in this page (and maybe duplicating the React-Redux part over there).

Can we get that material added in?

HershVar:
Update:

At this point, this page of the documentation has:

- A live codesandbox example with react, redux, react-redux in typescript
- Documentation explaining how to use redux with typescript
- Documentation explaining how to use react-redux with typescript

@markerikson 
I would like to confirm with you that redux-thunk's usage with typescript should live in this page. In my opinion, if I wanted to look for how to use redux-thunk with typescript I would go to its documentation first. I may be wrong here so I'd like to hear opinions on this matter.

markerikson:
I'd specifically like thunk examples in this document.  I don't think most people look at the docs in the thunk repo, and either way, the main Redux docs get a lot more traffic.

I'll try to take a look at the updates in the next couple days.

HershVar:
Thanks for the confirmation, will add that next on tuesday/wednesday.

docs/advanced/UsageWithTypescript.md
markerikson: FWIW, I personally do tend to write action creators as:

```js
function doTheThing(arg1, arg2) {
    return {
        type : DO_THE_THING,
        payload : {arg1, arg2}
    };
}
```

But maybe I'm not in the majority on that :)

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,280 @@
...
+  session: string
+  userName: string
+}
+```
+
+Note that we are exporting these interfaces to reuse them later in reducers and action creators.
+
+## Type Checking Actions & Action Creators
+
+We will be using TypeScript's enums to declare our action constants. [Enums](https://www.typescriptlang.org/docs/handbook/enums.html) allow us to define a set of named constants. Note that we are making a tradeoff here when we declare our types in a sperate file. In exchange for seperating our types into a seperate file, we get to keep our other files more focussed on their purpose. While this tradeoff can improve the maintainability of the codebase, it is perfectly fine to organize your project however you see fit.
```
markerikson:
```suggestion
We will be using TypeScript's enums to declare our action constants. [Enums](https://www.typescriptlang.org/docs/handbook/enums.html) allow us to define a set of named constants. Note that we are making a tradeoff here when we declare our types in a separate file. In exchange for separating our types into a separate file, we get to keep our other files more focused on their purpose. While this tradeoff can improve the maintainability of the codebase, it is perfectly fine to organize your project however you see fit.
```

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,280 @@
...
+  payload: Message
+}
+
+export type ChatActionTypes = SendMessageAction
+```
+
+Note that you can use TypeScript's Union Type to express multiple actions like so:
+
+```ts
+export type ChatActionTypes = SendMessageAction | DeleteMessageAction
```
markerikson: Could we just go ahead and have the "delete message" action be included in the example overall, so we don't have to show this separately?  Could add the "Note that" line as a comment above the union.

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,280 @@
...
+
+export function sendMessage(newMessage: Message) {
+  return {
+    type: ChatActions.SendMessage,
+    payload: newMessage
+  }
+}
+```
+
+Side note: to appreciate TypeScript's inference, here is what a more verbose version would look like:
```
markerikson: This doesn't feel _too_ much more verbose.  Could maybe just emphasize that the return type is what's being inferred, with a commented-out version of the function declaration line shown right above the actual line of code for comparison?

docs/advanced/UsageWithTypescript.md
```diff
@@ -0,0 +1,280 @@
...
+  system: systemReducer,
+  chat: chatReducer
+})
+
+export type AppState = ReturnType<typeof rootReducer>
+```
+
+## Usage with React-Redux
+
+While react-redux is a separate library from redux itself, it is commonly used with react. For this reason, we will go through how react-redux works with TypeScript using the same example used previously in this section.
```
markerikson: Capitalize React-Redux everywhere.

timdorr: Actually, it's "React Redux". No dash, that's just the package name.

CarloPalinckx:
Have you considered adding an "advanced" section where you go over things like inference? It currently feels a bit weird you are only using the return type of the root reducer and not the return types for action creators and chat/system reducers. You could stay away from inference at all in the "regular" part and then go all in on inference in that section. I feel like the "Note that you can use inference here" could be relevant to more places in the docs, but that would get pretty messy imo.

docs/advanced/UsageWithTypescript.md
eps1lon: I prefer to add the return type myself. It makes it more obvious what that action will actually be created. This in turn makes it easier once you refactor the payload of that action. With the less verbose example you might get the error in an unrelated file or none at all. With explicit annotation you get the error message where it originates.

CarloPalinckx: I generally think inference is more of a personal preference than anything else, it's not better or worse. Some people like verbose and others don't 😄. 

> or none at all

Should basically only occur when you are not consuming a property that was typed. For instance if you remove a payload from the return of an action creator (that is being infered) that was never used in the first place.

Cryrivers:
@CarloPalinckx This is also my point from beginning. You can check out my proposed solution, which utilises type inference as much as possible.

https://github.com/reduxjs/redux/pull/3201#issuecomment-436990456

HershVar:
@CarloPalinckx 
Thanks for the feedback, there's definitely multiple ways to add type checking with varying levels of granularity. I'm not 100% sure on how opinionated we should be and what the opinion should be (level of granularity used in type checking). That being said, I did mention as a note in the end that the example uses just one of many possible ways to type check. If there are specific suggestions we can go from there.


Regarding the addition of "usage with redux thunk" there is one thing I am trying to figure out right now. A thunk returns a function but calling the thunk via props in a component will either return nothing or some other value that is explicitly returned. If anyone has any recommendation on how to add type checking here I would love to hear suggestions. I think the goal here is to have type checking around the parameters and what calling that thunk will return.

docs/advanced/UsageWithTypescript.md
```diff
@@ -56,47 +56,45 @@ Chat Action Constants & Shape:
 // src/store/chat/types.ts
 
 export enum ChatActions {
```
danpantry: I'm still not sure about the use of this enum.

One of the problems of using an enum here is that it is only possible to have type correctness if you refer to the enum. That is, the following is a compiler error:


```typescript
const foo: SendMessageAction = {
  type: 'SEND_MESSAGE', // "SEND_MESSAGE" is not assignable to `ChatActions`
  payload: ...
}
```

Which feels surprising. Is the motivation for using an enum just to prevent duplication of string literals?

docs/advanced/UsageWithTypescript.md
```diff
@@ -56,47 +56,45 @@ Chat Action Constants & Shape:
...
   payload: Message
 }
 
-export type ChatActionTypes = SendMessageAction
-```
-
-Note that you can use TypeScript's Union Type to express multiple actions like so:
+interface DeleteMessageAction {
+  type: ChatActions.DeleteMessage
+  timestamp: number
```
danpantry: Shouldn't we encourage using [FSA](https://github.com/redux-utilities/flux-standard-action)? If so, `timestamp` should be `meta: { timestamp }` instead.

docs/advanced/UsageWithTypescript.md
HershVar: This is a valid concern, the motivation is definitely for preventing duplication of string literals. If one has to type the same string literal twice then there is a chance for an error/typo which typescript will consider valid (as far as I am concerned).

Regarding the issue you pointed out it does become a problem if you use actions with string literals. This doesn't become a concern if you strictly use action creators (assuming the action creators' type points to the enum).

docs/advanced/UsageWithTypescript.md
HershVar: If I am not mistaken the flux library is what suggested the `payload` and `meta` format for actions. Redux actions do not require a meta field, only the type field is required. I would like to keep actions and its nomenclature consistent with the rest of the documentation. Another point to take into consideration is that many people who learn Redux aren't coming from the flux library anymore.

markerikson: Not quite.  `payload` and `meta` are from the ["Flux Standard Actions" community standard](https://github.com/redux-utilities/flux-standard-action), which was developed by Andrew Clark.  While Redux doesn't require FSA structure, a very large percentage of Redux apps use an FSA-based approach.

HershVar: Thanks for clarifying, I'll take that as changing the actions in this example to follow FSA.

docs/advanced/UsageWithTypescript.md
danpantry: > there is a chance for an error/typo which typescript will consider valid (as far as I am concerned).

This is not true.

```ts
interface SendMessageAction {
  type: "SEND_MESSAGE",
  payload: ...
}

const foo: SendMessageAction {
  type: "foo", // compiler error: Type string is not assignable to "SEND_MESSAGE"
  payload: ...
}

const foo: SendMessageAction {
  type: "SEND_MESSAGE", // no problem
  payload: ...
}

function reducer(state: any, action: SendMessageAction | { type: unknown }) {
  switch (action.type) {
    case 'SEND_MESSAGE':
      // type is narrowed to SendMessageAction
      return state
  }
  return state
}
```

docs/advanced/UsageWithTypescript.md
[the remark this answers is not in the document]
nurpax: +1 for string literals.  Like @danpantry says, they give you great type safety while being easy to understand even for TS beginners.

docs/advanced/UsageWithTypescript.md
```diff
@@ -276,6 +276,48 @@ interface AppProps {
...
+
+Redux Thunk is a commonly used middleware for asynchronous orchestration. Feel free to check out its documentation [here](https://github.com/reduxjs/redux-thunk). A thunk is a function that returns another function that takes parameters `dispatch` and `getState`. We will define a type which describes this in a new file:
+
+```ts
+// src/types.ts
+
+import { Dispatch } from 'redux'
+import { AppState } from './store'
+
+export type ThunkType<T> = (dispatch: Dispatch, getState?: AppState) => T
```
nurpax: `redux-thunk` in fact has typings so perhaps it'd make more sense to use `ThunkAction` from [redux-thunk/index.d.ts](https://github.com/reduxjs/redux-thunk/blob/master/index.d.ts)?

It's great you've included redux-thunk in these docs.  Definitely belongs here.

markerikson:
So where do we think this PR stands now?

HershVar:
I think there are 2 points of discussion to resolve before considering this "ready".

1) Usage of enums; @danpantry pointed out an alternative to get rid of enums altogether. Out in the wild, based on my preliminary research, there doesn't seem to be a dominant preference for either method.

2) There seems to be two emerging ways to use typescript here; Using typescript as explicitly as possible or relying on typescript's inference. Right now this PR uses bits and pieces from both styles, some parts are explicit and some parts use inference. It is mentioned at the bottom of this docs' page that there are multiple ways to implement typescript but should we be leaning to one side here? Some time has passed since this specific discussion started so opinions may have changed.

markerikson:
Hmm.  How difficult would it be to show kind of "side-by-side" examples of both approaches?  What are the tradeoffs of each?

HershVar:
I think doing that may add some unnecessary noise to this page. Going down that path feels like we are teaching more about typescript instead of focusing on the main goal which is to teach how to use redux with typescript.

joshburgess:
@HershVar RE: enums vs string literals

People use both. The unique thing about enums here is that the values exist at both run-time and compile-time, but you can accomplish exactly the same thing with constant string literals. It just requires a little more boilerplate.

```
// the assertions here aren't necessary... just showing what gets inferred
const ACTION_A: 'ACTION_A' = 'ACTION_A'
const ACTION_B: 'ACTION_B' = 'ACTION_B'

type ActionTypes = { ACTION_A: ACTION_A, ACTION_B: ACTION_B }

export const ActionTypes: ActionTypes = {
  ACTION_A,
  ACTION_B,
}
```

is _almost_ the same thing as 

```
enum ActionTypes {
  ACTION_A = 'ACTION_A',
  ACTION_B = 'ACTION_B'
}
```

The only difference is that the former's values can't be accessed with dot notation, like `ActionTypes.ACTION_A`. You, instead, have to do `ActionTypes['ACTION_A']`. You _could_ replicate the dot notation functionality here if you used TypeScript's namespaces feature though... but not many people use it. (It's one of the unsupported features of TS when using Babel 7, as far as I know).

danpantry:
With regards to: 

>  Going down that path feels like we are teaching more about typescript instead of focusing on the main goal which is to teach how to use redux with typescript.

If that's the case - Just go with string literals. Enums are unique to TypeScript and I have never personally seen them used for this use case nor would I recommend them. Every JavaScript and TypeScript user is going to understand what a string literal is.

HershVar:
@joshburgess @danpantry 

Thanks, I'll make the appropriate changes as soon as I get the time this week. My schedule will finally free up starting next week so I plan to finish up this PR very soon.

JounQin:
Is that possible to use `thunk actions` without casting `thunkSendMessage` to `any` of `AppProps` in the [example](https://codesandbox.io/s/w02m7jm3q7)?

HershVar:
@JounQin 
From what I understand the `getState` param in redux thunk returns a method to get the entire state of the store. In that case we would have to use `AppState` type to get correct type checking and intelisense.

Kovensky:
Very importantly -- this _only works_ if you use `const` to declare the key (no `var` or `let`), and the initializer must be a string literal (template literals acceptable only if they are untagged and have no substitutions).

I kinda prefer the `enum` approach as it lets you use `CamelCase` instead of `SHOUTING_CASE` without worrying about naming collisions, but it's true that using `export`ed `const`s will be much better for bundle size as the identifier names can be minified; enum key names cannot.

---

You can make the type of the consts easier to use by also exporting an associated type (`export const FOO = 'FOO'; export type FOO = typeof FOO;`), which has an effect similar to how enum keys are their own type. I don't necessarily recommend this, though, it just makes it easier to use "FOO"'s type without having to know to write "typeof FOO".

The thing I think is a problem, though, is that you no longer get an automatic type with all const values.

For that, you need something like `export type Actions = Extract<typeof import('./types')[keyof typeof import('./types')], string>`, but that will not compile with CRA because babel's parser doesn't support `import()` types yet. The version that works with CRA is:

```ts
// you can put this in types.ts itself, it's a circular import but it's
// only used for its typeof, not for its value, so it's safe
import * as types from './types'

export type Actions = Extract<typeof types[keyof typeof types], string>
```

The `Extract` itself is only needed if you export any other concrete values from `types.ts`. If your only value exports are already strings you can use just the `typeof` expression.

You also also have to make sure none of your exports are actually typed `string`, otherwise `Actions` will collapse to `string`; and you need to be careful when _reexporting_ `Actions` as it cannot be directly reexported (it technically can, but it'll break with CRA / babel or with `--isolatedModules`). To reexport it, you need to alias it:

```ts
import { Actions } from './types'
export type Actions = Actions
```

nurpax:
How important is using

export const FOO = ’FOO’

for action types in the first place! They’re only used in reducers and action creators. I find that this pattern used to be needed on normal JS to prevent misspelling the stringly typed action names. But on TS, if your reducers are written with full static types, typoing say ’FOO’ to ’Foo’ will result in an error. 

So I’ve moved away from declaring const action type symbols in most of my new code.

timdorr:
OK, this has been open way way too long and looks pretty good to me. I'm going to merge it in. It won't get automatically added to the site, since it's just a single new file. But this opens the door for others to provide PRs on this file and critique via changes, instead of comments. 

Thanks a ton @HershVar for sticking with this!

markerikson:
I know this got merged already, but just saw an interesting-looking post and wanted to use this as a venue to ping people.  Does this offer any additional useful techniques that are worth showing?

https://medium.com/@ankeet.maini/type-redux-with-typescript-without-writing-any-types-5f96cbfef806

PR #3220, opened 2018-11-23
Fix a TypeError exception in real-world example

avioli:
Fix a TypeError exception, raised because wrong argument is passed to a function call.


timdorr:
Thanks!

PR #3228, opened 2018-11-28
docs: Add language identifier to all code blocks

ght:
This PR adds a language identifier to all the code blocks under `/docs` that were missing one.

There are two cases where a `js` identifier was missing, causing the block not to be highlighted.

Apart from those two, this PR adds a `sh` identifier to all blocks containing shell commands.

In case the latter is not wanted, let me know and I can change this to only add the two missing `js` identifiers.


timdorr:
Thanks!

PR #3232, opened 2018-11-30
Remove outdated Link, Add Redux Book

rwieruch:
Just went through the Redux resources page and saw an outdated link from my website. It shows how to use normalizr, but I feel guilty because I really need to update it. 

A better resource for learning Redux in React would be Taming the State in React though :) Would love to get it into the books section.


markerikson:
Sure!

PR #3238, opened 2018-12-04
Fix wrong anchor in Store documentation

BTMPL:
The anchor wasn't matching the target element so it wasn't scrolling.


markerikson:
Thanks!

PR #3279, opened 2018-12-10
improved isPlainObject function

yakotika:


timdorr:
Thanks, but this isn't really a case we need to cover.

sergeysova:
@timdorr Why not?

timdorr:
It isn't common to attempt to return `Math` or a built-in from an action creator and try to dispatch it. If you're doing that, you probably have more fundamental problems with your code :)

PR #3291, opened 2018-12-11
Left-align feature text

markerikson:


markerikson:
Awright, enough people seem to want this that we'll go with it for now.  Can always tweak it again later.

PR #3298, opened 2018-12-13
fix overflow bug, centre align footer on mobile

gc:
overflow bug: you should no longer see a horizontal scroll bar for no reason

see footer on mobile/narrow screen for other change

### DIFF
--- a/website/static/css/custom.css
+++ b/website/static/css/custom.css
@@ -147,6 +147,8 @@ header.postHeader:empty + article h1 {
 .nav-footer .sitemap a {
   /* increase link contrast */
   color: white;
+  margin: 0;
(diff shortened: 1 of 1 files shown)


timdorr:
I'm not seeing the horizontal scroll on mobile. Also, the footer shouldn't be center aligned, as that mismatches the general flow of the rest of the site (primarily the menu).

gc:
I'll revert the centering. Regarding the overflow, it does exist - it doesn't show on mobile devices because the footer items reposition into a column, leaving space for the margin to not overflow the page. Try making your page slowly bigger/smaller on desktop, cc @markerikson we spoke about it in reactiflux

PR #3302, opened 2018-12-15
Make single-object-returning functions more terse

nielsbom:
Returning single expression is possible without a return statement.

(Could be that Redux needs this specific syntax, I'm new to it)


markerikson:
No specific need for it, but I personally am not a fan of implicit object returns. Appreciate the PR, but I think we'll keep it the way it is.

nielsbom:
Would you think it useful to have a consistent style in the docs and related files?
In the examples I do see some implicit object literal returns.

PR #3303, opened 2018-12-16
More links to docs sites instead of repo markdown

nielsbom:
Actual API docs are a little nicer to read than a file in the repository


markerikson:
Thanks!

PR #3305, opened 2018-12-16
Improve example description

nielsbom:
The example doesn't show the state tree for the whole app, the description says so, that's confusing.


nielsbom:
Never mind, misunderstood it 😄

PR #3310, opened 2018-12-19
Documentation: Add dash to handle apostrophe in title

chrisself:
**Description**
The Performance FAQ's TOS contains a broken anchor link. "Won't calling 'all my reducers' for each action be slow?" currently links to:

https://redux.js.org/faq/performance#wont-calling-all-my-reducers-for-each-action-be-slow

However, with the section title containing an apostrophe, the generated id contains a dash, meaning the link should be:

https://redux.js.org/faq/performance#won-t-calling-all-my-reducers-for-each-action-be-slow


timdorr:
Thank-s!

PR #3312, opened 2018-12-21
Expand docs for using Immutable.js with selectors

msrose:
You can view the compiled markdown [here](https://github.com/msrose/redux/blob/2ed8ca5b7602fcdac78010079ea2a175ae1847d2/docs/recipes/UsingImmutableJS.md#your-selectors-should-return-immutablejs-objects).

The info on this doc page was really helpful for us standardizing our usage of selectors with Immutable.js at Top Hat, so I just wanted to expand on some of the points already made to afford a bit more clarity. 

We have also been using the suggested HOC for converting Immutable.js props to plain JS objects to avoid polluting our presentational components; we found the same HOC cropping up in various incarnations in our front-end projects, so we've released the NPM package [with-immutable-props-to-js](https://www.npmjs.com/package/with-immutable-props-to-js). Hoping other may find it useful so I've linked it in the docs as well.

### DIFF
--- a/docs/recipes/UsingImmutableJS.md
+++ b/docs/recipes/UsingImmutableJS.md
@@ -232,7 +232,11 @@ Do not, however, use Immutable.JS in your dumb components.
 
 ### Your selectors should return Immutable.JS objects
 
-Always.
+Always. This practice has several advantages:
+
+- It avoids unnecessary rerenders caused by calling `.toJS()` in selectors (since `.toJS()` will always return a new object).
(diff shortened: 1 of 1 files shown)


msrose:
Here's a link to the relevant section in the deploy preview: https://deploy-preview-3312--redux-docs.netlify.com/recipes/using-immutablejs-with-redux#your-selectors-should-return-immutablejs-objects

timdorr:
LGTM. Thanks!

PR #3317, opened 2018-12-29
Add explicit mention of Redux ES build

TimvdLippe:
For new users who want to consume Redux using ES modules in a browser,
the current import directs them to the Node-compatible build.

Therefore, add an explicit comment in the README pointing these users to
the `redux/es/redux.mjs` build that does work in a browser environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,8 @@ To specify how the actions transform the state tree, you write pure _reducers_.
 That's it!
 
 ```js
+// This assumes you are running in a Node environment. If you are running in a browser environment,
+// import from 'redux/es/redux.mjs' instead.
 import { createStore } from 'redux'
 
 /**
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -126,6 +126,8 @@ To specify how the actions transform the state tree, you write pure _reducers_.
 That's it!
 
 ```js
+// This assumes you are running in a Node environment. If you are running in a browser environment,
+// import from 'redux/es/redux.mjs' instead.
```
jenbuzz: Too avoid confusion and people copy-pasting "import from 'redux/es/redux.mjs'" it might be better to actually just write out the full line "import { createStore } from 'redux/es/redux.mjs'"

timdorr:
Thanks, but this is actually confusing to newer users. If they're using a module bundler, they won't think their in a "node" environment. They may also want to use CJS. The point of these blocks of code isn't to explain module systems to the user, it's to show a basic example of using Redux.

PR #3318, opened 2018-12-30
compose: use identity function as a default argument for reduce

ilyavf:
Minor improvement of the functional style for `compose` definition: 
- replaced an `if` block with an identity function as default argument for `reduce`.

Even though `compose` can be defined as one line using the identity function, but one-function case optimization still makes sense.

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -10,13 +10,9 @@
  */
 
 export default function compose(...funcs) {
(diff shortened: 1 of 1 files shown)


markerikson:
Appreciate the PR, but the current version works as-is, and there's no reason to change it yet again.  Thanks!

PR #3322, opened 2019-01-02
ServerRendering docs: remove ref to data-react-id

dunglas:
Since React 15, the `data-reactid` attribute is gone: https://reactjs.org/blog/2016/03/07/react-v15-rc1.html


timdorr:
Thanks!

PR #3330, opened 2019-01-14
Bugfix: Real world example RepoPage Component was broken due to calli…

asleepysamurai:
Real world example, RepoPage Component was broken due to calling loadData with props.fullName rather than props


timdorr:
Thanks!

PR #3340, opened 2019-01-27
Enable collapsible sidebar navigation for website

yangshun:
Docusaurus v1.7.1 adds a new feature of making sidebar navigation collapsible. This is useful for websites with a large list of contents such as Redux.

screen shot 2019-01-26 at 10 02 29 pm
screen shot 2019-01-26 at 10 02 14 pm


yangshun:
Not sure why CI is failing. The failures don't seem related to the change.

markerikson:
Yeah, the failure is due to formatting issues.

I like Prettier, but I'm not a fan of making builds fail if things aren't formatted perfectly.

I like this idea in general, and I was absolutely thinking about this when I saw the Docusaurus PR get merged.  My only question is whether this is actually a net benefit.

I do see that the NgRx docs do the same kind of thing.

Thoughts?

markerikson:
Okay, I like it.  Awesome, thanks!

yangshun:
> I like this idea in general, and I was absolutely thinking about this when I saw the Docusaurus PR get merged. My only question is whether this is actually a net benefit.

I was skeptical as well then we enabled it on Docusaurus and it grew on me. Thanks for merging this 😄

timdorr:
> I like Prettier, but I'm not a fan of making builds fail if things aren't formatted perfectly.

This is necessary to prevent bickering over semicolons, quotes, and other bikeshedding topics. I'll reenable it on master. Please don't disable this again.

markerikson:
Okay, okay :)

PR #3341, opened 2019-01-27
Fix documentation container alignment

yangshun:
## Before 

Note that the "Learning Resources" is off alignment from the left and right columns. There's custom CSS which changes the default top padding which messes with the alignment especially on mobile view:

screen shot 2019-01-27 at 11 17 26 am
screen shot 2019-01-27 at 11 17 30 am

## After

On mobile view, the header doesn't get hidden behind the navigation.

screen shot 2019-01-27 at 11 17 48 am
screen shot 2019-01-27 at 11 17 52 am

### DIFF
--- a/website/static/css/custom.css
+++ b/website/static/css/custom.css
@@ -12,10 +12,6 @@ header.postHeader:empty + article h1 {
   padding-bottom: 1em;
 }
 
-.mainContainer {
-  padding-bottom: 24px;
-}
-
 .post article a {
   /* add underlines to links in blocks of text for a11y */
   color: $primaryColor;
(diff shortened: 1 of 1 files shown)


timdorr:
This misaligns the front hero section vertically. It's not perfect right now either, but it's closer (45px at the top and 50px vs 75px at the bottom). We shouldn't be using padding as an exterior alignment mechanism, so it's all a bit screwy anyways.

I'd get rid of this customized class entirely. The padding-bottom override shouldn't be happening either. Less noticeable unless you're on a very short page.

I think we can expand the top height of the hero section to match the bottom now too. That's not _too_ whitespacey.

timdorr:
OK, that looks better. Thanks, @yangshun!

PR #3343, opened 2019-01-28
fix(docs): Fixes navigation for shortcode headings

ivandevp:
When accessing to docs of API methods such as [`store.dispatch()`](https://redux.js.org/api/store#dispatch), the title gets cover by the navbar, but clicking on the anchor shortcut, gets rendered without problems. This PR adds the class used by the anchor, in order to have the same behavior.


timdorr:
Thanks!

Issue #3313, opened 2018-12-25
Discussion: revamped docs content, structure, and teaching flow


markerikson:
Per discussion in #2590 , we'd like to rethink the Redux docs content and structure pretty much from the ground up.  Now that we've got the new docs site up and running, it's time to start brainstorming what that content should look like.  (There's already a bunch of discussion and notes in #2590, but I wanted to open a new issue specifically to brainstorm ideas.)

There's a few questions that we should try to answer:

- What specific topics should we try to cover in the docs?
- What should the new user reading flow look like?
- How should we organize the content?
- How do we approach teaching Redux?
- How do we integrate teaching and encouraging the use of `redux-starter-kit` into the new docs?
- What are the weaknesses and deficiencies in the current docs that we are trying to resolve?
- How do we handle teaching prerequisite concepts like immutability, JS array methods, and event emitters, in relation to the actual Redux concepts and usage?  How do we differentiate between teaching "plain" Redux usage, and usage with a UI layer like React-Redux?

In general, there's _lots_ of good Redux tutorials, articles, and resources out there.  I would like us to beg, borrow, and steal the best ideas and examples out there for use with the docs (with permission if appropriate, of course :)  )

Anyone is welcome to contribute to this discussion, but I'd particularly like to get involvement from other people who have experience writing material that teaches Redux or working on docs.  On that note, tagging:

@dceddia
@rwieruch 
@joshwcomeau 
@valentinogagliardi
@alexkrolick
@matthew-gerstman
@tylermcginnis
@gaearon 
@timdorr

markerikson:
> Do we really want to encourage plain redux usage anymore?

Not "encourage", exactly, but I think there's definitely value in teaching people what the bare-bones "by hand" usage looks like, so they understand what's happening.  That way the later abstractions will make more sense.

As part of that, much of what RSK does only really makes sense once you understand the annoyances of doing everything "the hard way":

- `configureStore` simplifies the dance around middleware setup, composing enhancers, etc
- `createReducer` lets you avoid switch statements (which apparently is a big deal to many people), and supplies the "magic" in the form of automatically using Immer, thus allowing you to drastically simplify reducers
- `createSlice` generates action types and action creators for you so you don't have to write them by hand

But those are all abstractions that really benefit from the user understanding what doing it by hand looks like, so they know what's going on.  [Quoting Dan](https://twitter.com/acemarke/status/901329101088215044?lang=en):

[...]

PR #3349, opened 2019-02-04
Wipe away old docs, add a new TOC

matthew-gerstman:
This deletes all of the previous documentation and creates a new table of comments based on @markerikson's suggestion in #3313. 

I also deleted all of the content on existing docs that need to be reorganized. Even if we end up copy/pasting them back, I think it'll be valuable for us to go through the pull request process so that **everything** gets a new set of eyes.


markerikson:
Mmm.... "wipe it away" is considerably more drastic than what I was picturing.  I think I know what you're getting at, but I was figuring we'd start by shuffling the existing content around.

Feel free to keep playing with things on this branch, though.

matthew-gerstman:
I propose we merge this as this is a "baseline" branch.

In the past I've worked on features where we built a prototype and then later build a production-ready version. When we did that we typed git init on a new repo. Even if parts of the prototype got copy/pasted over it forced us to rereview it and question if we were making the right decisions.

From reading through the docs there are endless caveats and notes that may or may not be relevant anymore. I think it'll be easiest if we give the community a giant empty playground to start writing. 

Our goal here should be to empower individuals to contribute as much as possible and I think a clean slate is an effective way to do that.

If you'd like I'm happy to take a leadership role on this and manage this process.

PR #3351, opened 2019-02-05
Unnecessary div with unused className

DenrizSusam:
The div with className "mainContainer" seems to be unnecessary.


timdorr:
It's needed. It aligns our hero vertically.

timdorr:
It's in the default CSS for Docusaurus: https://github.com/facebook/Docusaurus/blob/1fb1a9229376b2909fa2a79f3a504b122a60885b/v1/lib/static/css/main.css#L619-L623

PR #3354, opened 2019-02-13
reduce code

chennng:
without these code, it can also work.


timdorr:
Everything works just fine as-is, so I'd rather not try to optimize or change this function too much. Thanks for the PR, though!

PR #3357, opened 2019-02-18
Remove --save option as it isn't required anymore

stevemoser:
"As of npm 5.0.0, installed modules are added as a dependency by default, so the --save option is no longer needed. The other save options still exist and are listed in the documentation for npm install."

https://stackoverflow.com/a/19578808/142358


timdorr:
Thanks!

PR #3361, opened 2019-02-24
(docs) minor correction

JoviDeCroock:
`this.` does not work in this scenario.

Proof: https://codesandbox.io/s/w07qqwlkq5 --> replace store.asyncReducers with this.asyncReducers. In utils/createStore


markerikson:
Thanks!

PR #3363, opened 2019-03-01
npm

szb512:


markerikson:
Uh.... no idea what this is supposed to be.

szb512:
It's supposed to improve npm.

markerikson:
Improve _what_?

It's bad enough that the PR doesn't have any kind of a meaningful description attached, but on top of that, I have no idea what these files are or why they're relevant to Redux at all.

I'm happy to look at PRs that are actually meaningful improvements to Redux itself or the tooling around it, but random unsolicited PRs that appear to be full of junk are not helpful.  Please don't do this again.

PR #3366, opened 2019-03-07
refactor: simplify isPlainObject function

V1c70r:
Makes the isPlainObject function simpler.


timdorr:
Thanks, but we don't really need to optimize this further. We'll replace it with lodash whenever they get to 5.0.

PR #3367, opened 2019-03-08
Update UsingObjectSpreadOperator.md

mtpetros:
I think this warning on using the Object spread operator to clone objects is important. While, good style would encourage the use of flat objects in the reducer, not everyone is going to do that. Using the spread syntax in a reducer on a multidimensional data structure is opening up the door to state mutation.


timdorr:
Dunno why this was never merged. Thanks!

PR #3371, opened 2019-03-08
Iterate using for in instead of Object.keys and minor fixes

pvorona:
Even though this code has no impact on the overall performance of the application `for in` is faster in modern browsers (https://jsperf.com/object-keys-vs-for-in-with-closure/138). 
I propose this change only as readability improvement.


timdorr:
Not a huge deal, but why not. Thanks!

PR #3374, opened 2019-03-12
Minor updates to make this example more usable

plug-n-play:
1. Added comment to better understand need for setTimeout
2. Increased Async time to 3 - gives more option to play around - e.g. click Async-add button but still click other options; was NA in 1 second.
3. Called out the time being taken in the Button text


timdorr:
Same here. I appreciate the PR, but this doesn't really add much to the example.

PR #3375, opened 2019-03-12
Removed hard-coded initial state

plug-n-play:
This make `render()` call more understandable.


timdorr:
Thanks, but I don't think this needs tweaking.

PR #3389, opened 2019-03-31
Allow Redux to be imported as `type="module"`

Wildhoney:
We'd all love to import Redux as `type="module"` natively in the browser, however the `process.env` and the `symbol-observable` ponyfill make that problematic.

By checking for the existence of `symbol-observable` ourselves &mdash; rather than delegating it to a module unless we really need to &mdash; as well as checking for `typeof process` we can make it happen.

Redux is a simple enough library. Let's all be standards compliant, friends (or as close as we can at this point)! 🍻 

Discuss...

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -118,7 +118,10 @@ export default function combineReducers(reducers) {
   for (let i = 0; i < reducerKeys.length; i++) {
     const key = reducerKeys[i]
 
-    if (process.env.NODE_ENV !== 'production') {
+    if (
+      typeof process !== 'undefined' &&
+      process.env.NODE_ENV !== 'production'
+    ) {
       if (typeof reducers[key] === 'undefined') {
         warning(`No reducer provided for key "${key}"`)
       }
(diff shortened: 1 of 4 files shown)


timdorr:
There is already a mjs build provided by #3143

Wildhoney:
Thanks @timdorr 

I saw the `*.mjs` file but discounted it without testing it, as it's typically a Node.js convention. I've not seen `*.mjs` being used as a browser module before anywhere else.

Do you think we need to document its existence? As the [current installation section](https://github.com/reduxjs/redux#installation) in the README doesn't *appear* to make reference to it.

Could we also not unify the module files (`es/redux.js` and `es/redux.mjs`), which was my attempt with this PR? That way we could reference a single file in the `package.json` file. At the moment the `*.mjs` has no reference in the `package.json` &ndash; that coupled with no README mention makes it terribly difficult to locate.

timdorr:
The mjs file is built in production mode only, so it cannot replace the standard module, which supports both. 

Some documentation on it's existence would be good. There isn't yet a package.json field for it that we can use safely, so an entry in the docs would be a good start.

PR #3392, opened 2019-04-02
Duplicate statements placed to functions for memoize

daydarkln:
There are duplicates of comparisons and it can be placed in functions.


timdorr:
Thanks, but this doesn't really add anything and makes the code harder to read through.

PR #3394, opened 2019-04-04
fixed undefined id when adding a new todo item

changx03:
the action `addTodo` only passes `text`. So the `action.id` in reducer gets `undefined`. Fixed by set `id` to `state.length+1` inside reducer.


timdorr:
Nope, it gets set by the action creator automatically: https://github.com/reduxjs/redux/blob/master/examples/todos/src/actions/index.js#L4

PR #3395, opened 2019-04-04
Fix chat action creators in UsageWithTypescript

stam:
The comments in the chat action creators imply that `TypeScript infers that this function is returning DeleteMessageAction`, while the returned union type is not defined.

If the return type is not manually linked to the Union Type, I have no idea how the inference based on just the constant should work.


timdorr:
Thanks!

PR #3398, opened 2019-04-08
Add quizz application to examples

slim-hmidi:
The quizz is a simple application wich will helps beginners to start with redux and learn the basics.


markerikson:
Hey.  Thanks for taking the time to put this together.  At the moment, I don't think we're looking to add any more examples directly to the repo.  We might take another look at this when we make some progress on the docs revamp in #3313 , though.

PR #3400, opened 2019-04-10
Update docs to describe current version of Redux-ORM

haveyaseen:
Hey guys, this is a [requested](https://stackoverflow.com/questions/55475808/module-descriptors-concept-in-redux-orm/55548296#comment97876038_55548296) update to the state management section describing Redux-ORM. It's actually been outdated since version 0.9 (we're at v0.13.2 now). We also moved the repository to a GitHub organization.

To prevent confusion, I simplified some payload properties and removed the part about "queued updates" which is an implementation detail regarding performance. The paragraph that I added in the end is supposed to make the benefits of managed relationships a bit more obvious.

### DIFF
--- a/docs/recipes/structuring-reducers/UpdatingNormalizedData.md
+++ b/docs/recipes/structuring-reducers/UpdatingNormalizedData.md
@@ -195,80 +195,73 @@ This approach makes it very clear what's happening for the `"ADD_COMMENTS"` case
 
 ### Redux-ORM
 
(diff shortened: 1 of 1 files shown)


timdorr:
Cool, thanks!

docs/recipes/structuring-reducers/UpdatingNormalizedData.md
```diff
@@ -195,80 +195,73 @@ This approach makes it very clear what's happening for the `"ADD_COMMENTS"` case
...
 export class Comment extends Model {
   static get fields() {
-    return {}
+    return {
+      id: attr(),
+      text: attr(),
+      // Define a foreign key relation - one Post can have many Comments,
+      // at a field named "comments"
+      postId: fk({
+        to: 'Comment',
```
fgallinari: There is an error here that drove me nuts.
It should be 'Post' instead of 'Comment'.

haveyaseen: Yep, my bad! Opened a PR with a fix.

PR #3402, opened 2019-04-12
fix broken links on Using Immutable.JS docs

tengis:
Hi guys,

I've noticed some of the links broken on this page.

Thanks

PR #3403, opened 2019-04-12
Fix docs formatting

markerikson:


timdorr:
The formatting is handled by prettier. If you've got a problem with it, take it up with them :)

markerikson:
That's actually exactly the issue.  I just merged #3402 , and forgot to check Travis first.  That PR has some bad formatting in it or something, so technically Travis is failing due to Prettier complaining.

But, when I check out master and run Prettier myself, it fixes some _different_ files, these the one from that PR isn't touched, and this branch _still_ fails.  

So, there's some kind of mismatch.

timdorr:
Are you running the latest version of Prettier?

markerikson:
Hmm.  Haven't updated the installed packages in my clone of this repo for a while.

Let's see how it looks now.

markerikson:
That seems to be better.

PR #3405, opened 2019-04-14
Use correct name for middleware

zygocephalus:
`loggerMiddleware` can't be imported in the `configureStore` function example.
https://github.com/reduxjs/redux/pull/3405/commits/8c2e936efd11838027adf4519c3995477b99f636#diff-178bc9a75ee1d44e782570fb9152d5a0L115


timdorr:
Thanks, but this is a default export, not a named export. So, the import statement can call it whatever it would like.

PR #3408, opened 2019-04-18
Add bracket for comments

pingfengafei:
`reducer` should wrapped with `[]` just the same as `preloadedState` and `enhancer`


timdorr:
Thanks!

epeterson320:
@pingfengafei I don't think it should be wrapped with brackets. That's how you mark a parameter as optional. To create a store, the reducer isn't optional.

http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

timdorr:
Oh, I'm not super familiar with JSDoc syntax. I'll revert this.

PR #3410, opened 2019-04-21
Add Deox to ecosystem

thebrodmann:
I wonder if you all would consider [Deox](https://deox.js.org/) as a "staff pick". I got good feedback on it (and 48 stars on GitHub so far).


markerikson:
Appreciate the PR, but I don't think this is something we'd want to add to that list at the moment.  That said, please go ahead and submit a PR to the overall Redux Addons List at https://github.com/markerikson/redux-ecosystem-links .

PR #3411, opened 2019-04-23
Infer action types from combineReducers

appden:
This change allows for `combineReducers` to completely infer both the state and action types for its returned reducer.

From experience with large TypeScript projects, it's common to see that the action type is not explicitly specified, which results in `AnyAction` in the resulting reducer type. Unfortunately, this will propagate through the type inference for `createStore` resulting in `dispatch` being very weakly typed. This change alone causes a chain reaction of a more correctly (and strongly) typed project with regards to Redux.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -94,9 +94,9 @@ export function combineReducers<T extends ReducersMapObject<any, any>>(
   reducers: T
 ): Reducer<InferStateType<T>, InferActionTypes<InferReducerTypes<T>>>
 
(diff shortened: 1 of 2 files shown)


timdorr:
Ah, good use of the `infer` keyword. When these were last touched, it was prior to TS 2.8, so that kind of thing wasn't even possible. 

Looks like some other formatting things came up unrelated to this PR. I'll see about fixing that up for you.

PR #3412, opened 2019-04-23
Fix Redux-ORM related model name

haveyaseen:
Also explain options using comments

See https://github.com/reduxjs/redux/pull/3400#pullrequestreview-229254034


timdorr:
Thanks!

PR #3413, opened 2019-04-23
Added "redux implementation" link

jakoblind:


timdorr:
Thanks, but I don't think we need to expand this section right now. If anything, we need to shrink it down.

PR #3414, opened 2019-04-28
add batch dispatch

gaoyonggege:


markerikson:
Nope, not doing this for several reasons:

- This won't work with the rest of the ecosystem (especially middleware)
- There's many addons that implement this functionality already
- Filing PRs to change core behavior without any good explanation isn't a good approach.

This is really a duplicate of #1813 .  See that issue for further details.

PR #3415, opened 2019-04-29
test(typescript): fix invalid tsconfig.json

thorn0:
The entries in `paths` should be arrays. See the [schema](http://json.schemastore.org/tsconfig).


timdorr:
Weird that works, but good idea to update it. Thanks!

PR #3418, opened 2019-04-30
remove "obviously"

panckreous:
inspired by #2950 - I think using "obviously" here can be confusing by assuming a reader knows it's obvious (and why), vs just stating that it's the most common.


markerikson:
Thanks!

Issue #3422, opened 2019-05-08
[examples] Todo with Undo example -  TypeError: Cannot read property 'length' of undefined


ntedgi:
todo with undo -  example crush both locally and on sandbox
https://codesandbox.io/s/github/reduxjs/redux/tree/master/examples/todos-with-undo

also, appear in this page broken 
https://redux.js.org/introduction/examples#todos-with-undo

the steps to reproduce
basic npm install and start

<img width="724" alt="Screen Shot 2019-05-08 at 8 07 52" src="https://user-images.githubusercontent.com/31243793/57351499-c3fb2900-716a-11e9-909c-39d270d932e8.png">
<img width="790" alt="Screen Shot 2019-05-08 at 8 08 04" src="https://user-images.githubusercontent.com/31243793/57351501-c3fb2900-716a-11e9-89d2-9923ad843b6f.png">

ntedgi:
after more forth investgating, it seems like a redux issue related 
changing in the package.json reduix version to     **"redux": "^3.5.2",**
working fine 
 tested with this dependencies 

 "dependencies": {
    "prop-types": "^15.7.2",
    "react": "^16.8.6",
    "react-dom": "^16.8.6",
    "react-redux": "^7.0.2",
    "redux": "^3.5.2",
    "redux-undo": "^0.6.1"
  },

ntedgi:
#3423

ntedgi:
https://github.com/reduxjs/redux/issues/3422
solve the issue by refactoring the example to new redux-undo API @ #3423

PR #3423, opened 2019-05-08
#3422 - Todo with Undo example - TypeError: Cannot read property 'length' of undefined -  bug fix

ntedgi:
change redux-undo API usage

following redux-undo last API here

https://www.npmjs.com/package/redux-undo

i change the following usage in undoRedo.js and todos.js


ntedgi:
#3422

timdorr:
Thanks, there were other changes that were just formatting stuff, so I took out the important bits and put them on master.

PR #3425, opened 2019-05-08
"Generating Reducers" in ReducingBoilerplate.md - return state object…

ScottyMJacobson:
… instead of array

If the array is indeed incorrect, this should close https://github.com/reduxjs/redux/issues/3424


markerikson:
Closing per comment in the linked issue.

PR #3428, opened 2019-05-13
expose bindActionCreator function

pranayrauthu:
Hi,
I have made a small change to expose `bindActionCreator` function. Although the existing `bindActionCreators` function accepts a function as first argument. I feel it's nice to have `bindActionCreator` function for readability when using a single dispatcher.


markerikson:
Appreciate the PR, but this is unnecessary.  `bindActionCreators()` already accepts just one action creator _or_ an object full of them.

PR #3430, opened 2019-05-20
easy optimization

fixerror:
(recalculating the length of the array in "for")
optimization


markerikson:
Appreciate the PR, but we don't need to try to micro-optimize like this.  Also, the test run time is going to fluctuate from run to run, and even if you were micro-benchmarking just `combineReducers` itself, you'd be hard-pressed to see any change at all.

fixerror:
Thank you, I will look for other ways)

PR #3431, opened 2019-05-24
Add docs about `batch()` API to Performance FAQ

augbog:
I noticed in the v7 release we added support for the `batch()` API. I know there are docs planned for upcoming releases but thought it could be something to mention in the Performance FAQ. In the section where we talk about batching updates, we still recommend in the FAQ other packages i.e. `redux-batched-actions`, etc. 

I feel it would be good to give users a heads up in this section to take a look if upgrading and using the new `batch()` API would address their needs. I still left text relevant to users who have not upgraded yet though. 

If you feel this is unnecessary or if I can improve the docs in any way, feel free to reject or let me know :) We started using the `batch()` API and it's been great.


markerikson:
Thanks!  I'll tweak the wording a bit, and then merge this in.

augbog:
For sure :) thanks for all your hard work <3

PR #3436, opened 2019-05-29
[DOCS]: Fixing small typo

raybooysen:
The docs have the `keysToRemove` array set to a const but then later in the code try to reassign the variable to another array.


markerikson:
Thanks!

PR #3439, opened 2019-06-03
fix(docs): Fix header anchor link positioning

### LINKED ISSUE #3427: Docs anchor link hover issue

bug

When hovering over some anchor links (the chain link icon), the content jumps. For example:

ohno

https://redux.js.org/api/store#getState

The issue seems to be related to the headers that use anchor tags.

```md
### <a id='getState' class='anchor'></a>[`getState()`](#getState)
```

The link's text is a link to the correct hash, but the anchor link ends up looking like this:

```
https://redux.js.org/api/store#a-id-getstate-class-anchor-a-getstate-getstate
```

wgao19:
What does this PR do?
---
Fixes #3427

So I've dug a bit more around the header anchor. There are four `<a />`s here: 
image

The first and the third are not shown because there is no content inside.

The second one, one with class name `.hash-link`, is the anchor for the icon. And the fourth one, one without any class name, is the one for the header.

Here are the two CSS rules involved, both in custom.css:

```css
.post article blockquote a:hover,
.post article blockquote a:focus,
.post article blockquote a:active {
  position: relative;
}
```

and 

```css
.post article .hash-link {
  position: absolute
}
```

The `.hash-link` has `position: absolute` which give it the desired positioning. It gets broken when the first rule is in action because it has higher CSS specificity.

### DIFF
--- a/website/static/css/custom.css
+++ b/website/static/css/custom.css
@@ -31,7 +31,6 @@ header.postHeader:empty + article h1 {
   outline: none;
   text-decoration: underline;
   overflow: hidden;
-  position: relative;
   transition: outline-offset 0.2s ease-in-out;
 }
 


markerikson:
Love the "lines of explanation to lines of code changed" ratio on this one :)  Always a great sign.

PR #3444, opened 2019-06-07
Add opt-out - with deprecation warnings - to the bans on getState, subscription Handling and dispatch in the reducers

9SMTM6:
Its not always easy to convert legacy code you may not even have written yourself, and as long as the continued acceptance of these antipatterns doesnt hurt other efforts I'd suggest the option for an opt-out of these bans, so that we dont force people to use older versions just because they feel like theres higher priorities than just completely revamping their applications state management right now.

This brings no breaking changes and the 'options' object may be expanded to allow for further configuration in the future. The documentation is riddled with warnings, and a console.error is set off every time a store gets created with one of these bans lifted. @timdorr this should COMPLETELY satisfy your apparent need to remove this behavior, as long as you dont provide a better reason I can accept or everyone else tells me they dont want to enable other people with legacy code to use the new features of Redux 4, just because these people think that using these functions in the reducer, while very bad, isnt the worst issue their app has currently and doesnt legitimize immediate complete workover of their potentially huge application.

This options object has been purposefully coded in a way that allows for expansion in the future, thus this does NOT prevent future expansions of arguments. I'd reccomend putting these in that options-object, as theyre unlikely to get used frequently and in that way - because of the non-sequential nature of the object-parameters - dont worsen the already bad situation currently where you have a headache with allowing different configurations without horrible syntax like passing in multiple `undefined`.

If this gets accepted and people are interested in that I might bring another merge request later which would allow a 3rd parameter in reducers which is passed the global state - thats also mentioned in the issue #1568 mentioned by @timdorr, the thing is THATS CURRENTLY INCOMPATIBLE WITH YOUR OWN COMBINEREDUCERS - allowing for in-pattern usage of the global state without having to revert to store.getState, which makes store-injection in testing very hard. This too would not break normal usage of redux, and when you use object destructuring in the argument to limit the scope of parameters you inject from the global state it also shouldnt lead to a to large state potentially influencing the reducer.

### DIFF
--- a/docs/api/createStore.md
+++ b/docs/api/createStore.md
@@ -18,6 +18,8 @@ There should only be a single store in your app.
 
 3. [`enhancer`] _(Function)_: The store enhancer. You may optionally specify it to enhance the store with third-party capabilities such as middleware, time travel, persistence, etc. The only store enhancer that ships with Redux is [`applyMiddleware()`](./applyMiddleware.md).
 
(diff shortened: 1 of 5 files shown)


markerikson:
Please stop re-submitting this.

PR #3450, opened 2019-06-17
 remove distinctState() filter from example

deanban:
distinctState() filter, which is deprecated, is now auto applied.


timdorr:
Thanks!

PR #3452, opened 2019-06-18
fix: allow applyMiddleware to be called more than once

### LINKED ISSUE #3451: Allow applyMiddleware to be applied more than once

Bug

ApplyMiddleware can only be called once for a given store. If you have 2 instances of applyMiddleware in a single store, dispatch will not function as expected. Calling dispatch from the first group of applied middleware will not call middleware in the second batch. This is because the applyMiddleware function references a local dispatch function, rather than a single, canonical dispatch that lives on the store object. 

I would argue the expected behavior is applyMiddleware can be applied several times. The docs are unclear on this point beyond stating it should be the first enhancer applied as the middleware may be asynchronous.

Allowing multiple invocations of applyMiddleware enables store enhancers that wrap different stores. Using the enhancer pattern, complex applications can be broken into smaller, composable pieces that are easier to test and reason about. This pattern requires that each enhancer can separately call applyMiddleware.

Based on the history of the appllyMiddleware file, all versions of redux are affected. But because most people don't call applyMiddleware more than once, it probably has never come up before.

ironcoconut:
closes #3451

I wonder if the docs should be updated as well?

### DIFF
--- a/src/applyMiddleware.js
+++ b/src/applyMiddleware.js
@@ -31,11 +31,9 @@ export default function applyMiddleware(...middlewares) {
       dispatch: (...args) => dispatch(...args)
     }
     const chain = middlewares.map(middleware => middleware(middlewareAPI))
-    dispatch = compose(...chain)(store.dispatch)
+    store.dispatch = compose(...chain)(store.dispatch)
+    dispatch = (...args) => store.dispatch(...args)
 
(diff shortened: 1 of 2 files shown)


markerikson:
Thanks for the PR, but I don't see a need to do this.

PR #3462, opened 2019-07-01
docs: fix Store api sidebar links

FLGMwt:
This page has a table of contents *and* a sidebar which I guess is a bit tricky in docusaurus. I put the ###s back to plaintext, which made the sidebar prettier, then switched the ToC to use hashes that those linked to to ensure that that works too.

<3

|Before|After|
|---|---|
|image|image|


markerikson:
Thanks!

PR #3465, opened 2019-07-06
TypeScript: `Reducer`: allow extra parameters

OliverJAsh:
This is sometimes necessary, e.g.:

> … pass a custom third argument from a parent reducer if a child reducer needs additional data to calculate its next state.

https://redux.js.org/faq/reducers#how-do-i-share-state-between-two-reducers-do-i-have-to-use-combinereducers

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -58,11 +58,11 @@ export interface AnyAction extends Action {
  * @template S The type of state consumed and produced by this reducer.
  * @template A The type of actions the reducer can potentially respond to.
  */
-export type Reducer<S = any, A extends Action = AnyAction, R extends unknown[] = []> = (
-  state: S | undefined,
-  action: A,
-  ...rest: R
-) => S
+export type Reducer<
+  S = any,
+  A extends Action = AnyAction,
+  R extends unknown[] = []
+> = (state: S | undefined, action: A, ...rest: R) => S
 
 /**
  * Object whose values correspond to different reducer functions.
--- a/index.d.ts
+++ b/index.d.ts
@@ -58,9 +58,10 @@ export interface AnyAction extends Action {
  * @template S The type of state consumed and produced by this reducer.
  * @template A The type of actions the reducer can potentially respond to.
  */
-export type Reducer<S = any, A extends Action = AnyAction> = (
+export type Reducer<S = any, A extends Action = AnyAction, R extends unknown[] = []> = (
   state: S | undefined,
-  action: A
+  action: A,
+  ...rest: R
 ) => S
 
 /**


index.d.ts
```diff
@@ -58,10 +58,11 @@ export interface AnyAction extends Action {
...
  * @template A The type of actions the reducer can potentially respond to.
  */
-export type Reducer<S = any, A extends Action = AnyAction> = (
-  state: S | undefined,
-  action: A
-) => S
+export type Reducer<
+  S = any,
+  A extends Action = AnyAction,
+  R extends unknown[] = []
```
Andarist: out of curiosity - does using unknown here in this context gives anything over good old any? tried few things in TS playground and they seem pretty much the same in this regard because R is going to be widened as much as possible anyway and everything extends uknown (and any)

timdorr:
That documentation is a bit oddly-worded, but [this isn't something we do ourselves](https://github.com/reduxjs/redux/blob/master/src/combineReducers.js#L168) and would only be provided with middleware. I think they're referring to redux-thunk there, which does have this feature. 

For the code we provide and its types, things are correct as-is. For middleware libraries, they should provide their own type overrides for these extra args.

index.d.ts
OliverJAsh: Not sure it does in this case. I always just use `unknown` because it's safer. I've used `any` in the past and it's leaked into places I didn't want it to!

OliverJAsh:
@timdorr IIUC, the docs I quoted are describing a technique along the following lines:

``` ts
const childReducer = (state, action, myDependency) => {
  // …
}

const rootReducer = (state, action) => {
  return {
    childState: childReducer(state.childState, action, myDependency)
  }
}
```

We would like to import the `Reducer` type from Redux so we can correctly annotate our `childReducer` function. The type is not just used internally and in middleware libraries—it can be used by direct consumers of Redux too.

Without the `Reducer` built-in type supporting this, users will have to define their own `Reducer` type in order to be able to do this (which is what we're currently doing). That's fine, but doesn't it make sense for the built-in type to encompass possible and valid use cases?

Andarist:
IIUC the typing is actually correct - because it constrains `createStore` calls nicely (it rather shouldnt accept reducer accepting more than 2 arguments). It could be called `RootReducer` which would avoid confusion - the only unfortunate thing is that combineReducers is using this typing, which is also correct as it also doesnt allow extra args, so maybe it could be named smth like `RootReducerLike`.

This would complicate things though for questionable gains. Fortunately TS is more about structural typings rather than nominal one, so if you want to dispatch your root reducer to inner reducers with different paramaters list than the root one it seems reasonable to create a "copy" of the builtin Reducer type for this purpose which can cover this use case.

Just my 2 cents.

PR #3471, opened 2019-07-11
Add support for Object.create(null) as state object

Connormiha:
Redux throws error if i use `action` or `store` object from `Object.create(null)`.
Not a typical case. But nevertheless, maybe remove this limitation?


markerikson:
What is the actual use case where this would be beneficial for you?

Connormiha:
I think, it's of little use. Experimenting in one legacy project, I discovered this feature, when i tried migrate from `self-written redux-like` lib to `redux`. I do not insist on this change.

markerikson:
Yeah, while it's not a particularly complex change, I don't see an immediate reason to add this.  Thanks for the PR, though!

PR #3478, opened 2019-07-21
Fix broken link to database design book

lindboe:
The original link was to a site that is no longer up. The book is now hosted on a different site, so changed link to that.


markerikson:
Thanks!

PR #3480, opened 2019-07-24
Intellij Dependency

rummanhasnayeen:
.gitignore update to ignore Intellij IDEA project files.


timdorr:
This goes in your global gitignore file, not in the project's. This project isn't dependent on any IntelliJ IDE, so it's not part of it's resources.

PR #3481, opened 2019-07-24
Update ignoring vscode dependency

nafistiham:
ignoring .vscode file while source commits.


timdorr:
Same as #3480. This isn't a project dependency, so you should ignore this locally on your machine.

Issue #2979, opened 2018-05-06
typings for combineReducers infers overly action type from reducer map


ohjames:
**Do you want to request a *feature* or report a *bug*?**  

Bug

**What is the current behavior?**

The following minimal reproduction shows the reducer is incorrectly typed to accept `AnyAction`

```typescript
interface A1 { type: 'A1' }
interface A2 { type: 'A2' }
type Action = A1 | A2
const reducerMap = {
  s1: (s: number, a: Action) => 1,
  s2: (s: number, a: Action) => 2,
}
const reducer = combineReducers(reducerMap)
// the reducer call works even though it should only accept actions A1 and A2.
reducer({ s1: 0, s2: 0 }, { type: 'blah' })
```

**What is the expected behavior?**

The action types should be inferred from the action. If the typings are changed to this it works:

```typescript
type FunctionReturnType<T> = T extends (...args: any[]) => infer R ? R : never

type CombinedState<T extends object> = { [K in keyof T]: FunctionReturnType<T[K]> }

type ActionFromReducer<T extends object> = T[keyof T] extends (
  state: any,
  action: infer A,
  ...args: any[]
) => any
  ? A
  : never

type ReducerReturnValue<T extends object> = (
  state: CombinedState<T>,
  action: ActionFromReducer<T>,
) => CombinedState<T>

export function combineReducers<T extends object>(reducers: T): ReducerReturnValue<T>
```

But this relies on conditional types which were added in typescript `2.8`. 

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**

Redux 4.0.0

timdorr:
[We're using Typescript 2.8.](https://github.com/reactjs/redux/blob/master/package.json#L89) Feel free to submit a PR.

ohjames:
It would prevent anyone using earlier releases of TypeScript from using the official redux typings. It would also break for those explicitly specifying the state/action type parameters. You could argue that with this level of inference there's no longer a need to use those parameters, at the same time some might be using them to explicitly verify their reducer map function. I like it this way at least ;)

aikoven:
@ohjames Current typings support restricting action type, see [the test](https://github.com/reactjs/redux/blob/v4.0.0/test/typescript/reducers.ts#L58).

bloadvenro:
@ohjames thank you! Your solution was very helpful. I quickly modified it:

```
type ReducersMapObject = { [index: string]: Reducer<any, any> };

[...]

PR #3484, opened 2019-07-28
#2979 Add strict type inference overload for combineReducers.

ChrisAckerman:
* Strictly infers state shape and actions from the reducers object map.
* Updated tests and comments to match new strict type inference.
* Tested and working in typescript 2.8.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -73,6 +73,44 @@ export type ReducersMapObject<S = any, A extends Action = Action> = {
 }
 
 /**
+ * Infer a combined state shape from a `ReducersMapObject`.
+ *
+ * @template M Object map of reducers as provided to `combineReducers(map: M)`.
+ */
+export type StateFromReducersMapObject<M> = M extends ReducersMapObject
+  ? { [P in keyof M]: M[P] extends Reducer<infer S, any> ? S : never }
+  : never
+
+/**
+ * Infer reducer union type from a `ReducersMapObject`.
+ *
+ * @template M Object map of reducers as provided to `combineReducers(map: M)`.
+ */
+export type ReducerFromReducersMapObject<M> = M extends {
+  [P in keyof M]: infer R
+}
+  ? R extends Reducer<any, any>
+    ? R
+    : never
+  : never
+
+/**
+ * Infer action type from a reducer function.
+ *
+ * @template R Type of reducer.
+ */
+export type ActionFromReducer<R> = R extends Reducer<any, infer A> ? A : never
+
+/**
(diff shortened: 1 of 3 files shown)


ChrisAckerman:
This is a fix for issue #2979.

timdorr:
This is basically what #3411 was doing, but without the BC breakage. I think this is good, but if others want to chime in, feel free.

timdorr:
BTW, we're testing against 3.5.3 now, so compatibility with older versions isn't really of concern. I've found most people upgrade TS pretty regularly.

pfgray:
Should we be encouraging shapes of reducers that include any type other than `Action<any>` as the second parameter? I'm afraid that reducers with specific action types aren't reflective of reality, since redux sends _all_ actions to _all_ reducers (even "private" redux actions).

For example, suppose we have this seemingly innocuous reducer:
```ts
type CounterAction = IncrementAction | DecrementAction
type IncrementAction = {type: "increment", amount: number}
type DecrementAction = {type: "decrement", amount: number}

function counterReducer(state: number = 0, action: CounterAction): number {
  if(action.type === "increment") {
    return state + action.amount;
  } else {
    return state - action.amount;
  }
}
``` 

The `counterReducer` believes it will only ever receive `CounterAction`s, but we know this isn't true, since all redux reducers receive all actions! 

This particular reducer produces a state of `NaN`, since TS indicates it's safe to assume any `action` that's not `IncrementAction` is a `DecrementAction`.

ChrisAckerman:
@pfgray The intention here was to make the combined reducer consistent with the typing of the individual reducers.

You're right that maybe individual reducers _should_ use `AnyAction` instead of specific action types. If you did that, then this would faithfully keep that `AnyAction` type on the combined reducer.

But, if you do use more specific types for your reducer action parameters, then currently that typing is stripped by combining.

Personally, I do specify types for my actions. But, I do it knowing I can't use an implicit else/default in my reducer. I add the types for testing (so I know when I'm testing an expected action), and for action property types when writing the code that consumes the action.

pfgray:
> But, I do it knowing I can't use an implicit else/default in my reducer.

I think we should make that implicit assumption explicit in the type system! That way less astute developers (like myself!) don't get tripped up.

I think the right approach would be to remove the first signature of `combineReducers`, which would prevent specific action-typed reducers, and would make this change unnecessary.

timdorr:
So, aside from the AnyAction discussion, which is valid but better suited for another issue/PR, I think we're good here. Let's merge this in.

PR #3485, opened 2019-07-30
#2808 Preloaded state is now selectively partial (instead of deeply partial).

### LINKED ISSUE #2808: TypeScript: usage of `DeepPartial` for `preloadedState` is too deep

ChrisAckerman:
Fixes #2808 by using a specialized (but effectively NO-OP) type `CombinedState` to indicate that the `Reducer` returned by `combineReducers()` has a generic state type which should be considered shallowly partial.

_NOTE: Requires Typescript >= 3.5 to work correctly._

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -32,6 +32,47 @@ export interface AnyAction extends Action {
   [extraProps: string]: any
 }
 
+/**
+ * Internal "virtual" symbol used to make the `CombinedState` type unique.
+ */
+declare const $CombinedState: unique symbol
+
+/**
+ * State base type for reducers created with `combineReducers()`.
+ *
+ * This type allows the `createStore()` method to infer which levels of the
+ * preloaded state can be partial.
+ *
+ * Because Typescript is really duck-typed, a type needs to have some
+ * identifying property to differentiate it from other types with matching
+ * prototypes for type checking purposes. That's why this type has the
+ * `$CombinedState` symbol property. Without the property, this type would
+ * match any object. The symbol doesn't really exist because it's an internal
+ * (i.e. not exported), and internally we never check its value. Since it's a
+ * symbol property, it's not expected to be unumerable, and the value is
+ * typed as always undefined, so its never expected to have a meaningful
+ * value anyway. It just makes this type "sticky" when we cast to it.
+ */
+export type CombinedState<S> = { readonly [$CombinedState]: undefined } & S
+
+/**
+ * Helper to extract the raw state from a `CombinedState` type.
+ */
+export type UnCombinedState<S> = S extends CombinedState<infer S1> ? S1 : S
+
+/**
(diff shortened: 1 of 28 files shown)


index.d.ts
```diff
@@ -59,9 +100,9 @@ export interface AnyAction extends Action {
  * @template A The type of actions the reducer can potentially respond to.
  */
 export type Reducer<S = any, A extends Action = AnyAction> = (
-  state: S | undefined,
+  state: UnCombinedState<S> | undefined,
   action: A
-) => S
+) => UnCombinedState<S>
```
timdorr: Hmm, this is the only part I really have issue with, mainly from an ergonomics perspective. The type returned from the reducer is now this wrapped state type, not my original state. I know it is evaluated to be that original state, but won't it appear in TS error traces? That seems a bit confusing, especially for a central type.

dawnmist: I'd almost suggest renaming "UnCombinedState" to simply "State", which both reflects that this is the main state and hopefully be less confusing naming in error traces. Otherwise I'm delighted to see that there is now a way to type the PreloadedState properly.

ChrisAckerman: How about `ReducerState` since it is the state as normalized for input to and output from a `Reducer`?

ChrisAckerman: Never mind. I found a way to make the `[$CombinedState]` property optional, which means it doesn't need to be stripped out in `Reducer`, and `UnCombinedState` is no longer necessary at all.

timdorr:
It looks like we can get rid of `DeepPartial` as well. Since that would be a breaking change, can we at least remove it from the imports on the tests?

ChrisAckerman:
Unsure why prettier doesn't like `docs/recipes/UsingObjectSpreadOperator.md`... I didn't change it and prettier doesn't complain when I run it locally.

ChrisAckerman:
Ah, this branch was behind master, which had prettier formatting flaws in the UsingObjectSpreadOperator.md. Fixed the formatting and re-pushed.

timdorr:
@ChrisAckerman Do you want to resolve the conflicts here?

ChrisAckerman:
@timdorr Done.

timdorr:
Awesome, thanks!

PR #3487, opened 2019-07-31
Add import statement to visibilityFilter example

dalyhabit:
In this code example, SET_VISIBILITY_FILTER is used in a switch case, but it is never imported first. It is imported in the subsequent example code block in the docs. This is confusing, especially for users who aren't familiar with switch statements.


timdorr:
Thanks!

Issue #3488, opened 2019-07-31
State doesn't update when replaceReducers is called with combineReducers and one of the reducers passed to it is removed. 


anubhavgupta:
**Do you want to request a _feature_ or report a _bug_?**
bug

**What is the current behavior?**
The `combineReducers` function does't detect if one of the reducers passed to it is removed(via replaceReducers) and thus returns an old state.
  
```js
// Reducers
const foo = (state = { value: 'foo' } ) => state
const bar = (state = { value: 'bar' }) => state

const ACTION = { type: 'ACTION' };

const originalCompositeReducer = combineReducers({ foo, bar })
const store = createStore(originalCompositeReducer)
store.dispatch(ACTION)


const initialState = store.getState() // returns -> { 
 //                                                   foo: { value: 'foo' }, 
 //                                                   bar: { value: 'bar' } 
 //                                                }

// now remove `foo` reducer
store.replaceReducer(combineReducers({ bar }))
const nextState = store.getState() // returns -> { 
//                                                 foo: { value: 'foo' }, 
//                                                 bar: { value: 'bar' } 
//                                                }
//                      should have returned -> { bar: { value: 'bar' } }
expect(nextState).not.toBe(initialState) // --> this fails
```

Steps to reproduce: 
Open [JSFiddle Link](https://jsfiddle.net/f5er907y/2/):
1) Click "Dispatch action" button
-> 2 boxes gets rendered,  one of `foo` and other of `bar`. 
2) Click "Replace Reducer: remove `bar` reducer" button
-> Again, 2 boxes gets rendered,  one of `foo` and other of `bar`. 
**Expected**: Only one box should render, i.e. box of `bar` only.

**What is the expected behavior?**
The `combineReducers` function should detect if one or more reducers passed to it is removed(via replaceReducers) and thus should update the store with new state.

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**
Redux Version: V4.0.4
OS: MAC
Worked on previous versions: no.

anubhavgupta:
Raised a PR which will fix this issue.

lileilei:
```
test
redux@4.0.4
onClick={e=>{ 
      store.replaceReducer(combineReducers({ bar }))
 }}
 it didn't work！！

---------------------
redux@3.7.2  
is
success ！

markerikson:
Given that #3488 has been merged, this can be closed.

PR #3490, opened 2019-07-31
Fixed combineReducers changeDetection logic(#3488)

anubhavgupta:
## Fixes following issue:
State doesn't update when replaceReducers is called with combineReducers and one of the reducers passed to it is removed.(https://github.com/reduxjs/redux/issues/3488)

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -173,6 +173,8 @@ export default function combineReducers(reducers) {
       nextState[key] = nextStateForKey
       hasChanged = hasChanged || nextStateForKey !== previousStateForKey
     }
+    hasChanged =
+      hasChanged || finalReducerKeys.length !== Object.keys(state).length
     return hasChanged ? nextState : state
   }
 }
--- a/test/combineReducers.spec.js
+++ b/test/combineReducers.spec.js
@@ -192,6 +192,7 @@ describe('Utils', () => {
       expect(spy.mock.calls[0][0]).toMatch(
         /Store does not have a valid reducer/
       )
+
       spy.mockClear()
       console.error = preSpy
     })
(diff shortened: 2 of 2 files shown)


timdorr:
What if replaceReducers swaps out one reducer for another? I don't think length is sufficient for a check here.

anubhavgupta:
Are you suggesting about following use case?:
```js
const bar = (state = { value: 'bar' } ) => state
const baz = (state = { value: 'baz' } ) => state

const originalCompositeReducer = combineReducers({ bar })
const store = createStore(originalCompositeReducer)
store.dispatch(ACTION)
const initialState = store.getState() // returns ->  { bar: { value: 'bar' }}

// replacing `bar` reducer with `baz` reducer
store.replaceReducer(combineReducers({ baz })) // returns ->  { baz: { value: 'baz' }}

const nextState = store.getState()
expect(nextState).not.toBe(initialState)
```
I have tried it out and it works as expected. 
Should I add this test case in the PR?

anubhavgupta:
@timdorr Here is an exhaustive test case. It verifies the change detection logic when we replace one reducer with another.

Expected `has state changed` results when we switch each `combined reducer` listed in the first column with the each`combined reducer` listed in the first row.

| | { foo, bar }        |  { foo, bar, baz }           | { foo, baz }  |
| ------------- |:-------------:|:-----:|:-----:|
| **{ foo, bar }**      | same state | state chages |  state chages  |
| **{ foo, bar, baz }**  | state chages | same state | state chages |
| **{ foo, baz }** | state chages | same state | state chages |


```js

const foo = (state = { value: 'foo' }) => state
const bar = (state = { value: 'bar' }) => state
const baz = (state = { value: 'baz' }) => state

const combinedReducers = [
  combineReducers({ foo, bar }),      // reducer 1
  combineReducers({ foo, bar, baz }), // reducer 2
  combineReducers({ foo, baz }),      // reducer 3
];

// Expected `has state changed` results when we switch one reducer with another.
const hasStateChangedResults = [
  // eg: when we switch `reducer 1` with `reducer 2` the resulting state should change.
  // But if we replace `reducer 1` with `reducer 1`(same reducer) the resulting state should not change.
  //1->1  1->2  1->3
  [false, true, true],
  //2->1  2->2  2->3
  [true, false, true],
  //3->1  3->2  3->3
  [true, true, false]
];

// initialize store with any reducer
const store = createStore(combinedReducers[0])
store.dispatch(ACTION)

// run all test cases
for(let i=0;i < hasStateChangedResults.length; i++){
  for(let j=0; j < hasStateChangedResults[i].length; j++){
    store.replaceReducer(combinedReducers[i])
    const state1 = store.getState()
    store.replaceReducer(combinedReducers[j])
    const state2 = store.getState()
    expect(state1 !== state2).toBe(hasStateChangedResults[i][j])
  }
}
```
I have verified it with the PR code and it works.

timdorr:
OK, thanks for the exhaustive testing!

PR #3491, opened 2019-08-01
fix: isPlainObject compatible Object.create(null)

jinxin0112:
Compatibility of isPlainObject methods Object.create(null) like [lodash.isPlainObject](https://github.com/lodash/lodash/blob/master/isPlainObject.js)


timdorr:
Thanks, but this function is good enough as is. We don't care about every edge case, just the most common ones.

PR #3494, opened 2019-08-02
Fix links to some headings with special chars

douglasnaphas:
A couple of links in tables of contents were broken.

These were links where the URL fragment in the target was
auto-generated. When that happens, most special characters are removed
from the heading.

When the links in the tables of contents were written, the ones I am
fixing replaced some of the special characters with `-`, instead of
stripping them.

I tested the links on my fork on the [Performance](https://github.com/douglasnaphas/redux/blob/broken-links-quotes-in-headings/docs/faq/Performance.md) and [React Redux](https://github.com/douglasnaphas/redux/blob/broken-links-quotes-in-headings/docs/faq/ReactRedux.md) pages.

These screenshots show that the currently live auto-generated URLs
match what I am changing the links to:

Screen Shot 2019-08-01 at 8 58 09 PM

Screen Shot 2019-08-01 at 8 59 46 PM

### DIFF
--- a/docs/faq/Performance.md
+++ b/docs/faq/Performance.md
@@ -10,7 +10,7 @@ hide_title: true
 ## Table of Contents
 
 - [How well does Redux “scale” in terms of performance and architecture?](#how-well-does-redux-scale-in-terms-of-performance-and-architecture)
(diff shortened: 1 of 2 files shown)


markerikson:
Thanks!

PR #3495, opened 2019-08-02
Remove outdated rule on single quotes in docs

douglasnaphas:
In PR #1874, a mass replacement of angled single quotes ’ with ASCII
single quotes ' was approved. Since then, the docs have contained a
mixture of angled and ASCII single quotes:

```
$ git log -S’ --since="Mon Aug 1 16:45:58 2016 -0400" --pretty=oneline | wc -l
17
```

so it does not make sense to state a rule on them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,9 +70,7 @@ npm run test:watch
 
 ### Docs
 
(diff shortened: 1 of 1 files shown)


timdorr:
I'm not even sure why we have this paragraph. We don't enforce this. (AFAIK, Prettier can't do this for us yet). Let's just get rid of it entirely.

PR #3497, opened 2019-08-04
docs: fix store methods link

kon-shou:
### Problem

All links to [Store methods](https://github.com/reduxjs/redux/blob/master/docs/api/Store.md#store-methods) seem to be outdated.

### Solution

| before           | after                      |
|------------------|----------------------------|
| #dispatch       | #dispatchaction          |
| #subscribe      | #subscribelistener        |
| #replaceReducer | #replacereducernextreducer |


timdorr:
Thanks!

PR #3498, opened 2019-08-06
Minor language fix

joshbambrick:
A slight improvement to the language, because:
1. "anymore" is not word
1. the second usage is also ambiguous - does it mean "we don't use it fromhereonin" or "we no longer use it"?


timdorr:
Thanks!

PR #3503, opened 2019-08-14
Add ".idea" folder to .gitignore

deepfunc:
It is generated by JetBrains IDEs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ website/translated_docs
 website/build/
 website/node_modules
 website/i18n/*
+
+/.idea/


timdorr:
No, this belongs in your global gitignore file. We don't require the use of JetBrains IDEs, so we shouldn't make affordances for it at the project level.

deepfunc:
I think if someone like me using JetBrains IDEs to clone and open this project, will see ".idea" folder is untracked in git status. After all, not everyone should set it in global gitignore file. :)
Like this repository: [https://github.com/facebook/react/blob/master/.gitignore](https://github.com/facebook/react/blob/master/.gitignore)
@timdorr

timdorr:
Yes, you absolutely should. If you use that IDE, you should ignore it. It doesn't belong in any repo.

deepfunc:
> Yes, you absolutely should. If you use that IDE, you should ignore it. It doesn't belong in any repo.

react, angular, webpack, TypeScript, bootstrap, electron, and more repositories set it in own gitignore file.
Why?

PR #3504, opened 2019-08-14
scaffolding for typescript

cellog:
This PR sets up the project to use typescript. It does not yet change any of the actual functionality.

It adds:

1. `@babel/typescript` for compiling typescript to js using babel
2. `@typescript-eslint` and needed sub-packages for linting typescript
3. disables `no-unused-vars` in favor of `@typescript-eslint/no-unused-vars`, and adds a new strict rule exception - any variable whose name starts with '_' can be ignored as unused
4. `tsconfig.json` which controls the type validation one will see in VSCode, and the result of explicitly running `tsc`
5. a new `check-types` script that is added to `prepare` to check types prior to compilation

Once this is merged, the actual work of typescript conversion may commence


timdorr:
Looks like eslint-plugin-import got dropped from the lockfile somehow. Try another npm install and that should fix the CI failure.

JustFly1984:
Guys we are using `tsdx` to build our library at `@react-google-maps/api` and it prove to be reliable and produces very small bundle

markerikson:
@JustFly1984 : I've looked at `tsdx` a bit, and agree it seems cool.  That said, we've got an existing build setup, and I'd think it'll be easier to integrate TS into what we've got now (especially given our use of Babel) then it would be to switch build setups entirely.

JustFly1984:
Btw you don't need eslint-plugin-import for typescript, more than that - it just doesn't support TS

markerikson:
I just created a new `ts-conversion` branch and updated the PR to merge into that instead, so we can play around there without disturbing `master`.

cellog:
> Btw you don't need eslint-plugin-import for typescript, more than that - it just doesn't support TS

The PR addresses this issue. If `eslint-plugin-import` is removed, the build fails. It is part of the `react-app` preset. In `.eslintrc`, we disable it with:

```js
    'no-unused-vars': 'off',
```

and then replace it with the typescript version here:

```js
    '@typescript-eslint/no-unused-vars': [
      'error',
      {
        vars: 'all',
        args: 'after-used',
        ignoreRestSiblings: true,
        argsIgnorePattern: '^_' // ignore unused variables whose name is '_'
      }
    ]
```

I spent many long hours in a past project trying to figure this out :). If you have a better solution, let's make it happen!

cellog:
ah - I did not look at `.travis.yml`, I'll fix the linting errors in `examples` and push that later, then this should be good to go

cellog:
OK, so feedback needed here - the `todos-flow` example flips out because it is in flow, and not ts. Do we need to lint this example? If so, we will need to add some special case linting that excludes .ts and does flow linting.

timdorr:
I'm happy with just getting rid of that example. I really don't see it adding much value to Redux. It's for Flow users, not Redux users.

JustFly1984:
Well, actually you don’t need create-React-app preset either, you can use eslint-plugin-React, eslint-plugin-React-hooks, eslint-plugin-React-perf and that will be enough.

PR #3506, opened 2019-08-15
migrate bindActionCreators test to typescript

cellog:
This test was easy to migrate, all of the failures involved testing against type mis-matches that can only occur in javascript.


timdorr:
Thanks, Greg!

PR #3507, opened 2019-08-16
fix replaceReducer type

### LINKED ISSUE #3482: TypeScript defintion of StoreEnhancer does not allow state extension with replaceReducer

The `StoreEnhancer` TypeScript interface does not seem to have the appropriate type for the `replaceReducer` property of the enhanced store. I would consider it a bug in the type definition.

When implementing the `StoreEnhancer` interface with `ExtraState`, the type signature of `replaceReducer` on the "enhanced" store is coupled to the type of the wrapped reducer.

Given a `StoreEnhancer` of type

```ts
StoreEnhancer<Ext, ExtraState>
```

the returned store is of type

```ts
Store<S & StateExt, A> & Ext
```

with the `replaceReducer` property as such

```ts
(nextReducer: Reducer<S & ExtraState, A>) => void
```

Returning a store with a `replaceReducer` that accepts the original reducer gives the following type-error:

```
Type '<S, A extends Action<any> = AnyAction>(reducer: Reducer<S, A>, preloadedState?: DeepPartial<S> | undefined) => { replaceReducer: (nextReducer: Reducer<S, A>) => void; dispatch: Dispatch<A>; getState(): S & ExtraState; subscribe(listener: () => void): Unsubscribe; [Symbol.observable](): Observable<...>; }' is not assignable to type 'StoreEnhancerStoreCreator<{}, ExtraState>'.
      Type '{ replaceReducer: (nextReducer: Reducer<S, A>) => void; dispatch: Dispatch<A>; getState(): S & ExtraState; subscribe(listener: () => void): Unsubscribe; [Symbol.observable](): Observable<S & ExtraState>; }' is not assignable to type 'Store<S & ExtraState, A>'.
        Types of property 'replaceReducer' are incompatible.
          Type '(nextReducer: Reducer<S, A>) => void' is not assignable to type '(nextReducer: Reducer<S & ExtraState, A>) => void'.
            Types of parameters 'nextReducer' and 'nextReducer' are incompatible.
                Type 'S | undefined' is not assignable to type '(S & ExtraState) | undefined'.
                  Type 'S' is not assignable to type '(S & ExtraState) | undefined'.
                    Type 'S' is not assignable to type 'S & ExtraState'.
                      Type 'S' is not assignable to type 'ExtraState'.

```

The type check fails for the following code:

```ts
import {
  StoreEnhancer,
  Action,
  AnyAction,
  Reducer,
  createStore,
  DeepPartial
} from "redux";

interface State {
  someField: "string";
}

interface ExtraState {
  extraField: "extra";
}

const reducer: Reducer<State> = null as any;

function stateExtensionExpectedToWork() {
  interface ExtraState {
    extraField: "extra";
  }

  const enhancer: StoreEnhancer<{}, ExtraState> = createStore => <
    S,
    A extends Action = AnyAction
  >(
    reducer: Reducer<S, A>,
    preloadedState?: DeepPartial<S>
  ) => {
    const wrappedReducer: Reducer<S & ExtraState, A> = null as any;
    const wrappedPreloadedState: S & ExtraState = null as any;
    const store = createStore(wrappedReducer, wrappedPreloadedState);
    return {
      ...store,
      replaceReducer: (nextReducer: Reducer<S, A>): void => {
        const nextWrappedReducer: Reducer<S & ExtraState, A> = null as any;
        store.replaceReducer(nextWrappedReducer);
      }
    };
  };

}
```

See `src/index.ts` in the linked codesandbox example that implements the same function that would be expected to type-check, followed by another function of how it actually has to be done.

https://codesandbox.io/s/redux-store-enhancer-types-s6d3v

The example is based on the typescript test for the enhancer at [`test/typescript/enhancers.ts`](https://github.com/reduxjs/redux/blob/b6e1586d382b7e1f8f27776c3f422ef59efcd33a/test/typescript/enhancers.ts#L36) in this repo. The code doesn't execute (due to unsafe casts of null as any), but it is the type check that is of interest here.

When a store is created with a store enhancer that wraps a reducer and adds state, I would expect that `replaceReducer` on the returned store can be called with the original `rootReducer`.

```ts
const store = createStore(rootReducer, preloadedState, enhancer)
// ...
store.replaceReducer(rootReducer)
```

It would be the responsibility of the `enhancer` to appropriately replace the wrapped reducer. I.e return a store such as:

```ts
{
  ...store,
  replaceReducer: (nextReducer: Reducer<S, A>) => {
    store.replaceReducer(wrapReducer(nextReducer))
  },
}

```

Redux version: 4.0.4,
OS: Ubuntu 19.04
Browser: N/A
Did this work in previous versions of Redux?: Not that I know of

cellog:
Currently, the `replaceReducer` function assumes that the reducer state stays in the same shape. This does not make sense, as the purpose of `replaceReducer` is to change the store state.

However, typescript types are static, and replaceReducer changes the type at run-time.

To solve this issue, we either need to cast the existing store to the new store type, or transform the store.

As it turns out, a simple solution is to modify the signature for `replaceReducer` to return the existing store, and this gives us the ability to typecast to the new reducer type in a graceful way.

With this change, the `combineReducers` test passes when converted to typescript with minimal changes.

This *may* also fix #3482

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -669,9 +669,3 @@ export function compose<R>(
 ): (...args: any[]) => R
 
 export function compose<R>(...funcs: Function[]): (...args: any[]) => R
-
-export const __DO_NOT_USE__ActionTypes: {
-  INIT: string
-  REPLACE: string
-  PROBE_UNKNOWN_ACTION: () => string
-}
--- a/index.d.ts
+++ b/index.d.ts
@@ -326,7 +326,9 @@ export interface Store<S = any, A extends Action = AnyAction> {
    *
    * @param nextReducer The reducer for the store to use instead.
    */
-  replaceReducer(nextReducer: Reducer<S, A>): void
+  replaceReducer<NewState = S, NewActions extends A = A>(
+    nextReducer: Reducer<NewState, NewActions>
+  ): Store<NewState, NewActions>
 
   /**
    * Interoperability point for observable/reactive libraries.
@@ -667,3 +669,9 @@ export function compose<R>(
 ): (...args: any[]) => R
 
 export function compose<R>(...funcs: Function[]): (...args: any[]) => R
+
+export const __DO_NOT_USE__ActionTypes: {
+  INIT: string
+  REPLACE: string
+  PROBE_UNKNOWN_ACTION: () => string
+}
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -224,7 +224,7 @@ export default function createStore(reducer, preloadedState, enhancer) {
    * implement a hot reloading mechanism for Redux.
    *
    * @param {Function} nextReducer The reducer for the store to use instead.
-   * @returns {void}
+   * @returns {Store} The same store instance with a new reducer in place.
    */
   function replaceReducer(nextReducer) {
     if (typeof nextReducer !== 'function') {
--- a/src/createStore.js
+++ b/src/createStore.js
@@ -238,6 +238,7 @@ export default function createStore(reducer, preloadedState, enhancer) {
     // will receive the previous state. This effectively populates
     // the new state tree with any relevant data from the old one.
     dispatch({ type: ActionTypes.REPLACE })
+    return store
   }
 
   /**
@@ -284,11 +285,12 @@ export default function createStore(reducer, preloadedState, enhancer) {
   // the initial state tree.
   dispatch({ type: ActionTypes.INIT })
 
-  return {
+  const store = {
     dispatch,
     subscribe,
     getState,
     replaceReducer,
     [$$observable]: observable
   }
+  return store
 }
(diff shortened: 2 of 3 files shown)


timdorr:
We should probably add some tests against those new types.

cellog:
> We should probably add some tests against those new types.

yes, would that be in typescript.spec.js?

cellog:
@timdorr pushed a test, let me know if you think more is needed

timdorr:
That's a good enough start for me. Thanks!

timdorr:
Oh crap, I missed that this was _supposed_ to be against master. Welp, we can wait until the full TS conversion is done.

cellog:
No worries, I only did it against master because that was one of 2 options, and this one is just fine too

PR #3508, opened 2019-08-16
convert combineReducers test to typescript

cellog:
This PR converts the `combineReducers` test to typescript. The changes are mostly typing of parameters, but also requires 2 major additions to the typings. The first is export of `__DO_NOT_USE__ActionTypes` and the second is a change to the definition of `replaceReducers` which also requires a small change to `createStore`

The change to `createStore` is duplicated in PR #3507, which is against master, whereas this PR is against `ts-conversion`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -669,3 +669,9 @@ export function compose<R>(
 ): (...args: any[]) => R
 
 export function compose<R>(...funcs: Function[]): (...args: any[]) => R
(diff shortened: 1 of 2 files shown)


timdorr:
I'm fine with adding the __DO_NOT_USE to the types temporarily. That'll just go in the src files whenever we move the typedefs inline.

PR #3514, opened 2019-08-16
fix typescript config for rollup

cellog:
This allows files other than `index.ts` to be in typescript

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@ dist
 lib
 es
 coverage
+# used by rollup-plugin-typescript2
+.rpt2_cache/
 
 website/translated_docs
 website/build/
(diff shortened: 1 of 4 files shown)


nickmccurdy:
As far as I'm aware, we don't need this since the existing Rollup config in `ts-conversion` already compiles TypeScript to JavaScript. Types can be checked with `npm install`, `npm publish`, or `npm run check-types`.

cellog:
as much as I would like to not need this, rollup does not recognize `import 'blah'` if `blah` is in `blah.ts` (feel free to prove me wrong, I would much rather not have to do this)

nickmccurdy:
I fixed this by using `rollup-plugin-node-resolve`: https://github.com/reduxjs/redux/pull/3517

cellog:
closing as obsolete, thanks @nickmccurdy !

cellog:
re-opening, the fix doesn't quite work as the ES and CommonJS builds don't use `nodeResolve`

cellog:
bye again

PR #3517, opened 2019-08-17
TypeScript: Resolve TS extensions with node-resolve plugin

nickmccurdy:


nickmccurdy:
@cellog Please take a look, this allows files other than `index.ts` to use the `.ts` extension, without using the TypeScript plugin.

cellog:
this is great! I'll close mine

cellog:
unfortunately, this does not work. the `CommonJS` and `ES` builds do not use `nodeResolve`, so I am going to re-open the other PR that uses the typescript plugin

nickmccurdy:
I added it to all confirmations, so it should work. What’s the use case that’s failing?

cellog:
> I added it to all confirmations, so it should work. What’s the use case that’s failing?

The use case where I hand-copied to my own PR without adding it to the first 2 🤣

I'll sit down over there...

nickmccurdy:
Ah ok, so this is ready to merge?

cellog:
from my seat, this works as advertised, yes

nickmccurdy:
@timdorr Would you mind merging? Thanks!

PR #3518, opened 2019-08-18
TypeScript: Fix linter configuration for test files

nickmccurdy:


nickmccurdy:
@cellog Fixes CI failure after merging #3515 and #3516

.eslintrc.js
```diff
@@ -34,9 +34,11 @@ module.exports = {
 
   overrides: [
     {
-      files: 'test/**/*.js',
-      env: {
-        jest: true
```
nickmccurdy: I removed this because it's no longer necessary, as all the test files are now in TypeScript, and TypeScript will make sure the Jest globals are used properly

.eslintrc.js
```diff
@@ -34,9 +34,11 @@ module.exports = {
...
+      files: 'test/typescript/*.ts',
+      rules: {
+        'no-unused-expressions': 'off',
+        '@typescript-eslint/no-namespace': 'off',
+        '@typescript-eslint/no-unused-vars': 'off'
```
nickmccurdy: These rules aren't useful for the TypeScript tests because the code doesn't actually run, it's just meant to be typechecked

package.json
```diff
@@ -38,7 +38,7 @@
     "clean": "rimraf lib dist es coverage",
     "format": "prettier --write \"{src,test}/**/*.{js,ts}\" index.d.ts \"**/*.md\"",
     "format:check": "prettier --list-different \"{src,test}/**/*.{js,ts}\" index.d.ts \"**/*.md\"",
-    "lint": "eslint src test",
+    "lint": "eslint --ext js,ts src test",
```
nickmccurdy: Needed to lint TypeScript files

.eslintrc.js
timdorr: I might recommend putting a .eslintrc.js file in that path, as it becomes slightly easier to see that those files have overrides applied. Much like we already have a tsconfig.json in that path.

nickmccurdy: Fixed

test/typescript/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  rules: {
+    'no-unused-expressions': 'off',
+    '@typescript-eslint/no-namespace': 'off',
+    '@typescript-eslint/no-unused-vars': 'off'
+  }
+}
```
nickmccurdy: These rules aren't useful for the TypeScript tests because the code doesn't actually run, it's just meant to be typechecked

package.json
cellog: would you mind adding this to the examples:lint command also?

nickmccurdy: Fixed. I didn't realize there was another ESLint command. Thanks for catching it.

PR #3519, opened 2019-08-19
fix typescript config for rollup

cellog:
It turns out that even with the `nodeResolve` fixes, we still need the `typescript` plugin to build with rollup. This PR fixes that

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,6 @@ dist
 lib
 es
 coverage
-# used by rollup-plugin-typescript2
-.rpt2_cache/
 
 website/translated_docs
 website/build/
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@ dist
 lib
 es
 coverage
+# used by rollup-plugin-typescript2
+.rpt2_cache/
 
 website/translated_docs
 website/build/
--- a/package-lock.json
+++ b/package-lock.json
@@ -3378,6 +3378,17 @@
         "map-cache": "^0.2.2"
       }
     },
+    "fs-extra": {
+      "version": "8.1.0",
+      "resolved": "https://registry.npmjs.org/fs-extra/-/fs-extra-8.1.0.tgz",
+      "integrity": "sha512-yhlQgA6mnOJUKOsRUFsgJdQCvkKhcz8tlZG5HBQfReYZy46OwLcY+Zia0mtdHsOo9y/hP+CxMN0TU9QxoOtG4g==",
+      "dev": true,
+      "requires": {
+        "graceful-fs": "^4.2.0",
+        "jsonfile": "^4.0.0",
+        "universalify": "^0.1.0"
+      }
+    },
     "fs-readdir-recursive": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/fs-readdir-recursive/-/fs-readdir-recursive-1.1.0.tgz",
@@ -5144,6 +5155,15 @@
         }
       }
     },
+    "jsonfile": {
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/jsonfile/-/jsonfile-4.0.0.tgz",
+      "integrity": "sha1-h3Gq4HmbZAdrdmQPygWPnBDjPss=",
+      "dev": true,
+      "requires": {
+        "graceful-fs": "^4.1.6"
+      }
+    },
     "jsprim": {
       "version": "1.4.1",
       "resolved": "https://registry.npmjs.org/jsprim/-/jsprim-1.4.1.tgz",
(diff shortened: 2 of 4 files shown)


nickmccurdy:
Is this needed to fix the Babel type issue you mentioned? If so, I think we should go with your original approach and remove both nodeResolve and Babel (or at least the Babel Typescript preset). Otherwise we're going to have two tools compiling TypeScript syntax, which doesn't feel right to me.

cellog:
We need babel for sure. It just removes all typescript syntax and converts to Javascript. We use `tsc` independently to verify types in the build step. I suppose there must be some other way to do this that is better?

To answer your question directly, when I ran the branch I'm about to draft PR, it failed on the first TS syntax. Maybe you can grab that branch and play around with it to find the best config? You seem more versed in the options than I am, it would be good to get this just right.

cellog:
(#3520 is the draft PR @nickmccurdy)

nickmccurdy:
The TypeScript compiler (`tsc`) can generate JS, it's not just a typechecker. If we disable `noEmit`, the TypeScript compiler will emit JS, and we can either remove Babel or just remove the TypeScript Babel preset if we need to use JS features not supported in TS (unlikely, even JSX works in TS).

Thanks, I'll take a look at #3520 and let you know if I have better ideas. Sorry if my previous config PRs caused confusion, I'm less experienced with the Babel TypeScript preset so I'm not as aware of its limitations compared to the TypeScript compiler.

nickmccurdy:
I see what you mean, removing the TypeScript Rollup plugin causes unexpected token errors, but I'm not sure why (note that you have to disable JSX in tsconfig to this this first, or TS will think that the `<>` type assertion syntax is JSX). Meanwhile giving the extension list to Babel causes a circular dependency for some reason.

nickmccurdy:
I managed to fix the syntax error, but there's a circular dependency in #3520. If you want to use the TypeScript compiler, you cannot import the declaration file from src/index.ts, because it has the same path, which will cause the index file to import itself and crash with a stack overflow. It seems like the best way to fix this would be to move `index.d.ts` to `src/types.ts` and import that in `src/index.ts` without importing `src/index.ts`, `.`, or `..` in `src/types.ts`.

nickmccurdy:
Also I'm still confused about how the Babel issue was solved. Is the TypeScript compiler needed for this or can we remove it in favor of Babel?

.gitignore
```diff
@@ -5,6 +5,8 @@ dist
 lib
 es
 coverage
+# used by rollup-plugin-typescript2
+.rpt2_cache/
```
timdorr: I don't know why they do this by default. `node_modules/.cache` is the standard that a few other tools use now. 

Can we add `cacheRoot: path.resolve('./node_modules/.cache')` to the plugin options? Then we don't need ignore it specifically.

cellog: nice. added and pushed

nickmccurdy: Sent a PR upstream if anyone is interested: https://github.com/ezolenko/rollup-plugin-typescript2/pull/167

cellog:
ok, @nickmccurdy @timdorr I'm pretty sure this is the simplest working setup.

I tried changing whether babel had its extensions, and removing the nodeResolve. This fixed nothing. We need `rollup-plugin-typescript2`.

I think this is the best setup for now. Perhaps we'll find something better later

nickmccurdy:
Thanks for your help, I know this isn't easy. In the case that we need TypeScript, I'd rather remove Babel, as we don't seem to be using any part of it that TypeScript itself can't handle. Would you like me to make a PR trying that? We could merge this for now, assuming it works with your other PR.

cellog:
not at all, I enjoy this stuff :)

I prefer to keep both, as babel has much better support for transpiling features down. typescript is compiling to the latest javascript, and then babel refines it for us. I also imagine that in the future, support for ts will improve, and we can revisit it then.

The final call is @timdorr in this, my opinions are only opinions

cellog:
OK. I finally have some clarity on this.

The configuration will go through 2 stages.

1. While we still have an `index.d.ts`, the config in this PR will be used.
2. Once all the types are moved out, we can use a config that is more standard, and remove the typescript plugin.

So, this PR should be considered a step-along-the-path PR, but an essential one.

nickmccurdy:
Interesting, so our dependency on the TypeScript compiler is only because of how we're migrating the `index.d.ts` file?

cellog:
> Interesting, so our dependency on the TypeScript compiler is only because of how we're migrating the `index.d.ts` file? In that case I'm ok with merging this, but I'd prefer to switch to either only Babel or only TypeScript once it's fully migrated.

This is explicitly the plan I'd like to take.

My motivation is having human-readable diffs along the way, and a working config as we go. If this seems too git-focused, I'm OK doing everything-at-once, but it should make it easier to roll back if I/we make a mistake along the way. At least that's my theory :P

cellog:
so... anything need changing in this PR?

cellog:
no worries, thanks for taking a look!

PR #3520, opened 2019-08-19
Convert all the things to Typescript

cellog:
This is an incremental step. It converts the existing source to use Typescript, but using `index.d.ts` as the source for all types. I did it this way because the first branch I made had such extensive changes to the files, it assumed they were entirely new, and we lost the granular diffs this PR provides.

A follow-up PR would then begin extracting the types from `index.d.ts` gradually, until done, and then the last PR would be the configuration changes to `tsconfig.ts` in both `/` and `tests/typescript`. In `/` this will be turning on `.d.ts` generation, and in `tests/typescript` it will be changing the base to point at `src` instead of `index.d.ts`

~~This PR needs to be rebased after #3518 and #3519 are merged~~*EDIT: good to go!*


cellog:
ok @timdorr this is ready for review. If it is too complex, I'm happy to slice and dice to bring each file individually into typescript-land

cellog:
@timdorr ping again, wondering how you'd like to proceed?

timdorr:
I think we can do this in a big-bang. I'm also OK with there being broken tests in the ts-conversion branch temporarily. As long as it doesn't end up on master in a broken state, we're all good.

cellog:
> I think we can do this in a big-bang. I'm also OK with there being broken tests in the ts-conversion branch temporarily. As long as it doesn't end up on master in a broken state, we're all good.

This PR has some issues - I've learned a bit about TS, so I'm splitting them out into separate ones

timdorr:
To add some clarity: We can either remove workarounds for broken tests and leave them temporarily broken, or just annotate that something is a workaround that is to be replaced later (maybe as a `// TODO` comment with an eslint rule to fail those until removed).

Again, broken is fine while we're in-progress on this stuff.

cellog:
> To add some clarity: We can either remove workarounds for broken tests and leave them temporarily broken, or just annotate that something is a workaround that is to be replaced later (maybe as a `// TODO` comment with an eslint rule to fail those until removed).
> 
> Again, broken is fine while we're in-progress on this stuff.

I understand. However, because it is so difficult to debug broken types, I don't think it is as useful to break a whole bunch then try to fix them. I'm confident the work I'm doing now will be both correct and easier to evaluate.

cellog:
closing in favor of the smaller PRs

timdorr:
OK, I'm good with that approach too!

PR #3525, opened 2019-08-25
add strict action creator typing as an option

### LINKED ISSUE #3455: [bug: typescript] ActionCreator lose it's arguments type

https://github.com/reduxjs/redux/blob/beb1fc29ca6ebe45226caa3a064476072cd9ed26/test/typescript/actionCreators.ts#L13-L18

We declare the `addTodo` function that accepts one string argument `text`,
but we could pass any argument to it.

`const addTodoAction: AddTodoAction = addTodo(false, 0, null);`
👆👆👆 //no ts error!

There may be a problem `(...args: any[])`
https://github.com/reduxjs/redux/blob/c676a25525bd7e4f2bf407cf1a135f355a2c1d79/index.d.ts#L440-L442

Raise a type error

cellog:
This fixes #3455 

Basically, it modifies `ActionCreator` to accept a new, optional second template parameter. This parameter is a tuple, whose types will define the required parameters for the action creator.

Note that there is no support for optional parameters. Additionally, this will enforce on the calling side only, not the declaration side, so if the action creator does not define all of the required parameters, it will not complain. It will complain if any specified parameters's types don't match.

### DIFF
--- a/test/typescript/actionCreators.ts
+++ b/test/typescript/actionCreators.ts
@@ -10,16 +10,61 @@ interface AddTodoAction extends Action {
   text: string
 }
 
-const addTodo: ActionCreator<AddTodoAction, true, string> = (text: string) => ({
+// test strict typing of action creators
+// no arguments
+const addTodoNoArgs: ActionCreator<AddTodoAction, []> = () => ({
+  type: 'ADD_TODO',
+  text: 'something'
+})
+
+const AddTodoNoneArgs: AddTodoAction = addTodoNoArgs()
+// typings:expect-error
+addTodoNoArgs('test')
+
+// 1 argument
+const addTodo: ActionCreator<AddTodoAction, [string]> = (text: string) => ({
   type: 'ADD_TODO',
   text
 })
 
 const addTodoAction: AddTodoAction = addTodo('test')
 // typings:expect-error
-const oops: AddTodoAction = addTodo(false)
+addTodo(false)
 // typings:expect-error
-const oops2: AddTodoAction = addTodo('hi', 'too many arguments')
+addTodo('hi', 'too many arguments')
+
+// many arguments
+const addTodoManyArgs: ActionCreator<
+  AddTodoAction,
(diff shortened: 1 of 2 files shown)


test/typescript/actionCreators.ts
```diff
@@ -10,12 +10,73 @@ interface AddTodoAction extends Action {
   text: string
 }
 
-const addTodo: ActionCreator<AddTodoAction> = (text: string) => ({
+// test strict typing of action creators
+// no arguments
+const addTodoNoArgs: ActionCreator<AddTodoAction, []> = () => ({
```
cellog: an example of declaring an action creator that should be called with no parameters

test/typescript/actionCreators.ts
```diff
@@ -10,12 +10,73 @@ interface AddTodoAction extends Action {
...
+  type: 'ADD_TODO',
+  text: 'something'
+})
+
+const AddTodoNoneArgs: AddTodoAction = addTodoNoArgs()
+// typings:expect-error
+addTodoNoArgs('test')
+
+// 1 argument
+const addTodo: ActionCreator<AddTodoAction, [string]> = (text: string) => ({
```
cellog: an example action creator that takes 1 parameter

test/typescript/actionCreators.ts
```diff
@@ -10,12 +10,73 @@ interface AddTodoAction extends Action {
...
 })
 
 const addTodoAction: AddTodoAction = addTodo('test')
+// typings:expect-error
+addTodo(false)
+// typings:expect-error
+addTodo('hi', 'too many arguments')
+
+// many arguments
+const addTodoManyArgs: ActionCreator<
```
cellog: and an example accepting many arguments

timdorr:
I just got rid of the index.d.ts file, so this will need to be rebuilt.

cellog:
I am unconvinced that this solution is the right one. I'll comment on the original issue and get some consensus before I close/delete branch

timdorr:
Closing this out as it's quite old now.

PR #3528, opened 2019-08-29
Use new default cache root

nickmccurdy:
Fixes https://github.com/reduxjs/redux/pull/3519#discussion_r315841722 with an upstream patch instead of custom config


timdorr:
OSS rocks! 🤘

Issue #3500, opened 2019-08-10
Port to TypeScript


nickmccurdy:
**Do you want to request a _feature_ or report a _bug_?**

No

**What is the current behavior?**

Redux publishes a manually created index.d.ts file. This also affects reselect, redux-devtools, and redux-thunk.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar.**

Explanation of why this isn't recommended: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

To summarize, type declarations should either be generated from TypeScript source code or shared in DefinitelyTyped, which allows contributors to edit the types without having to edit the original package. This can also help alleviate maintenance and versioning issues regarding Redux users that write their code in TypeScript. Many issues about TypeScript compatibility could be moved to the dedicated DefinitelyTyped community if the types were separated, and the types could have breaking changes released without interfering with non-TypeScript users.

**What is the expected behavior?**

Redux either removes its type declaration (should be a breaking change) and moves them to DefinitelyTyped (recommended) or is ported to TypeScript (could be difficult and requires additional maintenance, but makes the types more accurate).

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**

3.4.0+ (all versions that publish index.d.ts)

jamestharpe:
I'd like to see redux ported to TypeScript because TypeScript is the only way to ensure types are always in sync.

Types that are always in sync prevent issues from being reported, improve the overall developer experience, and increase the overall potential for adoption of the library.

The general benefits of types are that they improve the quality and productivity of the developer experience, even for non TypeScript developers. Types help catch bugs, enable better tooling, and enhance understanding. 

Types provide these benefits most effectively when they perfectly match the codebase. Leaving types to a separate community all but ensures a lag between library changes and the related type definition changes, making it an overall less effective approach to providing types and therefore lessening the overall value of the library.

Regarding effort, because TypeScript is a superset of JavaScript the code base can be converted gradually. Gradual conversion means ongoing efforts are only minimally disrupted. Converting to TypeScript can also be done as a nonbreaking change, meaning it is also less effort for developers to upgrade to.

Finally, TypeScript it's self has a vibrant community. Conversion to TypeScript may garner additional contributions to help with the effort.

markerikson:
Relevant Twitter discussion thread:

https://twitter.com/acemarke/status/1160243663370371073

[...]

PR #3531, opened 2019-08-29
convert combineReducers to typescript

cellog:
This converts `combineReducers` to typescript. It supplants #3520 and is a step on the path to #3500


timdorr:
Dammit JS, you is stoopid.

PR #3535, opened 2019-08-29
export all types from index.ts

cellog:
This exports all of the types, a very important step to generating `index.d.ts`. It supplants #3520 and is a step on the path to #3500


cellog:
ah, need the config too. Closing for now

PR #3536, opened 2019-08-29
[Meta PR] Convert to TypeScript

### LINKED ISSUE #3500: Port to TypeScript

timdorr:
This is just a PR tracking our ts-conversion branch so we can have a quick overview of the status of things, commentable code, and a big green merge button to finish things up.

Closes #3500


markerikson:
Side request:

If we're gonna do a v5.0 anyway, can we consider tweaking the wording of this error to drop the phrase "async actions"?  I find that really ambiguous and dislike the term:

https://github.com/reduxjs/redux/blob/9c9a4d2a1c62c9dbddcbb05488f8bd77d24c81de/src/createStore.js#L185-L190

Related Twitter poll on the meaning of the phrase:

https://twitter.com/acemarke/status/1165674966060109825

timdorr:
Yeah, definitely. Maybe `Actions must be plain objects. Use custom middleware for actions that are functions or have side effects.`

markerikson:
Bikeshedding:

> Actions must be plain objects. If you need enable passing other kinds of values to dispatch, add middleware to your store setup ( https://redux.js.org/recipes/configuring-your-store ).

markerikson:
@timdorr : one other question.  Given the sequence of changes for this meta-PR, would it make sense to actually _merge_ it into `master` vs doing a squash?  Otherwise, I'm afraid we'd lose the history, because Git would likely see everything as being rewritten completely.

MichaelDeBoey:
@markerikson @timdorr Would also give credit to the everyone instead of only 1 person when squashing it.

cellog:
> @timdorr : one other question. Given the sequence of changes for this meta-PR, would it make sense to actually _merge_ it into `master` vs doing a squash? Otherwise, I'm afraid we'd lose the history, because Git would likely see everything as being rewritten completely.

had this exact thing happen on a series of major PRs I worked on in another repo, and they all went to the person who merged the PR for git blame/history. Not a huge deal, but it makes it harder for the mob with pitchforks to know who to pursue

timdorr:
Yes, I'll probably do a rebase and merge specifically, so the commits don't appear as a separate tree in the history.

src/applyMiddleware.ts
```diff
@@ -0,0 +1,79 @@
...
+  middleware2: Middleware<Ext2, S, any>,
+  middleware3: Middleware<Ext3, S, any>
+): StoreEnhancer<{ dispatch: Ext1 & Ext2 & Ext3 }>
+export default function applyMiddleware<Ext1, Ext2, Ext3, Ext4, S>(
+  middleware1: Middleware<Ext1, S, any>,
+  middleware2: Middleware<Ext2, S, any>,
+  middleware3: Middleware<Ext3, S, any>,
+  middleware4: Middleware<Ext4, S, any>
+): StoreEnhancer<{ dispatch: Ext1 & Ext2 & Ext3 & Ext4 }>
+export default function applyMiddleware<Ext1, Ext2, Ext3, Ext4, Ext5, S>(
```
ematipico: Is this the only way in typescript to tell that that is possible to pass an N number of middlewares?

jedmao: Unfortunately, yes, because in this case it's not an array with the same type throughout like a `string[]` or even a `Foo[]`. Middlewares can be a variety of many different types. It's really the `compose` function that's not type friendly.

That said, if you're willing to entertain a breaking change, we can fix this by doing something like [this TS playground](https://www.typescriptlang.org/play/#code/GYVwdgxgLglg9mABBAFgQxmAPAFQHwAUwccAXIjgJSIDeAUIoqJLAogEZoBOB19jjLgFMoILkmJwGiAL7TOXAHSoMSALyIsAVUIBHEAA9yW6mrzJ0mAvoOI0AZwqIAZIhPTho8R2505dCAR7KERhexAAGxCNFSsaO3IAcgBBRNlKZUswAnj2JIAhNJkM2Oz4iCSAYSLKXjowyKhFNEQAelbEAHkAaXqhcKjFdjaOnr6BpogRrt6GwYATaYBRACUVzpW6IA):

```ts
function chain<T>(foo: T) {
  function bar() {
    return foo
  }
  bar.chain = <U>(qux: U) => chain(qux as T & U)
  return bar
}
```
This way, you can apply multiple middlewares like so:

```ts
const result = chain({ a: 'A' }).chain({ b: 'B' }).chain({ c: 'C' })()
result.a // OK
result.b // OK
result.c // OK
result.d // ERROR
```

jedmao: You might also be able to pull off something [like this](https://www.typescriptlang.org/play/#code/JYOwLgpgTgZghgYwgAgCLAM4Ac5gQCwB4AVAPmQG8AoZZAE0xz3wC5kAKRMYAexDeIBKZAF5yABSg8AtpgiEAbj2B1SVAL5UqMAK4gE3Pshg8eJUuwB012XToAbCAHc4UCBgEBtALqC26bFwCc0oaZDcwHSgQSnpGINYOLl5+ZCFRCSlZDAhLNwweewUIdmFNTSoEPgwwZBqeN1FjU3YKZAAjVzYAcgAhAEEAJW7kdQAaWM6ALx6BgC0R9UEterdLBkDmVo6u5D6hxeEAeiPkAHkAaSpV3I2mAm3p2f6F0ePTy+uwBtv4rbaAI46AAePQAigBVAAah2QJ2QAFFBoMzoMqEA):

```ts
interface Dispatch<T> {
  dispatch: (action: T) => Promise<void>
}

function foo<T>(...middlewares: T[]): Dispatch<T> {
  return { dispatch: (action: T) => Promise.resolve() }
}

const store = foo({ bar: 'BAR' }, { baz: 'BAZ' })

store.dispatch({ bar: 'BAR' }) // OK
store.dispatch({ baz: 'BAZ' }) // OK
store.dispatch({ qux: 'QUX' }) // ERROR
```

jedmao: Otherwise, you're going to get into some pretty hairy generics and conditional types like this:

```ts
export default function applyMiddleware<M extends Middleware>(
  ...middlewares: M[]
): M extends Middleware<infer E, infer S, infer D>
  ? StoreEnhancer<E, S> & D
  : never {
  return ((createStore: M extends Middleware<infer E, infer S>
    ? StoreEnhancerStoreCreator<E, S>
    : never) => <
    A extends Action = AnyAction
  >(
    reducer: M extends Middleware<never, infer S> ? Reducer<S, A> : never,
    ...args: any[]
  ) => {
```

jedmao: I think it's easiest (for now) if you just cast the inner return type `as any` so at least the function signature isn't so messy.

```ts
export default function applyMiddleware<T extends Middleware>(
  ...middlewares: T[]
): T extends Middleware<infer E, infer S, infer D>
  ? StoreEnhancer<E, S> & D
  : never {
  return ((createStore: StoreEnhancerStoreCreator) => (
    reducer: any,
    ...args: any[]
  ) => {/* guts */}) as any
}
```

markerikson: Doesn't TS now have some kind of a "mapped types" capability or something like that? I know @nickmccurdy was talking about maybe using that to simplify Reselect's types.

jedmao: Not sure if [this is what](https://mariusschulz.com/blog/mapped-type-modifiers-in-typescript) you are talking about, but I fail to see how that would give us anything here.

markerikson: I'm probably not using the right concept name.  I think Nick was referring to the improved generic type inference in 3.4:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html

cellog: check this out: https://github.com/pirix-gh/medium/blob/master/types-curry-ramda/src/bonus.ts

I think @markerikson may be on to something, I found this 3 links deep from the issue referenced in the release notes

jedmao: Nice find! Of particular interest and relevancy is the [`compose`](https://github.com/pirix-gh/medium/blob/294725cd723b1a66cfb8b441c862a0d2d5bd184d/types-curry-ramda/src/bonus.ts#L228-L229) function.

cellog: it's GPL, so I am trying to learn from it and write new code from memory atm

src/types/middleware.ts
```diff
@@ -0,0 +1,31 @@
...
+ * logging actions, performing side effects like routing, or turning an
+ * asynchronous API call into a series of synchronous actions.
+ *
+ * @template DispatchExt Extra Dispatch signature added by this middleware.
+ * @template S The type of the state supported by this middleware.
+ * @template D The type of Dispatch of the store where this middleware is
+ *   installed.
+ */
+export interface Middleware<
+  _DispatchExt = {}, // TODO: remove unused component
```
jedmao: If you guys are going to break versions and go 5.0, now would perhaps be a good time to remove this line?

cellog: it's actively used in the definition of `applyMiddleware`, so we need to fix that first, I suspect

timdorr: Yep, we're going to get the existing types working with minor tweaks first (this PR) and then make improvements or major breaking changes in subsequent PRs.

src/applyMiddleware.ts
jedmao: @cellog see #3563

jedmao: @cellog see #3563

timdorr:
I'm going to hold off #3563 until after this, since it's an enhancement to the types. I think we have everything in place now. Let's see what the CI says...

jedmao:
CI is happy, but there's still an issue with type duplication and resolution that I fixed in https://github.com/reduxjs/redux/pull/3564

timdorr:
OK, one last thing I'd like to do is flatten the types to a single bundle. But I don't think that's critical to move on to the next stage: enhancement!

Thanks to everyone for your help thus far!

phryneas:
> OK, one last thing I'd like to do is flatten the types to a single bundle. But I don't think that's critical to move on to the next stage: enhancement!
> 
> Thanks to everyone for your help thus far!

Very cool. :fireworks: 

One thing though:

While I see you are already aware of the "single bundle", I just wanted to chime in on how important that can be - please don't do a release before having the single type bundle:

@babel/preset-typescript has severe problems with type re-exports ( https://github.com/babel/babel/issues/8361 or https://github.com/babel/babel-loader/issues/603 among dozens of others) that sometimes make libraries with multiple typings files & re-exports almost impossible to use with a webpack+babel-typescript setup.
If you are looking for solutions, there are some mentioned in https://github.com/palmerhq/tsdx/issues/80, but I haven't found anything I am happy with yet - if you find something better, please let me know :)

jedmao:
[`rollup-plugin-dts`](https://www.npmjs.com/package/rollup-plugin-dts) looks like a great solution!

PR #3537, opened 2019-08-29
export types, prepare TS config for migrating files

cellog:
This is the last of the first step for #3500 

woot!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,6 @@
-import {
+// types
+// store
+export {
   Dispatch,
   Unsubscribe,
   Observable,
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -1,9 +1,31 @@
+import {
```
timdorr: Can we just use `export Foo from 'foo'` syntax here? Or is there some sort of optimization thing I'm forgetting about?

cellog: yes, I re-organized to put type exports at the top, take a peek. Not sure if we can continue to use `export * from...` once the types are in the files, we shall see

PR #3540, opened 2019-08-30
extract store types from index.d.ts

cellog:
This extracts the common types that will be shared between `createStore` and `applyMiddleware` into a separate file


timdorr:
Can we create a `src/types` path and put it in there?

cellog:
done

timdorr:
OK, just renamed it to be a bit less redundant :)

PR #3541, opened 2019-08-30
extract action types from index.d.ts

cellog:
Extract all things action types from `index.d.ts`

Note: this PR also fixes an oversight in the extraction of store-related types, which is the declaration of observable in `types/store.ts`

### DIFF
--- a/src/types/actions.ts
+++ b/src/types/actions.ts
@@ -1,0 +1,61 @@
+/**
+ * An *action* is a plain object that represents an intention to change the
+ * state. Actions are the only way to get data into the store. Any data,
+ * whether from UI events, network callbacks, or other sources such as
+ * WebSockets needs to eventually be dispatched as actions.
+ *
+ * Actions must have a `type` field that indicates the type of action being
+ * performed. Types can be defined as constants and imported from another
+ * module. It's better to use strings for `type` than Symbols because strings
+ * are serializable.
+ *
+ * Other than `type`, the structure of an action object is really up to you.
+ * If you're interested, check out Flux Standard Action for recommendations on
+ * how actions should be constructed.
+ *
+ * @template T the type of the action's `type` tag.
+ */
+export interface Action<T = any> {
+  type: T
+}
+
+/**
+ * An Action type which accepts any other properties.
+ * This is mainly for the use of the `Reducer` type.
+ * This is not part of `Action` itself to prevent types that extend `Action` from
+ * having an index signature.
+ */
+export interface AnyAction extends Action {
+  // Allows any extra properties to be defined in an action.
+  [extraProps: string]: any
+}
+
+/* action creators */
+
+/**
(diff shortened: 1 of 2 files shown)


src/types/actions.ts
```diff
@@ -0,0 +1,61 @@
...
+ * store instance.
+ *
+ * If an action creator needs to read the current state, perform an API call,
+ * or cause a side effect, like a routing transition, it should return an
+ * async action instead of an action.
+ *
+ * @template A Returned action type.
+ */
+export interface ActionCreator<A> {
+  (...args: any[]): A
```
zanona, line 53: Hey, @cellog 
Woudn't it be a good idea allowing a declaration for the `args` type as second generic for `ActionCreator`, so we can be 'type-safer' when calling those? i.e: `setCategory('foo')` and `setCategory(false)` would be both valid? It may not cover all cases, but at least it would be a bit safer and always have a fallback to `any` in case it's not declared? Cheers

cellog: Yes, good catch

cellog: In your PR I recommend you default to unknown[] first instead of any[] and see if that works out

zanona: Thanks! There you go:
https://github.com/cellog/redux/pull/3
Perhaps I picked the wrong branch to merge. Please let me know if I should change it to `reduxjs/redux` instead :)

The only problem with defaulting to `unknown` is:
```ts
import type {Action,ActionCreator} from 'redux';

type FooAction  = Action<'SET_FOO'> & {value:string};
type FooActionCreator = ActionCreator<FooAction>;

const setFoo:FooActionCreator = value => ({type:'SET_FOO', >>value<<});
// Type 'unknown' is not assignable to type 'string'.
```
Whereas it wouldn't happen for any as it can be assigned to `string` in this case without a problem. So I guess, assigning to `unknown`, could make it a breaking change.

cellog: yes, definitely switch it to `reduxjs/redux` unless you want it to be inaccessible to the future :)

cellog: yes, it would be a breaking change in that it would force downstream types to be declared. Thus the above would have to be `ActionCreator<FooAction, [string]>`;

Keep in mind that

```ts
const setFoo:FooActionCreator = value => ({type:'SET_FOO', value });
```

is not good practice.

```ts
const setFoo = (value: string): FooAction => ({type:'SET_FOO', value });
```

is the best way to declare an action creator. `ActionCreator` should only be used in functions that want to accept an action creator as a parameter, or in other settings that consume them.

zanona: That's, great. Thanks for the suggestion. I will certainly keep that in mind as it makes sense.

farah: @zanona that suggestion for the extra type param was very neat. Makes redux-toolkit/typescript work like magic.

PR #3543, opened 2019-08-30
extract middleware types from index.d.ts

cellog:
also extract `Dispatch` into `store.ts` where it belongs


timdorr:
I'd argue PreloadedState belongs in store.ts too.

cellog:
> I'd argue PreloadedState belongs in store.ts too.

sounds good, I'll move it

PR #3548, opened 2019-08-30
use new reducer types from types/reducers.ts, store types from types/store.ts

cellog:


timdorr:
Conflict!

cellog:
> Conflict!

rebased and fixed

PR #3552, opened 2019-09-01
fixes to bindActionCreators overloads

cellog:
this adds overloads to `bindActionCreators` and fixes the import

### DIFF
--- a/src/bindActionCreators.ts
+++ b/src/bindActionCreators.ts
@@ -35,6 +35,25 @@ function bindActionCreator<A extends AnyAction = AnyAction>(
  * function as `actionCreators`, the return value will also be a single
  * function.
  */
+export default function bindActionCreators<A, C extends ActionCreator<A>>(
+  actionCreator: C,
+  dispatch: Dispatch
+): C
+
+export default function bindActionCreators<
+  A extends ActionCreator<any>,
+  B extends ActionCreator<any>
+>(actionCreator: A, dispatch: Dispatch): B
+
+export default function bindActionCreators<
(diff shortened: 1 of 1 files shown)


src/bindActionCreators.ts
```diff
@@ -35,6 +35,25 @@ function bindActionCreator<A extends AnyAction = AnyAction>(
  * function as `actionCreators`, the return value will also be a single
  * function.
  */
+export default function bindActionCreators<A, C extends ActionCreator<A>>(
```
timdorr: Please don't use single character generic names. I haven't gone through the existing types to fix this, but I want to start enforcing it now. Preferably via a linting rule.

nickmccurdy: I thought this was the de facto standard for naming generics

cellog: I agree with @timdorr, descriptive names are better if the generic is longer than 1 line.

As for this PR, it is a copy/paste from the existing types in index.d.ts (https://github.com/reduxjs/redux/blob/master/index.d.ts#L559)

How would you feel about doing a follow-up PR that renames everything once we've got the whole thing converted to TS? Suggested steps:

1. make it work as-is, with as few changes as possible
2. merge in #3524 (this is good to go, and will be much harder to rebase after the template variable renames)
3. rename all the things

markerikson: Yep!  I definitely want us to do incremental, history-traceable, step-by-step changes throughout this whole task.  And yes, I'm on board with making types more clearly named.

PR #3558, opened 2019-09-02
add overloads to compose.ts and applyMiddleware.ts

cellog:
Because `applyMiddleware` relies upon `compose`, these two need to be committed at the same time. The comment changes in `compose.ts` are because of subtle differences between the comments in `index.d.ts` and `compose.ts`. Where there was a conflict I chose the `index.d.ts` version


kevinaskin:
Hard to understand of `overloads`, and where `index.d.ts` is. Can you explains more please?

PR #3559, opened 2019-09-02
fix import inside enhancers typescript test

cellog:
This is a subtle problem. If we import `PreloadedState` from `../../index`, and `createStore` from `redux`, we end up with different versions of the declared `$CombinedState` symbol, and so the types never match.

Also, we should be consistent about always loading the types from our aliased `redux`


timdorr:
Should we be aliasing at all? I feel like direct linking would be better.

PR #3560, opened 2019-09-02
Fix TSDoc comments

jedmao:


src/applyMiddleware.ts
```diff
@@ -17,16 +22,42 @@ import { Reducer } from './types/reducers'
...
+ * @template S The type of the state supported by a middleware.
  */
-export default function applyMiddleware(
-  ...middlewares: Middleware[]
-): StoreEnhancer {
-  return (createStore: StoreCreator) => <S, A extends AnyAction>(
-    reducer: Reducer<S, A>,
-    ...args: any[]
-  ) => {
+function applyMiddleware(): StoreEnhancer
```
jedmao: These types are mirrored from index.d.ts.

src/createStore.ts
```diff
@@ -257,11 +257,7 @@ export default function createStore<S, A extends Action, Ext, StateExt>(
       throw new Error('Expected the nextReducer to be a function.')
     }
 
-    // TODO: do this more elegantly
```
jedmao: See line 84 for the "more elegant" approach, which allows any reducer to be assigned to `currentReducer`. It's all that can be done, really.

test/applyMiddleware.spec.ts
```diff
@@ -51,16 +51,11 @@ describe('applyMiddleware', () => {
     const spy = jest.fn()
     const store = applyMiddleware(test(spy), thunk)(createStore)(reducers.todos)
 
-    // the typing for redux-thunk is super complex, so we will use an as unknown hack
```
jedmao: This is why I installed `redux-thunk` as a dev dependency. I think it makes sense to ensure its compatibility here, seeing as the tests are using thunks.

test/applyMiddleware.spec.ts
```diff
@@ -51,16 +51,11 @@ describe('applyMiddleware', () => {
     const spy = jest.fn()
     const store = applyMiddleware(test(spy), thunk)(createStore)(reducers.todos)
```
jedmao: This `store` was not being typed properly, because the `thunk` wasn't typed properly. That's now fixed in this PR.

test/applyMiddleware.spec.ts
```diff
@@ -51,16 +51,11 @@ describe('applyMiddleware', () => {
...
 
-    // the typing for redux-thunk is super complex, so we will use an as unknown hack
-    const dispatchedValue = (store.dispatch(
-      addTodoAsync('Use Redux')
-    ) as unknown) as Promise<void>
-    return dispatchedValue.then(() => {
-      expect(spy.mock.calls.length).toEqual(2)
-    })
+    await store.dispatch(addTodoAsync('Use Redux'))
```
jedmao: Now that the `store` is typed properly, this `dispatch` call has no issues. The root of the problem has been fixed.

test/createStore.spec.ts
```diff
@@ -505,6 +505,7 @@ describe('createStore', () => {
 
   it('throws if action type is missing', () => {
     const store = createStore(reducers.todos)
+    // @ts-ignore
     expect(() => store.dispatch({})).toThrow(
```
jedmao: TypeScript won't let you call `store.dispatch` with a missing action type, now that the `store` is typed properly; thus, we need to `// @ts-ignore` this line to allow the test to run w/o compilation errors.

test/createStore.spec.ts
```diff
@@ -519,10 +520,10 @@ describe('createStore', () => {
 
   it('does not throw if action type is falsy', () => {
     const store = createStore(reducers.todos)
-    expect(() => store.dispatch({ type: false })).not.toThrow()
-    expect(() => store.dispatch({ type: 0 })).not.toThrow()
+    expect(() => store.dispatch({ type: false } as any)).not.toThrow()
```
jedmao: Again, because the `store` is properly typed, TS won't compile this, so I'm just casting `as any` to allow it to compile. Pretty common to see `as any` usage in tests. In fact, I usually change my `.eslintrc` to allow `any` in test files.

test/helpers/middleware.ts
```diff
@@ -1,12 +1,15 @@
...
-  ? AnyAction
-  : T extends Function
-  ? T
-  : never
+import { rootActionTypes } from './actionTypes'
+import { Middleware } from '../..';
 
-export function thunk({ dispatch, getState }: MiddlewareAPI) {
-  return (next: Dispatch) => <T>(action: ThunkAction) =>
+export const thunk: Middleware<
```
jedmao: Unfortunately, using `ThunkMiddleware` kept returning a type of `any` for some reason, so I resorted to this redux `Middleware` type instead, which gives us a properly-typed `thunk`. Worth looking into in the `redux-thunk` library why this is happening, but outside the scope of this PR.

timdorr:
We're still in the middle of the conversion. We want to:

1. Move all the types into the source verbatim. 
2. Get things building and outputting essentially the same as before. _<- We are here_
3. Make improvements to the types.

So, this PR is going to conflict for a bit while we get finished with step 2. I'd hold off until #3536 closes, at which point we'll be ready for improvements.

jedmao:
@timdorr I've limited this PR in scope to _just_ TSDoc comments. I'll hold off for the other stuff.

PR #3561, opened 2019-09-03
fully move away from index.d.ts

cellog:
this is it! after this, typescript conversion is complete


package.json
```diff
@@ -26,13 +26,12 @@
   "main": "lib/redux.js",
   "unpkg": "dist/redux.js",
   "module": "es/redux.js",
-  "typings": "./index.d.ts",
+  "typings": "./es/index.d.ts",
```
timdorr: This should still be output in the root of the package. The types are always the same.

tsconfig.json
```diff
@@ -10,11 +10,11 @@
...
     // "checkJs": true,                       /* Report errors in .js files. */
     "jsx": "react" /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */,
-    // "declaration": true /* Generates corresponding '.d.ts' file. */,
-    // "declarationMap": true /* Generates a sourcemap for each corresponding '.d.ts' file. */,
+    "declaration": true /* Generates corresponding '.d.ts' file. */,
+    "declarationMap": true /* Generates a sourcemap for each corresponding '.d.ts' file. */,
     "sourceMap": true /* Generates corresponding '.map' file. */,
     // "outFile": "./",                       /* Concatenate and emit output to single file. */
-    // "outDir": "."                          /* Redirect output structure to the directory. */,
+    "outDir": "./dist" /* Redirect output structure to the directory. */,
```
timdorr: `dist` is for UMD builds, not what this outputs.

package.json
cellog: OK, I'll see if I can make it do this and push a new commit

cellog: OK, I changed it to output type declarations into the `types/` directory and to use them for everything.

With this change, this PR will need to be merged before #3524

package.json
```diff
@@ -43,9 +43,9 @@
     "test": "jest",
     "test:watch": "npm test -- --watch",
     "test:cov": "npm test -- --coverage",
-    "build": "rollup -c",
-    "prepare": "npm run clean && npm run check-types && npm run format:check && npm run lint && npm test",
-    "check-types": "tsc",
```
timdorr: Can we keep this with `--noEmit` on it so that we still have a type check as a part of the CI process? I know this happens by proxy of the test suite, but an earlier, more-focused test will make the problems easier to see.

timdorr:
Also, we should add the `types` path to the gitignore and remove those generated files from the repo.

package.json
cellog: `"noEmit": true` is in `tsconfig.json` which is the same thing. Do you want me to add back in the `check-types` command, even though it will basically run `tsc` twice the same way?

timdorr: I would do this:

- Set `declaration: false` in the tsconfig.
- Make `build-types` run `tsc -d`
- Make `check-types` run `tsc --noEmit` to be explicit about it.
- Run `check-types` in the `prepare` script.

Another option: Are we using typescript-eslint? That does a type check as well, so that would be sufficient for checking them.

cellog: I think I found a solution similar to this, although the declaration: false is unneeded if we do noEmit. The `--emitDeclarationOnly` flag does what we need. take a look when you're free and see if it fits the bill.

I also added `types/` to the `clean` command, as we will definitely want that

tsconfig.json
```diff
@@ -42,9 +44,9 @@
     /* Module Resolution Options */
     "moduleResolution": "node" /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */,
     "baseUrl": "./" /* Base directory to resolve non-absolute module names. */,
-    "paths": {
-      "*": ["*", "types/*"]
-    } /* A series of entries which re-map imports to lookup locations relative to the 'baseUrl'. */,
+    // "paths": {
```
cellog: this is unnecessary because redux is a top-level dep, we don't import any other types

tsconfig.json
```diff
@@ -10,15 +10,17 @@
     // "allowJs": true /* Allow javascript files to be compiled. */,
     // "checkJs": true,                       /* Report errors in .js files. */
     "jsx": "react" /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */,
-    // "declaration": true /* Generates corresponding '.d.ts' file. */,
-    // "declarationMap": true /* Generates a sourcemap for each corresponding '.d.ts' file. */,
+    "declaration": true /* Generates corresponding '.d.ts' file. */,
+    "declarationMap": true /* Generates a sourcemap for each corresponding '.d.ts' file. */,
```
cellog: question: is this useful? It pops a sourcemap into the `types/` directory but I'm not sure anyone would use it?

package.json
```diff
@@ -26,26 +26,27 @@
   "main": "lib/redux.js",
   "unpkg": "dist/redux.js",
   "module": "es/redux.js",
-  "typings": "./index.d.ts",
+  "typings": "types/index.d.ts",
```
jedmao:
```suggestion
  "types": "types/index.d.ts",
```

Just pointing out that "types" is [what you see most places and is synonymous with "typings"](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package).

tsconfig.json
timdorr: Does VS Code pick this up when I hit Go To Definition? That's what I would assume to use it for.

I'm inclined to thing they're not useful, though.

package.json
timdorr: Yes, but `types` is the primary name. Other tools are more likely to implement that than `typings` (it's easy to miss the alias).

tsconfig.json
cellog: another question: Do we still need the `.d.ts` files? Can't we just distribute `src/` now?

package.json
jedmao: I agree. `types` is the primary name; hence, my suggestion to change it to "types" above.

tsconfig.json
timdorr: Not the individual ones, just the main `index.d.ts`. Rewinding back a bit, I would expect that to be in the package root like it is now.

We already put the `src` into the package verbatim.

cellog: so is there any reason we can't do `"typings": "src/index.ts"`?

cellog: > Not the individual ones, just the main `index.d.ts`. Rewinding back a bit, I would expect that to be in the package root like it is now.

I tried to do this, it isn't possible in an automatic way, because typescript automatically ignores the directory we tell it to pop its results into, probably to avoid infinite loops.

We can just manually `cp types/index.d.ts .` but it seems unnecessary to me. I haven't seen any other examples out there that keep any autogenerated files in the repo

Also, I checked, and no one is using `"types": "index.ts"` so I'm going to assume it's a bad idea

cellog:
I think we are ready? @timdorr anything I missed?

tsconfig.json
timdorr: Ah yes, we don't need definitions because we are fully TS'ed now.

package.json
timdorr: Sorry, I was reading your suggestion as the actual source. Whoops!

cellog:
ok, just rebased this, when it's done, should be good to go

tsconfig.json
jedmao: @timdorr "Go to Definition" doesn't work like that with sourcemaps.

@cellog `"types": "index.ts"` is definitely a bad idea, as the types are supposed to point to a `.d.ts` file. All TS files should be gone when you publish to npm.

@timdorr do you mean we don't need manually-typed definitions? Correct. Still need auto-generated ones though.

timdorr: We publish `src` now. We want that in the package so the source is always available to read through. It's small and I consider it an important feature of Redux.

Whether we use that for our types or not, it's going to stay in the package.

jedmao:
@cellog there is some type duplication in this PR that I fixed in https://github.com/cellog/redux/pull/1

PR #3563, opened 2019-09-04
use ts-toolbelt for 'compose' types

jedmao:
In response to https://github.com/reduxjs/redux/pull/3536#discussion_r320416690

[ts-toolbelt](https://pirix-gh.github.io/ts-toolbelt/)

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "rollup-plugin-terser": "^5.1.1",
     "rollup-plugin-typescript2": "^0.24.0",
     "rxjs": "^6.5.3",
+    "ts-toolbelt": "^3.8.63",
     "typescript": "^3.5.3"
   },
   "npmName": "redux",
--- a/src/applyMiddleware.ts
+++ b/src/applyMiddleware.ts
@@ -72,7 +72,7 @@ export default function applyMiddleware(
       dispatch: (action, ...args) => dispatch(action, ...args)
     }
     const chain = middlewares.map(middleware => middleware(middlewareAPI))
-    dispatch = compose<typeof dispatch>(...chain)(store.dispatch)
+    dispatch = compose(...chain)(store.dispatch as typeof dispatch)
 
     return {
       ...store,
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -1,7 +1,4 @@
-type Func0<R> = () => R
-type Func1<T1, R> = (a1: T1) => R
-type Func2<T1, T2, R> = (a1: T1, a2: T2) => R
-type Func3<T1, T2, T3, R> = (a1: T1, a2: T2, a3: T3, ...args: any[]) => R
+import { F, T } from 'ts-toolbelt'
 
 /**
  * Composes single-argument functions from right to left. The rightmost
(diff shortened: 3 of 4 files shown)


src/applyMiddleware.ts
```diff
@@ -69,7 +69,7 @@ export default function applyMiddleware(
       dispatch: (action, ...args) => dispatch(action, ...args)
     }
     const chain = middlewares.map(middleware => middleware(middlewareAPI))
-    dispatch = compose<typeof dispatch>(...chain)(store.dispatch)
+    dispatch = compose(...chain)(store.dispatch as typeof dispatch)
```
jedmao: The next step is to get the chain properly typed so we don't need to do this casting.

src/compose.ts
```diff
@@ -1,97 +1,24 @@
...
-/* rest */
-export default function compose<R>(
-  f1: (b: any) => R,
-  ...funcs: Function[]
-): (...args: any[]) => R
-
-export default function compose<R>(...funcs: Function[]): (...args: any[]) => R
-
-export default function compose(...funcs: Function[]) {
+const compose = <Fns extends F.Function<any[], any>[]>(
```
jedmao: I really wanted to use [ts-toolbelt's](https://pirix-gh.github.io/ts-toolbelt/) `F.compose` here, but it was swallowing types. This way, I can be more explicit and get an accurate return type.

src/compose.ts
```diff
@@ -1,97 +1,24 @@
...
+  ...funcs: Fns
+): ((
+  ...args: Parameters<Fns[T.Tail<Fns>['length']]>
+) => F.Return<T.Head<Fns>>) => {
   if (funcs.length === 0) {
     // infer the argument type so it is usable in inference down the line
-    return <T>(arg: T) => arg
+    // TODO: should probably throw an error here instead
+    return (<T>(arg: T) => arg) as any
```
jedmao: TypeScript really doesn't like this line, because it goes against the return type. Rather than introducing a union type, I just cast it `as any`, but I think it might make more sense to just throw an error instead. Thoughts?

cellog:
brilliant!

package.json
```diff
@@ -89,6 +89,7 @@
     "rollup-plugin-terser": "^5.1.1",
     "rollup-plugin-typescript2": "^0.24.0",
     "rxjs": "^6.5.2",
+    "ts-toolbelt": "^3.8.63",
```
timdorr: This isn't a dev dep if it's used in src.

jedmao: I think you're right. I was thinking because all the types and interfaces disappear, it won't need to be a normal dependency, but if the import is still there... yeah.

jedmao: Fixed in 6a76ebb. If it's any consolation, here's the minified bundle size:

jedmao:
Just rebased against master.

timdorr:
Looks like I broke master. I'll get that fixed up soon.

markerikson:
I'd prefer to avoid adding a dependency just for one lousy function if possible.  Can the compose function/type be copy-pasted or something?

jedmao:
The compose type is way more complex than a simple copy paste. It has dependencies that have dependencies. I also foresee this toolbelt coming in handy in future PRs for improving types.

I did have a copy paste branch before and it was significant enough that I welcomed the dependency approach more.

What is your main concern about the dependency? If it has anything to do with footprint, it has no footprint.

markerikson:
I generally want to minimize increasing the number of dependencies that users have to download.  Sure, it's "just one more dependency", but that's how we ended up with 1500 deps being pulled down for a basic CRA+TS app.  As it is, it annoys me that we have forced dependencies on `loose-envify` and `symbol-observable`.  I'm okay with dev dependencies, but ideally, Redux would have _zero_ app-level dependencies.

timdorr:
The good news is it ships zero actual JS and should get fully shook out by any bundler: https://bundlephobia.com/result?p=ts-toolbelt@3.8.70

Unfortunately, compose is exactly the kind of function that TS *wasn't* built for. You basically have to build a recursive state machine inside of the types. It's really complex stuff, even to trace through with the source available.

I suppose if we want to do anything, it would be to just remove the function entirely and tell folks to get it from something like Ramda instead.

jedmao:
Do we have any issues with the Apache 2 license here? Concerning copy/paste.

timdorr:
I don't think we need to copypasta into this repo. We can incur the dep because it has no impact on bundle sizes.

markerikson:
Agreed on the bundle size thing, but this is still going to require end consumers to pull down one more extra package, right?

timdorr:
Yes, and just one. ts-toolbelt has no deps of its own.

markerikson:
Semi-related tangent: what is `loose-envify` even _used_ for, and _are_ we still using it?  For that matter, if we're doing a major, is there any way we can drop `symbol-observable`?

jedmao:
Rebased

markerikson:
Just filed #3567 to discuss trying to drop all external dependencies.

joshburgess:
I opened a new PR attempting to rewrite `compose` from scratch: https://github.com/reduxjs/redux/pull/3568

Note that I did not modify this `dispatch = compose<typeof dispatch>(...chain)(store.dispatch)` line yet. So, the build isn't passing right now. The `as typeof dispatch` assertion did not fix the type error here with my implementation of `compose`. However, as @jedmao said, this is due to the way other things in this file are typed and not the way the `compose` function itself is typed.

src/compose.ts
joshburgess: I was able to make this type check in my implementation of `compose`. However, as I suggested in comments on that PR, I think you're right that this is strange anyhow. However, I think it should simply be statically not allowed by not providing overloads for `compose` that take 0 arguments rather than allowing it and throwing a runtime error.

I would go farther and say that passing compose a single function is also strange behavior and probably shouldn't be allowed, but I can see why you all might want to keep that due to the current behavior, whereas I doubt anyone has much use for or would miss it accepting 0 args.

jedmao:
I can do a separate copy :pasta: PR when I have some more time on Monday. Have a good weekend y'all!

jedmao:
Closing because we can't come to a consensus about the 3rd-party dependency.

PR #3564, opened 2019-09-05
Defer to rollup TS plugin to build types

jedmao:
Same as https://github.com/cellog/redux/pull/1


timdorr:
I don't want to name the index file something other than `index`. That's a Node standard I would like to keep. Please revert it back to index.d.ts.

test/typescript/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "strict": true,
     "baseUrl": "../..",
     "paths": {
-      "redux": ["types/index.d.ts"]
+      "redux": ["lib/redux.d.ts"]
```
timdorr: Is this needed at all? Where are we importing from the `redux` package rather than the source code? I could see someone dealing with an out of sync build and running tests that give a false positive/negative.

jedmao: In the tests, there are a number of imports directly from `"redux"`. I see no reason these can't be changed to relative paths though.

jedmao:
`index.js` is fine, so long as there are types in `index.d.ts`.

jedmao:
I think the only way to make you happy is to do some manual renaming after the rollup build.

jedmao:
@timdorr fixed in 3600edf. You'll notice/hate that there are 3 `move-file` commands in the `npm run build` script. This is necessary to ensure the `d.ts` files are renamed according to the rollup configuration.

jedmao:
@timdorr fixed in 17fa65a. You'll notice/hate that there are 3 `move-file` commands in the `npm run build` script. This is necessary to ensure the `d.ts` files are renamed according to the rollup configuration.

test/typescript/tsconfig.json
jedmao: See 17fa65a

package.json
```diff
@@ -26,24 +26,23 @@
   "main": "lib/redux.js",
   "unpkg": "dist/redux.js",
   "module": "es/redux.js",
-  "types": "types/index.d.ts",
+  "types": "lib/redux.d.ts",
```
timdorr: No, I want to call the input file `index.ts` and have the output file be `index.d.ts`, just like it is now.

test/typescript/actionCreators.ts
```diff
@@ -4,7 +4,7 @@ import {
   Dispatch,
   bindActionCreators,
   ActionCreatorsMapObject
-} from 'redux'
+} from '../..'
```
timdorr: Thanks. This could have caused a number of problems for people working on the code!

package.json
jedmao: Do you realize that d.ts files are supposed to be paired (adjacent) to similarly named js files? I'm not understanding the logic here, as editors won't understand the relationship either if they don't have the same base name. Because the roll-up configuration is not writing an index.is file.

jedmao: I'm not sure I'm articulating the problem well enough, but if someone were to import `redux/redux` it wouldn't get any type information along with it because there is no redux.d.ts file.

jedmao:
```
lib/index.d.ts
lib/redux.js
```
☝️ see the issue?

You are telling TypeScript users that the `index.d.ts` file relates to an `index.js` file that doesn't exist. Also, if you import directly from `lib/redux` (for some reason) you're telling TypeScript to look for a `redux.d.ts` file, which also doesn't exist. This paradigm is broken in both directions. 

The only way it does work is if you `import {} from 'redux'`, because the package.json file defines the relationship. I realize this is the most common use case, but does it not bother you that the others are completely broken?

timdorr: Gotcha. I was thinking about it wrong and also forgot about the flat bundles somehow. 

Lemme check this PR out locally and see what it actually spits out.

timdorr:
OK, made a change to only put out the declarations once, as they were showing up in each of the output types. Here's what things look like now:

```
 ⇶ ls es dist lib types
dist:
redux.js  redux.min.js

es:
redux.js  redux.mjs

lib:
redux.js

types:
applyMiddleware.d.ts  bindActionCreators.d.ts  combineReducers.d.ts  compose.d.ts  createStore.d.ts  index.d.ts  types/  utils/
```

I don't know how to get the declarations to be flat like the other output bundles, or if that's even really possible. I don't think it's super-necessary, so this is looking good as-is. Thanks for following up on this @jedmao!

jedmao:
Just so you know, this fixes duplication, but it still breaks direct imports or imports that go deeper than "redux".

e.g., `import {} from 'redux/redux'` or `from 'redux/foo'` if you care at all to support those types of import statements and preserve types.

timdorr:
We haven't supported direct imports since moving to flat bundles.

PR #3565, opened 2019-09-07
Move the examples to a separate repo

timdorr:
I'd like to move the examples out of the repo. They add a _lot_ of time to the build process and generally don't change significantly. They also don't require the source copy of Redux, so it's not like they are a canary for changes we make on master. 

Moving them to their own home means it's more reasonable to accept new additions, which is something we've been hesitant to do for a long while (years!). 

I've set them up in a new repo: [reduxjs/examples](https://github.com/reduxjs/examples)

Is this a terrible idea? Anything we should do besides updating the docs and creating a new README in the `examples` path (which is temporarily gone at the moment).


timdorr:
Also of note: 

- Typical build time today: 3:30-4:10
- This PR build time: 1:10

That's a 3-3.5x faster build time.

markerikson:
Thoughts and concerns:

- The approach shown here and in the other repo appears to blow away all the accumulated Git history.  Could we at least do some `git filter-branch` work to extract the `examples` folder history as the starting point of the new repo, so that we can maintain that history?
- We have a bunch of links in the docs, and there's probably also external links:
  - Yeah, we'd need to do a bunch of fixing up for our docs.
  - External links would result in a 404.  Can we at least leave placeholder README files that point to the new location?
- What additional examples would you want to have in the future repo?

markerikson:
Out of curiosity, what about the examples is slowing down the build process?

timdorr:
There are so many of them and they all require separate `npm ci` commands.

markerikson:
But why are we even building / testing them at all as part of the main lib?  Where are they getting kicked off?

timdorr:
https://github.com/reduxjs/redux/blob/master/.travis.yml#L5-L6

markerikson:
Oh.  Well, at a minimum, we should remove those lines from Travis.  No reason to be checking those on every commit.

timdorr:
Redid the examples repo with history intact: https://github.com/reduxjs/examples

markerikson:
I'm still really not seeing a point to this.  If we're concerned about Travis build times for some reason, let's just configure Travis to not worry about the examples on every build or something.

nickmccurdy:
I tried to figure out how to conditionally build certain examples based on the changed files using [Travis' environment variables](https://docs.travis-ci.com/user/environment-variables/), but unfortunately it seems difficult to do this consistently. For example `TRAVIS_COMMIT_HASH` doesn't only show the commits from the PR which would cause false positives where examples are build when they shouldn't be, and if multiple commits are pushed outside of a pull request `TRAVIS_COMMIT` would only show the latest commit which could cause examples to break without being rerun in CI.

I think the safest way to make our builds faster is to remove the examples (like with this pull quest, but possibly replacing the existing files with links to the new repository) and move the example scripts to the other repository. Alternatively, we might be able to speed up our multiple package installations using monorepos with yarn workspaces or lerna, but I don't have experience with them.

markerikson:
We never did move these out, and tbh I'd like to keep them in here. Closing.

PR #3566, opened 2019-09-07
Improve TypeScript types

jedmao:
The [TS enhancement stage got the green light 🚦](https://github.com/reduxjs/redux/pull/3536#issuecomment-529003240), so this is a PR towards that effort. Could be some more after this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
     "prettier": "^1.18.2",
     "rimraf": "^3.0.0",
     "rollup": "^1.20.3",
+    "redux-thunk": "^2.3.0",
     "rollup-plugin-babel": "^4.3.3",
     "rollup-plugin-node-resolve": "^5.2.0",
     "rollup-plugin-replace": "^2.2.0",
--- a/src/applyMiddleware.ts
+++ b/src/applyMiddleware.ts
@@ -1,8 +1,10 @@
 import compose from './compose'
 import { Middleware, MiddlewareAPI } from './types/middleware'
-import { AnyAction } from './types/actions'
-import { StoreEnhancer, StoreCreator, Dispatch } from './types/store'
-import { Reducer } from './types/reducers'
+import {
+  Dispatch,
+  StoreEnhancer,
+  StoreEnhancerStoreCreator
+} from './types/store'
 
 /**
  * Creates a store enhancer that applies middleware to the dispatch method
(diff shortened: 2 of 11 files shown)


markerikson:
Can you clarify and document what all is being changed?

test/helpers/actionCreators.ts
```diff
@@ -1,71 +1,89 @@
...
 
-export function addTodo(text: string): AnyAction {
+export function addTodo(text: string): AddTodo {
   return { type: ADD_TODO, text }
 }
 
-export function addTodoAsync(text: string) {
-  return (dispatch: Dispatch): Promise<void> =>
-    new Promise(resolve =>
+export function addTodoAsync(text: string): ThunkAction<any, any, {}, AddTodo> {
```
jedmao: Now that redux thunk is being imported, we can leverage the `ThunkAction` type to be much more explicit about what kind of action this is; thus, get properly-typed actions throughout this file.

test/helpers/middleware.ts
```diff
@@ -1,12 +1,13 @@
...
-  : T extends Function
-  ? T
-  : never
+import { rootActionTypes } from './actionTypes'
+import { Middleware } from '../..'
 
-export function thunk({ dispatch, getState }: MiddlewareAPI) {
-  return (next: Dispatch) => <T>(action: ThunkAction) =>
+export const thunk: Middleware<
+  ThunkDispatch<any, {}, rootActionTypes>,
```
jedmao: This first generic type isn't even used by `Middleware`.

test/helpers/middleware.ts
```diff
@@ -1,12 +1,13 @@
...
+  any,
+  ThunkDispatch<any, {}, rootActionTypes>
```
jedmao: `rootActionTypes` here limits dispatching only to these action types.

test/helpers/reducers.ts
```diff
@@ -18,9 +18,8 @@ export interface Todo {
   id: number
   text: string
 }
-export type TodoAction = { type: 'ADD_TODO'; text: string } | AnyAction
 
-export function todos(state: Todo[] = [], action: TodoAction) {
+export function todos(state: Todo[] = [], action: rootActionTypes) {
```
jedmao: Now that the `action` is properly typed as `rootActionTypes`, the switch statement will narrow down the types to the appropriate one. This type of stuff is what TypeScript is all about!

package.json
```diff
@@ -82,6 +82,7 @@
     "prettier": "^1.18.2",
     "rimraf": "^3.0.0",
     "rollup": "^1.20.3",
+    "redux-thunk": "^2.3.0",
```
jedmao: If we're going to be testing thunks, it only makes sense that this dev dependency would aid in that effort.

timdorr:
Do you think we could break this up into smaller PRs? It might be easier to look through that way. Either that or at least multiple commits.

jedmao:
Unfortunately, it's harder than it seems to break this one apart, as each piece is quite interconnected. If a 150-line PR is too big, I can close it.

timdorr:
@jedmao Can you give this a poke to fix the prettier error in `test/helpers/reducers.ts`?

jedmao:
Done!

src/applyMiddleware.ts
```diff
@@ -23,40 +26,17 @@ import { Reducer } from './types/reducers'
...
-): StoreEnhancer<{ dispatch: Ext1 & Ext2 & Ext3 & Ext4 & Ext5 }>
-export default function applyMiddleware<Ext, S = any>(
-  ...middlewares: Middleware<any, S, any>[]
-): StoreEnhancer<{ dispatch: Ext }>
-export default function applyMiddleware(
-  ...middlewares: Middleware[]
-): StoreEnhancer<any> {
-  return (createStore: StoreCreator) => <S, A extends AnyAction>(
-    reducer: Reducer<S, A>,
+export default function applyMiddleware<
```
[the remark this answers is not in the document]
jedmao: @timdorr now that I look at this again, I don't think we should merge this, because it requires each middleware to be of the same type. As we discussed at length in other threads, there seems to be no way around this `Ext1, Ext2, ...` business.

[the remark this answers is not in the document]
jedmao: Restored in ebe7423

jedmao:
@timdorr any updates?

ematipico:
@timdorr @markerikson any change to get an alpha release with these types?

timdorr:
No, I haven't had a chance to really think through them.

src/applyMiddleware.ts
```diff
@@ -23,40 +26,17 @@ import { Reducer } from './types/reducers'
...
+  S = any,
+  M extends Middleware = Middleware
+>(
+  ...middlewares: M[]
+): StoreEnhancer<
+  M extends Middleware<any, any, infer D> ? { dispatch: D } : never,
+  S
+> {
+  return (createStore: StoreEnhancerStoreCreator<any>) => (
```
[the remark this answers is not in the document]
timdorr: I'm not groking why the StoreEnhancer's `Ext`ension here would be an `any`. We know what apply middleware does (it has a very specific API) and it doesn't do any actual extension of the store's properties. So, at the very least, shouldn't it be `{}`?

timdorr:
Closing this out because we never got good consensus. I would recommend re-opening with a smaller scope of work.

PR #3568, opened 2019-09-08
Rewrite `compose` utility function

joshburgess:
I saw you discussing pulling in `ts-toolbelt` to rewrite the `compose` utility in https://github.com/reduxjs/redux/pull/3563 and asking for alternative solutions on Twitter. I don't think you need to pull in a utility library to do so this, and `ts-toolbelt`'s solution seemed more complex than necessary, utilizing a bunch of custom type-level helper utilities. So, I decided to take a shot at rewriting `compose` from scratch, using no external dependencies & avoiding `any` entirely. This is what I came up with.

Important notes:
- There are no behavior changes other than a change to the generic type params used in various overloads. When explicitly providing generic params for overloads accepting multi-arg functions, you must now pass in a tuple of types for the generic instead of only the type. This is accomplished via the combination of `A extends unknown[]` & `(...args: A)`. This actually fixes a bug that was present before which (accidentally, I believe) restricted these multi-arg functions to take in arguments all of the same type, which isn't necessary. You can see the changes I'm talking about reflected in the updated test.

- I did not change the actual logic in the implementation body of the function. We could rewrite it without `reduce` in a potentially more performant way, but I wanted to keep the changes in this PR restricted to how the types work so that it would be easier to review. Note: There are some changes in the implementation, but they are just related to type signatures. Also, we could use `Function` where I'm using `Fn` and `unknown[]` where I'm using `Parameters<typeof b>` if you don't like what I did with these. It would still work the same way.

- I think what's in the PR now is an improvement and could be merged as is (after fixing the line passing `...chain` to `compose` that's currently causing the build to fail), but, while I kept the actual functionality/behavior the same in this PR, I think it's worth considering making some further changes beyond what's been done so far in order to simplify what's required to properly type this function. I'll add more info on this in a followup comment.

### DIFF
--- a/src/applyMiddleware.ts
+++ b/src/applyMiddleware.ts
@@ -72,7 +72,7 @@ export default function applyMiddleware(
       dispatch: (action, ...args) => dispatch(action, ...args)
     }
     const chain = middlewares.map(middleware => middleware(middlewareAPI))
-    dispatch = compose<typeof dispatch>(...chain)(store.dispatch)
+    dispatch = compose(...chain)(store.dispatch as typeof dispatch)
 
     return {
       ...store,
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -4,13 +4,23 @@
  */
 interface Fn extends Function {
   <A extends unknown[]>(...args: A): unknown
+  <A>(...args: A[]): unknown
 }
 
 /**
- * A type-level utility that extracts the type of a function's input params as
- * a tuple, inferring `unknown` for any input params assigned to generics
+ * A type-level utility function to
  */
-// type Params<F> = F extends (...args: infer A) => unknown ? A : never
+type Tail<Fns extends Fn[]> = Fns extends <A>(
+  a: A,
+  ...rest: infer Rest
+) => unknown
+  ? Rest
+  : never
+
+/**
+ * A type-level utility function to
+ */
+type Last<Fns extends Fn[]> = Fns[Tail<Fns>['length']]
 
 /**
  * Composes single-argument functions from right to left. The rightmost
@@ -102,6 +112,10 @@ export default function compose<
   fbc: (...b: B) => C[0],
   fab: (...args: A) => B[0]
 ): (...args: A) => F
+// generic type signature for any number of functions
+export default function compose<Fns extends Fn[]>(
+  ...funcs: Fns
+): (...args: Parameters<Last<Fns>>) => ReturnType<Fns[0]>
 // generic base case type signature and function body implementation
 export default function compose<Fns extends Fn[]>(...fns: Fns): Fn {
   const len = fns.length
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -1,7 +1,16 @@
-type Func0<R> = () => R
-type Func1<T1, R> = (a1: T1) => R
-type Func2<T1, T2, R> = (a1: T1, a2: T2) => R
-type Func3<T1, T2, T3, R> = (a1: T1, a2: T2, a3: T3, ...args: any[]) => R
+/**
+ * A generic representation of a function exposing more type information than
+ * the built-in `Function` type, allowing extraction of its input param types
+ */
+interface Fn extends Function {
+  <A extends unknown[]>(...args: A): unknown
+}
+
+/**
+ * A type-level utility that extracts the type of a function's input params as
+ * a tuple, inferring `unknown` for any input params assigned to generics
+ */
+type Params<F> = F extends (...args: infer A) => unknown ? A : never
 
 /**
  * Composes single-argument functions from right to left. The rightmost
@@ -13,90 +22,99 @@ type Func3<T1, T2, T3, R> = (a1: T1, a2: T2, a3: T3, ...args: any[]) => R
  *   to left. For example, `compose(f, g, h)` is identical to doing
  *   `(...args) => f(g(h(...args)))`.
  */
-export default function compose(): <R>(a: R) => R
+// when given no args and no generic type params, infer args as tuple
+export default function compose(): <A extends unknown[]>(...a: A) => A[0]
+// allow specifying type param of args tuple
+export default function compose<A extends unknown[]>(): (...a: A) => A[0]
+// when given a single function, just return that function
+export default function compose<A extends unknown[], B>(
+  fab: (...args: A) => B
+): (...args: A) => B
+// standard case, given 2 functions
+export default function compose<A extends unknown[], B, C>(
+  fbc: (b: B) => C,
+  fab: (...args: A) => B
+): (...args: A) => C
+// standard case, given 3 functions
+export default function compose<A extends unknown[], B, C, D>(
+  fcd: (c: C) => D,
+  fbc: (b: B) => C,
+  fab: (...args: A) => B
+): (...args: A) => D
+// standard case, given 4 functions
+export default function compose<A extends unknown[], B, C, D, E>(
+  fde: (d: D) => E,
+  fcd: (c: C) => D,
+  fbc: (b: B) => C,
+  fab: (...args: A) => B
+): (...args: A) => E
+// standard case, given 5 functions
+export default function compose<A extends unknown[], B, C, D, E, F>(
+  fef: (e: E) => F,
+  fde: (d: D) => E,
+  fcd: (c: C) => D,
+  fbc: (b: B) => C,
+  fab: (...args: A) => B
+): (...args: A) => F
+// extra overloads allowing functions other than the right-most function
+// to take in more than one argument, though the extra arguments go unused
+// 2 multi-arg functions
+export default function compose<A extends unknown[], B extends unknown[], C>(
+  fbc: (...b: B) => C,
+  fab: (...args: A) => B[0]
+): (...args: A) => C
+// 3 multi-arg functions
+export default function compose<
+  A extends unknown[],
+  B extends unknown[],
+  C extends unknown[],
+  D
+>(
+  fcd: (...c: C) => D,
+  fbc: (...b: B) => C[0],
+  fab: (...args: A) => B[0]
+): (...args: A) => D
+// 4 multi-arg functions
+export default function compose<
+  A extends unknown[],
+  B extends unknown[],
+  C extends unknown[],
+  D extends unknown[],
+  E
+>(
+  fde: (...d: D) => E,
+  fcd: (...c: C) => D[0],
+  fbc: (...b: B) => C[0],
+  fab: (...args: A) => B[0]
+): (...args: A) => E
+// 5 multi-arg functions
+export default function compose<
+  A extends unknown[],
+  B extends unknown[],
+  C extends unknown[],
+  D extends unknown[],
+  E extends unknown[],
+  F
+>(
+  fef: (...e: E) => F,
+  fde: (...d: D) => E[0],
+  fcd: (...c: C) => D[0],
+  fbc: (...b: B) => C[0],
+  fab: (...args: A) => B[0]
+): (...args: A) => F
+// generic base case type signature and function body implementation
+export default function compose<Fns extends Fn[]>(...fns: Fns): Fn {
+  const len = fns.length
 
-export default function compose<F extends Function>(f: F): F
-
-/* two functions */
-export default function compose<A, R>(f1: (b: A) => R, f2: Func0<A>): Func0<R>
-export default function compose<A, T1, R>(
-  f1: (b: A) => R,
-  f2: Func1<T1, A>
-): Func1<T1, R>
-export default function compose<A, T1, T2, R>(
-  f1: (b: A) => R,
-  f2: Func2<T1, T2, A>
-): Func2<T1, T2, R>
-export default function compose<A, T1, T2, T3, R>(
-  f1: (b: A) => R,
-  f2: Func3<T1, T2, T3, A>
-): Func3<T1, T2, T3, R>
-
-/* three functions */
-export default function compose<A, B, R>(
-  f1: (b: B) => R,
-  f2: (a: A) => B,
-  f3: Func0<A>
-): Func0<R>
-export default function compose<A, B, T1, R>(
-  f1: (b: B) => R,
-  f2: (a: A) => B,
-  f3: Func1<T1, A>
-): Func1<T1, R>
-export default function compose<A, B, T1, T2, R>(
-  f1: (b: B) => R,
-  f2: (a: A) => B,
-  f3: Func2<T1, T2, A>
-): Func2<T1, T2, R>
-export default function compose<A, B, T1, T2, T3, R>(
-  f1: (b: B) => R,
-  f2: (a: A) => B,
-  f3: Func3<T1, T2, T3, A>
-): Func3<T1, T2, T3, R>
-
-/* four functions */
-export default function compose<A, B, C, R>(
-  f1: (b: C) => R,
-  f2: (a: B) => C,
-  f3: (a: A) => B,
-  f4: Func0<A>
-): Func0<R>
-export default function compose<A, B, C, T1, R>(
-  f1: (b: C) => R,
-  f2: (a: B) => C,
-  f3: (a: A) => B,
-  f4: Func1<T1, A>
-): Func1<T1, R>
-export default function compose<A, B, C, T1, T2, R>(
-  f1: (b: C) => R,
-  f2: (a: B) => C,
-  f3: (a: A) => B,
-  f4: Func2<T1, T2, A>
-): Func2<T1, T2, R>
-export default function compose<A, B, C, T1, T2, T3, R>(
-  f1: (b: C) => R,
-  f2: (a: B) => C,
-  f3: (a: A) => B,
-  f4: Func3<T1, T2, T3, A>
-): Func3<T1, T2, T3, R>
-
-/* rest */
-export default function compose<R>(
-  f1: (b: any) => R,
-  ...funcs: Function[]
-): (...args: any[]) => R
-
-export default function compose<R>(...funcs: Function[]): (...args: any[]) => R
-
-export default function compose(...funcs: Function[]) {
-  if (funcs.length === 0) {
-    // infer the argument type so it is usable in inference down the line
-    return <T>(arg: T) => arg
+  if (len === 0) {
+    return <A extends unknown[]>(...args: A): A[0] => args[0]
   }
 
-  if (funcs.length === 1) {
-    return funcs[0]
+  if (len === 1) {
+    return fns[0]
   }
 
-  return funcs.reduce((a, b) => (...args: any) => a(b(...args)))
+  // `Params<typeof b>` is equal to `unknown[]` in the below type signature,
+  // but, `Params<typeof b>`, arguably, more clearly conveys intent
+  return fns.reduce((a, b) => (...args: Params<typeof b>) => a(b(...args)))
 }
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -10,7 +10,7 @@ interface Fn extends Function {
  * A type-level utility that extracts the type of a function's input params as
  * a tuple, inferring `unknown` for any input params assigned to generics
  */
-type Params<F> = F extends (...args: infer A) => unknown ? A : never
+// type Params<F> = F extends (...args: infer A) => unknown ? A : never
 
 /**
  * Composes single-argument functions from right to left. The rightmost
@@ -114,7 +114,7 @@ export default function compose<Fns extends Fn[]>(...fns: Fns): Fn {
     return fns[0]
   }
 
-  // `Params<typeof b>` is equal to `unknown[]` in the below type signature,
-  // but, `Params<typeof b>`, arguably, more clearly conveys intent
-  return fns.reduce((a, b) => (...args: Params<typeof b>) => a(b(...args)))
+  // `Parameters<typeof b>` is equalavelnt to `unknown[]` in the below type,
+  // signature, but `Parameters<typeof b>` more clearly conveys intent
+  return fns.reduce((a, b) => (...args: Parameters<typeof b>) => a(b(...args)))
 }
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -8,19 +8,27 @@ interface Fn extends Function {
 }
 
 /**
- * A type-level utility function to
+ * A type-level utility to get the type of the tail of a tuple of functions
  */
-type Tail<Fns extends Fn[]> = Fns extends <A>(
-  a: A,
-  ...rest: infer Rest
-) => unknown
-  ? Rest
-  : never
+type Tail<Fns extends Fn[]> = (
+  ...fns: Fns
+) => unknown extends <A>(a: A, ...rest: infer Rest) => unknown ? Rest : never
 
 /**
- * A type-level utility function to
+ * A type-level utility to get the type of the last function from a tuple of
+ * functions
  */
-type Last<Fns extends Fn[]> = Fns[Tail<Fns>['length']]
+type Last<Fns extends Fn[]> = Fns['length'] extends 0
+  ? never
+  : Fns[Tail<Fns>['length']]
+
+/**
+ * A type-level utility to get the return type of the composition of a tuple of
+ * functions
+ */
+type Composed<Fns extends Fn[]> = (
+  ...args: Parameters<Last<Fns>>
+) => ReturnType<Fns[0]>
 
 /**
  * Composes single-argument functions from right to left. The rightmost
@@ -113,9 +121,7 @@ export default function compose<
   fab: (...args: A) => B[0]
 ): (...args: A) => F
 // generic type signature for any number of functions
-export default function compose<Fns extends Fn[]>(
-  ...funcs: Fns
-): (...args: Parameters<Last<Fns>>) => ReturnType<Fns[0]>
+export default function compose<Fns extends Fn[]>(...funcs: Fns): Composed<Fns>
 // generic base case type signature and function body implementation
 export default function compose<Fns extends Fn[]>(...fns: Fns): Fn {
   const len = fns.length
--- a/test/compose.spec.ts
+++ b/test/compose.spec.ts
@@ -104,7 +104,8 @@ describe('Utils', () => {
     })
 
     it('returns the first given argument if given no functions', () => {
-      expect(compose<number>()(1, 2)).toBe(1)
+      expect(compose<[number, number]>()(1, 2)).toBe(1)
+      expect(compose()('zero', 1, 2)).toBe('zero')
       expect(compose()(3)).toBe(3)
       expect(compose()(undefined)).toBe(undefined)
     })


joshburgess:
- If you look at the overloads, you will see a set of them under this comment:

```
// extra overloads allowing functions other than the right-most function
// to take in more than one argument, though the extra arguments go unused

...overloads here...
```

I had to add these overloads because various tests were passing in functions, other than the right-most function, which took in multiple params. This behavior was allowed previously due to the types (and I've maintained that behavior in this refactor), but it actually goes against what the JSDoc comment description said: 	Composes single-argument functions from right to left. The rightmost
 * function can take multiple arguments ...`. Personally, I would use the type system to not allow the user to give `compose` functions that take multiple arguments, since the implementation completely ignores the extra args anyway. I think the way it works right now is a bit misleading. If we removed this odd behavior & changed those tests, we could get rid of those extra overloads and have simpler typings for `compose`.

- If we consider further breaking changes, does the right-most `compose` really need to take in multiple arguments? This is abnormal versus how `compose` functions work in most other utility libraries that have it. I understand you might want to support it to make things easier to use, but the typings would be much simpler & more straight forward if all functions passed to compose just took a single argument, and the user could still accomplish the same things by just partially applying some arguments to the right-most function using wrapper functions & currying instead. They have to do this already with the other (non-right-most) functions anyway since only the first arg passed to them gets used. It's worth thinking about, IMO.

- Why allow `compose` to take in no arguments and 1 argument? Again, this behavior is a little strange compared to what you typically expect from a `compose` utility. It seems to be written in a way that is too dynamic and overly complicated, IMO. While it was possible to provide valid typings for the current behavior, these corner cases made things more complicated & required more overloads. Conceptually, it doesn't really make much sense to allow this function to take 0 or 1 args. Not restricting this in the type system could actually be worse for the user by allowing them to accidentally do things they may have not intended to do, IMO. Passing one function just gives you that same function right back, which isn't very useful. Worse, passing no functions to `compose` gives you a back a function taking any number of arguments and returning the first argument it's passed. If memory serves me correctly, this is different than even what `ramda`'s `compose` function does. I believe it just returns itself when given no args. Either way, both of these things are a little strange and not very useful. 

- Regardless of whether any of the above things change or not, why does the library offer only a right-to-left composition utility and not left-to-right? It would be pretty straight forward to add a complementary `pipe`/`flow`/`composeFlipped`/whatever-you-want-to-call-it function that worked the same way as `compose` but in the opposite direction. I prefer right-to-left composition in languages where it's normal & there are no problems with type inference, but TypeScript (and Flow) generally has a much easier time with left-to-right inference. Even pretty serious functional programming libraries, like `fp-ts`, have decided to deprecate right-to-left versions of these types of utility functions & promote the use of left-to-right versions instead due to stronger inference capabilities. I'm not necessarily suggesting to remove `compose` altogether, but if you keep it, why not add a complementary left-to-right composition function to go along with it for people who would rather use that?


Anyway, these are just some thoughts I had while/after implementing the changes for this PR. I know most of the things mentioned above are a result of code that was written before any of the current maintainers were maintaining the project, but, considering you're already rewriting in TypeScript and will have some breaking changes anyway, I think it's worth thinking about some of these things and possibly making some changes in an effort to  simplify behavior & make things easier to statically type.

markerikson:
Thanks for filing this!

To answer your last question briefly, Redux's `compose` is really only used for two cases: store enhancers and HOCs.  In both cases, the reading order matches the nesting:

```js
const composedEnhancer = compose(
    applyMiddleware(thunk, logger),
    batch(),
    devTools()
);

const composedComponent = compose(
    connect(mapState, mapDispatch),
    intl(),
)(MyComponent);
```

There's no need for varying argument orders here.

joshburgess:
Note: The build is currently broken due to this line: 

`dispatch = compose<typeof dispatch>(...chain)(store.dispatch)`

This will need to be fixed. I think part of the problem is related to other types in this file, but there's also the fact that `chain` is an Array of any length here, whereas this PR's implementation of `compose` only allows overloads for up to 5 functions.

markerikson:
Hmm. Can it be made generic for an arbitrary number of functions? Realistically I don't expect too many, but that was part of the genesis for pulling in `ts-toolbelt` in the first place (see discussion at https://github.com/reduxjs/redux/pull/3536#discussion_r320398474 and following ).

joshburgess:
@markerikson I'm not sure I follow about the reading order matching the nesting, but I can see why a left-to-right version or eliminating the 1-arg case would complicate things since the dev tools' `compose` is also right-to-left & has the same 1-arg handling as the current behavior. I think, ideally, neither the devtools' `compose` or this one would allow 0 or 1 args and people would just not wrap their code with `compose` in those cases, but I can see how coordinating this with the dev tools maintainer would be difficult. So, it's probably not worth trying to change.

About my other suggestions: I think the most important one is the first about not allowing the non-right-most functions to take more than one argument. I can't think of a good reason for allowing this, since the implementation ignores them anyway. I think it would be best to just not allow people to pass these kinds of functions to `compose` at all.

joshburgess:
> Hmm. Can it be made generic for an arbitrary number of functions?

I'll look into it. I also haven't played with `ts-toolbelt` at all. I hadn't heard of it until seeing you all talk about it. I'll try it out.

markerikson:
Yeah, lemme phrase it this way: there's a ton of existing code that we don't want to break, `compose` has a limited number of use cases with Redux, and we're looking for practicality over idealism or purity.

Agreed that multi-args probably aren't relevant, but I don't know how we'd prevent that at the JS level (and not entirely sure how you'd do it at the TS level).  Not terribly critical.

joshburgess:
Okay, @markerikson I updated the PR with a new overload based on what @jedmao was doing in the other PR, but without using `ts-toolbelt`. I replaced my `Params` util with the native `Parameters`, which I didn't know existed before (I usually defined something like this myself in my own TS projects), and then used that along with the native `ReturnType` and some new type-level utils to add a new generic overload for any number of functions.

However, I didn't remove any of the other overloads, as I feel like this should be a last resort sort of solution. Typically, you'd still want the help of the inference of the other overloads. This is really only needed for that one spot where `compose` is used internally in `applyMiddleware`. Maybe, that bit of code should be rewritten to not use `compose` instead?

Also, we could raise the number of functions from up to 5 to up to 10 to cover more ground, if you think it's necessary? It obviously adds a good bit of length to the definition, but it's fairly mechanical.

> Agreed that multi-args probably aren't relevant, but I don't know how we'd prevent that at the JS level (and not entirely sure how you'd do it at the TS level). Not terribly critical.

I think this would be pretty easy. From the TS perspective, it would just require removing the overloads I created, the ones under this comment:

```
// extra overloads allowing functions other than the right-most function
// to take in more than one argument, though the extra arguments go unused
```

and if you wanted to change the JS behavior also, the reduce call in the implementation body would just need to be changed to not pass extra args through after the first iteration. I think this would do it:

```typescript
  return fns.reduce((a, b, i) => (...args: Parameters<typeof b>) =>
    a(b(...(i === 0 ? args : [args[0]])))
  )
```

cellog:
I am unclear how this is better than what we have now - the point of the rewrite was to remove the overloads, this just replaces them? Can you explain what makes this better than the existing implementation? I am sure you already have, but I was not able to parse it out from the thread above, and perhaps a simple summary would help me

nickmccurdy:
If we're going to have this generic type signature anyway, why not delete all the others? ```ts
export default function compose<Fns extends Fn[]>(...funcs: Fns): Composed<Fns>

```

**Comment by markerikson** [comment]
I'm inclined to agree with Greg.  The original genesis of the question was if there was a way to handle an arbitrary number of arguments, generically, _without_ hand-specifying specific numbers of parameters in a copy-paste manner.  At the moment, this just looks like it rewrites the manual overloads, and doesn't simplify / replace them.

**Comment by markerikson** [comment]
@OliverJAsh pointed to his `pipe-ts` lib as an alternative implementation:

https://github.com/unsplash/pipe-ts

However, that one _also_ has hardcoded overloads, although it goes up to 9.  (*but why not 11? :)*)  He thinks that an arbitrary implementation isn't possible atm:

https://twitter.com/OliverJAsh/status/1170753063709224961

**Comment by joshburgess** [comment]
> If we're going to have this generic type signature anyway, why not delete all the others?
> 
> export default function compose<Fns extends Fn[]>(...funcs: Fns): Composed<Fns>

Like I said, this should be a last resort. It doesn't provide the same inference as the overloads because of the nature of how it works. All it does is create the return type of the composed function by getting the params of the right-most function and the return type the left-most function:

```
type Composed<Fns extends Fn[]> = (
  ...args: Parameters<Last<Fns>> // params of right-most function
) => ReturnType<Fns[0]> // return type of left-most function
```

This says nothing about the types of the functions you're passing in though. You basically lose all inference help for functions in between the first and the last and now must manually annotate everything. In this PR, try commenting out all of the overloads except for the generic one you suggested & the base case with the implementation body. Then, in the tests, find a `compose` call taking in at least 3 function, chose a function in the middle and replace it with an unannotated version (defined above or defined inline). You'll find that you have no type inference at all and the variables in your unannotated function get inferred as `any`.

This makes sense, because the generic implementation `function compose<Fns extends Fn[]>(...funcs: Fns): Composed<Fns>` knows nothing at all about this function or how types are propagated through each function. It only knows what the result of composing all of them together should be.

Also, there are problems with the old overloads. I called out an example in the original post: `This actually fixes a bug that was present before which (accidentally, I believe) restricted these multi-arg functions to take in arguments all of the same type, which isn't necessary.` Because of the way generics were being used, restrictions were being applied in ways that didn't need to be there.

For example, in this test: `expect(compose<number>()(1, 2)).toBe(1)`, the overload for this case forced all params to be of type `number`, but neither this case or any of the other cases that take multiple params, actually need to do this. They should support params of different types. That's why using the `A extends unknown[]` + `(...args: A)` style of overloads is better. It provides better inference & more flexibility with overload signatures.

Another thing to try... Underneath the above mentioned test, try adding this line:

`expect(compose()('zero', 1, 2)).toBe('zero')`

You will get a type error because there is no overload that properly infers arguments without manually/explicitly providing generics for this case. This isn't true with the overloads in my PR. They are able to provide correct inference without explicitly annotating for many cases.


I think people are looking for an easy, simple solution where one doesn't exist. Thoroughly typing a generic, variadic version of `compose` with good inference has been a problem in TS for a very long time, and it's still not really solved in a small, elegant, generic way by any library right now. As @jedmao noticed, you lose type information with ts-toolbelt's `Compose` utility (you'll even see in comments and commit messages that they already know they have trouble with propagating generics in various situations), and you lose inference in various ways here with the params of last function + return type of first function approach.

This is why libraries like `fp-ts` (which has some of the best inference capabilities of any library with utilities like these) take the pragmatic/practical approach of just specifying the exact overload signatures for a reasonable amount of arguments (usually 5-10). A function with a ton of boilerplate of carefully written overloads that guides you along & helps you to prevent errors through proper inference & type propagation ends up being a lot more helpful/useful than a small, elegant solution that loses type information & is unable to provide you proper inference.

Yes, requiring a ton of manually written overloads sucks, I know, but, unfortunately, as far as I know, it's the only way to really make things work the way they should in TS.

**Comment by nickmccurdy** [comment]
If it's possible, it would be better to use mapped tuple spread types so we can make this type safe *and* homomorphic.

**Comment by markerikson** [comment]
That's what I _thought_ I was suggesting in the other thread when I first brought this up :)

**Comment by joshburgess** [comment]
> If it's possible, it would be better to use mapped tuple spread types so we can make this type safe and homomorphic.

Hmm... I think I would need to see an example. I'm not tracking what you mean.

Don't get me wrong, I would love if someone were able to solve this problem without inference problems & tons of overloads, but I haven't seen it yet.

**Comment by jedmao** [comment]
To be clear, the ts-toolbelt actually uses like 40 overloads under the hood, so hiding that in a separate library does make the code more readable, FWIW.

**Comment by nickmccurdy** [comment]
I checked the types from Underscore and Ramda, and they don't do this. As far as I can tell, this is not possible with TypeScript yet. I can do the following the a fixed number of arguments to homomorphically detect the return type in TypeScript 3.4, but since spread types aren't Turing complete I can't generalize this to variable arguments without breaking the homomorpism:
```ts
function compose<A, B, C>(f: (arg: A) => B, g: (arg: B) => C): (arg: A) => C {
    return x => g(f(x));
}
```

nickmccurdy:
`Compose.sync` from `ts-toolbelt` seems like an improvement as it can infer the arg types as the last function's args and the return type as the first function's return, but it doesn't seem to match the types in between so I don't know if you can call this a homomorphism: https://github.com/pirix-gh/ts-toolbelt/blob/3b072630/src/Function/Compose.ts#L75

joshburgess:
@nickmccurdy Yes, that's what I was saying. It loses type information/there is no inference for the function args. It only knows how to create the return type, but this requires you to annotate all functions, and leaves open the possibility for masked mistakes.

> To be clear, the ts-toolbelt actually uses like 40 overloads under the hood, so hiding that in a separate library does make the code more readable, FWIW.

What do you mean? `ts-toolbelt`'s `Compose` does not provide explicit overloads. It works essentially in the same way as the `Composed` helper I defined here and has exactly the same problems.

I'm just reiterating the things I've already said in previous comments, but just to be extra sure, I just created a new branch, installed `ts-toolbelt` and used its `F.Function` and `F.Compose` utils to define `compose` and looked over the tests, and all the same issues are still present.

I went ahead and modified `expect` assertions + left a bunch of comments to try to show the problems with using this approach as the sole type definition without overloads.

You can find the branch here:
https://github.com/joshburgess/redux/tree/ts-toolbelt-compose

`compose` function defined in terms of `ts-toolbelt`'s `F.Function` & `F.Compose`:
https://github.com/joshburgess/redux/blob/ts-toolbelt-compose/src/compose.ts

Test file with modifications & comments showing the problems with this approach:
https://github.com/joshburgess/redux/blob/ts-toolbelt-compose/test/compose.spec.ts

This is why using explicit overloads is better, IMO. It gives you much stronger inference, and, thus, is much safer. A type like `F.Compose` is not safe on its own. It only really makes sense as a last resort fall back on when no other overloads can be used to try to give the user a small amount of type information.

cellog:
> > To be clear, the ts-toolbelt actually uses like 40 overloads under the hood, so hiding that in a separate library does make the code more readable, FWIW.
> 
> @jedmao What do you mean? `ts-toolbelt`'s `Compose` does not provide explicit overloads. It works essentially in the same way as the `Composed` helper I defined here and has exactly the same problems.
> 
> I'm just reiterating the things I've already said in previous comments, but just to be extra sure, I just created a new branch, installed `ts-toolbelt` and used its `F.Function` and `F.Compose` utils to define `compose` and looked over the tests, and all the same issues are still present.
> 
> I went ahead and modified `expect` assertions + left a bunch of comments to try to show the problems with using this approach as the sole type definition without overloads.
> 
> You can find the branch here:
> https://github.com/joshburgess/redux/tree/ts-toolbelt-compose
> 
> `compose` function defined in terms of `ts-toolbelt`'s `F.Function` & `F.Compose`:
> https://github.com/joshburgess/redux/blob/ts-toolbelt-compose/src/compose.ts
> 
> Test file with modifications & comments showing the problems with this approach:
> https://github.com/joshburgess/redux/blob/ts-toolbelt-compose/test/compose.spec.ts
> 
> This is why using explicit overloads is better, IMO. It gives you much stronger inference, and, thus, is much safer. A type like `F.Compose` is not safe on its own. It only really makes sense as a last resort overload to fall back on when no other overloads can be used to try to give the user a small amount of type information, because that's better than nothing (I suppose).

Perhaps we aren't using `ts-toolbelt` properly?

https://github.com/pirix-gh/ts-toolbelt/blob/master/src/Function/Compose.ts#L45

This code clearly infers the returns and parameter types of every intermediate function. I have to run, but I will write more later

joshburgess:
> Perhaps we aren't using ts-toolbelt properly?

I think you're right. Sorry, I didn't read those comments before. I see they are using mapped types to try to infer more information via the tuple's indexes. That looks interesting.

> This code clearly infers the returns and parameter types of every intermediate function. I have to run, but I will write more later

I don't think that's true. You can see in the example that he has annotated all of those functions manually.

I updated with a new branch using the F.Composer + F.Compose approach like the example in the comment there showed.

New branch here:
joshburgess/redux@ts-toolbelt-composer-compose 

`compose`:
https://github.com/joshburgess/redux/blob/ts-toolbelt-composer-compose/src/compose.ts

Test file:
https://github.com/joshburgess/redux/blob/ts-toolbelt-composer-compose/test/compose.spec.ts

I had to update the test file to change some functions, as this way of defining `compose` does not allow any of the functions to take in multiple arguments. So, even if this worked 100% without issues, you wouldn't be able to use it directly.

However, it looks like it's still unable to properly propagate types through the function args, as you can see the things that were `any` and `never` before are still the same.

nickmccurdy:
Explicit overloads are not homomorphic. I think we’re better off with TS-toolbelt’s compose function.

cellog:
OK, I did some experimenting, you can see the results [here](https://www.typescriptlang.org/play/index.html#code/KYDwDg9gTgLgBDAnmYcD6AxArgOwMYwCWEOcAvHNvkSQFC2iSwLKoCGVBxOAPAApxQMYDgAmAZzhsciANoBdclJkAaOACVBIYWMnTES-QD5ytOObgAKSwDo7bKAHNxALjh8AlORPqvAMnROGhx6Rmh4JBQ4AElhKDZggFk2MCUAbzMLAHIAWgAWAAYst1kstDQstVyAZgBOSrhcwob8gqqcrPkVTPMa+pKmovbqgA4Gvpa69s7ui0ac0eK4UoX64YB2cYWxtW3prp756s2B7a3qgDZzzd3j-dnshavT4-OAVnOr24v7w5qPl5fI55d6TN6-OY1EEvD7DarnEG3PIQx7VeEvRFHABM53ht2qKN6CxxLzxRwAjOccbcsYSKUsVtVqUchvTbuS6TUii9Ke0sWsWZM2vMZn8cvyGVytlidvMJbt+ZzxWNTvK5Tc5bLlUqsSdlrkZdKgQaNeL1jrnvqzdLYXKgeKfiKDpDxQCrVjjeLMQbba6ddD3baDWSDZivTr0e7veLmcGWliCU6HkSsST3SHxby5cyYzrKarY+LWQas5mddz3VncuSBQbWUWlTXJUWtuTZdWBTka42VVam+1yRrq1q2429Ss263PYOWoPG5aJ0Ou0Gu-byY7cqKXeS3RPp9Gd7PwUmxeSAxOV+SM2fZ8iT9vIxOD4Wr7PE5vnY9yWmJ9eq12c2-Rt8z7F9iy7UtyQ5e8vwrCd-1rLt63JYUP2TeZ+j7cD2xyTtaiVXsVkQpctRGJVx1yHDPVNc0YKJBdciXFd7Q3DpPyJXdck9aNfWPNCxXPXIVwzMM734l1HyaaVzkmHVm2vaSFUbZtY3AyDOTg6txirVlUKKeQ-hA0pi1jLNoKyABqLdsh-LIqxDZlaUaKz2MaSTY29Ml3xc9CskEkMg0xMSfL+TjvWNW0+JCyEGKDIcgQ3aLsgo412w1WjLOs3pCKyIda1lMjnKyxpMNKdsUPGAV8KK1y7M02soO01kUKqJLekagYKqqb9tLM8y2saHrOqrV9uuZICat819OtM71RrgK9WuKuzBPmuygxvNQzyW2rD06ubjUPLaouW9dJSOwahzOrbEtOijrsG8qNTnSbDJymduoa2VR1eyEm069ssWLfsFuqzLaqB87ayxKtIbUIGdt8mHJThxpU3GZH4f65b0YGTG0YcxynPBpH3NjLFvQTBoE0Rv4KZRhyg3p+HgpxzjmbR40sVtbnachD0UaZocBfh26IYokW0cBjVdT5myct1DHodlGU5d6CU8fbaoTIFRVfuybWUdrap7NZbW1bcoyskNqomXGE2GhNi3rdsh3bZDO21CZZ20UlT23M8rzvOW6p-I9oNQ8d1nauqTjI9t41Y8dk6Y4YpPbaHS5HbF3zXgGLPba1jU7n13pFnzzPjdlUYfdK6322aW2BSmUvGmafPa3KcZG7gQpna7gYu6qIe4HEGAoEIHBHDUHAsAAWwAI2AKBpjgAAfRoinXmraAAX1CcBwhYKJYmXhJuHIWRDgAa2ARAIAAMxiOJz5IZIwHQ2-76f0-4iSFJ0JjwnlPdCs9F7L18h0beWQt4b0yrQAyDBD7MEiKgDA0A54JEkBQLIOAsjQPEFkA+TAIisEoBghIPBohaB0BIZ+Z9ghqAfnPeAQgRB0PQVATBMBxAmAoBkSEhC3DRFkFiAykI8HCNkNUAyu9ZDMJgIggA9AAKhUXAAAKvEHA4gH4YKkHAMBS8oBwEnjACAyg4BqIAAa-1fjgaxaizDqMsOYuAWBxCoBsXwKAwAABujiVFqBsQAOSEIEtQARvEQHEIEjwzi4AAAEwAODYHPOAISEAWPHtIXRGCEmJN8TALAUAdFWJUbYl+wRAkFNAGksAAAbYACTrGtJ4Qkwgc8SFwDSNEXecAH5QAgOkrIPCcjmIgA0peDSYBEKsQk1BpilDRBsHY4IAB5B+PAYFZBMEopRywABEXZDlqEOeSU5cBDkFEucKa5hzxEqIWWQnAQhlk2DCdoHghATBzHMPso5NyzlYkuecy50FDkWQec8qIYBfF+PeT4-x3zfl-IBbIY5IKzlAquSc3YEKcjQvmVY4+qAcCvO0IimJPAKUwFRXAAFFlyQwrJXC-xVLxA8DZX4+lAKuwtNac4pRSDumLLWdwTZPBMlsN0KPcek9HB8MONK7Q7DJBf0fvQv+3B36HAAPxavse-WQITxEWGEVUnVKRShd0QWEFBZC+DUuoTKuh4qSBKrmJw7hVCqh4KMMQo+iyNFsEIA0ngGiaFquUHIeQnqLDWDsDYGAbgNFeDICYfQXhXWSGsAAC2AGwUQbh9BqCTTAUNDS3CTwfsvTRIaw3pszTIeJcwDUaIbQ0w4bhXl+OXoGh1URkgwDwHmpQcDRDAAfmwLAMy5l-IXYu6BnTGnADniIMZRh51Lp3dAngOQV1NPXTgHh27d2LrgTmnIW7Djnr+XA-dOaz13rmJegAjlgNgDTCEDtISfMQoAlBAIVdvIxdaN7iEQIvSZIqg1kIANJ305esqNsqIALwAFbAAIPG8wGqn4oYCNEADIBYODtQAAIQgJMwtpAKBaKwKgDeGAv2eN-aSzRUBGPmAoMy+1f60GsdQBQAo7HFkAFFVW6B4AAQXJKhuhpa4AyaxApvQMhcPLDk4oHNyxe3LzNeYA1LHv2oABfoGAeaQP6ZMYQSQ07TPduU-J3TKnb1wHbVx5pC63AmbY-xjj0QulHo3ZyuTamY1qBUxF4wpg5iSdoWF6CKmTC6YY95iwnnGNOb880gLEmP2sdky5qTinVDKdU7p2LZBDiWFk0YSwTblMRfC1l1AvmhPZtK7m+rjXvDNdc6ptrcAOumdbZlzj2W5ijbY2RgTMQwsxfK+Fqr5XuGjoi8OjbODJ3TtnbMvhPS-i7ZnXOv5bgEtqtk2oOYcmA2QkPWu0L16lhBdXcenh13zB3b+Ferd5q4CXekzJm7FgfuQn3Y9j7QiYjBaeyepLUX7vZEfd1-B5gLvdeK6D5TyPejAEK9+9H03AeE8R38mTAa5HrbzQZfLZD1lQB4BRkriW4BUZo9INQFHKvdfZ9Rpp0hDsCIsAUNwIuF1i8oEJ9z5hySpq84cfecx5dHaXVL9LsuFoK6mxYfeciWfyFkDzu1yD5sAGVgBNIIIhxAyGIvoawwQNQiQluIDUDTzbCRtuNBO-t3Z6RDiyHg6Y0gtvkNGHkOLrXqv4Na6lzZpXshojIeD10OAiQPfe7zZHvewekM8HWbn2giz4M4AgAAdxwO6nA79w9KEk3gBpWBJ08Hw4a-+H9GhdwDYsp14hCBED7TX+vFBLfW5gOHqhlq34AOWEphf5WlPg172QsJjhz7D5nzgUfcBx-YcnwXmv781CyEX+7yLl+PwBtoKo9RAARYA4g8ATyXuq8vVe4B+K-YxyQmq2DlJgZQBOJWLCr05RAhLzzGLYJq7ZDAZTxLAS5-JZBfoNJLDeqULH7Wpl6V7V7b515Ibp5ZCEJ7IHK4AXz-4NINJyrAKOBa6ZQY7kJcKYH4HWr96D6EBb4MLcDh5EEkEMpkE4AUFPwWQWDwF0FLqbiMEYEwDT7cGz5gAmrAAb5D7AAj6EFVD8EArkEkBwCao5BiHypTz0FbwjZME+pYGKE7J8G7JK5-BAGIH0GoHoEUKyGWHB4f54HyG14pC8F+oB7aFCG6GUHUFAH0FWTSGuFyHaoKGyDsGqHqF25EH+oCHuJBGkCaqiHmBhGSFQJmEyHRFGrWrr6b5qHb5+GNApGBHCFwAGHZFQH9qSGmG+ZRHuHWH+F45wD65iZr4NFQCSoqps5AGaaDHRqQHgJQDiClAOFG4oFUGdCpGSYoA6KcGoAV6D5jpAHiD6pwC26FGd6yD7425H6sFd5n5L4XEX4hJqD6BxpxpObiE9FRAAAyIgjglmEabuCgTCLCEWMhMBuCAeFAsBvQMO4xxikqGipQTSU8lmnQnRgJqa0JbxcJsi8iLCpuoqZCzxbAY8nxq2sammUJrxsJeaEalaEaRgHRdOZuHGAAEoWqIPiXzvoAoJpiSe8WSRoqlnzqJm2oYv4o0QDlCQUJiXBlEHsS6nzjXppgUdEJobYeAagOoMAMUqUjwBgH8bgFwB6nFhYJqbpomvYE4K4DGgoE1jWnWr4DseoE5onkqRkkIFQhFjKXqeYO4SIuSPIIoACngCQEQLPE-qYpasYQ6T4sMnZsAJKgIASZpjGXzuGXPJGd8jgLWiYpTjsTJk5nwE8agObkYY4AUebhFuIT8awnzv8cLn8AvI4UuqrhzoLiEOrm4MWbplkA5p4s+gusZjLuejrhluYHIkDhIDwKWXvrnhIksP3vsRKlsubkYAiTDubnngoogosgAMLDKQCeIYA4Dm6ID4AallJVZBDcDfG7ERbt67m8JulwAckfGdpHm8I8ls75m0EFEh4BBjmAmdHGY6Jp6pHRBPyYJgAoCiAca+JsCjrADgWWaoDsJOYcDanBA8CyCpF5rSCiANIga4DXyeFwCOAiDLyEB4DiBqAV6hoBmOADLQBwBwoQDblfpa4qlqm8DXlxHUqfJuHb6Sqfk0EKoLmR4vnRqWkmIAAaWuBqaQquSmUuYlcijOPAw5nKuFnhagYlVJgOWOqluB55GlE5i6PagpUAhw6eLFJSbF-505NevFcAX5BZC59xcwN+G5W5MSwAu5MmkGh515MWp5JA55IeumV5Oi7JKJXJFJ15C5EWb5CqH5dl-FIClRth-JiZkZkq7F8GcagFwFKQYFEFha0FsFBaggYgiF-lvAaFAKGFYg2FU8aReFuBBFRFE8pF5FlFIGeiJi9FjFXaS6aVnikq5l6p7F05XFM5JAtl9ltBglcawlsqolcAElS6UlMl5WclClTOylPAOlVe6lml21u1OAeljlWuRlfaJlcwZlqpFlT5HFnKNlWyfF4hjlnRN+DpiQEAk6SgxBB5eA+CcCuJf1RCt+aicAm5XSWAwgcAFeGFEQJVk8YAUNehT8NiEN25wAgSo8eaEAs64FS8BSKS8Q6SvlbifpXS7lhNqS6Sc8X1qA3leASiQN+AVgeqZA1iv1+AWQ1ira6iiSdScOAq1i7SJKrS1iQqc2HG6N7lTOvlJ5yFZ56etN31umn131OCDNQJIJjQmtZhaQweoecAIV4gUe4NblO5e5f1T5ag8GRgu8vkzN-14uBtk8Rtd8mq15pt0tFtXlVt15NtdteeytwAyiYN3tXiFSFVgSkg5ihFcFl1fNRNaSlAZSZN5t3mfNRSFlkgNiUdIBfNAtq6QtIt6iK6R8aQGA-SgywyjQYyEyUyVusyziCS+yByQFcA98WAUgvicADS1G18nVtFL+haVFpK9mtF1i5NGN4tJKrdcA1i4dUA1iMNYa1B0FeA18NFtmOASN8AAQC96dy9cFehUNu9CSk6TeDgqAD8CtuhU97lT5fxNgFVbJtgxpzgvmNgi9T5RgHgn94dP9zdJKfpOi8AABFAlgbAquQBTW1iAAJGkFA7vDPeoiA2PHAHgEoJYHgKruIeaf1rIDg48hgyQOgwvFgwvLgwWU1rIJQ48gkvfZ4tg2oAvDcR4JYHkFiLzS3QCjJmIKYk-J3d3agH3RAAPfVV1ZYo7fPWAIQCgEfawOfdhg0lfQMrfaQIw8AI-bphgM-eo6-Umg4B-ZQF-enbLTolUI7bsn-SY9-f7Y0FY-dk8sA6Q2A4YAzVYFAz2n0bAwg0gygyQ6A3AOQxQNI5YJQ24OITQ3Qww64xg+439VYDg5EwWfgxmssEQ0A+omwBRYPiQxTUw3gCw2wxw1wxLQ6QA3LSyS-UrXTZtnUxrcDVWYIsDWYW-TYEY6aXwNTaqcvJyjiXiVFb-f1sNbwAyUWj-QiVY24O05024N08Tb05MTwAM7IUM01gNVo5s0NTdeqeM0yUMwuUHXTYgjfdQBfJo9o9U-o3GrMyaW4HY6FTY5U6FdrYQE-JYAAISdM2AwmcleBZ2lJWCGMmlNadOyCimHCAukA-O+KiBYB4DADWBnNFMYMlK+InoVVNZIFzDvNWCfMosAu7MaPosboVVa7QvAvv3iBNYovYOkuYvqPjaDnxL7y0BoMRBKCaPWCmlRP9ZIGUuyAm17xqCQPeMTFqC8vUP8tQvEvz0IPiC7xuB+PIN7zxLsukM0a-MQCOCuKWB8RWZZAeDqu0BAA).

Note that the copy-pasta to make this work is pretty insane (first of all).

Also, note the pure simplicity of the compose function, no types need be defined at all.

Second, it demonstrates that in fact the types are correctly inferred for intermediate functions that explicitly define their types.

However, I think I see the disconnect now. This definition of compose expects the intermediate functions to explicitly declare their parameter types. The redux implementation has been more strict, requiring all of the intermediate functions to have the same signature (all middlewares, for instance, or all store enhancers).

This then raises a simple question: is this bad? What is the issue with requiring middlewares and enhancers to be explicitly typed?

jedmao:
We could definitely save a lot of nonsense if we explicitly typed the middlewares.

timdorr:
I assume when people need to compose a set of general functions, their first thought isn't to reach for Redux.

But perhaps we can be explicit about this by renaming the function to `composeEnhancers` (and re-exporting under the `compose` name for BC).

joshburgess:
> Second, it demonstrates that in fact the types are correctly inferred for intermediate functions that explicitly define their types.

I think people were (are?) misunderstanding the things I wrote for some reason. We are talking about two different things here. You're saying that it's able to infer the correct return type for `compose`, but I've already said & shown that in the previous posts. I'm talking about the functions you actually give to `compose` as arguments.

The reason why I'm advocating for overloads is specifically because the functions being passed to `compose` can not be properly inferred, which is unsafe and allows `any`s to sneak in. The tests file in my last post shows this. In my view, that means that this solution is inherently worse than explicitly providing overloads which provide much better inference and safety.

> This then raises a simple question: is this bad? What is the issue with requiring middlewares and enhancers to be explicitly typed?

I think so, yes, because it could provide much better inference. The issue isn't whether or not explicitly typing things is bad. It's whether or not the function can allow you to write the intermediate function arguments inline without explicitly annotating, which is very much possible with a different implementation, like what's in this PR. In my opinion, disregarding this just to try to achieve an elegant one type signature that doesn't cover all cases and has worse type inference is a bad tradeoff.

If you all disagree with that, then there's nothing more I can say, but I think that is contrary to what most people using TypeScript want & expect.

jedmao:
I agree. If we can provide better inference and not swallow types along the way, even if that means an insane number of overloads, the user experience should be the most important goal here.

cellog:
This makes sense to me Josh, thanks for the extra clarity. I think the key difference is in assumptions.

I am assuming that our `compose` function has only 2 limited uses: composing middlewares and composing store enhancers, as this is the two ways it should be used. In both cases, functions will never be defined inline to the actual compose definition.

If we assume it should be general purpose, then my assumptions break down.

A larger problem is the definition of `applyMiddleware`, which also currently is a forest of overloads. To properly redefine that, we will need to write our own definition of the compose types anyways.

In any case, I think we are getting closer to a solution, even if we ultimately decide not to touch the existing definitions. Understanding the beast we are up against is equally important.

The key takeaway I have from all of this is that typescript really needs to be improved drastically before it can truly handle the expressiveness JavaScript supports in a clear way, and that is a rather large asterisk on the statement "Typescript provides type safety for JavaScript" that we should all be cognizant of.

Now, before I get too philosophical, I wonder if we can take a step back and answer a couple of basic questions:

1. What are the limits we expect to place upon the inputs to `composeEnhancers` (h/t to @timdorr this is a suggestion I like)?
2. How about to `applyMiddleware`?

I have answers but rather than bias the results I would like to hear from others first.

If, for example, we want a maximum number of middlewares or of store enhancers to support, that would create a limit we can work with. Or, if we expect all params to be the same signature, that also gives us a useful limit. Some of this is implicit (all middlewares should have same sig) but let's make it explicit.

Then, we should design types that support these limits.

jedmao:
We could also introduce two separate compose functions that are meant for either scenario, which might allow more freedom in how each is typed.

markerikson:
The other common use case I've seen is composing HOCs.

timdorr:
I think React Redux should provide its own compose for that, then. That would let whomever is writing the typings for React Redux also avoid this mess, because an HoC has a specific and consistent type (`(args: T) => (comp: React.Component): React.Component => comp`)

FireyFly:
I'm guilty of using `compose` from Redux for non-Redux things (most commonly composing HOCs, be it including React Redux `connect` or not). My reasoning for that has been that it's provided by Redux anyway, and [the documentation][1] I think quite clearly states/suggests that it's just ye olde functional-programming `compose` anyway.

[1]: https://redux.js.org/api/compose#composefunctions

I don't really have a strong opinion on the issue topic as a whole, but at least to me it's made sense to just reach for it as a general-purpose `compose` in a project that already depends on Redux anyhow. One potential concern is that if the function remains exported as `compose` but the TS typesig restricts it to Redux uses (of composing store enhancers etc), then that might add friction to migrating a project to TypeScript from vanilla JS.

jedmao:
Honestly, I think compose should be an external dependency from Redux, but all the compose functions on npm seem to be not what we want, let alone typed properly.

Everyone seems to have a different idea about the implementation of compose. Looks like they use promises pretty commonly too.

markerikson:
It's part of our current API, everyone depends on it, and we're not going to remove it for v5.

timdorr:
Why not? It's the perfect time to make a breaking change.

cellog:
I'd recommend deprecating it for a while before removing it

cellog:
or (sorry to not have thought of this before I hit send), re-export it from somewhere else and deprecate it. Make the external dep a peer dep, so that when it is finally removed, the user still has the peer dep

markerikson:
I _strongly_ oppose removing it.  

The compose function we have works.  It's just a question of the best way to type it.

There's no reason to remove it, and we would break tens of thousands of applications by removing this function.

I'm fine with fiddling with type declarations for v5, but the point of the conversion was that we would _not_ change any actual JS functionality.

nickmccurdy:
To be honest, I don't think we have enough TypeScript contributors to keep the compose types up to date at the level that it should be, especially considering that new TypeScript will add advanced new type features that we should adapt to. Compose should be removed so we can get the best types from another library. Simply adding types to our existing JS code isn't enough, TypeScript won't be able to infer or verify the types.

timdorr:
I'm not saying remove the function itself, just rename it to its specific use (`composeEnhancers`) and type it as such.

cellog:
OK. Here's what I propose: we do nothing. The types, as they are, work fine. The only goal here was to improve them. AFAIK, no one has had trouble with the existing types.

Perhaps later Typescript will make this easy. Until then, I don't see any obvious benefits to changing things, and many disadvantages.

jedmao:
I know for certain there are people importing `redux` _just_ for its `compose` function. They aren't even using Redux! I think that's a pretty good indicator that it doesn't _solely_ belong in the `redux` package. If `redux` re-exports it, that makes more sense, but then it turns into "another dependency," which is [a different conversation](https://github.com/reduxjs/redux/issues/3567) entirely.

This conversation has surely turned into a 🐇 🕳 , so I'm with @cellog. I'll be closing my PRs.

nickmccurdy:
We should delete compose to let it be maintained by other packages, then Redux users can install the other packages and non-Redux users won't need to install Redux just for compose.

markerikson:
**NO.**

Again, I am not going to break the Redux public API, even in a major, just for the sake of some notion of "letting people bring their own `compose` function".

ematipico:
Why removing an API just because a tool (TypeScript) is not able to do what's needed? 
There should be a good reason to deprecate => remove an API. Reasons around the library itself, not a tool. Right? Or did I miss something?

jedmao:
The argument to extract the compose function outside of the library has nothing to do with TypeScript, which means it's out of scope from this thread.

markerikson:
Here's my stance atm:

- We are ***not*** removing `compose`
- We are ***not*** renaming it
- We are ***not*** making it an error to pass in 1 argument
- If there are any other legitimate improvements we can make to the types ***without pulling in other libraries***, I'm interested, whether it be based on multiple overloads or some kind of mapped tuple whatevers

Otherwise, we'll just stick with what we've got and I'll close this.

timdorr:
I'm just going to close this regardless. There are some potential avenues to explore, but it doesn't sound like this PR is one of them.

PR #3570, opened 2019-09-09
Simplify applyMiddleware

jedmao:
Breaking up #3566 into smaller pieces.

This PR simplifies the types of `applyMiddleware` by replacing all the overloads with a single generic function signature that _infers_ the union of any/all `Dispatch` types provided into a single `StoreEnhancer`. This implementation has the added benefit that you can provide as many middlewares as you want w/o maxing out at just 5.

### DIFF
--- a/src/applyMiddleware.ts
+++ b/src/applyMiddleware.ts
@@ -1,8 +1,10 @@
 import compose from './compose'
 import { Middleware, MiddlewareAPI } from './types/middleware'
-import { AnyAction } from './types/actions'
-import { StoreEnhancer, StoreCreator, Dispatch } from './types/store'
-import { Reducer } from './types/reducers'
+import {
+  Dispatch,
+  StoreEnhancer,
+  StoreEnhancerStoreCreator
+} from './types/store'
 
 /**
  * Creates a store enhancer that applies middleware to the dispatch method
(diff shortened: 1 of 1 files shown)


src/applyMiddleware.ts
```diff
@@ -1,8 +1,11 @@
...
-import { AnyAction } from './types/actions'
-import { StoreEnhancer, StoreCreator, Dispatch } from './types/store'
-import { Reducer } from './types/reducers'
+import {
+  Middleware,
+  StoreEnhancer,
+  Dispatch,
+  MiddlewareAPI,
+  StoreEnhancerStoreCreator
+} from '.'
```
timdorr: I'd rather keep these as direct imports so readers know where they are actually found in the source code.

src/applyMiddleware.ts
```diff
@@ -23,40 +26,17 @@ import { Reducer } from './types/reducers'
...
-): StoreEnhancer<any> {
-  return (createStore: StoreCreator) => <S, A extends AnyAction>(
-    reducer: Reducer<S, A>,
+export default function applyMiddleware<
+  S = any,
+  M extends Middleware = Middleware
+>(
+  ...middlewares: M[]
+): StoreEnhancer<
+  M extends Middleware<any, any, infer D> ? { dispatch: D } : never,
```
timdorr: Man, I really like `infer`.

timdorr:
Looks good and _much_ easier to read through this way. Thanks, Jed!

src/applyMiddleware.ts
```diff
@@ -23,40 +26,17 @@ import { Reducer } from './types/reducers'
...
-  ...middlewares: Middleware<any, S, any>[]
-): StoreEnhancer<{ dispatch: Ext }>
-export default function applyMiddleware(
-  ...middlewares: Middleware[]
-): StoreEnhancer<any> {
-  return (createStore: StoreCreator) => <S, A extends AnyAction>(
-    reducer: Reducer<S, A>,
+export default function applyMiddleware<
+  S = any,
+  M extends Middleware = Middleware
```
timdorr: Why does this need to be generic? In what case would our middleware not be a `Middleware`? Either we don't need this, or we need to add a test for this generic.

jedmao: If it's not generic, I can't _infer_ off of it; rather, I do get errors if I try to do `typeof middlewares extends Middleware<any, any, infer D>`.

```
Type '(createStore: StoreEnhancerStoreCreator<any, never>) => <S = any, A extends Action<any> = AnyAction>(reducer: Reducer<S, A>, ...args: any[]) => any' is not assignable to type 'StoreEnhancer<never, S>'.
  Type '<S = any, A extends Action<any> = AnyAction>(reducer: Reducer<S, A>, ...args: any[]) => any' is not assignable to type 'StoreEnhancerStoreCreator<never, S>'.
    Type 'any' is not assignable to type 'never'.ts(2322)
```

It will always be a `Middleware`, which is why I said it must `extend Middleware`, but there's no reason someone can't create an interface which extends `Middleware` and use that too.

jedmao: Also, we already have tests against this generic in the existing `applyMiddleware.spec.ts` which pass just fine for some reason; though, I am seeing red underlines in my editor (even before my changes here). I've fixed those red-line errors in other pieces of my work, but once I started breaking things apart, it's a bit harder to compose these things into smaller chunks w/o some overlap.

jedmao:
Let me give a stab at some tests.

jedmao:
Nope, this doesn't work the way I thought. The tests prove that. Side note, it doesn't work this way before my changes either. Should I submit a PR with breaking tests to show you what I mean? Or is my assumption incorrect?

```ts


  it('combines Dispatch types from all middlewares', () => {
    const foo: Middleware<any, any, Dispatch<{ type: 'foo' }>> = ({
      dispatch
    }) => _next => _action => dispatch({ type: 'foo' })
    const bar: Middleware<any, any, Dispatch<{ type: 'bar' }>> = ({
      dispatch
    }) => _next => _action => dispatch({ type: 'bar' })

    const store = createStore(x => x, applyMiddleware(foo, bar))

    store.dispatch({ type: 'foo' }) // OK
    store.dispatch({ type: 'bar' })  // OK
    store.dispatch({ type: 'baz' }) // not OK
  })
```
The part TS doesn't like is `applyMiddleware(foo, bar)`, because `bar` isn't of the same type as `foo`. Need to rethink this a bit, I think.

PR #3577, opened 2019-09-28
Expand storeAPI for middlewares to include the extended store properties

petuomin:
Expands the API to the store the middlewares receive. 

Changes the API creation so that it explicitly removes the `subscribe` and `replaceReducer` from the store instead of explicitly picking `getState`. This allows the middlewares to use the extended properties from the store.

### DIFF
--- a/src/applyMiddleware.ts
+++ b/src/applyMiddleware.ts
@@ -1,7 +1,7 @@
 import compose from './compose'
 import { Middleware, MiddlewareAPI } from './types/middleware'
 import { AnyAction } from './types/actions'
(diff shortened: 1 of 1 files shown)


timdorr:
I'm not sure I agree with this. If there are additional properties being put on the store, then it sounds like an enhancer, not a middleware. What's the use case here?

markerikson:
This was discussed way back in the early days and explicitly decided against.  Don't have the link atm, but I don't see a reason to do this.

I'm going to pre-emptively close this.  That said, if you have a specific use case, please let us know here in this thread, and we can discuss alternatives.

PR #3578, opened 2019-09-29
Fixed Grammatical Errors.

walter-ind:
<h1>Description</h1>

1. It seems that the verb <strong>support</strong> does not agree with the subject. So, the verb form was changed to <strong>supports</strong>.

2. The preposition <strong>out</strong> was looking unnecessary in this context. So, I have removed it.

3. The preposition <strong>with</strong> looks unnecessary in this context. So, I have removed that also.

Thanks a lot!!!


markerikson:
Appreciate the PR, but the wording is correct as-is.

walter-ind:
> Appreciate the PR, but the wording is correct as-is.

Okay! It was nice going through your repository. 

Thankyou once again.

PR #3582, opened 2019-10-08
redux action type including init/replace types

Mazuh:
related to https://github.com/reduxjs/redux/issues/3580

not sure if it's worth it, but I was here studying typescript/redux and thought this might help :) idk

### DIFF
--- a/src/createStore.ts
+++ b/src/createStore.ts
@@ -8,7 +8,7 @@ import {
   Observer,
   ExtendState
 } from './types/store'
-import { Action } from './types/actions'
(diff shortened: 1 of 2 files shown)


markerikson:
Appreciate the PR, but we explicitly don't want to expose the `INIT` action type.  That's internal, and should not be directly depended on by user code.

Not quite sure what @timdorr meant by his comment, exactly.

Mazuh:
I see. If it's intentional then there's no use for this. Closing it. Thanks for the quick feedback, tho.

PR #3583, opened 2019-10-08
increase test cov of reducers test helpers and observable return

Mazuh:
### summary

not so important, but I was running some tests and saw these warnings.

- for some tests, like for `dispatchInMiddle`, I saw some helpers unreachable lines (cause exceptions were being tested) and led to these uncovered warning.
- another helper, `id`, was function with an unused branch, that was an unused default value for a param
- and finally, there was an uncovered function on store `observable` return

### before

Screenshot from 2019-10-08 14-12-25

### now

Screenshot from 2019-10-08 14-41-10

(there's still an uncovered line for a warning shown on redux.js bundle)

### DIFF
--- a/test/createStore.spec.ts
+++ b/test/createStore.spec.ts
@@ -470,9 +470,7 @@ describe('createStore', () => {
       )
     ).toThrow(/may not dispatch/)
 
-    expect(() =>
-      store.dispatch(dispatchInMiddle(() => {}))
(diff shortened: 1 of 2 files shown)


timdorr:
Can you run Prettier on your code? That will get it to pass CI.

Mazuh:
Thx, CI is passing now!

timdorr:
Thanks!

PR #3585, opened 2019-10-16
Added currentState arg in listener function

uxitten:
We can get current state from listener function:

```js
//log.js
function log(currentState){
  console.log(currentState);
}
export default log

//index.js
import log from 'log';

store.subscribe(log);
```

And we don't need import `store` in `log.js` file

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ let store = createStore(counter)
 // Normally you'd use a view binding library (e.g. React Redux) rather than subscribe() directly.
 // However it can also be handy to persist the current state in the localStorage.
 
-store.subscribe(() => console.log(store.getState()))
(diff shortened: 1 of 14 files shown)


uxitten:
@markerikson I was not convinced

markerikson:
I appreciate the time and effort put into the PR, but it's not something we plan to change.

uxitten:
@markerikson You're welcome

timdorr:
Just increase the listener's order like this:
```js
// log.js
const log = store => () => {
  console.log(store.getState())
}

export default log

// index.js
import log from 'log'

store.subscribe(log(store))

uxitten:
@timdorr Awesome, thank's

PR #3590, opened 2019-10-28
Code refactoring

KirankumarAmbati:
1. Removed unused code
2. Converted class components to functional components


timdorr:
Thanks, but I don't think this is really necessary. We're also looking to move to Docusaurus 2 soon-ish, so this will all get overwritten anyways. 

If you have some time, we could use help with rewriting the docs. Check out the other issues for that.

PR #3604, opened 2019-10-29
fix: remove dead link for tutorial in docs: learning resources

cchanxzy:
Unfortunately the link for this tutorial seems to be dead, and there doesn't seem to be a replacement.

Issue: https://github.com/reduxjs/redux/issues/3562


markerikson:
Looks like it actually got moved to https://github.com/pshrmn/notes/blob/master/redux/redux.md .

cchanxzy:
oh, that's good then! In that case I'll change the PR so it's for updating the URL instead.

PR #3627, opened 2019-11-18
improve performance of ensureCanMutateNextListeners

artalar:
Small improvements for preventing unnecesary calls of `ensureCanMutateNextListeners`

### DIFF
--- a/src/createStore.ts
+++ b/src/createStore.ts
@@ -102,6 +102,7 @@ export default function createStore<
   let currentListeners: (() => void)[] | null = []
   let nextListeners = currentListeners
   let isDispatching = false
+  let dispatchNesting = 0
 
   /**
    * This makes a shallow copy of currentListeners so we can use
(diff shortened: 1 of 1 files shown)


artalar:
CI failing is not my fault 👀

markerikson:
Fixed the formatting issue on `master`.

How exactly does the change here improve performance?  What benchmarks have you run to see if there's an improvement?

artalar:
Currently `currentListeners.slice()` calls at each `dispatch`, but it is unnecesary if it dispatch is not nesting. It is correct?

artalar:
@markerikson I writing some simple [benchmarks.](https://github.com/artalar/redux/blob/artalar/perf-improves-bench/bench/index.js) It checks exactly the case solved by this PR - subscribe / unsubscribe after each dispatch.

изображение

timdorr:
Finally sat down to look through this. I don't think it really warrants the complexity increase for a 0.3 millisecond speed boost. I'd rather make things easier to read in this case. Thanks anyways for looking at it!

PR #3631, opened 2019-11-23
Dropping symbol-observable dependency, realizing internal implementation

### LINKED ISSUE #3567: Investigate dropping all external dependencies
We currently have two external dependencies:  https://github.com/zertosh/loose-envify and https://github.com/benlesh/symbol-observable .

`loose-envify` appears to be used in `package.json`, but only as some kind of polyfill / build plugin for Browserify:

https://github.com/reduxjs/redux/blob/85024d431ca765c843561be6792edeb7890d5fce/package.json#L102-L106

`symbol-observable` is used in `createStore` as a polyfill for `Symbol.observable`, to allow compatibility for treating the Redux store as an observable:

https://github.com/reduxjs/redux/blob/85024d431ca765c843561be6792edeb7890d5fce/src/createStore.ts#L1

https://github.com/reduxjs/redux/blob/85024d431ca765c843561be6792edeb7890d5fce/src/createStore.ts#L331-L333

I would like to strongly consider dropping both of those dependencies if at all possible.

I question whether we need to have any specific support for Browserify in here.  If there's a build step that needs to be done, let the end users take care of it.  No reason to have an installation dep just for that.

`symbol-observable` boils down to [this 15-line file for the polyfill](https://github.com/benlesh/symbol-observable/blob/e56e765476fd96dcb430fd44d936bc543d4d97ba/es/ponyfill.js) plus another [15 lines for figuring out what the global object is](https://github.com/benlesh/symbol-observable/blob/e56e765476fd96dcb430fd44d936bc543d4d97ba/es/index.js).

@davidkpiano suggests [rolling our own `symbol-observable` if really needed](https://twitter.com/DavidKPiano/status/1170185751901794304).

Finally, we are currently talking about [adding `ts-toolbelt` as a dependency for its `compose` function](https://github.com/reduxjs/redux/pull/3563).  I would like us to seriously investigate whatever other alternatives are out there to adding a dep, even if it's one that ultimately compiles away.

r3dm1ke:
Partialy fixes #3567 
Removed `symbol-observable` from `package.json` and added a polifyll to `/src/utils/symbol-observable`, similar to RxJs implementation.

_Sorry for the mess, I do not know what is wrong with my pc and why it keeps messing up the formatting._

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7102,11 +7102,6 @@
         "has-flag": "^3.0.0"
       }
     },
-    "symbol-observable": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/symbol-observable/-/symbol-observable-1.2.0.tgz",
-      "integrity": "sha512-e900nM8RRtGhlV36KGEU9k65K3mPb1WV70OdjfxlG2EAuM1noi/E/BaW/uMhL7bPEssK8QV57vN3esixjUvcXQ=="
-    },
     "symbol-tree": {
       "version": "3.2.4",
       "resolved": "https://registry.npmjs.org/symbol-tree/-/symbol-tree-3.2.4.tgz",
--- a/package.json
+++ b/package.json
@@ -50,8 +50,7 @@
     "examples:test": "cross-env CI=true babel-node examples/testAll.js"
   },
   "dependencies": {
-    "loose-envify": "^1.4.0",
-    "symbol-observable": "^1.2.0"
+    "loose-envify": "^1.4.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.6.0",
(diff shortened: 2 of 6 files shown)


timdorr:
I'm not seeing a package-lock.json update. Can you include that?

timdorr:
Are you using a recent version of npm? I would bump up to the latest (6.13, I think?), as the package-lock.json should have sha512 hashes in it.

r3dm1ke:
> 
> 
> Are you using a recent version of npm? I would bump up to the latest (6.13, I think?), as the package-lock.json should have sha512 hashes in it.

`npm -v` reports 6.13.1
I did some research, and it seems there is a bug in npm with no known solution yet. There is a workaround to do `npm cache clean -f` and regenerate `package-lock.json` but it solves it only temporarily, until next generation. Do you want me to proceed with it?

timdorr:
I fixed it up for you now.

Thanks for the PR!

PR #3632, opened 2019-11-23
Fix rollup-plugin-babel usage

Andarist:
Most likely when you have migrated to TS you have missed that Babel has actually stopped transpiling files - because by default it only operates on .js, .jsx, .es6, .es, .mjs extensions and .ts is not one of them.

So this fixes dist files from being es6+ to es5.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -6,6 +6,7 @@ import { terser } from 'rollup-plugin-terser'
 
 import pkg from './package.json'
 
+const extensions = ['.ts']
 const noDeclarationFiles = { compilerOptions: { declaration: false } }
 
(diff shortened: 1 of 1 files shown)


rollup.config.js
```diff
@@ -6,7 +6,8 @@ import { terser } from 'rollup-plugin-terser'
 
 import pkg from './package.json'
 
-const noDeclarationFiles = { compilerOptions: { declaration: false } };
+const extensions = ['.ts', '.js']
```
timdorr: We actually have no .js files anymore. Let's keep it to just .ts.

Andarist: ok, changed it to just `.ts`

timdorr:
Awesome. Thanks!

PR #3634, opened 2019-11-24
Docs: remove bad import

staff0rd:
## Checklist

https://redux.js.org/recipes/usage-with-typescript#type-checking-reducer

## What is the problem?
`SystemActions` import is not required, nor is it in `types.ts`.
Remove `SystemActions` from `import`

### DIFF
--- a/docs/recipes/UsageWithTypescript.md
+++ b/docs/recipes/UsageWithTypescript.md
@@ -190,11 +190,7 @@ Type checked system reducer:
 ```ts
 // src/store/system/reducers.ts
 
-import {
-  SystemState,
-  SystemActionTypes,
-  UPDATE_SESSION
-} from './types'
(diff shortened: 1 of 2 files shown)


markerikson:
Good catch.  Looks like the file isn't formatted right, though.  If you can fix that, I can merge it.

staff0rd:
how can i see what the actual format error is? This PR just deletes one line...

markerikson:
Yeah, you wouldn't think that would cause any errors :)

Re-run Prettier with `npm run format`.  You can confirm it's okay by running `npm run format:check` and verifying no files are listed as badly formatted.

staff0rd:
@markerikson it's odd that that test appears in the second commit?

timdorr:
No worries. The change there is minor, so I'm good with it. Thanks for the docs fix!

PR #3636, opened 2019-11-24
Add PURE annotations

Andarist:
Similar to what got added to react-redux - it lets minifiers to remove those 3 small calls when `redux` stays unused in the bundle.


timdorr:
Thanks!

PR #3641, opened 2019-11-27
fix: index.ts error: Cannot find name 'process'

heyanfeng:
fix: index.ts error: Cannot find name 'process'


timdorr:
Hmm, I haven't run into this myself, but it can't hurt. Thanks!

Issue #3595, opened 2019-10-29
[Docs Rewrite] Meta-Issue: Tutorials


markerikson:
This is a tracking issue to cover all tasks related to rewriting and updating the "Tutorials" section.

## Tasks

- [x] Define goals and outline for new tutorial sequence (refs: #2590 , #3313 , #2933)
- [x] Come up with better sample apps for tutorials
- [x] Restructure docs sidebar: add "Tutorials" category, move "Basic" and "Advanced" under that
- ~~Add "Barebones" tutorial~~
- [x] Rewrite "Basic" tutorial
  - [x] Rewrite "Usage with React" page completely
- ~~Renamed "Advanced" to "Standard" and rewrite~~
  - [x] Move "Middleware" page to "Thinking in Redux > Design"
- [x] Add "Real World" tutorial
- [x] Add a "Next Steps" page (#2626)

markerikson:
I was just browsing the analytics for the Redux docs.  Here's the top 10 pages for the last week in pageviews:

1. /: 27.6K
2. Getting Started: 25.9K
3. Actions: 17.1K
4. Reducers: 15.5K
5. Usage w/ React: 14.7K
6. Style Guide: 12.8K
7. Tutorial: 12.1K
8. Example: 8.7K
9. Async Actions: 8.5K
10. Store API: 8.1K

So, this tells me that we've got a lot of folks actively learning React, and we should prioritize rewriting the tutorials.  (And, that we need to get the tutorials _right_.)

Before we can do that, we need to have an idea how we're going to rewrite them.


## Previous Thoughts

Here's my prior thoughts from #3313 : 

> I want to have a series of tutorials that add increasing levels of complexity.  These might be based on or tied into the existing projects in the "examples" folder.  The exact project types and implementations should be carefully thought out to help get across the specific concepts we want to teach for each one.
> 
> I'd like to keep these tutorials as focused as possible.  That means skipping stuff like "three-phase actions", writing action constants inline instead of as separate variables and files, etc.  I also want to start with as much stuff in a single file to begin with, and only split it into multiple files later, at which point I want to use "feature folders" instead of "folder-by-type".  I also want to hold off on showing action creators right away.  (Related: I ran [a Twitter poll on what folder structure the tutorials and examples should show by default](https://twitter.com/acemarke/status/1078332610609401856).)
> 
> Possible sequence:
> 
> - **Barebones**
>     - Goal: show the absolute basic mechanics of the store API, no other libs or anything
[...]

PR #3642, opened 2019-11-27
#3595, Barebones Tutorial

richardcrng:
# Barebones tutorial (part of #3595 )

## PR Type

* Adds a new category, '**Tutorials**'
* Adds a new page under that category, '**Barebones Tutorial**'

## Checklist

  - [[Docs Rewrite] Meta-Issue: Tutorials](#3595 )

Tutorial

Newcomers to Redux (who might also be relatively new to Javascript)

Basic Javascript

They should know:
* What an action is
* What a reducer is, in the context of Redux
* How they could use an action payload
* What an action creator is
* How to create a Redux store
* How to use `store.getState()` and `store.dispatch`

* Actions are plain Javascript objects that have a `type` property
* Reducers have the signature `(state, action) => newState` and should cause no mutations or side-effects
* An action can be given a `payload` to provide additional information beyond `type` to a reducer
* An action creator is a function that returns an action
* `createStore` takes a reducer function as an argument and returns a store which uses that reducer
* `store.getState()` retrieves the current state held by the store
* `store.dispatch(action)` updates the store's state by passing its current state and `action` to the store's reducer

## Other notes

This page is quite long. I think that, if used, it would be more suited to having 'Barebones Tutorial' as a subcategory under 'Tutorials', with a single page then dedicated each to Actions, Reducers, Payloads, Action creators, and the Store, but wanted to check what thoughts were on that before I made that change.

It also might not be the style of thing that you want on the docs - I think the material is probably more naturally suited to a video-ish thing. No offence taken if that's the decision, as I can reuse this elsewhere.

### DIFF
--- a/docs/tutorials/BarebonesTutorial.md
+++ b/docs/tutorials/BarebonesTutorial.md
@@ -554,13 +554,207 @@ elevatorReducer(
 )
 ```
 
-Or, more typically, we can create the action at the point that we need it, i.e. as we're passing it into the reducer:
+Or, more typically, we can create the action at the point that we need it, i.e. call the action creator as an argument of the reducer:
 
 ```js
 elevatorReducer(
+  { floorNumber: 4, passengers: ['Arthur the Aadvark', 'Bugs Bunny', 'Charlie Chaplin', 'Dastardly Dan'] },
+  boardPassenger('Eddie the Eagle') // executing this action creator returns an action!
+)
+```
+
+Isn't that much nicer?
+
+```diff
+elevatorReducer(
+  { floorNumber: 4, passengers: ['Arthur the Aardvark', 'Bugs Bunny'] },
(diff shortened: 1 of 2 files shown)


markerikson:
Thanks for putting this together!  Given the holiday week and some upcoming business travel, I may not have time to look through it right away, but I'll make sure I get to it in the near future.

richardcrng:
@markerikson, can I nudge you on this? (I'll quite happily reuse this elsewhere with no offence taken if you don't want it for the official docs)

markerikson:
I really appreciate the PR and the effort you've put into this, but I'd like to hold off on it for now.  Given how critical the tutorial section is, I think we need to put some serious thought and planning into the exact structure, sequence, and content of the tutorials as a whole, so that it all combines into a coherent whole.

It's worth keeping this PR around as a reference for when we do that planning, though.

timdorr:
Closing since we have the new essentials guide.

PR #3648, opened 2019-12-03
Fix "ReducerFromReducersMapObject" typing

ngolin:
```ts
import { Reducer, ReducerFromReducersMapObject } from './reducers'

type TodoReducer = Reducer<
  string[],
  { type: 0; id: number } | { type: 1; text: string }
>
type FilterReducer = Reducer<number, { type: 2; filter: number }>
type MapObject = { todo: TodoReducer; filter: FilterReducer }
type State = {
  todo: string[]
  filter: number
}
type Action =
  | { type: 0; id: number }
  | { type: 1; text: string }
  | { type: 2; filter: number }

function reducersTest(_reducer: Reducer<State, Action>) {}
let reducer!: ReducerFromReducersMapObject<MapObject>
// Error: Argument of type 'Reducer<string[], { type: 0; id: number; } | { type: 1; text: string; }> | Reducer<number, { type: 2; filter: number; }>'
// is not assignable to parameter of type 'Reducer<State, Action>'
reducersTest(reducer)
```

Please code review and merge PR.

### DIFF
--- a/src/createStore.ts
+++ b/src/createStore.ts
@@ -6,7 +6,8 @@ import {
   StoreEnhancer,
   Dispatch,
   Observer,
-  ExtendState
+  ExtendState,
+  UnpackedState
 } from './types/store'
 import { Action } from './types/actions'
 import { Reducer } from './types/reducers'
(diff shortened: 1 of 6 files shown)


timdorr:
I moved your code from the issue in here, because we don't need two things open to track one problem.

timdorr:
Looking through this now, I'm not sure what this is for. `ReducerFromReducersMapObject` is an internal type solely for `combineReducers`. It's not meant to be used in other contexts. 

Thanks, but this doesn't seem to be necessary.

PR #3658, opened 2019-12-14
Add actual example implementation for StoreEnhancer to TS extraMethods test

Andarist:
I was looking through tests to see how this thing can be typed correctly and found only this implementation-less test which was already a little bit helpful for me, but with the actual implementation would be even more... so I've added one for future explorers 😉


timdorr:
Thanks!

PR #3663, opened 2019-12-23
Fix API reference link

jmm:
---
---

^^ This content was auto populated.

## Checklist

  - _link issue here_

README

## What is the problem?

The current API reference link doesn't work.

Updates the link to one that works, but I don't know if the old is supposed to and something else is wrong (not doing a redirect or loading an index it should be).


timdorr:
Thanks!

Issue #3674, opened 2020-01-07
[Docs Rewrite] Create "Quick Start" page


markerikson:
## What docs page is being added or updated?

- **Section**: Introduction
- **Page**: Quick Start

### What kind of content category is this page (tutorial, how-to, explanation, reference)?

The "Quick Start" page is a form of tutorial, but intended to demonstrate the "fastest" / "easiest" / "correct" way to use Redux without explaining _why_ things work this way.  

### Who is the intended target audience?

Beginners who know nothing about Redux, yet.

#### What knowledge are we assuming they have?

- JS syntax: ES6, `async/await`, object spreads
- React: components, hooks
- Other: HTTP / AJAX, `fetch`

### What are the intended results or takeaways from reading this page?

Readers should be able to mimic the demonstrated patterns to create a basic React+Redux app that does some CRUD work.

### What is the most critical info they should learn?

- Redux is a tool for managing state outside of your React components
- The Redux Toolkit package is the simplest and recommended way to use Redux
- Key React+Redux APIs (`configureStore`, `createSlice`, `useSelector` / `useDispatch`)
- Immutability is vital, but either requires hand-written updates or use of RTK+Immer "magic"

## Notes and Suggestions

- Need to make it as "hands-on" and interactive as possible
- Use https://jsonplaceholder.typicode.com/ as a fake but working backend.  (Note: `PUT / POST / DELETE` don't actually change anything.)
- Add "Prerequisites" section at the start of the page.  Don't waste time explaining other things here.
- Key references / examples:
    - https://www.apollographql.com/docs/react/get-started/#apollo-boost
    - https://www.taniarascia.com/getting-started-with-react/
    - https://www.taniarascia.com/getting-started-with-vue/
    - https://vuejs.org/v2/guide/
    - https://nextjs.org/learn/basics/getting-started
- Styling: https://taniarascia.github.io/primitive/

markerikson:
So hey, this is done now :)

PR #3675, opened 2020-01-07
[Docs Rewrite] Create the "Quick Start" page

markerikson:
## PR Type

Adds the new "Quick Start" page

## Checklist

  - #3674 

- **Section**: Introduction
- **Page**: Quick Start

The "Quick Start" page is a form of tutorial, but intended to demonstrate the "fastest" / "easiest" / "correct" way to use Redux without explaining _why_ things work this way.  

Beginners who know nothing about Redux, yet.

- JS syntax: ES6, `async/await`, object spreads
- React: components, hooks
- Other: HTTP / AJAX, `fetch`

Readers should be able to mimic the demonstrated patterns to create a basic React+Redux app that does some CRUD work.

- Redux is a tool for managing state outside of your React components
- The Redux Toolkit package is the simplest and recommended way to use Redux
- Key React+Redux APIs (`configureStore`, `createSlice`, `useSelector` / `useDispatch`)
- Immutability is vital, but either requires hand-written updates or use of RTK+Immer "magic"

## Status

Hacked together this initial skeleton of a page.  So far I've done:

- Figured out how to use MDX with Docusaurus
- Added reusable components for `<Note>` (including multiple styles based on the Cypress.io docs) and `<DetailedExplanation>`
- Created an empty "Quick Start" page
- Added some initial "what is this and why would I use it?" content
- Added instructions on creating a project

I got stuck with some yak shaving when I realized that using the `diff` highlighting format leaves the `+-` diff signs as copyable characters, and ideally I'd like them _not_ copied or selectable.  I tried playing with `user-select: none` and the `:first-letter` selector, but it's not quite doing what I want.

Putting this up for ongoing feedback as we work through this.

/cc @taniarascia , @bpas247

### DIFF
--- a/docs/introduction/QuickStart.mdx
+++ b/docs/introduction/QuickStart.mdx
@@ -48,6 +48,12 @@ Redux does this by asking you to:
 - Write separate functions called **_"reducers"_** that know how to update the state inside the store
 - Follow specific rules for how your code can update and read the state
 
+<Note type="error">
+
+**TODO** needs more of a sales pitch. what problems does this solve? how does it help, especially vs plain React?
+
+</Note>
+
 ### What does Redux include?
 
 The core Redux library consists of a few API functions. The most important API is the `createStore` function.
--- a/docs/introduction/QuickStart.mdx
+++ b/docs/introduction/QuickStart.mdx
@@ -1,0 +1,8 @@
+---
+id: quick-start
+title: Quick Start
+sidebar_label: Quick Start
+hide_title: true
+---
+
+# Quick Start
--- a/docs/introduction/QuickStart.mdx
+++ b/docs/introduction/QuickStart.mdx
@@ -5,68 +5,173 @@ sidebar_label: Quick Start
 hide_title: true
 ---
 
-import { TestComponent } from '../components/TestComponent'
+import Tabs from '@theme/Tabs'
+import TabItem from '@theme/TabItem'
+
 import { DetailedExplanation } from '../components/DetailedExplanation'
 import { Note } from '../components/Note'
 
-export const TC2 = ({ children }) => {
-  return <div style={{ color: 'red' }}>{children}</div>
-}
-
 # Quick Start
-
-Test content here 2
-
-<TestComponent />
-
-<TC2>
-
-- **Bullet 1**
-- Bullet 2
-
-</TC2>
-
-<DetailedExplanation>
-
-Test content
-
-```js
-// code block here
-const testFunc = () => {
-  console.log('blah!')
-}
-```
-
-</DetailedExplanation>
-
-A note:
 
 <Note>
 
-**Note**: Info 1
+### What You'll Learn
+
+- What Redux is, and why you should use it
+- How Redux relates to React
+- How to set up and start using Redux quickly
+- What a small working React+Redux app looks like
+
+### Prerequisites
+
+There are several things you should know in advance before learning Redux:
+
+- Basic familiarity with HTML & CSS, and the DOM.
+- Basic knowledge of JavaScript and programming.
+- Familiarity with ES6 syntax and features.
+- Basic knowledge of React, including components, props, and state.
+- Node.js and `npm` installed globally.
 
 </Note>
 
-<Note type="success">
+## What Is Redux?
 
-**Note**: Success 1
+**Redux** is a predictable state container for JavaScript apps. It acts as the data layer inside an application.
 
-</Note>
+It helps you write applications that behave consistently and can run in different environments (client, server, and native).
+It also helps you to write code that is testable, and enables you to better understand when, where, why, and how the data in your
+application has changed.
 
-<Note type="info">
+Redux does this by asking you to:
 
-**Note**: Info 2
+- Place some data (**_"state"_**) into a special object called a **_"store"_**
+- Write separate functions called **_"reducers"_** that know how to update the state inside the store
+- Follow specific rules for how your code can update and read the state
 
-</Note>
+### What does Redux include?
 
-<Note type="warning">
-
-**Note**: Warning
-
-</Note>
+The core Redux library consists of a few API functions. The most important API is the `createStore` function.
 
 <Note type="error">
 
-**Note**: Error
+**TODO** figure out what else to put here, or if this should be dropped
 
 </Note>
+
+### How does Redux work with a UI?
+
+Redux, by itself, is a plain JavaScript library. You can use Redux together with React, or with any other UI view library.
+
+To help make this possible, there are additional bindings libraries that connect together a Redux store and your UI components. Each different
+UI framework has its own specific bindings library. For example, the [React-Redux library](https://react-redux.js.org) allows your React components to read values from a Redux store,
+and tell the store to calculate an update to the state inside.
+
+### What is Redux Toolkit?
+
+Redux itself is small and unopinionated. We also have a separate addon package called [Redux Toolkit](https://redux-toolkit.js.org), which includes some opinionated defaults
+that help you use Redux more effectively, and includes packages that we think are essential to building a Redux app.
+It's our official recommended approach for writing Redux logic. As with the Redux core, you can use Redux Toolkit with any UI layer.
+
+## Goals
+
+We're going to create a small application with React and Redux together. The app will be a simple employee database, and we'll learn:
+
+- How to set up Redux with React
+- The data flow lifecycle of a Redux app
+- How to write Redux logic that updates data, and shows the updated data in the UI
+- How to create, update, view, and delete users (employees) from the system
+- How to make API calls for each of the above actions
+
+## Setup and Installation
+
+### Create a React Project
+
+We'll start by creating an empty React project. You can do this in one of two ways: creating a local project on your own computer with the
+Create-React-App tool, or creating a project in the CodeSandbox web app.
+
+#### Create a Local Project with Create-React-App
+
+[Create-React-App](https://create-react-app.dev/) is the official tool for creating new React projects.
+
+Create a new React project named `redux-employee-example` by running this command:
+
+```bash
+npx create-react-app redux-employee-example
+```
+
+Then, install the required Redux packages, using either the NPM or Yarn package manager:
+
+<Tabs
+  defaultValue="npm"
+  values={[
+    { label: 'NPM', value: 'npm', },
+    { label: 'Yarn', value: 'yarn', }
+  ]
+}>
+
+<TabItem value="npm">
+
+```bash
+npm add @reduxjs/toolkit react-redux
+```
+
+</TabItem>
+
+<TabItem value="yarn">
+
+```bash
+yarn add @reduxjs/toolkit react-redux
+```
+
+</TabItem>
+
+</Tabs>
+
+#### Create an Online Project with CodeSandbox
+
+[CodeSandbox.io](https://codesandbox.io) is an online IDE that lets you create entire web applications in your browser, without needing to
+install anything locally.
+
+Create a new React project sandbox by browsing to the [CodeSandbox "Create Sandbox" page](https://codesandbox.io/s/) and selecting the official React template.
+
+Once the sandbox has been created, click the "Add Dependency" button on the left, search for `@reduxjs/toolkit`, add it, and then do the same for `react-redux`.
+
+### Create a Redux Store
+
+The React project starts with a single `<App>` component that shows some "hello" text on the page. Our first goal is to create a Redux store instance
+that holds some data inside, and update the `<App>` component to read that data and show it instead.
+
+To create a Redux store, start by updating the `index.js` file to import the `configureStore` function from the `@reduxjs/toolkit` package.
+
+```diff
+import React from "react";
+import ReactDOM from "react-dom";
++import {configureStore} from "@reduxjs/toolkit";
+
+import App from "./App";
+
+const rootElement = document.getElementById("root");
+ReactDOM.render(<App />, rootElement);
+```
+
+In order to actually create a Redux store, we need to define our first "reducer" function. Reducers are functions that know how to
+calculate a new state value when asked. A Redux store must always be given a single reducer function when it is created, which we call the
+"root reducer" to say that it is the starting point for any other reducer functions inside.
+
+For now, we can create a small root reducer function that always returns a string, like this:
+
+```diff
+import React from "react";
+import ReactDOM from "react-dom";
+import {configureStore} from "@reduxjs/toolkit";
+
+import App from "./App";
+
++function rootReducer() {
++    return "Hello Redux!";
++}
+
+
+const rootElement = document.getElementById("root");
+ReactDOM.render(<App />, rootElement);
+```
--- a/docs/introduction/QuickStart.mdx
+++ b/docs/introduction/QuickStart.mdx
@@ -5,4 +5,68 @@ sidebar_label: Quick Start
 hide_title: true
 ---
 
+import { TestComponent } from '../components/TestComponent'
+import { DetailedExplanation } from '../components/DetailedExplanation'
+import { Note } from '../components/Note'
+
+export const TC2 = ({ children }) => {
+  return <div style={{ color: 'red' }}>{children}</div>
+}
+
 # Quick Start
+
+Test content here 2
+
+<TestComponent />
+
+<TC2>
+
+- **Bullet 1**
+- Bullet 2
+
+</TC2>
+
+<DetailedExplanation>
+
+Test content
+
+```js
+// code block here
+const testFunc = () => {
+  console.log('blah!')
+}
+```
+
+</DetailedExplanation>
+
+A note:
+
+<Note>
+
+**Note**: Info 1
+
+</Note>
+
+<Note type="success">
+
+**Note**: Success 1
+
+</Note>
+
+<Note type="info">
+
+**Note**: Info 2
+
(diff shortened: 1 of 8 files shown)


docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+### What does Redux include?
+
+The core Redux library consists of a few API functions. The most important API is the `createStore` function.
+
+<Note type="error">
+
+**TODO** figure out what else to put here, or if this should be dropped
+
+</Note>
```
nickmccurdy: Personally I don't think this section is needed in a quick start, since it's specifically for getting something quickly rather than overviewing the whole API, which I think the rest of the docs introduce well enough

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,177 @@
...
+  values={[
+    { label: 'NPM', value: 'npm', },
+    { label: 'Yarn', value: 'yarn', }
+  ]
+}>
+
+<TabItem value="npm">
+
+```bash
+npm add @reduxjs/toolkit react-redux
```
nickmccurdy:
```suggestion
npm install @reduxjs/toolkit react-redux
```

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+The React project starts with a single `<App>` component that shows some "hello" text on the page. Our first goal is to create a Redux store instance
+that holds some data inside, and update the `<App>` component to read that data and show it instead.
+
+To create a Redux store, start by updating the `index.js` file to import the `configureStore` function from the `@reduxjs/toolkit` package.
+
+```diff
+import React from "react";
+import ReactDOM from "react-dom";
++import {configureStore} from "@reduxjs/toolkit";
```
nickmccurdy: I find it a little confusing to skip straight from having configureStore to having a reducer. I think this code snippet should add a minimal example call with an empty reducer function, then the next snippet can add the example reducer function.

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,177 @@
...
+import App from "./App";
+
++function rootReducer() {
++    return "Hello Redux!";
++}
+
+
+const rootElement = document.getElementById("root");
+ReactDOM.render(<App />, rootElement);
+```
```
nickmccurdy: I like the idea of introducing the root reducer as soon as the store is configured, but I think it might be more clear to a beginner what the reducer is used for if there's a more real-world example state here. For example, it could have `return { firstName: "Homer", lastName: "Simpson" }` and then later add actions/reducers/slices for updating the name fields, showing the relationship between the initial state and how Redux stores are updated as they're developed.

markerikson: My intent is to have the first step simply be `() => "Hello Redux"` + `const helloText = useSelector(state => state)`, and get that to show up on screen instead of the initial text in `<App>`.  Then, once they've seen the initial connection sequence, we can move on to filling out actual app logic.

markerikson:
Finished the "Setup and Installation" part of the page.

I'm feeling very split on the content so far.  I think what I've got here is decent, but at the same time I'm worried it's too low-level / introductory, and taking too long to get to anything "useful" and "real".

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,337 @@
...
+
+### Prerequisites
+
+There are several things you should know in advance before learning Redux:
+
+- Basic familiarity with HTML & CSS, and the DOM.
+- Basic knowledge of JavaScript and programming.
+- Familiarity with ES6 syntax and features.
+- Basic knowledge of React, including components, props, and state.
+- Node.js and `npm` installed globally.
```
bpas247: should we link to external resources here? In case the end user doesn't know these pre-reqs.

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,337 @@
...
+```
+
+If you look at the console in your browser's DevTools (or in CodeSandbox), you should see that the words `"Hello Redux!"` were printed.
+
+### Set Up the Provider Component
+
+We've got a Redux store with its state inside, but now we need to find a way to use that state inside our `<App>` component.
+
+It's _possible_ that we could move the store into a separate file, import that store into the `App.js` file, and call `store.getState()`. But, as the
+app gets bigger, we'd have to import the store into _every_ file. That's not a good idea.
```
bpas247: I'm not sure how useful this information is to a beginner? Just comes off as more information than needed IMO

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,337 @@
...
+  rootElement
+)
+```
+
+### Show the State in the App Component
+
+The last step for the initial setup is to update our `<App>` component to read the state value from the Redux store.
+
+We've already seen that the React-Redux `<Provider>` component makes the store available to any components inside. To go along with that, React-Redux
+includes a React hook called `useSelector`, which React function components can call to retrieve the state values from the store.
```
bpas247: We should link to the React hooks API docs here. Might also be worth mentioning that React hooks is a pre-req for this page, since the end user could be new to the React hooks API as well.

website/sidebars.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   docs: {
     Introduction: [
       'introduction/getting-started',
+      'introduction/quick-start',
```
bpas247: The placement of this page is a bit awkward. Are we assuming that the end users are going to directly navigate to the "Quick Start" page rather than reading the "Getting Started" page first? Since both of these pages start out with the same information, and have a lot of duplicate information.

Also a bit strange that we go over installation in this page, but then the next page the user might navigate to is the "Installation" page.

docs/introduction/QuickStart.mdx
markerikson: Mmm... yes and no.  My thought was to justify why React-Redux's API is needed at all.

website/sidebars.js
markerikson: Yeah, agree that there's overlap, and that needs to be sorted out.  For now, my focus is on the actual content of this page.

docs/introduction/QuickStart.mdx
markerikson: Yep, I'm just trying to get _a_ list in place for now.

docs/introduction/QuickStart.mdx
bpas247: I guess I'm just worried about this "Quick Start" page having more content than it needs, since in my mind the goal of this page is to provide as minimal information as possibly needed to get started with Redux and React.

If you feel like this content fits that goal then we should definitely keep it 😃

website/package.json
```diff
@@ -6,8 +6,8 @@
     "deploy": "docusaurus deploy"
   },
   "dependencies": {
-    "@docusaurus/core": "2.0.0-alpha.39",
-    "@docusaurus/preset-classic": "2.0.0-alpha.39",
+    "@docusaurus/core": "2.0.0-alpha.40",
+    "@docusaurus/preset-classic": "2.0.0-alpha.40",
```
bpas247: Outta curiosity, is there a specific reason that these are getting version bumps? Might be better to branch out these version bumps to a separate PR.

DylanVann:
I think you should get the users to have some simple working code and get them interested/excited about it before going into a bunch of explanations and prerequisites and stuff.

---

### 1. Create a Create React App app:

```bash
npx create-react-app redux-counter-example
```

### 2. Install `@reduxjs/toolkit-react`

```bash
yarn add @reduxjs/toolkit-react
```

### 3. Paste this example into `index.js`

```tsx
import React from "react"
import { render } from "react-dom"
import {
  Provider
  useSelector,
  useDispatch,
  createAction as createEvent,
  configureStore as createStore,
  createReducer,
} from "@reduxjs/toolkit-react"

const onIncrementClicked = createEvent("incrementClicked")
const onDecrementClicked = createEvent("decrementClicked")

const counterReducer = createReducer(0, builder =>
  builder
    .addCase(onIncrementClicked, state => state + 1)
    .addCase(onDecrementClicked, state => state - 1),
)

const store = createStore({
  reducer: {
    counter: counterReducer,
  },
})

function App() {
  const count = useSelector(state => state.counter)
  const dispatch = useDispatch()
  return (
    <div>
      <button onClick={() => dispatch(onDecrementClicked())}>-</button>
      {count}
      <button onClick={() => dispatch(onIncrementClicked())}>+</button>
    </div>
  )
}

render(
  <Provider store={store}>
    <App />
  </Provider>,
  document.getElementById("root"),
)
```

Open the [Redux DevTools](https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd) and try (there should be some gifs showing how to do this stuff):
- Clicking the buttons to see events.
- Viewing the state changes that happen on the events.
- Using time travel debugging to see earlier states of the app.

---

Basically I think "Quick Start" should be small and quick. It would be fine to have lots of links at the end. Or just a link at the end going to a more in depth tutorial after people are excited about having something working.

The idea of `@reduxjs/toolkit-react` would be to make it so people only have to install one package. It would just look like:

```tsx
export * from 'redux'
export * from 'react-redux'
export * from '@reduxjs/toolkit'
```

It would be nice to have a toggle somewhere on the page to change the code to TypeScript and change the CRA command to use the TS template.

taniarascia:
@DylanVann In general, I agree - less is more. However, instead of "paste the all this code into index.js", I think it would be better to break it into chunks.

We'll be creating a simple counter application. First, make`index.js` file and import all dependencies.

```js
import React from "react"
import { render } from "react-dom"
import {
  Provider
  useSelector,
  useDispatch,
  createAction as createEvent,
  configureStore as createStore,
  createReducer,
} from "@reduxjs/toolkit-react"
```

Next, create two **events** - one for incrementing, one for decrementing.

```js
const onIncrementClicked = createEvent("incrementClicked")
const onDecrementClicked = createEvent("decrementClicked")
```

Then we'll create a **reducer**.

```js
const counterReducer = createReducer(0, builder =>
  builder
    .addCase(onIncrementClicked, state => state + 1)
    .addCase(onDecrementClicked, state => state - 1),
)
```

...etc.

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,337 @@
...
+ReactDOM.render(<App />, rootElement)
+```
+
+In order to actually create a Redux store, we need to define our first "reducer" function.
+
+<Note type="success">
+
+#### Key Concept: "Reducers"
+
+A **_reducer_** is a function that know how to calculate a new state value when asked. A Redux store requires a single reducer function
```
nickmccurdy:
```suggestion
A **_reducer_** is a function that knows how to calculate a new state value when asked. A Redux store requires a single reducer function
```

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,337 @@
...
+when it is created, which we call the "root reducer" to say that it is the starting point for any other reducer functions inside.
+
+Every reducer is called with two arguments: the current `state`, and an object called an **`action`**. We'll look at actions a bit later.
+
+Reducers must always follow certain rules inside. As a quick summary:
+
+- Reducers must always return a state value, never `undefined`
+- Reducers can only decide what the next state value should be by looking at the `state` and `action` arguments
+- Reducers are not allowed to make changes to the `state` argument, or anything else. They can only create the new state value by making copies of the originals, and updating the copies.
```
nickmccurdy: This might be confusing to Redux Toolkit users (and we recommend it to beginners), as mutations are supported via Immer and are recommended in the docs.

docs/introduction/QuickStart.mdx
nickmccurdy: I've seen a lot of beginners try to use vanilla Redux in React apps and they get very confused with how to handle the store. I agree with Mark here that it's important for us to explain why React Redux is helpful right away.

Of course, if we have a guide for non-React users in the future, this wouldn't be necessary.

docs/introduction/QuickStart.mdx
taniarascia: "A reducer is a function that calculates a new state value."

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,337 @@
...
+  return (
+    <div className="App">
+      <h1>{helloText}</h1>
+    </div>
+  )
+}
+```
+
+We import the `useSelector` function from React-Redux, and then create a function called `selectHelloText`. It receives whatever state is in the Redux
+store as an argument. For now, we'll just return that entire state value.
```
nickmccurdy: I've seen beginners frequently return the entire state when they don't need to in `mapStateToProps`, causing unnecessary rerenders and defeating the purpose of this API. I understand that. There's just one value in the store in this case, but I think we should at least note that only the state the component needs should be returned, as Redux optimizes rerenders based on the result.

nickmccurdy:
Sorry I fucked up that last review, those comments are supposed to be part of my suggested changes

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,337 @@
...
+
+Create a new React project sandbox by browsing to the [CodeSandbox "Create Sandbox" page](https://codesandbox.io/s/) and selecting the official React template.
+
+Once the sandbox has been created, click the "Add Dependency" button on the left, search for `@reduxjs/toolkit`, add it, and then do the same for `react-redux`.
+
+### Create a Redux Store
+
+The React project starts with a single `<App>` component that shows some "hello" text on the page. Our first goal is to create a Redux store instance
+that holds some data inside, and update the `<App>` component to read that data and show it instead.
+
```
dceddia: I think it'd be good to show a full working React example before adding the Redux store. Maybe something with state.

Presumably, landing people on this page have used React a little. Ideally they have, anyway. Starting with a React example and "refactoring it to Redux" would:

- help a person who already knows React feel grounded, adding new knowledge on top of old
- help a person who has never seen React realize that they should go learn that first (provide a link to official tutorial) and that Redux builds on top of React

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,337 @@
...
+<Note type="success">
+
+#### Key Concept: "Store"
+
+A Redux **_store_** is a special object that holds some data inside, such as a list of users that was fetched from the server.  
+The data inside the store object is called **_state_**.
+
```
dceddia: Could add some kind of visual here to hammer home the idea that the store holds the state.

Nit below: "**Each** Redux store..." kind of implies that an app might have more than one. "A" or "Every" might be better?

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,337 @@
...
+```
+
+In order to actually create a Redux store, we need to define our first "reducer" function.
+
+<Note type="success">
+
+#### Key Concept: "Reducers"
+
+A **_reducer_** is a function that know how to calculate a new state value when asked. A Redux store requires a single reducer function
+when it is created, which we call the "root reducer" to say that it is the starting point for any other reducer functions inside.
```
dceddia: Maybe skip talking about "root reducers" until later. Get one working first before revealing that there can be a tree of them.

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,337 @@
...
+
+Every reducer is called with two arguments: the current `state`, and an object called an **`action`**. We'll look at actions a bit later.
```
dceddia: Mention here that the reducer "returns the new computed state".

Could also add a few words to describe actions so it's not gnawing at them - "...an object called an **action** that describes how to update the state"

docs/introduction/QuickStart.mdx
dceddia: Yeah this is a good point. Later on, it would be good to show "here's the hard way (ignoring that immer is installed)" and "here's the easy way because of immer".

docs/introduction/QuickStart.mdx
dceddia: If this stays, it'd be good to explain _why_ "That's not a good idea." - it will get messy? hard to manage? but there's an easier way?

docs/introduction/QuickStart.mdx
```diff
@@ -0,0 +1,337 @@
...
+The last step for the initial setup is to update our `<App>` component to read the state value from the Redux store.
+
+We've already seen that the React-Redux `<Provider>` component makes the store available to any components inside. To go along with that, React-Redux
+includes a React hook called `useSelector`, which React function components can call to retrieve the state values from the store.
+
+<Note type="success">
+
+#### Key Concept: "Selectors"
+
+A **_selector_** is a function that takes the Redux store state as an argument, and returns some part of that store state.
```
dceddia: It quite literally "selects" a portion of the state.

markerikson:
Haven't given up on this, but I decided I ought to take a step back and take notes on several excellent React+Redux tutorials (and some totally non-Redux-related tutorials) to get a feel for how they handle explanations first.  Then I'll try to take some inspiration from those and take another stab at this.

docs/introduction/QuickStart.mdx
markerikson: Yep, this is still my #1 concern with trying to explain use of RTK in general, and especially in this page.

markerikson:
Should have closed this a while ago.

PR #3676, opened 2020-01-07
Move correct deep state update example to the top

vcarl:
---
---

## Checklist

  - _link issue here_

Opened this purely from GitHub, can follow up if this there's support for making this change.

- **Section**: Structuring Reducers
- **Page**: Immutable Update Patterns

## What is the problem?

This is a nit, but this is my go-to page to direct folks at for examples of immutable updates (regardless of whether they're using Redux). As it stands right now, the first 2 things they see are broken code, which isn't immediately apparent from a quick skim. This seems likely to introduce some confusion.

It moves the correct example so it's the first thing on the page. Since the other examples don't have subtly-broken variations, I think this is a minimal change that will have a slight positive impact.


timdorr:
Yep, I agree with putting that first. Thanks!

PR #3679, opened 2020-01-14
Fix combineReducers typings.

gilbsgilbs:

Fixes a bug.

It fixes a typing issue I and some other folks have with the `combineReducers` function: https://github.com/reduxjs/redux/issues/2709

## Checklist

  - https://github.com/reduxjs/redux/issues/2709 (it is closed, but many people are still affected)
- [ ] Have the tests been updated to match the changes in the PR?

Unfortunately, I failed to create a simple reproduction example. Still, it happens that TypeScript would fail to infer the proper generic type to `combineReducers` function. In my case, I'm doing something like this:

```ts
interface FooState {…}
interface AppAction extends Redux.Action<string> {}
interface FooAction0 extends AppAction {…}
interface FooAction1 extends AppAction {…}
type FooActions = FooAction0 | FooAction1;

function fooReducer(state: FooState, action: FooActions): FooState {
  return state;
}

// Type error:
// Argument of type 'Reducer<CombinedState<{ fooReducer: FooState; }>, FooActions>' is not assignable to parameter of type 'Reducer<CombinedState<{ fooReducer: FooState; }>, AnyAction>'.
combineReducers({ fooReducer: fooReducer });

// works
combineReducers<{fooReducer: FooState}>({ fooReducer: fooReducer });
```

I would expect TypeScript to infer a generic type automatically.

It removes explicit `any`s which appears to help TypeScript to infer a type. I'm not sure why though :man_shrugging: . Could this be a bug in TypeScript?

More specifically, the first generic type of `ReducersMapObject` is a state which is already `any` by default, and the second generic type is `A extends Action` which is `Action` by default and not `any`. So the fact that the second generic type is `any` confuses TS for some reason.

### DIFF
--- a/src/combineReducers.ts
+++ b/src/combineReducers.ts
@@ -130,7 +130,7 @@ export default function combineReducers<S>(
 export default function combineReducers<S, A extends Action = AnyAction>(
   reducers: ReducersMapObject<S, A>
 ): Reducer<CombinedState<S>, A>
-export default function combineReducers<M extends ReducersMapObject<any, any>>(
+export default function combineReducers<M extends ReducersMapObject>(
   reducers: M
 ): Reducer<
   CombinedState<StateFromReducersMapObject<M>>,
--- a/src/types/reducers.ts
+++ b/src/types/reducers.ts
@@ -36,7 +36,7 @@ export type Reducer<S = any, A extends Action = AnyAction> = (
  *
  * @template A The type of actions the reducers can potentially respond to.
  */
-export type ReducersMapObject<S = any, A extends Action = Action> = {
+export type ReducersMapObject<S = any, A extends Action = AnyAction> = {
   [K in keyof S]: Reducer<S[K], A>
 }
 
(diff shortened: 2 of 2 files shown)


timdorr:
Should ReducerMapObject's second arg default be `AnyAction` instead?

gilbsgilbs:
> Should ReducersMapObject's second arg default be AnyAction instead?

It sounds reasonable to me. I have checked against my local project though and it doesn't improve the type inference in this case.

I also see another usage of `ReducersMapObject<any, any>` in `StateFromReducersMapObject`. Should I also update this one to `ReducerMapObject` without specifiying generics?

gilbsgilbs:
@timdorr I don't want to be "that guy" you know, but is there anything I can do to help getting this merged and released anytime soon? :wink:

jonjaques:
Can confirm that this fixes a regression we experienced recently. Specifically, our thunk dispatch type was conflicting w/ a third party reducer (TS complaining about actions missing `type` & `payload`. 

Thanks @gilbsgilbs!

timdorr:
Only thing this needs is some tests!

timdorr:
I guess I'll (finally) merge this in without tests. Thanks!

Charles-Chiakwa:
Be sure you're using connect from the right package not "http2"

PR #3685, opened 2020-01-23
Fix README.md's broken link

take:


timdorr:
Actually, a small change there to point to the Basic Tutorial page. We lost the index pages on the move to Docusaurus 2.

PR #3687, opened 2020-02-02
Docs CSS fix: show pointer only on clickable part of details

roryokane:
## Checklist

  - no need; it’s a tiny change
  - N/A – I see no lint rules for docs CSS

- **Section**: Style Guide
- **Page**: Style Guide

## What is the problem?

The cursor turns to `pointer` when hovering over non-clickable parts of `details` elements.

In the CSS, make the cursor turn to `pointer` only when hovering over the `summary` element, which is the clickable part.


markerikson:
Thanks!

PR #3697, opened 2020-02-11
inside a reducer function, it is called parameters

KennethKinLum:
a reducer is a function. Inside a function, the passed in values are not arguments. They are parameters.

We call a function by optionally supplying arguments, but once inside of a function, they are called parameters.

Or 
never mutate the arguments passed in.

---
---

## Checklist

  - _link issue here_

- **Section**:
- **Page**:


timdorr:
A reducer is just a function. A function in JavaScript has arguments. This is correct as-is.

Issue #3692, opened 2020-02-07
Add detailed explanation for "Treat Reducers as State Machines"


bolatovumar:
## What docs page is being added or updated?

This bit: https://redux.js.org/style-guide/style-guide/#treat-reducers-as-state-machines

- **Section**: #treat-reducers-as-state-machines
- **Page**: Redux Style Guide

### What updates should be made to the page?

Add detailed explanation for "Treat Reducers as State Machines" section

### Do these updates change any of the assumptions or target audience? If so, how do they change?

No, but it would make it easier to understand with a code example provided.

bolatovumar:
I'm willing to submit a PR for this if acked.

markerikson:
Yeah, this ties into #3598 .  I know @davidkpiano had offered to try writing that page.

davidkpiano:
@bolatovumar I can help review, if you'd like! If I were to write it, it would basically be a distillation of this: https://dev.to/davidkpiano/redux-is-half-of-a-pattern-1-2-1hd7 (ignore the article title)

bolatovumar:
@davidkpiano if you are on it then that's great. Your example from the article is great. Your article what got me to open this issue as the docs weren't anywhere as clear as your writeup.

PR #3700, opened 2020-02-16
Update style-guide.md

davidkpiano:
Add detailed docs for #3692

## PR Type

## Checklist

  - #3692 

https://redux.js.org/style-guide/style-guide/#treat-reducers-as-state-machines

Adding detailed explanation for "Treat Reducers as State Machines"

No

### DIFF
--- a/docs/style-guide/style-guide.md
+++ b/docs/style-guide/style-guide.md
@@ -286,9 +286,83 @@ It's a bit more typing, but it results in the most understandable code and state
 
 ### Treat Reducers as State Machines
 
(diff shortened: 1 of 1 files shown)


markerikson:
Sweet, thanks!

My one stylistic nitpick is that I'm generally trying to move our examples away from hardcoded constants and `switch` statements, but other than that, nice explanation with a good example.

I briefly debated asking for a side-by-side comparison of implementing the same logic via both a "state-first" and "action-first" approach, but we can save that for a longer docs page.  I want to keep the "Style Guide" page shorter and focused overall.

davidkpiano:
> I'm generally trying to move our examples away from hardcoded constants and switch statements

I'd like to do that; how would you suggest I refactor my example with that in mind?

markerikson:
Erm... that's a good question, really.

Problem is, that's primarily a `createSlice` thing, which is of course action-type-oriented, and going for any kind of a state machine abstraction is probably more distracting than helpful for the code example in this specific page.

Totally fine with showing, say, "baby's first FSM abstraction" in the larger docs page (as well as the obvious mentions of the small and full XState libs as full-blown solutions in that page as well).

PR #3707, opened 2020-02-19
Use a proper copyright symbol, en-dash, and current year in the footer

markjaquith:
Currently:

Screen Shot 2020-02-18 at 10 19 53 PM

With this change:

Screen Shot 2020-02-18 at 10 15 37 PM

(The "end-year" is dynamic, using the same code that Docusaurus uses on their own site.)


markerikson:
Thanks!

PR #3708, opened 2020-02-19
Update writing tests page to use React Testing Library

MatanBobi:
---
name: "Updated Documentation Content - Recipes - Writing Tests"
about: Updating content in an existing docs page - Recipes - Writing Tests
---

## PR Type

**Update an existing page [Writing tests](https://redux.js.org/recipes/writing-tests)**

## Checklist

  There is no issue at the moment

- **Recipes**:
- **Writing Tests**:

I changed the code example for React component to use React Testing Library instead of Enzyme, as the React team also suggests to use React Testing Library over Enzyme.
Also changed the class component to a functional component.

No, this is a technical change.


MatanBobi:
@kentcdodds anything you think I should add regarding the explanation of React Testing Library?

kentcdodds:
Yeah, I have done thoughts. I'll comment later today.

docs/recipes/WritingTests.md
```diff
@@ -258,46 +258,38 @@ describe('todos reducer', () => {
 
 A nice thing about React components is that they are usually small and only rely on their props. That makes them easy to test.
 
-First, we will install [Enzyme](http://airbnb.io/enzyme/). Enzyme uses the [React Test Utilities](https://facebook.github.io/react/docs/test-utils.html) underneath, but is more convenient, readable, and powerful.
+First, we will install [React Testing Library](https://testing-library.com/docs/react-testing-library/intro). React Testing Library is built on top of React-DOM and React-DOM/test-utils, but is more convenient, readable, and powerful.
```
kentcdodds: React Testing Library is not built on top of react-dom/test-utils (it only uses `act`) and all it uses from react-dom is the `render` method.

docs/recipes/WritingTests.md
```diff
@@ -258,46 +258,38 @@ describe('todos reducer', () => {
...
 
 ```sh
-npm install --save-dev enzyme
+npm install --save-dev @testing-library/react
```
kentcdodds: I'd also recommend having people install `@testing-library/jest-dom` if you're suggesting people use `jest`.

docs/recipes/WritingTests.md
```diff
@@ -258,46 +258,38 @@ describe('todos reducer', () => {
...
-We will also need to install Enzyme adapter for our version of React. Enzyme has adapters that provide compatibility with `React 16.x`, `React 15.x`, `React 0.14.x` and `React 0.13.x`. If you are using React 16 you can run:
-
-```sh
-npm install --save-dev enzyme-adapter-react-16
-```
-
-To test the components we make a `setup()` helper that passes the stubbed callbacks as props and renders the component with [shallow rendering](http://airbnb.io/enzyme/docs/api/shallow.html). This lets individual tests assert on whether the callbacks were called when expected.
+To test the components, we `render` them into the DOM and pass stubbed callbacks as props, then we assert wheter the callbacks were called when expected.
 
 #### Example
```
kentcdodds: This example is not a connected component. It doesn't have anything to do with redux, so I'm a bit puzzled why it's here.

One other thing I'd note about this page is the section on "Connected Components" should be changed to *not* recommend exporting the unconnected component, and instead recommending utilizing the `wrapper` option for React Testing Library's `render` function. And to take it further, people should make their own render method that does this automatically: https://testing-library.com/docs/react-testing-library/setup#custom-render

docs/recipes/WritingTests.md
```diff
@@ -311,46 +303,41 @@ can be tested like:
...
-  const enzymeWrapper = shallow(<Header {...props} />)
-
-  return {
-    props,
-    enzymeWrapper
-  }
-}
-
 describe('components', () => {
   describe('Header', () => {
```
kentcdodds: Personally I think having these `describe` blocks are useless.

- `components`: yes, I'm obviously testing components in this file
- `Header`: yes, this is the test file for the header, I'm obviously testing the Header component.

I'd recommend ditching `describe` in favor of simply `test`: https://kentcdodds.com/blog/avoid-nesting-when-youre-testing

docs/recipes/WritingTests.md
```diff
@@ -311,46 +303,41 @@ can be tested like:
...
     it('should render self and subcomponents', () => {
-      const { enzymeWrapper } = setup()
+      const mockAddTodo = jest.fn()
+      const { container, getByText, getByPlaceholder } = render(
+        <Header addTodo={mockAddTodo} />
+      )
+      expect(getByTestId(container, 'header')).toBeInTheDocument()
+      expect(getByText('todos')).toBeInTheDocument()
+      expect(getByPlaceholder('What needs to be done?')).toBeInTheDocument()
```
kentcdodds: Here's how I'd write this:

```javascript
// the screen variable comes from the React Testing Library import: https://testing-library.com/docs/dom-testing-library/api-queries#screen
render(<Header addTodo={mockAddTodo} />)
expect(screen.getByText(/todos/i)).toBeInTheDocument()
expect(getByPlaceholder(/what needs to be done/i)).toBeInTheDocument()
```

I'd suggest the same kind of changes to the other tests (note the absense of the test ID query, those aren't recommended: https://testing-library.com/docs/guide-which-query)

But actually I don't see much value in this test by itself. These assertions are covered by the other tests... More on this: https://kentcdodds.com/blog/write-fewer-longer-tests

docs/recipes/WritingTests.md
```diff
@@ -487,8 +474,6 @@ In some cases, you will need to modify the `create` function to use different mo
...
 
-- [Shallow rendering](http://airbnb.io/enzyme/docs/api/shallow.html): Shallow rendering lets you instantiate a component and effectively get the result of its `render` method just a single level deep instead of rendering components recursively to a DOM. Shallow rendering is useful for unit tests, where you test a particular component only, and importantly not its children. This also means that changing a child component won't affect the tests for the parent component. Testing a component and all its children can be accomplished with [Enzyme's `mount()` method](http://airbnb.io/enzyme/docs/api/mount.html), aka full DOM rendering.
+- [React Test Utils](https://reactjs.org/docs/test-utils.html): ReactTestUtils makes it easy to test React components in the testing framework of your choice. Used by React Testing Library.
```
kentcdodds: Just to be clear, the only thing that's used by React Testing Library is the `act` function. RTL does not use anything else from it.

docs/recipes/WritingTests.md
MatanBobi: I sure agree with you that this test has no meaning, It was just a refactor to the test that was already there so I didn't feel comfortable deleting it :)

docs/recipes/WritingTests.md
MatanBobi: The only problem with using `wrapper` is that AFAIK it should be a comopnent, meaning I can't give it an `initialState` for example since it doesn't pass any props to `React.createElement` but it's children.. Am I missing something? It means that I'd probably wouldn't want to place it in a setup file. 
The `wrapper` option is a good option for other providers like shown in the example exactly, but for redux, I'm not sure it's so useful to what people would want to test..
The example in https://testing-library.com/docs/example-react-redux is more suitable and I think it can fit well in the docs here. WDYT?

MatanBobi:
@kentcdodds It sure helps, thanks :)
I left a few comments and made some changes, would appreciate your review.

docs/recipes/WritingTests.md
MatanBobi: btw, thought of changing the `it`s to `test`s but It meant changing the whole document and I'm not sure everyone's onboard

MatanBobi:
Hi @kentcdodds any chance for a quick review of the changes?
Thanks!

docs/recipes/WritingTests.md
```diff
@@ -365,60 +344,37 @@ Consider the following `App` component:
...
+import { Provider } from 'react-redux'
+import App from '../../containers/App'
+import { initialState, reducer } from './reducer.js'
+
+function renderWithRedux(
+  ui,
+  { initialState, store = createStore(reducer, initialState) } = {}
+) {
+  return render(<Provider store={store}>{ui}</Provider>)
 }
```
kentcdodds: Could we recommend the use of the `wrapper` option as well as suggest people follow the instructions in the setup docs here so they don't have to worry about the redux provider in any of their tests? https://testing-library.com/docs/react-testing-library/setup

MatanBorenkraoutNielsen: I was wondering about that, what will be the best way to give it an `initialState` in my test? The only option I could think of is adding an `initialState` as one of the options and pass it to the `AllTheProviders` component.. WDYT? 
Btw, I can also update the Recipe for React-Redux in RTL's docs if you'd like, since I used it to write my example here and it's also outdated.
https://testing-library.com/docs/example-react-redux

MatanBobi: I was wondering about that, what will be the best way to give it an `initialState` in my test? The only option I could think of is adding an `initialState` as one of the options and pass it to `AllTheProviders` component.. WDYT? 
Btw, I can also update the Recipe for React-Redux in RTL's docs if you'd like, since I used it to write my example here and it's also outdated.
https://testing-library.com/docs/example-react-redux

kentcdodds: The `renderWithRedux` you have here is basically the exact same thing you would have as your custom render. So you'd still be able to provide an additional argument. Here's a better example:

```javascript
function render(
  ui,
  {
    initialState,
    store = createStore(reducer, initialState),
    ...renderOptions
  } = {},
) {
  function Wrapper({children}) {
    return <Provider store={store}>{children}</Provider>
  }
  return rtlRender(ui, {wrapper: Wrapper, ...renderOptions})
}
```

MatanBobi: Got it, fixed it :)
Thanks for that!
I can update RTL's docs accordingly to contain the example you wrote here for the React-Redux recipe if you think we should do it.

kentcdodds: That would be super. Thank you!

MatanBobi:
Hi @timdorr / @markerikson , can you please have a look at this one? :)
Thanks!!

joshribakoff:
A lot of the merit of using RTL over Enzyme is about making it an integration test, which is not only entirely possible in Enzyme, but may also be preferable. I think we should show both options & highlight the pros & cons, which is what I was going to show in https://github.com/reduxjs/redux/pull/3799

Can we please make this an unbiased documentation, and avoid "endorsing" one library over the other? 

It is a disservice to the community to "switch it", there are users with tens or hundreds of thousands of tests written in Enzyme, the official answer shouldn't be "rewrite all those tests", when there are perfectly valid approaches to getting the same verification with Enzyme.

joshribakoff:
Also tagging this to this original issue, so others can avoid starting PRs that may partially duplicate or conflict with ongoing work in other's PRs: https://github.com/reduxjs/redux/issues/1462

joshribakoff:
Also this PR updates the "docs", but not the "examples" which could be a point of confusion

kentcdodds:
I think there's value in coming to a consensus on a testing tool and the direction of the community is obvious. The React team itself recommends using Testing Library throughout the documentation and only makes a single mention of enzyme (which was held over from the old docs).

I know I'm biased, but there's an obvious trend in the community and I think it's a good trend. Having both would just present another "choice" for people in a situation where they already have to make so many choices as it is.

joshribakoff:
The callout you're referring to is this? https://reactjs.org/docs/test-utils.html

I don't think that merits erasing Enzyme from the Redux docs.

Some of the merits of using Enzyme are best explained in Micheal Feather's book "working effectively with legacy code". Sometimes, in your programming endeavorers, you may need to "break the rules", for example retrofitting tests on legacy code written by someone who left the company before you joined. Feather's makes a metaphor this is like "scaffolding" at a construction site, its not pretty, but it can be wonders for execution efficiency of teams.

Enzyme has tools to do whatever you want, and RTL doesn't necessarily prevent all the foot guns Kent points out with Enzyme, for example, when using RTL it is still possible to mock things out excessively & write tests that give false confidence & are not resilient to refactors.

I agree that RTL as a whole has done a lot for the community, and there is value in juxtaposing the 2 libraries in the docs, but I think the Redux docs should avoid "forcing" that tool (by erasing Enzyme from the docs). Maybe they could each have their own page, similar to how one would expect Redux docs to showcase both Flow & Typescript examples... This is a classic holy war, and one of the selling points of Redux has always been that its agnostic & portable.

markerikson:
I'm going to side with Kent here. 

Showing one specific tool in the docs does not "erase" any other tools from the ecosystem. However, we are not obligated to have docs pages showing how to use Tool X with every other possible tool that exists out there. Ideally, we focus the docs on the axial tools, patterns, and practices that we currently recommend, and those recommendations may change over time. 

At this point, I would not recommend use of Enzyme. I also would not recommend use of Immutable.js or Flow. The docs content should reflect that. 

There's still hundreds of pages out there that show how to use Enzyme, so it's not like removing it from our docs will limit anyone's ability to use it.

joshribakoff:
> There's still hundreds of pages out there that show how to use Enzyme

My experience is that most people using Enzyme are falling into the anti patterns Kent preaches about. I've wanted to switch large code bases to RTL, but it hasn't always been feasible. I think it would be great to bring more visibility to using the better patterns  RTL popularized, for Enzyme users. After all, the React docs still link to both. My experience has been that most of the info out there about integration testing in React is dominated by RTL users, and it is hard to find info about doing it "properly".

While I also loathe the ecosystem fragmentation in React ecosystem, this sort of just fragments things further into 3rd party blog posts.

Also am I looking at the same data as you guys? https://www.npmtrends.com/enzyme-vs-react-testing-library

Screen Shot 2020-06-14 at 9 12 18 AM

kentcdodds:
Extends that graph to two years and see the trajectory. Those graphs are misleading anyway because inertia is hard to beat. Take a look at download numbers for old tools like getting and gulp and they're still downloaded way more than you might think despite being older.

joshribakoff:
If either of us could see the future, we'd be day trading not arguing testing libraries on GH ;)

If the decision is to endorse RTL & erase Enzyme from the docs, I'm disappointed. Its a blow to the majority of React-Redux users who still use Enzyme. Enzyme's tools can be incredibly powerful, and its unfortunate to outright steer people away from it & to a product that has a commercial aspect to it, being Kent's for profit training videos...

kentcdodds:
The insinuation that I'm only pushing testing library because I want to make money is naive, insulting, and ridiculous.

I'll let the maintainers do whatever they want here. I'm no longer participating in this conversation if insults like that will be allowed.

MatanBobi:
Hi @joshribakoff, It's a shame this PR went towards this direction.
The only thing you're missing here is that @kentcdodds wasn't the one opening or leading me towards opening this PR, I did that on my own with a strong thought the RTL has a better approach to testing react components than enzyme does.
I really believe that this is the right step for this documentation or I wouldn't have taken the time to create this PR.
I'm sad to see that you've decided to write something like that towards someone like @kentcdodds who clearly shows the exact opposite to what your claiming right here each and every day with his help to this community.

joshribakoff:
To be clear I am not attacking Kent Dodds in any way, or at least that is not my intent. While I appreciate all that Kent Dodds does, and am envious, I still find it unfortunate there are *tons* of beginners out there tasked with working on an Enzyme code base, who will now find it harder to find guidance.

Enzyme does not prescribe or recommend any approach, that is a common misconception about RTL vs Enzyme holy war, and exactly why I push back here. You can do the same approach with Enzyme, plus more.

MatanBobi:
> To be clear I am not attacking Kent Dodds in any way, or at least that is not my intent. While I appreciate all that Kent Dodds does, and am envious, I still find it unfortunate there are _tons_ of beginners out there tasked with working on an Enzyme code base, who will now find it harder to find guidance.
> 
> Enzyme does not prescribe or recommend any approach, that is a common misconception about RTL vs Enzyme holy war, and exactly why I push back here. You can do the same approach with Enzyme, plus more.

As @markerikson mentioned, this isn't the only place to find Enzyme documentation at and removing it from here isn't a blow as you wrote IMO..
About "You can do the same approach with Enzyme, plus more", that's exactly the problem. I've been writing Enzyme tests for plenty of times, and had they broke almost in every refactor. The "plus more" is exactly where I don't want the plus. Having one way to do something, guides you towards the right way of doing it. Writing tests with RTL and following the concepts it dictates creates resilient and "user like" tests.
I guess we can go on and on with this talk, and as Kent said, this is the maintainers choice what to do with this PR.
I'd love to see it being pushed and if I need to edit the examples section also, will definitely do.

joshribakoff:
> . Having one way to do something, guides you towards the right way of doing it.

This is where I disagree, as I wrote above it is still very much possible to write over-specified tests with RTL. There is no such thing as "right" and "wrong" in programming, only tradeoffs.

> it dictates creates resilient and "user like" tests.

You can still mock out components to behave differently than they behave in production. Your users may not about a lot of things, that you still want to verify. For example, users do not care about metrics or logs, but you may want to verify these things. I would argue testing is not as simple as "test the user facing behavior". That may work for simpler CRUD apps, but by all means is not a one size fits all approach

> I'd love to see it being pushed

I still don't see the harm in highlighting the pros & cons & showing both... We can still recommend RTL for beginners, right?

> this isn't the only place to find Enzyme documentation at 
>  I've been writing Enzyme tests for plenty of times, and had they broke almost in every refactor. 

I get cognitive dissonance reading this. If these docs had better examples, maybe people would write better Enzyme tests that don't break. I also see the community writing bad Enzyme tests. The part where I disagree is that constraining the API of the testing tool is a silver bullet. If the docs did a better job of explaining the pros & cons, people will come to a solution that is right for them.

markerikson:
This conversation isn't going anywhere useful, and tbh it's making me _less_ inclined to have anything Enzyme-related in here.

Look. These are not the Enzyme docs. These are not the RTL docs.  These are the Redux docs. The goal here is to show how some viable approaches and guidelines for how to test Redux-specific logic (reducers, thunks, etc), and Redux-connected React components.  It's not intended to be an all-encompassing article on testing best practices, a deep dive into various test tools, or a treatise on how to work with legacy code.

The RTL-related content here appears to be primarily how to set up a Redux store and wrap the RTL `render` method accordingly.  That seems good to me.

I'm going to go ahead and merge this PR.  If someone wants to submit a PR to (re-)add some links to resources on testing React-Redux components with Enzyme to the page, go ahead and we can discuss that aspect there.

PR #3720, opened 2020-03-04
Bump docusaurus version

imbsky:


timdorr:
Thanks for nuking that extra yarn.lock file!

imbsky:
👍

PR #3721, opened 2020-03-04
Run format script

imbsky:
The CI fails because the current master has code that is not formatted with prettier. This PR fixes that.


imbsky:
The current CI fails only with the commit before this PR, and should probably work correctly with the next PR.

timdorr:
Thanks!

imbsky:
👍

PR #3736, opened 2020-03-19
Remove return type from TS docs

stefanoeb:
---
---

## Checklist

- **Section**: usage-with-typescript
- **Page**: type-checking-actions--action-creators

## What is the problem?
The documentation mention taking advantage of TS type inference for the returns of the actions, but we are still typing the return on this code block

Remove the types from the function return to simplify. I just did a fresh project doing the exact same and it works alright


timdorr:
It's actually not attempting to say the inference is for what the return type is at all, just that it's able to infer which of the unioned types of `ChatActionTypes` is the return type. I get this is kind of a confusing topic, but I think this is fine as is.

PR #3751, opened 2020-04-15
fix click bug like as 01dc17e0821

yukihane:

- **Section**:

Basic Tutorial

- **Page**:

Usage with React > [`components/TodoList.js`](https://redux.js.org/basics/usage-with-react#componentstodolistjs)

## What is the problem?

When todo item clicking, wrong record is processed.

Use `todo.id` instead of `index`,  like as "[Example: Todo List](https://redux.js.org/basics/example#componentstodolistjs)" page, 01dc17e0821.


timdorr:
Thanks!

PR #3753, opened 2020-04-15
Update ExampleRedditAPI.md

rishabh-husky:
---
---

## Checklist

  - _link issue here_

- **Section**: containers/AsyncApp.js
- **Page**: Advanced Tutorial -> Example: Reddit API

## What is the problem?
Is there any need to  call  `this.props.dispatch(fetchPostsIfNeeded(nextSubreddit))` in `handleChange(nextSubreddit)` , since, `componentDidUpdate()` is already going to dispatch this action after **selectedSubreddit** state in store is changed ?

**Note:** I previously created similar [PR](https://github.com/reduxjs/redux/pull/3752). But, was choosing incorrect function for modification
Removes `this.props.dispatch(fetchPostsIfNeeded(nextSubreddit))` from `handleChange(nextSubreddit)`

### DIFF
--- a/docs/advanced/ExampleRedditAPI.md
+++ b/docs/advanced/ExampleRedditAPI.md
@@ -245,7 +245,6 @@ class AsyncApp extends Component {
 
   handleChange(nextSubreddit) {
     this.props.dispatch(selectSubreddit(nextSubreddit))
-    this.props.dispatch(fetchPostsIfNeeded(nextSubreddit))
   }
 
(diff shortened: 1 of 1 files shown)


timdorr:
While not strictly needed, this will be slightly more performant as-is because it will fetch the posts preemptively. There's no harm in handling it this way.

rishabh-husky:
Thanks.

PR #3757, opened 2020-04-27
Explicit type argument any can be removed in "bindActionCreators"

deepfunc:
Explicit type argument any can be removed in "bindActionCreators.ts".

## Checklist


timdorr:
Thanks!

PR #3758, opened 2020-04-28
refactor: use isPlainObject to judge actionCreators.

deepfunc:
Use "isPlainObject" to judge actionCreators.

I think should use "isPlainObject" to judge it, and also fit with "import * as ActionCreators from". 

## Checklist


timdorr:
We don't really care that it's a plain object there. It can be something more esoteric. All that matters is the `for in` loop works.

Issue #3762, opened 2020-05-05
[Docs] Dark Theme?


B-Reif:
By default, docusaurus's classic theme [provides a Dark mode](https://v2.docusaurus.io/docs/theme-classic/#dark-mode). It looks like the Redux docusaurus config [has explicitly disabled](https://github.com/reduxjs/redux/blob/e002c3ad759453e26a90b4f35721773bd67bde88/website/docusaurus.config.js#L10) the dark mode. 

It looks like Redux is already [using the Docusaurus classic theme](https://github.com/reduxjs/redux/blob/e002c3ad759453e26a90b4f35721773bd67bde88/website/docusaurus.config.js#L117), with some custom styles. What would it take to enable the docusaurus dark mode?

markerikson:
There was some prior discussion on this when we transitioned to DS v2 for Redux Toolkit, with the conclusion that dark theme didn't mesh well for Redux's colors.  See https://github.com/reduxjs/redux-toolkit/pull/247#issuecomment-555455165 .

I'm not against having a dark theme, but I have no time to put into it myself and no expertise on that.  My focus is on reworking our content.

If someone else can figure out what a reasonably correct set of dark theme colors would look like, we can look at getting that in, but it'll have to be a community-driven thing.

timdorr:
I think the version we started with had an incomplete version of it? If you want to make a PR to enable it, go right ahead.

B-Reif:
I'll investigate. Is there a set of brand guidelines for Redux colors? Some of the custom css will have to change to maintain good contrast ratios. I'll try just lightening them for now.

timdorr:
Nope, it's always been made up as we go along. Whatever looks good :)

PR #3765, opened 2020-05-05
[Docs] add dark mode variables (issue #3762)

B-Reif:
---
name: "Add Dark Mode"
about: Adding variables and changing Docusaurus config to enable dark mode.
---

## Checklist

  - #3762

## What is the problem?
Viewing redux docs at 1am in light mode hurts my eyes.

- Modifies the docusaurus config to enable the classic theme's built-in dark mode.
- Adds new dark-mode-specific variable overrides to the custom Docusaurus css.
- Changes the 'recommendation' colors from hard-coded hex values to some color variables provided by Docusaurus.
- Fixes a css bug with the front page to correctly style the Feature svgs with the base font color (was previously typoed to an incorrect variable name and rendered as #000).

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -7,7 +7,7 @@ module.exports = {
   organizationName: 'reduxjs',
   projectName: 'redux',
   themeConfig: {
-    disableDarkMode: true,
+    disableDarkMode: false,
     prism: {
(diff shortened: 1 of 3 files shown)


timdorr:
LGTM. Let's just go with it and folks can chime in with PRs if there are minor things to fix.

PR #3766, opened 2020-05-09
Tighten applyMiddleware types

Methuselah96:
---
name: "Tighten applyMiddleware types"
about: More correctly types applyMiddleware
---

Fixes types

The `applyMiddleware` enhancer should not be allowed to accept any arguments, since it is only guaranteed to accept `preloadedState?` as its second argument.

Also the `createStore` parameter should not be `StoreCreator` since it doesn't allow for enhancers. It should match [the parameter type for `StoreEnhancor`](https://github.com/reduxjs/redux/blob/master/src/types/store.ts#L259-L261).

## Checklist

  - Simple enough that it probably doesn't need an issue.

It's implied that more than 2 arguments are accepted by this enhancer, when it should it only be called with 2 arguments.

TypeScript should produce an error if trying to pass more than the two defined arguments.

It tightens the arguments' types.

### DIFF
--- a/src/applyMiddleware.ts
+++ b/src/applyMiddleware.ts
@@ -1,7 +1,7 @@
 import compose from './compose'
 import { Middleware, MiddlewareAPI } from './types/middleware'
 import { AnyAction } from './types/actions'
-import { StoreEnhancer, StoreCreator, Dispatch } from './types/store'
+import { StoreEnhancer, Dispatch, PreloadedState, StoreEnhancerStoreCreator } from './types/store'
 import { Reducer } from './types/reducers'
 
 /**
@@ -55,11 +55,11 @@ export default function applyMiddleware<Ext, S = any>(
 export default function applyMiddleware(
   ...middlewares: Middleware[]
 ): StoreEnhancer<any> {
-  return (createStore: StoreCreator) => <S, A extends AnyAction>(
+  return (createStore: StoreEnhancerStoreCreator) => <S, A extends AnyAction>(
     reducer: Reducer<S, A>,
-    ...args: any[]
+    preloadedState?: PreloadedState<S>
   ) => {
-    const store = createStore(reducer, ...args)
+    const store = createStore(reducer, preloadedState)
     let dispatch: Dispatch = () => {
       throw new Error(
         'Dispatching while constructing your middleware is not allowed. ' +
--- a/src/applyMiddleware.ts
+++ b/src/applyMiddleware.ts
@@ -1,7 +1,12 @@
 import compose from './compose'
 import { Middleware, MiddlewareAPI } from './types/middleware'
 import { AnyAction } from './types/actions'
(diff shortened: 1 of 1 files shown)


Methuselah96:
This was the only usage of (`StoreCreator`)[https://github.com/reduxjs/redux/blob/master/src/types/store.ts#L215-L236] within Redux. Do we still want to keep that type or can it be removed?

timdorr:
Can we leave the `args` spread? There's historical reason for that in #2201.

Methuselah96:
Wow, that's a lot of historical context to dig through. However, I still think we should get rid of the `args` spread.

I'll start with https://github.com/reduxjs/redux/pull/1302 since that seems like the most authoritative reasoning. The first problem is that in the example given, the enhancers are applied in the wrong order and should actually be flipped to be like this instead:
```
const finalCreateStore = applyMiddleware(thunk, logger)(createStore)
const store = finalCreateStore(
  reducer,
  initialState,
  DevTools.instrument()
)
```
This will apply `DevTools.instrument()` first and the `applyMiddleware()` afterward which will make it work correctly (I reproduced the example not working here: https://codesandbox.io/embed/optimistic-torvalds-vl1qm?fontsize=14&hidenavigation=1&theme=dark). Do we really expect people to mix the two different styles of enhancers _and_ be able to apply them in the right order?

The main issue I have is that enhancers are not expected to pass along an enhancer as their third argument. The above example is the _only_ pattern that will work (`applyMiddleware` is used to create a `finalCreateStore` and then enhancers are passed into it as a third argument). If any other enhancer was substituted for `applyMiddleware` or if `applyMiddleware` was passed into a `finalCreateStore` created with a different enhancer it would not work. No user can depend on an enhancer passing along a third `enhancer` argument in its `createStore` function and there's no reason why `applyMiddleware` should support it. If we don't expect other enhancers to pass along a third `enhancer` argument, why do we expect `applyMiddleware` to do so?

I think the other issues mentioned actually bolster my claim. All of the issues that were linked to in the description of the issue you linked to (https://github.com/reduxjs/redux/issues/2028, https://github.com/reduxjs/redux/pull/2128, https://github.com/reduxjs/redux/issues/2131, https://github.com/reduxjs/redux/issues/2200) all agree with this change and would not have been created if this change were applied. 

As a newcomer to the code I found the signature of `applyMiddleware` confusing, and it seems obvious that others have thought so as well, since it seems incorrect and/or unnecessary. I think it would help clear up confusion to have `applyMiddleware` just support the intended function signature for enhancers instead of supporting an arcane use case that only works in very few specific cases and relies on knowing implementation and historical details.

markerikson:
Given that we're looking at doing this all in a v5.0 major anyway, I'd be fine with dropping whatever vague bits of support exist for that legacy syntax.

timdorr:
Yeah, let's just do this.

PR #3770, opened 2020-05-09
Re-enable TypeScript tests

Methuselah96:
---
name: "Re-enable TypeScript tests"
about: Testing is important
---

Internal change.

See https://github.com/reduxjs/redux/issues/3500#issuecomment-542363238 and following. I tried out `tsd`, but I did not like it as much. The `expect()` method syntax made it more cumbersome to assert in some scenarios. It also wasn't very well documented. I had to read the source code to figure out how to use it. So, I think sticking with `typings-tester` is preferable for now.

Note, I also needed to add an import in `types\store.ts` in order to get the global `Symbol.observable` definition working in the type outputs.

## Checklist

  - No, this is an internal change.

Types can break.

The ability to test types.

Re-enables types testing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4782,7 +4782,8 @@
         },
         "kind-of": {
           "version": "6.0.2",
-          "resolved": "",
+          "resolved": "https://registry.npmjs.org/kind-of/-/kind-of-6.0.2.tgz",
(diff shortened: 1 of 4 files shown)


timdorr:
Ah, we had somehow missed this. Thanks for catching and fixing it!

PR #3771, opened 2020-05-09
Increased performance of combineReducer result function

whiteand:
---
name: "Improvements of performance of combineReducer and combineReducer's result"
about: Improvements of performance of combineReducer and combineReducer's result
---

It's just a refactoring with improvements of performance

It's not necessary to check errors in a positive case on each call to function combination.
I do not know which kind of issue I should create for this improvement. Because this is not a bug.

## Checklist

  - _link issue here_

It's not a new feature or bugfix. So I do not know what to write here.

### DIFF
--- a/src/combineReducers.ts
+++ b/src/combineReducers.ts
@@ -142,10 +142,12 @@ export default function combineReducers(reducers: ReducersMapObject) {
   for (let i = 0; i < reducerKeys.length; i++) {
     const key = reducerKeys[i]
 
-    if (process.env.NODE_ENV !== 'production') {
-      if (typeof reducers[key] === 'undefined') {
-        warning(`No reducer provided for key "${key}"`)
-      }
+    if (
+      process.env.NODE_ENV !== 'production' &&
+      typeof reducers[key] === 'undefined'
+    ) {
+      warning(`No reducer provided for key "${key}"`)
+      continue
     }
 
     if (typeof reducers[key] === 'function') {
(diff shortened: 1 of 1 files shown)


timdorr:
Does this have any measurable performance improvement?

whiteand:
> Does this have any measurable performance improvement?

No, I do not did measurements. Is it not clear that 

```javascript
function f() {
  // code
}
```

is faster than

```javascript
function f() {
  if (someError) {
    throw someError
  }
 // code as in the example above
}
```

whiteand:
Also I think that this code is not optimal:

```javascript
for (let i = 0; i < arr.length; i++) {
  const elem = arr[i]
  if (typeof elem === 'undefined') {
     // something
  }
  if (typeof elem === 'function') {
   // something else
  }
}
```
It should be slower than:
```javascript
for (let i = 0; i < arr.length; i++) {
  const elem = arr[i]
  if (typeof elem === 'undefined') {
     // something
    continue
  }
  if (typeof elem === 'function') {
   // something else
  }
}
```

If `arr` will contain N `undefined` elements. We will not execute `if (typeof elem === 'function')` N times.

timdorr:
combineReducers shouldn't be called that often. If it's your bottleneck, you have other issues. I'm only seeing a 7% difference in my system. It's not really worth changing right now. Thanks anyways!

whiteand:
> combineReducers shouldn't be called that often. If it's your bottleneck, you have other issues. I'm only seeing a 7% difference in my system. It's not really worth changing right now. Thanks anyways!

The main improvement was not about improvement of combineReducer performance But about improvement of performance of reducer created by combineReducer (That is called after each action dispatch)

Thank you for your, review.

PR #3775, opened 2020-05-13
[Docs] fix typo in code example of Usage With React section

Alexsey:
  - _link issue here_

- **Section**: Basic Tutorial
- **Page**: Usage With React 

## What is the problem?

`Link` component should return `span` with no `onClick` listeners in case if it's **inactive**. In case if it's **active** it should return an `a` with `onClick` listener

It's fixing the code

Notes: in *Basic Tutorial* -> *Example: Todo List* `List` implementation is a little bit different and doesn't suffer from this bug


timdorr:
Thanks!

PR #3776, opened 2020-05-13
Fix type of next parameter in StoreEnhancer type

Methuselah96:
---
name: "Fix type of next parameter in StoreEnhancer type"
about: Fix type of next parameter in StoreEnhancer type
---

Fix a bug. Resolves https://github.com/reduxjs/redux/issues/3768.

Without this PR the type for `next` in `StoreEnhancer` is incorrect. It assumes that `next` will create a store with the current enhancers `Ext` and `StateExt` already applied to store. Instead, `next` returns a store with `NextExt` and `NextStateExt` generic parameters since the enhancer should work when composed with any other enhancer. It then expects the `StoreEnhancer` to return a new store creator that will create a store with `Ext` and `NextExt`, and `StateExt` and `NextStateExt`.

Note that I had to remove the `ExtendState` type in order to make this work. This type was introduced in https://github.com/reduxjs/redux/pull/3524, but I wasn't able to get the types working while using `ExtendState`. I think the types are more clear without it anyway.

## Checklist

  - https://github.com/reduxjs/redux/issues/3768

Steps to reproduce are in https://github.com/reduxjs/redux/issues/3768. There's also a new test for checking the types on stores created from composed enhancers. This test fails in `master`.

The types should allow for a generic `NextExt` and `NextStateExt`.

It adds those generics.

### DIFF
--- a/src/types/store.ts
+++ b/src/types/store.ts
@@ -3,20 +3,6 @@ import { Reducer } from './reducers'
 import '../utils/symbol-observable'
 
 /**
- * Extend the state
- *
- * This is used by store enhancers and store creators to extend state.
- * If there is no state extension, it just returns the state, as is, otherwise
- * it returns the state joined with its extension.
- *
- * Reference for future devs:
- * https://github.com/microsoft/TypeScript/issues/31751#issuecomment-498526919
- */
-export type ExtendState<State, Extension> = [Extension] extends [never]
-  ? State
-  : State & Extension
-
-/**
  * Internal "virtual" symbol used to make the `CombinedState` type unique.
  */
 declare const $CombinedState: unique symbol
@@ -134,11 +120,7 @@ export type Observer<T> = {
  * @template A the type of actions which may be dispatched by this store.
  * @template StateExt any extension to state from store enhancers
  */
-export interface Store<
-  S = any,
-  A extends Action = AnyAction,
-  StateExt = never
-> {
+export interface Store<S = any, A extends Action = AnyAction, StateExt = {}> {
   /**
    * Dispatches an action. It is the only way to trigger a state change.
    *
@@ -172,7 +154,7 @@ export interface Store<
    *
    * @returns The current state tree of your application.
    */
-  getState(): ExtendState<S, StateExt>
+  getState(): S & StateExt
 
   /**
    * Adds a change listener. It will be called any time an action is
@@ -217,7 +199,7 @@ export interface Store<
    * For more information, see the observable proposal:
    * https://github.com/tc39/proposal-observable
    */
-  [Symbol.observable](): Observable<ExtendState<S, StateExt>>
+  [Symbol.observable](): Observable<S & StateExt>
 }
 
 /**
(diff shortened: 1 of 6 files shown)


test/typescript/enhancers.ts
```diff
@@ -13,12 +13,17 @@ function dispatchExtension() {
...
+  }> = createStore => (reducer, preloadedState) => {
     const store = createStore(reducer, preloadedState)
-    return {
+
+    function replaceReducer<NewState, NewActions extends Action>(
+      nextReducer: Reducer<NewState, NewActions>
+    ) {
+      store.replaceReducer(nextReducer)
+      return newStore
+    }
```
Methuselah96, line 24: Many of the enhancer tests had incorrect `replaceReducer` methods. Previously they were actually returning the old store without the enhancer extensions applied instead of returning the new store with the enhancers applied. The fix in the types caught this bug.

test/typescript/enhancers.ts
```diff
@@ -294,3 +328,63 @@ function finalHelmersonExample() {
   // typings:expect-error
   newStore.getState().wrongField
 }
+
+function composedEnhancers() {
```
Methuselah96, line 332: This is a new test that would fail in `master`.

test/typescript/enhancers.ts
Methuselah96: Also note that the boilerplate that is now required for any enhancer is significantly larger since `replaceReducer` has to return the store. If https://github.com/reduxjs/redux/pull/3772 gets merged, this boilerplate would reduce significantly.

samarmohan:
Can you fix the conflicts?

markerikson:
I'll try to circle back and actually look at this in the near future.  My main observation atm is that this is gonna go onto `master`, and we have no active plans to actually _release_ the code in `master` at this point.  It's the 4.x branch that matters for the foreseeable future.

Methuselah96:
@markerikson No problem, don't mean to guilt you into looking into this but the old types are objectively wrong. I'd be happy to create a separate PR to target 4.x.

Methuselah96:
Let me know when you're ready to look at either one and I can resolve merge conflicts.

src/types/store.ts
```diff
@@ -259,13 +245,13 @@ export interface StoreCreator {
  * @template Ext Store extension that is mixed into the Store type.
  * @template StateExt State extension that is mixed into the state type.
  */
-export type StoreEnhancer<Ext = {}, StateExt = never> = (
-  next: StoreEnhancerStoreCreator<Ext, StateExt>
-) => StoreEnhancerStoreCreator<Ext, StateExt>
-export type StoreEnhancerStoreCreator<Ext = {}, StateExt = never> = <
+export type StoreEnhancer<Ext = {}, StateExt = {}> = <NextExt, NextStateExt>(
+  next: StoreEnhancerStoreCreator<NextExt, NextStateExt>
+) => StoreEnhancerStoreCreator<NextExt & Ext, NextStateExt & StateExt>
```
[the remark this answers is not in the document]
Methuselah96: Not sure why it doesn't show the old lines but here it is for context:
```ts
export type StoreEnhancer<Ext = {}, StateExt = never> = (
  next: StoreEnhancerStoreCreator<Ext, StateExt>
) => StoreEnhancerStoreCreator<Ext, StateExt>
```

markerikson:
I'd _like_ to try to do some general Redux org issue triage and cleanup over the next couple weeks, so yeah, getting this knocked out would be high on the priority list.

PR #3780, opened 2020-05-18
Change example code for merge approach

SDP190:
For just adding new items to state (Comments in our example) deep merge for unchanged state properties is unnecessary, just return a new state object with references to old items, and for new items either shallow copy action`s items or ( - if action items should not be used - ) deep merge only for that added items.

However the use case for deepmerge  all properties, maybe if we wanna have a single logic to handle also scenarios where we have an UPDATE action which makes changes to existing items by supplying items ('post comments' in this example) which include only updated properties, so the reducer then needs to deep-merge those updated properties from each item ('comment') with their other unchanged properties.

Thanks for the PR!

### DIFF
--- a/docs/recipes/structuring-reducers/UpdatingNormalizedData.md
+++ b/docs/recipes/structuring-reducers/UpdatingNormalizedData.md
@@ -22,7 +22,9 @@ function commentsById(state = {}, action) {
   switch (action.type) {
     default: {
       if (action.entities && action.entities.comments) {
-        return merge({}, state, action.entities.comments.byId)
+        return { ...state, ...action.entities.comments.byId}
+        // or if storing shallow copy of action items is not recommended:
+        // return merge({ ...state}, action.entities.comments.byId)
       }
       return state
     }


markerikson:
I'm not sure I agree with this change.

Per the `_.merge` docs:

> This method is like _.assign except that it recursively merges own and inherited enumerable string keyed properties of source objects into the destination object. Source properties that resolve to undefined are skipped if a destination value exists. Array and plain object properties are merged recursively. Other objects and value types are overridden by assignment. Source objects are applied from left to right. Subsequent sources overwrite property assignments of previous sources.

As far as I can tell, that means that I could potentially supply a partial item, and the fields would get merged into the existing item (_hopefully_ immutably, although I haven't tried it myself).

Changing it to be shallow here is a considerable change in behavior, and I don't think I understand the reasoning for changing this example.

SDP190:
@markerikson thanks for responding to this. 
You wrote:

> As far as I can tell, that means that I could potentially supply a partial item, and the fields would get merged into the existing item (hopefully immutably, although I haven't tried it myself).

That's exactly what I meant, that approach is only useful if you plan to use that logic for updating existing items with partial items, as I wrote:

> However the use case for deep-merge all properties, maybe if we wanna have a single logic to handle also scenarios where we have an UPDATE action which makes changes to existing items by supplying items ('post comments' in this example) which include only updated properties, so the reducer then needs to deep-merge those updated properties from each item ('comment') with their other unchanged properties.


However to clearify my issue, since the page in proceeding paragraph is stating:  _We'll use the example of adding a new Comment to a Post_. 
And then in next statement it implies that we must use deep-merging: [quote] _...One approach is to merge the contents of the action into the existing state. In this case, we need to do a deep recursive merge, not just a shallow copy. The Lodash merge function can handle this..._

So that's the confusion, because if only for adding new comments, why deep clone? So we have to clarify in documentation that deep-merge is not required besides for specific cases.

And indeed the PR should've include changes to documentation itself not for example code.

Do you agree @markerikson?

SDP190:
Anything on this?

timdorr:
OK, after looking through this in more detail, this doesn't appear to be the right change. In the example code, a deep merge may not be strictly needed, but this may be the case for other scenarios. For instance, if we decorate the comments with additional data in later, we may not want to duplicate that effort for all comments when only one was added.

Thanks, but I think we should leave this as-is.

SDP190:
@timdorr, If you don't wanna change the example code, maybe consider doing a PR on the documentation itself clarifying the point when deep-merge is needed?!

PR #3785, opened 2020-05-20
Clearify merge approach example for updating normalized data

SDP190:
Instead of only: _"Adding comments to a post"_ we actually describe how to Add/Update (-can be done with both: Simple Merging & Slice Reducer Composition solutions) or Delete (-which can be done with Slice Reducer Composition solution only-). So we should rather say: _"We'll use the example of handling mutations for Comments on a Post"_

Instead of: _"we need to do a deep recursive merge, not just a shallow copy"_, which is not correct because we don't have to, if we don't plan to have update actions with only partial items, we should rather say:  _"we can use deep recursive merge, not just a shallow copy, to allow for actions with partial items to update stored items."_

Thanks for the PR!

### DIFF
--- a/docs/recipes/structuring-reducers/UpdatingNormalizedData.md
+++ b/docs/recipes/structuring-reducers/UpdatingNormalizedData.md
@@ -7,13 +7,13 @@ hide_title: true
 
 # Managing Normalized Data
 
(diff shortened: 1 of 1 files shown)


timdorr:
Yep, that's better. Thanks!

PR #3786, opened 2020-05-25
Avoid unnecesary prototype loop

WebReflection:
## Background

People at redux-toolkit copy/pasted this utility in there, and [they are not welcoming a PR](https://github.com/reduxjs/redux-toolkit/pull/581) because of that, but the `while` loop in here is completely unnecessary, as the comparison is made on the first `proto` only.

## Improvements

  * _O(1)_ instead of _O(n)_
  * safe fallback for `Object.create(null)` cases
  * fast lane for `new Class()` derived objects
  * better performance
  * smaller code size

As this change has zero downsides, unless I am missing some explicit reason to `while` loop instead of never using `getPrototypeOf` more than twice, I hope it'll get in, so that others might copy and paste this new version.

Best Regards.

Thanks for the PR!

### DIFF
--- a/src/utils/isPlainObject.ts
+++ b/src/utils/isPlainObject.ts
@@ -3,12 +3,9 @@
  * @returns True if the argument appears to be a plain object.
  */
 export default function isPlainObject(obj: any): boolean {
-  if (typeof obj !== 'object' || obj === null) return false
-
-  let proto = obj
-  while (Object.getPrototypeOf(proto) !== null) {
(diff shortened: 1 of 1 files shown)


timdorr:
Same reasoning here as your linked PR. You can read the background in #2599.

WebReflection:
If ever interested in actually improving this little part of the equation, I've fixed all Travis CI related issues.

Best Regards.

PR #3797, opened 2020-06-09
refactor(types/actions): allow type-setting args on ActionCreator

zanona:
Following up https://github.com/reduxjs/redux/pull/3541#discussion_r437355966

We are allowing Action creators to be typed by declaring their return type as well as the argument type so function calls keep the expected parameter types.

---
name: "allow type-setting args on ActionCreator"
about: improving type declarations on Redux
---

It sort of adds a new feature, but it's opt-in for the type usage defaulting to the same as the previous behaviour. 

Allow developers to use Redux provided typings in order to implement ActionCreators

## Checklist

  - _link issue here_

### DIFF
--- a/docs/Glossary.md
+++ b/docs/Glossary.md
@@ -66,7 +66,7 @@ The base dispatch function _always_ synchronously sends an action to the store's
 ## Action Creator
 
 ```js
-type ActionCreator<A, P = any> = (...args: P[]) => Action | AsyncAction
+type ActionCreator<A, P extends any[] = any[]> = (...args: P) => Action | AsyncAction
 ```
 
 An _action creator_ is, quite simply, a function that creates an action. Do not confuse the two terms—again, an action is a payload of information, and an action creator is a factory that creates an action.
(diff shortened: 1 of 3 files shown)


zanona:
I'm not quite sure if there are tests for TS typings. Please let me know if I can add those in, if necessary. Thanks

timdorr:
Because it's a rest parameter, this is enforcing a uniform type for every argument. However, it would be possible to support non-uniform types like so:

```ts
interface ActionCreator<A, P extends any[] = any[]> {
  (...args: P): A
}
```

That would allow you to separately type each argument. 

The counter-point would be that you must pass an array type, such as `ActionCreator<MyAction, string[]>`. But I think that's OK to get more fine-grained control there.

zanona:
> The counter-point would be that you must pass an array type, such as `ActionCreator<MyAction, string[]>`. But I think that's OK to get more fine-grained control there.

:clap: That's a very nice idea! I'd guess it could look easier to the eye, having it declared as `ActionCreator<MyAction, [string]>`?

So we could even have `ActionCreator<MyAction, [string, number, string[]]>` :sunglasses:

Thanks for the review.

timdorr:
OK, I'm cool with this as-is. Thanks!

PR #3802, opened 2020-06-22
Make the script snippet of basic example consistent

vincent-pli:
The `Link.js` in 
https://github.com/reduxjs/redux/blob/master/docs/basics/ExampleTodoList.md#componentslinkjs
and 
https://github.com/reduxjs/redux/blob/master/docs/basics/UsageWithReact.md#componentslinkjs
is not consistent, moreover in latter, the logic is not correct:

should be:
```js
if (active) {
    return <span>{children}</span>
}
```
not:
```js
if (!active) {
    return <span>{children}</span>
}
```

Thanks for the PR!

### DIFF
--- a/docs/basics/UsageWithReact.md
+++ b/docs/basics/UsageWithReact.md
@@ -150,23 +150,17 @@ export default TodoList
 import React from 'react'
 import PropTypes from 'prop-types'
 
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

PR #3807, opened 2020-07-04
StrictReducer type is provided for reducers without initial state

sgrishchenko:
---
name: "\U0001F41B Bug fix or new feature"
---

_feature_

Sometimes it isn't possible to provide initial state for reducer, and this state should be provided externally. This changes should help in this cases.

## Checklist

  - _link issue here_

## New Features
StrictReducer type is provided for reducers without initial state. Also `createStore` overload is supported for reducer without initial state.

If you want to create reducer without inittial state, you can use type shortcut for it.
Also, if you use reducer in `createStore`, that don't accept undefined as input state, you shoul provide initial state for it.

I think docs changes aren't needed for this PR, because it is type checking improvement, Typescript should help you if something goes wrong.

### DIFF
--- a/src/createStore.ts
+++ b/src/createStore.ts
@@ -9,7 +9,7 @@ import {
   ExtendState
 } from './types/store'
 import { Action } from './types/actions'
-import { Reducer } from './types/reducers'
+import { Reducer, StrictReducer } from './types/reducers'
 import ActionTypes from './utils/actionTypes'
 import isPlainObject from './utils/isPlainObject'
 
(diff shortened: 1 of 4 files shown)


timdorr:
A reducer is expected to be able to handle an undefined state. This type would violate one of the expected behaviors of Redux, so it shouldn't be included in the library.

sgrishchenko:
For example, I have [this store configuration](https://codesandbox.io/s/nervous-flower-7olov)
What kind of initial state I should provide for `counterReducer`?

Methuselah96:
@sgrishchenko But you never pass `counterReducer` into `createStore`, right?

sgrishchenko:
@Methuselah96 yes, but what if I want to write some kind of integration test?

Methuselah96:
@sgrishchenko If you passed `counterReducer` directly into `createStore` in a test, then the test would fail. The counter state would be initialized to `undefined`. You either need to handle `undefined` or test it differently.

sgrishchenko:
@Methuselah96 If I would use curent implementation of redux it will fail. But if I use Typescript + my changes in this PR, I would receive compilation error, that I try to call `createStore` without second parameter `preloadedState`. Do you see?

Methuselah96:
@sgrishchenko Thanks, I understand your use case now. The change is valid (as in, everything will work), but it's not the normal use case and it sounds like @timdorr doesn't think it should be officially supported as per his earlier comment.

sgrishchenko:
@Methuselah96 Yea, I see, but thank you for discussing. To be honest, I wanted to declare `StrictReducer` type here, make another PR to redux-toolkit to support such kind of reducers in `createReducer` helper and re-use `StrictReducer` there. But it seems it is not use case for this library...

PR #3811, opened 2020-07-14
Add Storybook example to Writing Tests

msutkowski:
A handful of people have recently asked in Discord as well on StackOverflow about how to setup Storybook when using hooks instead of `connect`. When attempting to help these people, I noticed there weren't any real concrete examples available. The Storybook docs themselves rely on the concept of 'container components' and expect that every component is going to be passed props a la `<TaskList tasks=[{...}] />`, which isn't typically going to be the case as the usage of hooks increases. This PR offers a solution to that question and is based on the CRA redux template and Redux-Toolkit.

A TypeScript version of this can also be offered, and I can throw in an example repo as well if desired.

Thanks!

---
name: "\U0001F4D6 New/Updated Documentation Content"
---

## PR Type

This PR adds new content to the `Recipes > Writing Tests` page.

## Checklist

  - _link issue here_

- **Section**: Recipes
- **Page**: Writing Tests

This is an example of how to utilize hook-based components with Storybook. There are no concrete examples of how to do this currently that are easily findable. 

All users

None

The goal is to introduce Storybook and show a basic implementation using the current CRA redux template. The reader should understand that a story needs to be wrapped with a Provider just like their App does. They should also see how they can seed data for a component that strictly uses `useSelector` and `useDispatch` instead of being passed props directly.

A story needs access to the store, and once established, the component testing with Storybook can be expanded to fully integrate with RTL/enzyme/etc

### DIFF
--- a/docs/recipes/WritingTests.md
+++ b/docs/recipes/WritingTests.md
@@ -499,7 +499,6 @@ Next, we'll create a `withStoreProvider` [`decorator`](https://storybook.js.org/
 // ~repo/.storybook/decorators.js
 import React from 'react'
 import { Provider } from 'react-redux'
-import { ConfigureStoreOptions } from '@reduxjs/toolkit'
 import { createStore } from '../src/app/store'
 
 export function withStoreProvider(options) {
--- a/docs/recipes/WritingTests.md
+++ b/docs/recipes/WritingTests.md
@@ -469,19 +469,15 @@ In some cases, you will need to modify the `create` function to use different mo
 
 Integrating [Storybook](https://storybook.js.org/) with your redux-based application is very straightforward. The main requirement is to make sure that your stories are wrapped with `Provider` so that the components have access to the store.
 
+To get started, make sure you have followed the Storybook [automatic setup docs](https://storybook.js.org/docs/guides/guide-react/#automatic-setup).
+
 > ##### Note on `connect`ed components
 >
-> If your components are all `connect`-based and not yet utilizing hooks, the follow example is still relevant to you, but you can always pass props into the component instead of relying on setting the `preloadedState` as shown below. There is a complete example of that approach in the [Storybook data documentation](https://www.learnstorybook.com/intro-to-storybook/react/en/data/).
+> If your components are all `connect`-based, the following example is still relevant to you, but you can always pass props into the component instead of relying on setting the `preloadedState` as shown below. There is a complete example of that approach in the [Storybook data documentation](https://www.learnstorybook.com/intro-to-storybook/react/en/data/).
 
-To get started, make sure you have Storybook installed. Using the `cli` is the simplest method and will automatically configure the recommended settings for your project:
+First, we'll slightly modify the store to use a `createStore` helper that leverages [Redux-Toolkit's](https://redux-toolkit.js.org/api/configureStore) `configureStore`.
 
-```sh
-npx -p @storybook/cli sb init
-```
-
-First, we'll slightly modify the store to use a `createStore` utility that leverages [Redux-Toolkit's](https://redux-toolkit.js.org/api/configureStore) `configureStore`.
-
-```diff js
+```diff
 // store.js
 
 import { configureStore } from '@reduxjs/toolkit'
@@ -528,7 +524,7 @@ export function withStoreProvider(options) {
 
 Finally, we'll setup a few stories and decorate them with `withStoreProvider`:
 
-```ts
+```js
 // src/stories/0-Counter.js
 import React from 'react'
 import { Counter } from '../features/counter/Counter'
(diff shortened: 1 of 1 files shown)


docs/recipes/WritingTests.md
```diff
@@ -465,8 +465,82 @@ it('passes dispatch and getState', () => {
...
+// Export a store creator helper for usage by the stories
+export const createStore = options =>
+  configureStore({
+    reducer: {
+      counter: counterReducer
+    },
+    ...options
+  })
+
+export const store = createStore()
```
phryneas, line 493: It is not clear on first read that this is meant to be the `store` to be used by that app AND the `createStore` method to be used by storybook.

Maybe format it as `diff` instead of `js` to make clear that it is a recommended change to an existing file?

phryneas:
Another thing that could be added is that all stories sharing the same decorator will share the same store.

So if the decorator is added application-wide, all stories will share the same store instance, if it is added on a file level, all stories from a file will share it, and if a story has it's own decorator, that story will have isolated state.

All three models might be useful depending on usage.

docs/recipes/WritingTests.md
msutkowski: Good idea. Will update this.

msutkowski:
> Another thing that could be added is that all stories sharing the same decorator will share the same store.
> 
> So if the decorator is added application-wide, all stories will share the same store instance, if it is added on a file level, all stories from a file will share it, and if a story has it's own decorator, that story will have isolated state.
> 
> All three models might be useful depending on usage.

Agreed on this as well. I didn't know specifically how much detail to go into. Another point would be mentioning initializing storybook via the cli - similar to how setting up `jest` or RTL is mentioned.

I could also plug that hooking up `msw` is extremely simple for storybook usage if someone is testing components that rely on async thunks, but I'm not too sure if that is out of the scope of this brief overview. I could definitely use additional input on this @markerikson @phryneas before I do another pass on this, if you don't mind :)

docs/recipes/WritingTests.md
```diff
@@ -465,8 +465,103 @@ it('passes dispatch and getState', () => {
 
 In some cases, you will need to modify the `create` function to use different mock implementations of `getState` and `next`.
 
+### Testing with Storybook
+
+Integrating [Storybook](https://storybook.js.org/) with your redux-based application is very straightforward. The main requirement is to make sure that your stories are wrapped with `Provider` so that the components have access to the store.
+
+> ##### Note on `connect`ed components
+>
+> If your components are all `connect`-based and not yet utilizing hooks, the follow example is still relevant to you, but you can always pass props into the component instead of relying on setting the `preloadedState` as shown below. There is a complete example of that approach in the [Storybook data documentation](https://www.learnstorybook.com/intro-to-storybook/react/en/data/).
```
bpas247, line 474:
```suggestion
> If your components are all `connect`-based and not yet utilizing hooks, the following example is still relevant to you, but you can always pass props into the component instead of relying on setting the `preloadedState` as shown below. There is a complete example of that approach in the [Storybook data documentation](https://www.learnstorybook.com/intro-to-storybook/react/en/data/).
```
I'm also not sure if the `and not yet utilizing hooks` portion is necessary here

docs/recipes/WritingTests.md
```diff
@@ -465,8 +465,103 @@ it('passes dispatch and getState', () => {
...
+
+To get started, make sure you have Storybook installed. Using the `cli` is the simplest method and will automatically configure the recommended settings for your project:
+
+```sh
+npx -p @storybook/cli sb init
+```
```
bpas247, line 480: Linking to their [setup docs](https://storybook.js.org/docs/guides/guide-react/#automatic-setup) is probably sufficient enough

docs/recipes/WritingTests.md
```diff
@@ -465,8 +465,103 @@ it('passes dispatch and getState', () => {
...
+
+First, we'll slightly modify the store to use a `createStore` utility that leverages [Redux-Toolkit's](https://redux-toolkit.js.org/api/configureStore) `configureStore`.
+
+```diff js
```
bpas247, line 484: I'm not sure if the `js` portion actually adds any extra highlighting, so it could be removed

docs/recipes/WritingTests.md
```diff
@@ -465,8 +465,103 @@ it('passes dispatch and getState', () => {
...
+  const store = createStore(options)
+  return storyFn => {
+    return <Provider store={store}>{storyFn()}</Provider>
+  }
+}
+```
+
+Finally, we'll setup a few stories and decorate them with `withStoreProvider`:
+
+```ts
```
bpas247, line 531:
```suggestion
```js
```

docs/recipes/WritingTests.md
msutkowski: Good catch, and agreed :)

markerikson:
This is very stale.

@msutkowski , if you feel like redoing this, feel free!

Issue #3812, opened 2020-07-14
Redux Essentials Tutorial Part 4 has a type error


samwightt:
## What docs page needs to be fixed?

- **Updating Post Entries**:
- **Redux Essentials, Part 4: Using Redux Data**:

## What is the problem?
On the **Redux Essentials, Part 3: Basic Redux Data Flow** page, the user is told to create this initial state containing integer ids:

```
const initialState = [
  { id: 1, title: 'First Post!', content: 'Hello!' },
  { id: 2, title: 'Second Post', content: 'More text' }
]
```

However, later on in the tutorial, the user is told to create a `SinglePostPage` that compares a route parameter (a string) with each post id. Because a triple equals sign is used and type conversion doesn't happen, the original data will always show the not found page. New data works just fine, but old data does not.

## What should be changed to fix the problem?

The selector code should be:

```
const post = useSelector(state =>
    state.posts.find(post => post.id.toString() === postId)
)
```

markerikson:
Y'know, I think I fixed this in the app code and just forgot to update the snippets in the page.  I changed the IDs to strings:

https://github.com/reduxjs/redux-essentials-example-app/commit/c783509a6da54bb21066f20c760f2afdc7203f2c

```js
const initialState = [
  { id: '1', title: 'First Post!', content: 'Hello!' },
  { id: '2', title: 'Second Post', content: 'More text' },
]
```

@samwightt 	, can you submit a follow-up PR to drop the `.toString()` and update the displayed snippets to match what's in that commit?

samwightt:
Yep sure!

PR #3813, opened 2020-07-14
Fix type error in Redux Essentials.

samwightt:
---
name: "Fix error in Redux Essentials tutorial"
about: Fixes a type error in the Redux Essentials tutorial
---

This PR fixes a bug in the Redux Essentials Tutorial, fixing issue #3812.

New users who go through the Redux Essentials tutorial will have code that might not work as expected if this fix is not added.

## Checklist

    - See #3812.

See above.

When the user visits the route `/posts/1` or `/posts/2`, the not found screen appears.

The post for id `1` or id `2` should display.

This PR adds a `.toString()` call in a comparison function to fix a type error where the selector was comparing an integer and a string (which will always return false).

### DIFF
--- a/docs/tutorials/essentials/part-4-using-data.md
+++ b/docs/tutorials/essentials/part-4-using-data.md
@@ -55,7 +55,7 @@ export const SinglePostPage = ({ match }) => {
   const { postId } = match.params
 
   const post = useSelector(state =>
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

PR #3816, opened 2020-07-15
Change package module field to point to the .mjs file

pndewit:
I would classify it as a bug, but there could be other opinions about it. It's probably a breaking change though.

I've tried to find a similar issue/PR, but couldn't find it and I was wondering if it was ever even discussed. This may be closed if so, but if not:

This PR is basically just an idea (and I am new to Redux so please bear with me), let me know what you think.

When this PR is accepted there is no more need for `import ... from 'redux/es/redux.mjs';` imports when using ES modules in a browser environment, these can be shortened to: `import ... from 'redux';`.

## Checklist

- [ ] ~~Is there an existing issue for this PR?~~ No
- [ ] Have the docs been updated to match the changes in the PR? Not yet
- [ ] Have the tests been updated to match the changes in the PR? No
- [ ] Have you run the tests locally to confirm they pass? No

For what it's worth, I did do a functional check in my personal project using ES modules with Redux and this change fixed my `process is not defined` issues.

`import ... from 'redux';` loads `redux/es/redux.js` which then shows a (probably) well-known error: `process is not defined`.

That it works without the error.

The generated `.mjs` file has filtered out usages of `process`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   ],
   "main": "lib/redux.js",
   "unpkg": "dist/redux.js",
-  "module": "es/redux.js",
+  "module": "es/redux.mjs",
   "types": "types/index.d.ts",
   "files": [
     "dist",


timdorr:
Thanks, but given that `"module"` isn't part of the official spec anyways, I don't think we want to change this right now. It is intended for bundlers to use, which will have the ability to handle `process.env` stuff.

One thing we could look into is a conditional export. Newer versions of Node can make use of this, which would be the ideal way to handle a dual package: https://nodejs.org/api/esm.html#esm_conditional_exports

PR #3817, opened 2020-07-16
Only apply mapped types to un-branded types

invliD:
This is the same as #3805, except it's targeting master instead of 4.x.


timdorr:
Thanks!

PR #3825, opened 2020-07-26
Add hightlight to Tutorials/Using Redux Data

dvpfran:

- **Section**: Tutorials
- **Page**: Redux Essentials, Part 4: Using Redux Data

## What is the problem?
The line 482 doesn't have the highlight. For people who are writing code manually maybe they don't notice that they need to add that line.
code
Add: **// highlight-next-line**


markerikson:
Thanks!

PR #3829, opened 2020-07-30
docs: overview-concepts obj reference

joshSummerhays:
---
---

## Checklist

  - https://github.com/reduxjs/redux/issues/3828

- **Section**: tutorials
- **Page**: overview-concepts

## What is the problem?
`const obj1 =` in the "In order to update values immutably, your code must make copies of existing objects/arrays, and then modify the copies" section should be `const obj =` as it is referenced below it as just `obj`
change `const obj1` = to `const obj =`


markerikson:
Thanks!

PR #3832, opened 2020-07-31
Fix website readme

deminoth:
## Checklist

## What is the problem?

- Redux website have `package-lock.json` and do not have `yarn.lock`. Shell commands in the readme should use `npm`.

- Docusaurus requires an `USE_SSH=true` env. variable to use SSH, not an `USE_SSH=1`. Please see https://v2.docusaurus.io/docs/deployment#environment-settings

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -5,13 +5,13 @@ This website is built using Docusaurus 2, a modern static website generator.
 ### Installation
 
 ```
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks, but this is just the default readme from Docusaurus. You can ignore it.

PR #3840, opened 2020-08-07
Fix navbar wrapping between 997 and 1157px

### LINKED ISSUE #3839: Docs header overflows between 997 and 1157px

- **Section**: Header
- **Page**: All

## What is the problem?

The header content is overflowing when the browser width is between 997px and 1157px.
Tested on the latest version of Chrome and Safari.
Screenshot 2020-08-07 at 10 48 15

The easiest fix would be to set `flex-wrap: nowrap;` on `.navbar__inner`.

I'm happy to do a quick pull request for this.

MirceaUngureanu:
---
name: Bug fix\
about: Fixing a styling problem with Redux website
---

Fix a bug. Resolves #3839

Without this PR the navbar items will wrap between 997 and 1157px.

## Checklist

  - Resolves #3839
~~[ ] Have the docs been updated to match the changes in the PR?~~
~~[ ] Have the tests been updated to match the changes in the PR?~~

### What is the current behaviour, and the steps to reproduce the issue?
Website navbar overflows between 997 and 1157px. Can be viewed on any browser.

### What is the expected behaviour?
Navbar items shouldn't wrap if this behaviour breaks layout and impedes user interaction. 

PR will modify the `flex-wrap` rule on the navbar to keep items on one line.

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -114,6 +114,10 @@ a:visited {
   color: var(--ifm-color-primary);
 }
 
+.navbar .navbar__inner {
+  flex-wrap: nowrap;
+}
+
 .navbar .navbar__items {
   flex: 1 1 auto;
 }


markerikson:
Thanks!

PR #3853, opened 2020-08-12
changed heading from h4 to h3

klequis:
---
---

## Checklist

  - no

- yes

- **Section**: Reducers and Loading Actions
- **Page**: part-5-async-logic.md

## What is the problem?

Not sure but looks like the section titled "Reducers and Loading Actions" should be a h3 but is formatted as a h4.

Changed 
#### Reducers and Loading Actions
to


markerikson:
I _think_ I really did intend it as an h4, but it could work this way too.  Thanks!

PR #3878, opened 2020-09-05
Master event loop correspondence

2Shaun:
---
name: "\U0001F4D6 New/Updated Documentation Content"
---

## PR Type

Updates an existing page.

## Checklist

- **Section**: Essentials
- **Page**: 1

I think there should be more emphasis on the correspondence with the familiar event loop to mitigate the terminology learning curve.

They assume a newcomer to Redux, which I believe is already the target audience for this page.

### DIFF
--- a/docs/tutorials/essentials/part-1-overview-concepts.md
+++ b/docs/tutorials/essentials/part-1-overview-concepts.md
@@ -254,7 +254,7 @@ const addTodo = text => {
 
 #### Reducers
 
-A **reducer** is a function that receives the current `state` and an `action` object, decides how to update the state if necessary, and returns the new state: `(state, action) => newState`.
+A **reducer** is a function that receives the current `state` and an `action` object, decides how to update the state if necessary, and returns the new state: `(state, action) => newState`. **You can think of a reducer as an event listener which handles events based on the received action (event) type.**
 
 :::info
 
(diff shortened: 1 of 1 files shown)


markerikson:
Hey, thanks for filing this.  I know we chatted about it a bit over on Twitter.

I'm not sure I agree with some of the changes here, though.

First off, the phrase "event loop" is wrong here.  I think you mean "event _emitter" instead.

Second, the table really doesn't belong in the "what you've learned", because it's not recapping something that was already taught earlier in the page.

Third, if anything, I feel like this maybe goes a bit more in the "data flow" section.

tomoshaughnessy:
Assuming you were to agree with the change, do you think the reducer comparison belongs here? It seemed to fit with the other bolded comparisons in the doc. 

As for the table placement, I agree with what you're saying. Honestly, the 'event loop' phrase choice came from a quick web search that returned [this Wikipedia article](https://en.wikipedia.org/wiki/Event_loop), which seemed to describe a generic model for events. I can push with 'event emitter' used instead, but I'd have to reread the "data flow" section to figure out placement.

markerikson:
The phrase "event loop" _specifically_ refers to a tight `while` loop where a central driver application pulls events off a FIFO queue and executes them.  Concrete examples of this are the Win32 event loop (handled by `WndProc` functions in user app code), or in the JS world, the JS interpreter loop where every DOM event, timer event, etc is handled by adding events to the queue and executing them sequentially.

One of the best explanations of this is the talk ["What is the event loop, anyway?"](https://www.youtube.com/watch?v=8aGhZQkoFbQ), which walks through how the JS event loop works.

This is a completely different than an event _emitter_, where arbitrary parts of an app can subscribe to some future event, and the emitter object runs all listeners when that event is triggered.

At the moment, I think I'm inclined to leave that "reducer" sentence alone, but just drop the table.

2Shaun:
@markerikson Have you thought about this any further? I think it would help to relate the reducer to something familiar for new comers.

By the way, I did watch that talk and the event loop task queue definitely cleared up a couple of bugs I've had in the past.

markerikson:
Sure, this works.

Issue #3600, opened 2019-10-29
[Docs Rewrite] Meta-Issue: Understanding Redux section


markerikson:
This is a tracking issue for all work related to creating a new "Understanding Redux" section

## Goal

Pasting from my original writeup in #3313 :

> This section has several objectives:
>
> - Teach the "prerequisites" needed for Redux, including JS array methods, immutability, etc
> - Teach people how to "think in Redux" (multiple reducers listening to one action, treating actions more as events than "setters", etc)
> - Some details on how Redux works internally
> - Info on when you should use Redux, and why Redux patterns exist. (This should mostly be cribbed from Dan's "You Might Not Need Redux", my "Idiomatic Redux: The Tao of Redux" posts, and Justin Falcone's "What's So Great About Redux?").

## Tasks

- [ ] "Background Concepts" section
  - [ ] Outline content
  - [ ] Move material from "Structuring Reducers"
- [ ] "Thinking in Redux" section
  - Outline content
- [ ] "How does Redux Work?" section
  - [ ] Outline content
- [ ] "History and Design" section
  - [ ] Outline content

wgao19:
Would love to work on these also. But I feel these sections are more subjective than others so maybe will work on them later. Putting up a few resource links first

- [Getting Started with Redux](https://egghead.io/courses/getting-started-with-redux)
- [Building React Applications with Idiomatic Redux](https://egghead.io/courses/building-react-applications-with-idiomatic-redux)
- [Idiomatic Redux: The History and Implementation of React-Redux](https://blog.isquaredsoftware.com/2018/11/react-redux-history-implementation/)
- [Dan's twitter thread](https://twitter.com/dan_abramov/status/1039570011986321408?lang=en) on preferring _how to “think in it”_ over _explaining it in terms of API_

markerikson:
Two great resources on how and why to model actions as events:

- https://github.com/dmmulroy/talks/blob/master/event-driven-redux/slides.pdf
- https://youtu.be/K6OlKeQRCzo?t=2626 / https://rangle.slides.com/yazanalaboudi/deck#/

markerikson:
Totally can't forget @modernserf 's great post on [What's So Great about Redux?](https://www.freecodecamp.org/news/whats-so-great-about-redux-ac16f1cc0f8b/).

markerikson:
A very good summary of the "business logic" approach:

https://twitter.com/FwardPhoenix/status/952972062058074112

[...]

PR #3889, opened 2020-09-24
Create Thinking in Redux category and move some pages in it

sbaudray:
---
name: :book: Thinking in Redux
about: Adding a category and moving some pages to it
---

## PR Type

It moves some pages to a new category

## Checklist

  - https://github.com/reduxjs/redux/issues/3594

- **Section**: Added Thinking in Redux
- **Page**: 
    - introduction/motivation -> thinking-in-redux/motivation
    - introduction/three-principles -> thinking-in-redux/three-principles
    - introduction-prior-art -> thinking-in-redux/prior-art


website/_redirects
```diff
@@ -65,11 +65,11 @@
 
 # Old old Gitbook links?
 /docs/introduction/CoreConcepts.html        /introduction/core-concepts
```
markerikson, line 67: :exclamation: These are the very old legacy Gitbook links.  We do need to update these, but we _also_ need to add new redirects from the _current_ URLs as well.

Can you start a new comment-block-ish section somewhere labeled "2020 site reorganization" or soemthing along those lines, and add new redirect entries there from `/introduction/` to `/thinking-in-redux/` for these pages?

markerikson:
Hey, thanks for filing this!

Got a couple items I'd like to discuss around intended organization before this is ready.

First, the other relevant PR here is #3600 , which talks about adding an "Understanding Redux" section that would have categories inside it like "Thinking in Redux" and "History and Design".

Second, given that, I'd like to see this new "Thinking in Redux" category as a sub-category.

Third, I'm not yet 100% sure I want to name the parent category "Understanding Redux", but I'm not immediately sure what a better option is.  Suggestions?

markerikson:
Having thought about it further, I'm leaning towards:

- New parent category, "Understanding Redux", at `/understanding/`
  - New child category, "Thinking in Redux", at `/understanding/thinking-in-redux/`
    - Motivation
    - Three Principles
  - New child category, "History and Design"
    - Prior Art

website/_redirects
sbaudray: It should be better now

markerikson:
Awright, this looks pretty good.  Confirmed the redirect for the "Motivation" page works correctly. Let's get this in.

markerikson:
Hah. And of course, as soon as I hit the "Merge" button, I remembered that I wasn't 100% sure on having the "Understanding Redux" category be right below the "Introduction".

@sbaudray : Tell you what.  Can you create a second PR that:

- Moves the "Understanding" category to be after "Recipes" (ie, 4th in the sidebar list)
- Moves the "Glossary" page from the "Other" category to "Understanding > Thinking in Redux"
- Moves the "Troubleshooting" page from the "Other" category to "Recipes"
- Removes the "Other" category entirely

From there, we can think about how to reorganize the "Introduction" and "Recipes" categories further.

PR #3896, opened 2020-09-28
Small fix to tutorials/essentials/part-2-app-structure

maxsynnott:
The tutorial currently says to select the last "counter/increment" entry however the tutorial talks about and shows the last "counter/incrementByAmount" entry as seen directly below the changed line.


markerikson:
Yep, good catch. Thanks!

PR #3899, opened 2020-10-02
docs: Add an example of type checking a middleware function

mszabo-wikia:
---
---

## PR Type

It updates an existing page.
## Checklist

- **Section**: Recipes
- **Page**: Usage With TypeScript

It was pointed out in a Reactiflux conversation last night that the TypeScript
examples provided by Redux currently do not demonstrate how to type a middleware
function. This patch adds a brief section to the docs that shows how to define a
typed middleware.

This recipe assumes an entry-level knowledge of Redux and TypeScript and the new example aims to remain in line with these expectations.

### DIFF
--- a/docs/recipes/UsageWithTypescript.md
+++ b/docs/recipes/UsageWithTypescript.md
@@ -230,6 +230,26 @@ const rootReducer = combineReducers({
 export type RootState = ReturnType<typeof rootReducer>
 ```
 
+### Type Checking Middlewares
+
+A middleware is a higher-order function that composes a dispatch function to return a new dispatch function.
+To create a type checked middleware function, we can leverage the `Middleware` interface provided by Redux together with our store type (as defined in the previous example):
+
+```ts
+// src/middleware/example.ts
+
+import { Middleware } from 'redux'
+
+import { RootState } from '../store'
+
+export const exampleMiddleware: Middleware<
+  {},
+  RootState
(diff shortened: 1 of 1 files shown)


docs/recipes/UsageWithTypescript.md
```diff
@@ -230,6 +230,26 @@ const rootReducer = combineReducers({
...
+
+```ts
+// src/middleware/example.ts
+
+import { Middleware } from 'redux'
+
+import { RootState } from '../store'
+
+export const exampleMiddleware: Middleware<
+  {},
```
timdorr, line 246: We may want to briefly explain what this generic is for, as it's not that obvious from this example.

timdorr (approved):
This looks good as-is. If want to address the Dispatch extension generic, we can, but I don't think that's a blocker.

docs/recipes/UsageWithTypescript.md
mszabo-wikia: Sure, I can add a note. Would you prefer it in a standalone commit or squashed?

timdorr: Either is fine. We're going to squash it all anyways when it gets merged.

mszabo-wikia: 👍  3c0023ac

timdorr:
Thanks!

PR #3905, opened 2020-10-15
Fix: Confusing part in the tutorial

0000marcell:
---
---

## Checklist

  - _link issue here_

- **Section**: tutorials/essentials
- **Page**: part-3-data-flow

## What is the problem?
The tutorial goes on a limb to explain how to create a new project than
sudenly go back to the tutorial demo app like if it was the same thing

Add aditional information, making it clear for the reader that now we're talking about the demo app and not the generated one


markerikson:
Hey, appreciate the PR, but I'm afraid I don't think this really improves the current wording.

PR #3906, opened 2020-10-16
Fix naming when thunk-ifying AddPostForm.js.

chrzhang:
---
---

## Checklist

  - _link issue here_

- **Section**: Async Logic and Data Fetching
- **Page**: Checking Thunk Results in Components

## What is the problem?
Naming of function being replaced is inaccurate.

Correctly names function to be replaced with thunk.


timdorr:
Thanks!

PR #3918, opened 2020-10-25
Redux toolkit tree view

aquaductape:
---
---

## PR Type
This updates the `example/tree-view` to using Redux Toolkit.

However enzyme testing files are removed.

I wasn't sure how to update the enzyme testing files, and also reading discussion https://github.com/reduxjs/redux/pull/3708 lead me to the conclusion that testing with enzyme with redux is tricky. 

- **Section**: https://redux.js.org/introduction/examples#tree-view


markerikson:
Thanks!  Will take a look through this and leave some comments

examples/tree-view/package.json
```diff
@@ -1,29 +1,33 @@
 {
-  "name": "tree-view",
-  "version": "0.0.1",
+  "name": "redux-toolkit-tree-view",
```
markerikson, line 2: :hand: Let's leave this as `"redux-tree-view"` - I don't think it's necessary to specify the "toolkit" part

examples/tree-view/src/Node/Node.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react'
```
markerikson, line 1: :hand: Let's rename the folder itself from `Node` to `nodes`.  `Node.js` is fine.

examples/tree-view/src/Node/Node.js
```diff
@@ -0,0 +1,66 @@
...
+import { useDispatch, useSelector } from 'react-redux'
+import {
+  addChild,
+  createNode,
+  deleteNode,
+  removeChild,
+  increment
+} from './treeSlice'
+
+const Node = React.memo(({ id, parentId }) => {
```
markerikson, line 11: :question: Not sure if passing an anonymous function to `React.memo()` this way correctly gives the nested component a display name in the DevTools. Can you check that?  If not, we'd want to extract this to its own separate variable first and then pass it to `memo(0`.

examples/tree-view/src/Node/Node.js
```diff
@@ -0,0 +1,66 @@
...
+  const dispatch = useDispatch()
+
+  const { childIds, counter } = node
+
+  const handleIncrementClick = () => {
+    dispatch(increment(id))
+  }
+
+  const handleAddChildClick = () => {
+    dispatch(createNode())
```
markerikson, line 22: :hand: Can we try to figure out a way to dispatch _one_ action that has both these results?  Not sure what the necessary reducer logic would be, but we do recommend trying to [dispatch fewer actions in a row](https://redux.js.org/style-guide/style-guide#avoid-dispatching-many-actions-sequentially) and have [many reducers handle one action](https://redux.js.org/style-guide/style-guide#allow-many-reducers-to-respond-to-the-same-action).

examples/tree-view/src/Node/Node.js
```diff
@@ -0,0 +1,66 @@
...
+    dispatch(addChild(id))
+  }
+
+  const handleRemoveClick = () => {
+    dispatch(removeChild({ nodeId: parentId, childId: id }))
```
markerikson, line 27: :hand: Ditto here

examples/tree-view/src/Node/Node.js
```diff
@@ -0,0 +1,66 @@
...
+    dispatch(deleteNode(id))
+  }
+
+  const renderChild = ({ childId, id }) => {
```
markerikson, line 31: :hand: Does this really need to be defined as a separate function?  I'd say it's more idiomatic to have an inline callback for `map()`

examples/tree-view/src/Node/treeSlice.js
```diff
@@ -0,0 +1,67 @@
...
+import generateTree from "../generateTree";
+
+const { tree, idCount } = generateTree();
+
+let nextId = idCount;
+const initialState = tree;
+
+const getAllDescendants = (tree, id) => {
+  const descendants = [id];
+  const run = (id) => {
```
markerikson, line 11: :hand: Hmm. Writing this as a recursive function _works_, but it's a bit unusual.  Is there another way to do this? Maybe writing it as a queue that we keep pushing into?

examples/tree-view/src/Node/treeSlice.js
```diff
@@ -0,0 +1,67 @@
...
+      const nodeId = action.payload;
+      state[nodeId].counter++;
+    },
+    createNode(state) {
+      nextId++;
+      state[nextId] = { id: nextId, counter: 0, childIds: [] };
+    },
+    deleteNode(state, action) {
+      const nodeId = action.payload;
+      const ids = getAllDescendants(state, nodeId);
```
markerikson, line 40: :question: Is there a specific need to move this logic outside the reducer? It's not wrong to do so, and naming things can help provide clarity, but it also can make it harder to follow what's going on.

examples/tree-view/src/Node/treeSlice.js
```diff
@@ -0,0 +1,67 @@
...
+
+      deleteMany(state, ids);
+    },
+    addChild(state, action) {
+      const nodeId = action.payload;
+      const childId = nextId;
+      state[nodeId].childIds.push(childId);
```
markerikson, line 47: :speech_balloon: I haven't even looked at the old code, but I guarantee this is _wayyyy _better than the hand-written immutable update version :)

examples/tree-view/src/Node/treeSlice.js
```diff
@@ -0,0 +1,67 @@
...
+const treeSlice = createSlice({
+  name: "tree",
+  initialState,
+  reducers: {
+    increment(state, action) {
+      const nodeId = action.payload;
+      state[nodeId].counter++;
+    },
+    createNode(state) {
+      nextId++;
```
markerikson, line 35: :exclamation: This is a side effect, and shouldn't happen in a reducer. But, if you add a `prepare` callback for this reducer, we can do the exact same line in the prepare callback and then put the new ID value into the action itself.

examples/tree-view/src/Node/treeSlice.js
```diff
@@ -0,0 +1,67 @@
...
+    },
+    deleteNode(state, action) {
+      const nodeId = action.payload;
+      const ids = getAllDescendants(state, nodeId);
+
+      deleteMany(state, ids);
+    },
+    addChild(state, action) {
+      const nodeId = action.payload;
+      const childId = nextId;
```
markerikson, line 46: :exclamation: Should also put the next ID value into the action via a `prepare` callback here

examples/tree-view/src/actions/index.js
```diff
@@ -1,33 +0,0 @@
-export const INCREMENT = 'INCREMENT'
-export const CREATE_NODE = 'CREATE_NODE'
-export const DELETE_NODE = 'DELETE_NODE'
-export const ADD_CHILD = 'ADD_CHILD'
-export const REMOVE_CHILD = 'REMOVE_CHILD'
-
-export const increment = (nodeId) => ({
```
markerikson, line 7: :speech_balloon: yayyy, goodbye!

examples/tree-view/src/reducers/index.js
```diff
@@ -1,65 +0,0 @@
...
-      return {
-        ...state,
-        childIds: childIds(state.childIds, action)
-      }
-    default:
-      return state
-  }
-}
-
-const getAllDescendantIds = (state, nodeId) => (
```
markerikson, line 38: :speech_balloon: Ahh, so those functions existed in the original version, got it

examples/tree-view/src/reducers/index.spec.js
```diff
@@ -1,163 +0,0 @@
-import deepFreeze from 'deep-freeze'
-import reducer from './index'
-import { increment, createNode, deleteNode, addChild, removeChild } from '../actions'
-
-describe('reducer', () => {
```
markerikson, line 5: :hand: We should be able to test the reducers in basically the same way we did before, since those are just `const actual = reducer(testState, action); expect(actual).toEqual(expected);`

In fact, if you switch the dispatched actions to the ones generated by `createSlice`, these reducer tests _should_ still pass.

examples/tree-view/src/containers/Node.spec.js
```diff
@@ -1,90 +0,0 @@
-import React from 'react'
-import { shallow } from 'enzyme'
-import ConnectedNode, { Node } from './Node'
-
-function setup(id, counter, childIds, parentId) {
```
markerikson, line 5: :hand: Can we replace this file with some tests written with React Testing Library?

markerikson (changes requested):

examples/tree-view/src/Node/Node.js
aquaductape: You're correct, in the React Dev Tools, the Memo Component is labeled as Anonymous.

What should I name it? Is MemoNode okay?
```js
const MemoNode = React.memo(function Node({ id, parentId }) { ...
```

examples/tree-view/src/Node/Node.js
aquaductape: I could wrap them in a batch. Or simplify into a single reducer for each, addChildToNode (does createNode and addChild), removeNodeByParent (does removeChild and deleteNode). If I do that, should I still keep the specific reducers for testing purposes, or remove them and change the tests.

```js
// BEFORE
removeChild(state, action) {
  const { childId, nodeId } = action.payload
  const filteredChildIds = state[nodeId].childIds.filter(
    id => id !== childId
  )
  state[nodeId].childIds = filteredChildIds
},
deleteNode(state, action) {
  const nodeId = action.payload
  const ids = getAllDescendants(state, nodeId)

  deleteMany(state, ids)
}

// AFTER
removeNodeByParent(state, action) {
  const { childId, parentId } = action.payload
  const filteredChildIds = state[parentId].childIds.filter(
    id => id !== childId
  )
  // remove child
  state[parentId].childIds = filteredChildIds

  // delete Node(s)
  const ids = getAllDescendants(state, childId)

  deleteMany(state, ids)
},
```
```js
// BEFORE
createNode: {
  reducer: (state, action) => {
    const nextId = action.payload
    state[nextId] = { id: nextId, counter: 0, childIds: [] }
  },
  prepare: () => {
    nextId++
    return { payload: 'node_' + nextId }
  }
}
addChild: {
  reducer: (state, action) => {
    const { nodeId, nextId } = action.payload
    const childId = nextId

    state[nodeId].childIds.push(childId)
    state[childId] = { id: childId, childIds: [], counter: 0 }
  },
  prepare: nodeId => {
    return { payload: { nodeId, nextId: 'node_' + nextId } }
  }
},

// AFTER
addChildToNode: {
  reducer: (state, action) => {
    // create Node
    const { nextId, id } = action.payload
    state[nextId] = { id: nextId, counter: 0, childIds: [] }

    // add child
    state[id].childIds.push(nextId)
  },
  prepare: id => {
    nextId++
    return { payload: { id, nextId: 'node_' + nextId } }
  }
},
```

examples/tree-view/src/Node/treeSlice.js
aquaductape: You're right that it does look strange, but it's the fastest that I written so far https://jsbench.me/w0kgrpkfl0/1, it compares the previous one which was an array reduce, breadth first search (is that what you mean by writing as a queue).

Here's one where it pushes ids into a queue. From my online research, js implementations of BFS use array.shift, but it slows it down, so I used  a index to retrieve the value instead, so it's much faster but increases space complexity.
```js
const getAllDescendantIds = (tree, id) => {
  const queue = [id, ...tree[id].childIds]
  const descendants = []
  let indexPointer = 0

  while (indexPointer < queue.length) {
    const currentItem = queue[indexPointer]
    descendants.push(currentItem)
    queue.push(...tree[currentItem].childIds)
    indexPointer++
  }

  return descendants
}
```

examples/tree-view/src/reducers/index.js
aquaductape: So it's okay for me to leave it as a separate function?

examples/tree-view/src/containers/Node.spec.js
aquaductape: Did my best, tried to mock reducer actions, but couldn't figure it out

examples/tree-view/src/Node/Node.js
markerikson: That seems reasonable.

examples/tree-view/src/Node/Node.js
markerikson: Yeah, I think I like that change.  Update tests to match whatever logic we've got in here when we're done.

examples/tree-view/src/reducers/index.js
markerikson: Eh, sure.

examples/tree-view/src/containers/Node.spec.js
slim-hmidi: @markerikson  @aquaductape why we need to test the reducers and actions in the app? I think testing the component behavior correctly will check the reducers functions in the background (and the components tests will executes our reducers and we should get the corresponding response according to that).
You don't think that we will add some implementation details to the tests? (for a simple example it is ok but can confuse readers later and need to write tests for every reducer and actions in a big app and the bundle will be huge).

References: 
https://kentcdodds.com/blog/testing-implementation-details
https://kentcdodds.com/blog/avoid-the-test-user

aquaductape: Thanks for sharing @slim-hmidi, and your concerns are valid. Testing isn't my forte and would like to hear what @markerikson has to say.

aquaductape:
@markerikson any updates? :)

aquaductape:
:'(

PR #3923, opened 2020-10-28
refactor: Explicit type argument any can be removed.

deepfunc:
Explicit type argument any can be removed in "src/types/reducers.ts".


timdorr:
Thanks!

PR #3930, opened 2020-11-02
docs: Fix broken url in docs

saintmalik:
## Summary

Easy Navigation and easy reading for others 

(Fix broken url in MigratingToRedux.md and Troubleshooting.md)

### DIFF
--- a/docs/recipes/MigratingToRedux.md
+++ b/docs/recipes/MigratingToRedux.md
@@ -17,7 +17,7 @@ We don't want to lock you in!
 
 Your process will look like this:
 
-- Create a function called `createFluxStore(reducer)` that creates a Flux store compatible with your existing app from a reducer function. Internally it might look similar to [`createStore`](../api/createStore.md) ([source](https://github.com/reduxjs/redux/blob/master/src/createStore.ts)) implementation from Redux. Its dispatch handler should just call the `reducer` for any action, store the next state, and emit change.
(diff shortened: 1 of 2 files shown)


markerikson:
Tell you what - can you specifically link to the file versions from the `v4.0.5` tag tree instead?

saintmalik:
> Tell you what - can you specifically link to the file versions from the `v4.0.5` tag tree instead?

I don't understand you well, can you explain better what you mean

markerikson:
@saintmalik the file links in there have two issues:

- They're to the TS version of `createStore`, which hasn't actually been published in a real version yet, and might be harder to read for people who aren't familiar with TS
- Since it's pointing to `master`, that file might change at some point in the future

So, it's better to link to a specific version of the file at a known point in history.  In this case, I'm suggesting that we link to the file as it existed when we published the `v4.0.5` tag, which you can link to this way:

https://github.com/reduxjs/redux/blob/v4.0.5/src/createStore.js

saintmalik:
> @saintmalik the file links in there have two issues:
> 
> * They're to the TS version of `createStore`, which hasn't actually been published in a real version yet, and might be harder to read for people who aren't familiar with TS
> * Since it's pointing to `master`, that file might change at some point in the future
> 
> So, it's better to link to a specific version of the file at a known point in history. In this case, I'm suggesting that we link to the file as it existed when we published the `v4.0.5` tag, which you can link to this way:
> 
> https://github.com/reduxjs/redux/blob/v4.0.5/src/createStore.js

Okay, I will do that

saintmalik:
> https://github.com/reduxjs/redux/blob/v4.0.5/src/createStore.js

Done, made the chnages

markerikson:
Thanks!

PR #3936, opened 2020-11-04
Always rebuild docs on Netlify

smikitky:
Hi, while I was preparing a Japanese translation of the Redux docs using the fork-based approach, I noticed that Netlify refuses to rebuild and deploy the forked site even after I pushed some translated docs under `/docs`. The deploy log says:

```
12:27:22 PM: No changes detected in base directory. Returning early from build.
12:27:22 PM: Failed during stage 'checking build content for changes': Canceled build due to no content change
```

And this is due to [the change Netlify made in October 2019](https://docs.netlify.com/configure-builds/file-based-configuration/#ignore-builds):

> For sites connected to a Git repo after October 3rd, 2019, Netlify tries to determine if there are any changes in the site's base directory by comparing the last known version of the files within that directory. If no change was detected, the build system skips the build, returning early from the build process.

This PR fixes this problem. It adds the `ignore` attribute in `netlify.toml` so that the site will be rebuilt on _every_ push even when the contents of the `/website` directory have not changed. (This uses `/bin/false`, a UNIX command that just exits with the status code `1`.)

According to the Netlify docs, this doesn't affect how Netlify builds the original docs site (redux.js.org), but it helps new forks in the future.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,3 +2,4 @@
   base    = "website"
   publish = "website/build"
   command = "npm run build && cp _redirects ./build"
+  ignore  = "false"
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,4 +2,4 @@
   base    = "website"
   publish = "website/build"
   command = "npm run build && cp _redirects ./build"
-  ignore = "git diff --quiet HEAD^ HEAD -- ../docs/ ."
+  ignore  = "git diff --quiet HEAD^ HEAD -- ../docs/ ."
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,4 +2,4 @@
   base    = "website"
   publish = "website/build"
   command = "npm run build && cp _redirects ./build"
-  ignore  = "false"
+  ignore = "git diff --quiet HEAD^ HEAD -- ../docs/ ."


timdorr:
We should still check specific paths to determine if we need to rebuild. For code-only PRs, this is going to create a lot of noise from the comments and commit statuses. I would use the suggested command from their docs to limit it to checking only the docs and website paths. 

Also, this isn't going to affect this repo because it was connected before that cutoff date.

smikitky:
@timdorr  You are correct that this PR does *not* affect how *this* site (redux.js.org) is built by Netlify, simply because it was connected to Netlify before 2019. This PR is about *future translation forks* like ko.redux.js.org.

I do not mean to change how Netlify currently builds *this* repo. What I want to do is to make (future) forks work just like *this* repo. Currently, Netlify always makes a branch preview for each PR of *this* repository, even if it's a code change (`/src`). While I know this behavior is fixable, this PR is not for changing it. (And this behavior is not that bad because most incoming PRs are about docs anyway.)

In addition, this PR won't create "a lot of noise from the comments". Nothing will change for *this* repo. Netlify has never made more than one comment for each PR (for example see #3770).

netlify.toml
```diff
@@ -2,3 +2,4 @@
   base    = "website"
   publish = "website/build"
   command = "npm run build && cp _redirects ./build"
+  ignore  = "false"
```
timdorr, line 5:
```suggestion
  ignore  = "git diff --quiet HEAD^ HEAD -- docs/ website/"
```

This is what I mean we should use. This prevents code-only PRs from being built by Netlify, but also should resolve your issues on your fork.

smikitky: I agree, this should fix the problem of new forks, too. I simply did not intend to change the current behavior of *this* site, but this should be an improvement. Let me check this on my fork first.

smikitky: Git-diff runs on the base directory (`/website`), so the correct command was:

```
ignore = "git diff --quiet HEAD^ HEAD -- ../docs/ ."
```

Otherwise this approach seems to be working fine on my fork.

timdorr: Ah, missed that. Thanks!

timdorr:
OK, I'm good with this. Thanks for digging in with me!

PR #3946, opened 2020-11-09
Toggle todo in doc

mehagar:
## Checklist

  - _link issue here_

- **Section**: Style Guide
- **Page**: Style Guide: Best Practices

## What is the problem?

I was struggling to understand the logic in this part of the docs. Why are we setting the id to what it already is here instead of toggling the todo's `completed` state like elsewhere in the docs?

Update the example so that it's `completed` flag is toggled.


timdorr:
Thanks!

PR #3947, opened 2020-11-10
Add a complex example with typescript, hooks, sagas

kazurus:
We have a lot of different examples, but we don't have a complex example with typescript

---
about: Adding a new example with typescript, hooks, sagas
---

## PR Type

Not yet

## Checklist

  - _link issue here_


timdorr:
Thanks, but we're not looking to add more examples right now.

PR #3950, opened 2020-11-11
Fix a word to an actually used one

falinsky:
---
---

## Checklist

  - _link issue here_

- **Section**: Loading Users
- **Page**: Redux Essentials, Part 5: Async Logic and Data Fetching

## What is the problem?
Incorrectly mentioned the word `User` instead of `author` that is actually used.

Changed `User` to `author`


timdorr:
Thanks!

PR #3953, opened 2020-11-16
Fix typo: There should be to ways...

maxalmonte14:
Hi Redux maintainers, I think the right text should be _"There should be **two** ways to..."_

---
---

## Checklist

- **Section**: defining-requirements
- **Page**: docs/tutorials/fundamentals/part-3-state-actions-reducers.md

## What is the problem?

Missing letter in a word, leading to confusion

Just fix the typo


markerikson:
Thanks!

PR #3963, opened 2020-12-08
[Tutorial Docs] Changing PostAuthor userId property to consume a userId as opposed to a user

tylerthedeveloper:
---
---

## PR Type

Update

## Checklist

  - _link issue here_

- **Section**:
redux/docs/tutorials/essentials

- **Page**:
part-4-using-data.md

## For Adding New Content 

n/a

tutorial

new redux users

generic routing practices with regards to parameter passing

Best practice on generic routing practices with regards to parameter passing

see above

change holding user as a user object to userID

### DIFF
--- a/docs/tutorials/essentials/part-4-using-data.md
+++ b/docs/tutorials/essentials/part-4-using-data.md
@@ -572,7 +572,7 @@ export const PostAuthor = ({ userId }) => {
 
 Notice that we're following the same pattern in each of our components as we go. Any component that needs to read data from the Redux store can use the `useSelector` hook, and extract the specific pieces of data that it needs. Also, many components can access the same data in the Redux store at the same time.
 
-We can now import the `PostAuthor` component into both `PostsList.js` and `SinglePostPage.js`, and render it as `<PostAuthor userId={post.user} />`, and every time we add a post entry, the selected user's name should show up inside of the rendered post.
+We can now import the `PostAuthor` component into both `PostsList.js` and `SinglePostPage.js`, and render it as `<PostAuthor userId={post.userId} />`, and every time we add a post entry, the selected user's name should show up inside of the rendered post.
 
 ## More Post Features
 
@@ -605,7 +605,7 @@ Since we can't just put a `Date` class instance into the Redux store, we'll trac
             date: new Date().toISOString(),
             title,
             content,
-            user: userId,
+            userId: userId,
           },
         }
       },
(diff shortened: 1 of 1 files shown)


markerikson:
Hey, appreciate the PR.  Not sure if this is going to be correct, though.  The examples need to match the app source code.  Is this fixing an error where the examples _don't_ match the source, or is this trying to rename the fields to indicate to the reader that "this is an ID, not an object"?

tylerthedeveloper:
I haven't gone through part 5 yet but after a cursory review it doesn't like it would break anything - i can report back shortly to confirm this.

It is more of a description change than an error, as there are a few cases where there is a mapping of user : userId which could be avoided if they were one in the same - in renaming the fields it makes more sense by using id to indicate the type of (lookup) relationship, as opposed to the reader thinking its the object itself. 
I think its also important because this could be extended to where the the author component would make an API call to get the latest user info and compare it, which would also rely on the userId

docs/tutorials/essentials/part-4-using-data.md
```diff
@@ -464,7 +464,7 @@ const postsSlice = createSlice({
             title,
             content,
             // highlight-next-line
-            user: userId
+            userId: userId
```
timdorr, line 467: This can be shorthanded now:

```suggestion
            userId
```

docs/tutorials/essentials/part-4-using-data.md
```diff
@@ -605,7 +605,7 @@ Since we can't just put a `Date` class instance into the Redux store, we'll trac
             date: new Date().toISOString(),
             title,
             content,
-            user: userId,
+            userId: userId,
```
timdorr, line 608: This one too:

```suggestion
            userId,
```

tylerthedeveloper:
@timdorr 
There's a 3rd commit here - https://github.com/reduxjs/redux/pull/3963/commits/f46f65535ea6a78d43ae68820ca1801351bf66fd

Please let me know if you want me to make a separate PR for that?

docs/tutorials/essentials/part-4-using-data.md
tylerthedeveloper: agreed

docs/tutorials/essentials/part-4-using-data.md
tylerthedeveloper: agreed

tylerthedeveloper:
@timdorr  / @markerikson 

Although I agree with my suggestion, because of the API used in the subsequent tutorial page, this will be a breaking change.
The 2 options then would be 
- leave a not in existing doc commenting on why its user and not userId, in that its needed for a future API
- accept the proposed change and then we would also have to make additional changes on the next page as well, effectively reverting this change

obviously the first option seems easier, if desired 👍

PR #3964, opened 2020-12-09
[Tutorial Docs] Updating Initial state to account for expected changes in UI

tylerthedeveloper:
---
---

## Checklist

  - https://github.com/reduxjs/redux/issues/3967

- **Section**:
redux/docs/tutorials/essentials

- **Page**:
part-4-using-data.md

## What is the problem?

We change the UI to state that we will expect posts to have a reactions property in the ReactionButtons component, but we both don't check for null or update the initial state. 

In my PR i follow the same process used earlier in the tutorial with both the user and the date property on posts, by making sure to update the initial state objects of posts to have a reactions object property.

### DIFF
--- a/docs/tutorials/essentials/part-4-using-data.md
+++ b/docs/tutorials/essentials/part-4-using-data.md
@@ -803,6 +803,40 @@ export const ReactionButtons = ({ post }) => {
 
 Now, every time we click a reaction button, the counter should increment. If we browse around to different parts of the app, we should see the correct counter values displayed any time we look at this post, even if we click a reaction button in the `<PostsList>` and then look at the post by itself on the `<SinglePostPage>`.
 
+Whenever adding new properties to state or state objects, it is important to make sure to also update initial state if there are any defuault objects. Just like before when we added user and date to our initial state of posts, we also need to add 'initial reactions' as well.
+
+```jsx title="features/posts/postsSlice.js"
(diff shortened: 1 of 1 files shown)


tylerthedeveloper:
Added issue - https://github.com/reduxjs/redux/issues/3967

docs/tutorials/essentials/part-4-using-data.md
```diff
@@ -803,6 +803,40 @@ export const ReactionButtons = ({ post }) => {
 
 Now, every time we click a reaction button, the counter should increment. If we browse around to different parts of the app, we should see the correct counter values displayed any time we look at this post, even if we click a reaction button in the `<PostsList>` and then look at the post by itself on the `<SinglePostPage>`.
 
+Whenever adding new properties to state or state objects, it is important to make sure to also update initial state if there are any defuault objects. Just like before when we added user and date to our initial state of posts, we also need to add 'initial reactions' as well.
```
Shrugsy, line 806:
```suggestion
Whenever adding new properties to state or state objects, it is important to make sure to also update initial state if there are any default objects. Just like before when we added user and date to our initial state of posts, we also need to add 'initial reactions' as well.
```

Corrects a spelling error

markerikson:
Should be resolved by #4153 .

PR #3973, opened 2020-12-18
[Docs: Rephrase] Explain correctly what a thunk is

Sakai-san:
---
---

## Checklist

  - [issue 3972](https://github.com/reduxjs/redux/issues/3972)

- **Section**: Recipes
- **Page**: Usage With TypeScript

## What is the problem?

In the paragraph [Usage with Redux Thunk](https://redux.js.org/recipes/usage-with-typescript#usage-with-redux-thunk), the concept of thunk is wrongly explained.

> A thunk is a function that returns another function that takes parameters dispatch and getState.

This definition corresponds to a **thunk action creator** but not to a thunk.
Explain correctly what a thunk is.

### DIFF
--- a/docs/recipes/UsageWithTypescript.md
+++ b/docs/recipes/UsageWithTypescript.md
@@ -324,7 +324,7 @@ export default connector(MyComponent)
 
 ## Usage with Redux Thunk
 
-Redux Thunk is a commonly used middleware for writing sync and async logic that interacts with the Redux store. Feel free to check out its documentation [here](https://github.com/reduxjs/redux-thunk). A thunk is a function that returns another function that takes parameters `dispatch` and `getState`. Redux Thunk has a built in type `ThunkAction` which we can use to define types for those arguments:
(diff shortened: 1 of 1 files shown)


msutkowski:
I don't believe the current copy needs any changes. Per MDN, [`Any unit of code that can be evaluated to a value is an expression` ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators#Expressions) - which would be inaccurate in this instance, as a thunk doesn't necessarily evaluate to anything.

Sakai-san:
> I don't believe the current copy needs any changes. Per MDN, [`Any unit of code that can be evaluated to a value is an expression` ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators#Expressions) - which would be inaccurate in this instance, as a thunk doesn't necessarily evaluate to anything.

@msutkowski Thanks for your remark. I ll close the PR in this case. Just before doing so, I just mention here how [redux-thunk documentation](https://github.com/reduxjs/redux-thunk#whats-a-thunk) explains what a thunk is :  

Screenshot 2020-12-19 at 00 58 52

timdorr:
This seems fine to me. Thanks!

PR #3981, opened 2021-01-01
New function: subscribeProperties

vporton:
I added a new function `subscribeProperties`.

It is like `subscribe` but subscribes to changes only of a particular set (represented as an array) of properties of a state that is an object.


markerikson:
Appreciate the PR, but we're not looking to change the core API at this point in time.  You can always define this as your own store enhancer, though!

PR #3986, opened 2021-01-05
[Enhancement] Update isPlainObject

bcrazydreamer:
**[Enhancement] Update isPlainObject**

Update isPlainObject with a more optimal solution `(50% more faster then current one)`


Methuselah96:
Does this work across realms?

PR #3987, opened 2021-01-05
[Enhancement] Update isPlainObject

bcrazydreamer:
**[Enhancement] Update isPlainObject**

Update isPlainObject with a more optimal solution `(50% more faster then current one)`


timdorr:
Thanks, but this is already fast enough and doesn't need any further improvements or changes.

PR #3993, opened 2021-01-14
Added instructions for installing the package with yarn in the README.md file.

Prashoon123:
---
---

## PR Type

Update an existing page.

## Checklist

  - _link issue here_

Added instructions for how to install this package with yarn.

They make life easier for developers who use yarn as their package manager.


timdorr:
Thanks, but we don't need to iterate out every install command for every package manager. What about pnpm or dry? npm is still the bundled package manager with Node, so everyone has it. They may not use it, but it's most helpful to newbies to use the tools they have most readily available.

Prashoon123:
Ok @timdorr, it's our choice I thought this will be helpful but you have a very valid explanation. Thank you!! I hope to add some helpful PR in the near future!!!

PR #4002, opened 2021-01-25
Update egghead links in Learning Resources

utkarshsingh99:
## Checklist

      No 

- **Section**: Introduction
- **Page**: LearningResources

The link for a specific redux tutorial is not updated. It directs to a page that later asks the users to go to the new tutorial page. It would be better if the new link is added to save time and be on the safer side (just in case the older link is closed by the author).
No, they do not change any assumptions.


timdorr:
Thanks!

PR #4006, opened 2021-01-27
[Docs]: Add runonclick=1 param to CSB embeds

msutkowski:
---
---

Stems from an issue over on RTK - https://github.com/reduxjs/redux-toolkit/issues/877. This is an easy way to keep the CSB embed functionality in place while not freezing up some user's devices.


timdorr:
Thanks!

PR #4014, opened 2021-02-14
4.0.5 with frame dispatch v2

etaylib:
Thanks for the PR!


markerikson:
Hi. Looking at this PR, there's multiple reasons why we would not accept something like this:

- There's no description of what this is supposed to be or why you submitted it
- The PR includes a bunch of build artifacts, which should _definitely_ not be checked in
- It also actually tries to bump the release version
- We've already got a bunch of changes in the main branch that are waiting for release in a v5.0 build
- The actual logic changes here appear to be related to modifying the core store's notification behavior.  That can be implemented as a store enhancer, and does not need to be added to the store itself.

PR #4026, opened 2021-03-05
fix: Export type of $CombinedState

JacobLey:
---
name: Export type of $CombinedState
about: Support latest version of Typescript
---

Fix a bug with lastest version of Typescript

Exporting the result of `combineReducers` currently fails with latest version of Typescript, requires access to `$CombinedState`.

See `Bug Fixed -> What is the current behavior, and the steps to reproduce the issue?` for in-depth explanation

Only the type is exported, so it is not possible a consumer thinks the value is actually usable (no risk)

## Checklist

  - None, but can create if needs to be discussed further

## New Features
Typeof `$CombinedState` is exported an available to consumers

Adds support for Typescript exporting the result of `combineReducers` with latest version of typescript

None

Upgrade to [typescript v2.4.2](https://github.com/microsoft/TypeScript/releases/tag/v4.2.2)

Create a file `example.ts` with the following content:
```
import { combineReducers, $CombinedState } from 'redux';

export default combineReducers({})
```

Typescript will emit an error on:
`Default export of the module has or is using private name '$CombinedState'.`

No error should be emitted

By explicitly exporting the typeof `$CombinedState`, Typescript will be able to properly handle an exported `combineReducers`

Admittedly this mostly feels like an issue with Typescript (the type is already available to import) but it is an otherwise safe, non-breaking change so it is worth fixing at the Redux level

### DIFF
--- a/src/types/store.ts
+++ b/src/types/store.ts
@@ -20,6 +20,7 @@ export type ExtendState<State, Extension> = [Extension] extends [never]
  * Internal "virtual" symbol used to make the `CombinedState` type unique.
  */
 declare const $CombinedState: unique symbol
+export type { $CombinedState }
 
 /**
  * State base type for reducers created with `combineReducers()`.


markerikson:
The first issue here is that this PR won't actually fix anything atm. `master` is currently an unreleased conversion of the Redux core to TS that will (maybe, eventually, someday) become Redux 5.0.  

The current version is 4.0.5, which is still written in JS with attached typedefs.  That's on the `4.x` branch.

That said, I don't fully understand what the problem is, and would appreciate seeing an issue filed with a reproduction.

timdorr:
Hmm, is this something that changed in TS 4.2? I'm not seeing any reference to it. Our own internal use of these types doesn't produce these kinds of errors. 

It sounds like something is wrong with your build environment. Exporting a private symbol (which is only to ensure a distinct type for inference purposes) shouldn't be needed here, nor is it desired because we don't want folks relying on it or messing with it in some way. This isn't part of our API surface.

PR #4031, opened 2021-03-08
fix: Declare "EmptyObject" interface to wrap $CombinedState

JacobLey:
---
name: Declare EmptyObject interface to wrap $CombinedState
about: Adding interface for $CombinedState appeases latest version of Typescript, allows exporting result of `combineReducers`
---

Fix a bug with lastest version of Typescript

No changes to exports are included, but the internal change from type->interface allows Typescript to correctly reference/export usage of `$CombinedState`

## Checklist

  - closes https://github.com/reduxjs/redux/issues/4030
  - Previous PR, different attempt at solution https://github.com/reduxjs/redux/pull/4026

None, just maintains support of typescript with latest 4.2.x releases

None

Upgrade to typescript v2.4.2

Create a file `example.ts` with the following content:

```
import { combineReducers, $CombinedState } from 'redux';

export default combineReducers({})
```
make sure `declarations: true` is in your `tsconfig.json`

Typescript will emit an error on:
`Default export of the module has or is using private name '$CombinedState'.`

Also see https://github.com/reduxjs/redux/pull/4029 for minimal reproducible issue

No error should be emitted

`combineReducers<S>()` returns a type of `Reducer<CombinedState<S>>`, and the `CombinedState<S>` is the type in question in this PR

https://www.typescriptlang.org/docs/handbook/advanced-types.html#interfaces-vs-type-aliases
Digging a bit into how Typescript works, but basically a `type` is an alias. So any usage of a `type` that internally uses `$CombinedState` means the resulting declaration must also use `$CombinedState`. Since that value is not exported, Typescript fails.

Using an interface however does not need direct access to `$CombinedState`, so the exporting of `combineReducers` will magically succeed.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -53,16 +53,17 @@ declare const $CombinedState: unique symbol
  * typed as always undefined, so its never expected to have a meaningful
  * value anyway. It just makes this type distinquishable from plain `{}`.
  */
-export type CombinedState<S> = { readonly [$CombinedState]?: undefined } & S
+interface EmptyObject {
+  readonly [$CombinedState]?: undefined
+}
+export type CombinedState<S> = EmptyObject & S
 
 /**
  * Recursively makes combined state objects partial. Only combined state _root
  * objects_ (i.e. the generated higher level object with keys mapping to
  * individual reducers) are partial.
  */
-export type PreloadedState<S> = Required<S> extends {
-  [$CombinedState]: undefined
-}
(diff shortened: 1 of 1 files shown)


timdorr:
It looks like the issue is more that we're saying `readonly [$CombinedState]?: undefined` extends just `[$CombinedState]: undefined`. So, just a signature mismatch that's no longer allowed.

In any case, this both resolves it and keeps it from being a likely issue in the future. Thanks!

PR #4053, opened 2021-04-01
[Docs Rewrite] Remove "Using Immutable.js" page

AaronPowell96:
---
Name: :book: New/Updated Documentation Content
About: Adding a new docs page, or updating content in an existing docs page
---

## PR Type

## Checklist

  - https://github.com/reduxjs/redux/issues/4052

- **Deleted**: docs/recipes/UsingImmutableJS.md
- **Updated**: docs/FAQ.md
- **Updated**: docs/api/createStore.md
- **Updated**: docs/faq/ImmutableData.md
- **Updated**: docs/faq/ReactRedux.md
- **Updated**: docs/introduction/Ecosystem.md
- **Updated**: docs/recipes/README.md
- **Updated**: docs/recipes/Troubleshooting.md
- **Updated**: docs/recipes/structuring-reducers/BeyondCombineReducers.md
- **Updated**: docs/recipes/structuring-reducers/StructuringReducers.md
- **Updated**: docs/understanding/history-and-design/PriorArt.md
- **Updated**: website/_redirects
- **Updated**: website/sidebars.js

Updated the recommendation of Immutable.js to be Immer, I have assumed these libraries are hot-swappable and replacements for each other. It is critical to confirm this as otherwise, wording may need to be changed.

___________________________________

***Please see below key comments on changes I have made that should be questioned.***

### docs/FAQ.md - 
Deleted references to using Immutable.JS with Redux.
Changed "Do I have to use Immutable.JS" to "Do I have to use Immer" - Is this correct still?

### docs/faq/ImmutableData.md -
- JavaScript Immutability presentation PDF - Does mention Immutable.js, however, 
it also provides solid understandings of immutability so I think it is worth keeping it?

- Under "What approaches are there for handling data immutability? Do I have to use Immer?", 
I have kept the original Immutable.JS description, are these libraries swappable and the description is still correct?

### docs/recipes/structuring-reducers/BeyondCombineReducers.md - 
- Here, I have removed a direct reference to Immutable.JS but kept the "combineReducers" description about immutable maps in general.
- I swapped the description of using Slice reducers to recommend RTK that uses a createReducer utility, is this acceptable?

### website/_redirects -
- I have added redirects as I have seen fit, however, I couldn't seem to test them locally? (Tested in Netlify deploy preview and seems to work).
- I have separated them out from other redirects for now as they are very specific, please advise on their location.


markerikson:
Thanks for filing this!

First observation: no, Immer and Immutable.js are _not_ swappable for each other directly. They both operate in the same space (providing ways to immutably update values, and preventing accidental mutations), but they take very different approaches:

- Immutable.js is a collection of data types like `Map` and `List`, implemented using a specialized internal data structure. Operations like `Map.set()` return new instances immutably, instead of mutating the existing instance.  These data types are not compatible with plain JS data directly. Instead, you have to convert to/from plain JS data.
- Immer is a single `produce` function that uses Proxies to wrap your original state in a "draft" value. You can "mutate" that draft, and Immer tracks all the attempted changes. It then replays those mutations immutably to return the desired result, same as if you'd done a bunch of nested object spreads and array slices by hand.  Immer works specifically with plain JS data.

So, same problem space, very different approaches. If you're using Immutable.js already, it takes a bunch of work to migrate over to using Immer instead.

docs/api/createStore.md
```diff
@@ -50,7 +50,7 @@ console.log(store.getState())
 
 - Don't create more than one store in an application! Instead, use [`combineReducers`](combineReducers.md) to create a single root reducer out of many.
 
-- It is up to you to choose the state format. You can use plain objects or something like [Immutable](https://facebook.github.io/immutable-js/). If you're not sure, start with plain objects.
+- It is up to you to choose the state format. You can use plain objects or something like [Immer](https://github.com/immerjs/immer). If you're not sure, start with plain objects.
```
markerikson, line 53: :hand: This paragraph should be rewritten, along the lines of "Redux state is normally plain JS objects and arrays", and maybe just drop the "other formats" aspect.

docs/faq/ImmutableData.md
```diff
@@ -431,18 +430,18 @@ Note that, conversely, if the values in your props object refer to mutable objec
 - [Building Efficient UI with React and Redux](https://www.toptal.com/react/react-redux-and-immutablejs)
 - [ImmutableJS: worth the price?](https://medium.com/@AlexFaunt/immutablejs-worth-the-price-66391b8742d4#.a3alci2g8)
 
-## What approaches are there for handling data immutability? Do I have to use Immutable.JS?
+## What approaches are there for handling data immutability? Do I have to use Immer?
 
-You do not need to use Immutable.JS with Redux. Plain JavaScript, if written correctly, is perfectly capable of providing immutability without having to use an immutable-focused library.
```
markerikson, line 436: :speech_balloon: Yeah, this entry reads okay based on those changes.

docs/faq/ImmutableData.md
```diff
@@ -464,9 +463,7 @@ Operating on JavaScript objects and arrays in an immutable way can be slow, part
 
 Remember, to change an immutable object, you must mutate a _copy_ of it, and copying large objects can be slow as every property must be copied.
 
-In contrast, immutable libraries such as Immutable.JS can employ sophisticated optimization techniques such as [structural sharing](https://www.slideshare.net/mohitthatte/a-deep-dive-into-clojures-data-structures-euroclojure-2015) , which effectively returns a new object that reuses much of the existing object being copied from.
-
-For copying very large objects, [plain JavaScript can be over 100 times slower](https://medium.com/@dtinth/immutable-js-persistent-data-structures-and-structural-sharing-6d163fbd73d2#.z1g1ofrsi) than an optimized immutable library.
+In contrast, immutable libraries such as Immer can employ sophisticated optimization techniques such as [structural sharing](https://www.slideshare.net/mohitthatte/a-deep-dive-into-clojures-data-structures-euroclojure-2015) , which effectively returns a new object that reuses much of the existing object being copied from.
```
markerikson, line 466: :hand: Immer doesn't use any "sophisticated data structures". It _does_ say it uses "structural sharing", but Immer uses that phrase in a different way than the Immutable.js docs do.

We should probably drop the link to that "Clojure data structures" article, and I don't think the phrase "sophisticated optimization techniques" applies here.  Really, all it's doing is copying only the parts of the object structure that need to be changed, and reusing the ones that aren't changing.

docs/introduction/Ecosystem.md
```diff
@@ -361,13 +351,6 @@ const newData = update(myData, {
 })
 ```
 
-**[mariocasciaro/object-path-immutable](https://github.com/mariocasciaro/object-path-immutable)** <br />
```
markerikson, line 364: :speech_balloon: Good catch!

docs/introduction/Ecosystem.md
```diff
@@ -376,17 +359,6 @@ const newState = dotProp.set(state, `todos.${index}.complete`, true)
...
-**[gajus/redux-immutable](https://github.com/gajus/redux-immutable)** <br />
-combineReducers equivalent that works with Immutable.js Maps
-
-```js
-const initialState = Immutable.Map()
-const rootReducer = combineReducers({})
-const store = createStore(rootReducer, initialState)
-```
-
 **[eadmundo/redux-seamless-immutable](https://github.com/eadmundo/redux-seamless-immutable)** <br />
```
markerikson, line 362: :hand: We might as well remove this `seamless-immutable`-related lib too, since Immer obsoletes that also.

docs/recipes/structuring-reducers/BeyondCombineReducers.md
```diff
@@ -7,13 +7,13 @@ hide_title: true
 
 # Beyond `combineReducers`
 
-The `combineReducers` utility included with Redux is very useful, but is deliberately limited to handle a single common use case: updating a state tree that is a plain Javascript object, by delegating the work of updating each slice of state to a specific slice reducer. It does _not_ handle other use cases, such as a state tree made up of Immutable.js Maps, trying to pass other portions of the state tree as an additional argument to a slice reducer, or performing "ordering" of slice reducer calls. It also does not care how a given slice reducer does its work.
+The `combineReducers` utility included with Redux is very useful, but is deliberately limited to handle a single common use case: updating a state tree that is a plain Javascript object, by delegating the work of updating each slice of state to a specific slice reducer. It does _not_ handle other use cases, such as a state tree made up of immutable Maps, trying to pass other portions of the state tree as an additional argument to a slice reducer, or performing "ordering" of slice reducer calls. It also does not care how a given slice reducer does its work.
```
markerikson, line 10: :hand: We can leave the reference to "Immutable.js Maps" here, because that statement is still accurate.

docs/recipes/structuring-reducers/BeyondCombineReducers.md
```diff
@@ -7,13 +7,13 @@ hide_title: true
...
 
-## Using slice reducers with Immutable.js objects
+## Using slice reducers with Immer objects
 
-Since `combineReducers` currently only works with plain Javascript objects, an application that uses an Immutable.js Map object for the top of its state tree could not use `combineReducers` to manage that Map. Since many developers do use Immutable.js, there are a number of published utilities that provide equivalent functionality, such as [redux-immutable](https://github.com/gajus/redux-immutable). This package provides its own implementation of `combineReducers` that knows how to iterate over an Immutable Map instead of a plain Javascript object.
+Since `combineReducers` currently only works with plain Javascript objects, an application that uses an Immer Map object for the top of its state tree could not use `combineReducers` to manage that Map. [Redux Toolkit](https://redux-toolkit.js.org/) includes a [createReducer utility](https://redux-toolkit.js.org/api/createReducer) that applies updates immutably.
```
markerikson, line 16: :hand: The changes to this paragraph aren't really correct.  Let's just delete this entire heading + paragraph completely - none of it matters at this point.

docs/recipes/structuring-reducers/StructuringReducers.md
```diff
@@ -21,7 +21,7 @@ It is vital that these Prerequisite Concepts are **thoroughly understood** befor
 
 #### [Prerequisite Concepts](PrerequisiteConcepts.md)
 
-It's also important to note that some of these suggestions may or may not be directly applicable based on architectural decisions in a specific application. For example, an application using Immutable.js Maps to store data would likely have its reducer logic structured at least somewhat differently than an application using plain Javascript objects. This documentation primarily assumes use of plain Javascript objects, but many of the principles would still apply if using other tools.
+It's also important to note that some of these suggestions may or may not be directly applicable based on architectural decisions in a specific application. For example, an application using Immer Maps to store data would likely have its reducer logic structured at least somewhat differently than an application using plain Javascript objects. This documentation primarily assumes use of plain Javascript objects, but many of the principles would still apply if using other tools.
```
markerikson, line 24: :hand: Honestly, this whole page needs to get moved somewhere else and redone :)  But for now, let's just rewrite the paragraph as:

> Standard Redux architecture relies on using plain JS objects and arrays for your state. If you're using an alternate approach for some reason, the details may differ based on your approach, but many of the principles will still apply.

docs/understanding/history-and-design/PriorArt.md
```diff
@@ -15,7 +15,7 @@ Redux was inspired by several important qualities of [Flux](https://facebook.git
...
 
-Another important difference from Flux is that **Redux assumes you never mutate your data**. You can use plain objects and arrays for your state just fine, but mutating them inside the reducers is strongly discouraged. You should always return a new object, which is easy with the [object spread operator proposal](../../recipes/UsingObjectSpreadOperator.md), or with a library like [Immutable](https://facebook.github.io/immutable-js).
+Another important difference from Flux is that **Redux assumes you never mutate your data**. You can use plain objects and arrays for your state just fine, but mutating them inside the reducers is strongly discouraged. You should always return a new object, which is easy with the [object spread operator proposal](../../recipes/UsingObjectSpreadOperator.md).
```
markerikson, line 18: :hand: Let's rewrite this a bit:

```suggestion
Another important difference from Flux is that **Redux assumes you never mutate your data**. You can use plain objects and arrays for your state just fine, but mutating them inside the reducers is strongly discouraged. You should always return a new object, which can be done using the [object spread operator](../../recipes/UsingObjectSpreadOperator.md) or [the Immer immutable update library](https://immerjs.github.io/immer/).
```

docs/understanding/history-and-design/PriorArt.md
```diff
@@ -27,11 +27,11 @@ For what it's worth, Flux's creators [approve](https://twitter.com/jingc/status/
 
 Unlike Redux, Elm is a language, so it is able to benefit from many things like enforced purity, static typing, out of the box immutability, and pattern matching (using the `case` expression). Even if you don't plan to use Elm, you should read about the Elm architecture, and play with it. There is an interesting [JavaScript library playground implementing similar ideas](https://github.com/paldepind/noname-functional-frontend-framework). We should look there for inspiration on Redux! One way that we can get closer to the static typing of Elm is by [using a gradual typing solution like Flow](https://github.com/reduxjs/redux/issues/290).
 
-### Immutable
+### Immer
 
-[Immutable](https://facebook.github.io/immutable-js) is a JavaScript library implementing persistent data structures. It is performant and has an idiomatic JavaScript API.
+[Immer](https://immerjs.github.io/immer/) is a JavaScript library implementing persistent data structures. It is performant and has an idiomatic JavaScript API. Immer operates on normal JavaScript objects and arrays.
```
markerikson, line 32: :hand: Immer does _not_ use "persistent data structures".  

I would actually say let's revert the changes to this paragraph entirely, because Redux _was_ inspired by Immutable.js

docs/understanding/history-and-design/PriorArt.md
```diff
@@ -27,11 +27,11 @@ For what it's worth, Flux's creators [approve](https://twitter.com/jingc/status/
...
 
-Immutable and most similar libraries are orthogonal to Redux. Feel free to use them together!
+Immer and most similar libraries are orthogonal to Redux. Feel free to use them together!
```
markerikson, line 34: :hand: but let's rewrite this sentence:

```suggestion
(Note that while Immutable.js helped inspire Redux, today we recommend [using Immer for immutable updates instead](../../style-guide/style-guide.md#use-immer-for-writing-immutable-updates).)
```

markerikson (changes requested):

docs/introduction/Ecosystem.md
AaronPowell96: There is also "rtfeldman/seamless-immutable" under "Data Structures", should we remove that too?

docs/recipes/structuring-reducers/BeyondCombineReducers.md
AaronPowell96: Deleted, sorry about that!

AaronPowell96:
Thanks for the explanation! 
I have made all the changes requested except this comment here: https://github.com/reduxjs/redux/pull/4053#discussion_r605795450 
I closed the comments as "resolved" but then realised that might be something you need to do? Hopefully, all is okay!

markerikson:
Y'know, let's just delete _everything_ under the "Immutable Data" header in the "Ecosystem" page except for Immer. None of those other libs are relevant any more now that Immer exists.

(and yeah, marking comments as "resolved" is fine if you've made the relevant changes)

docs/introduction/Ecosystem.md
AaronPowell96: Removed everything under the "Immutable Data" header, except Immer (Changed the link to Immer as well).

markerikson:
Heh, I _love_ pasting in old URLs into previews and watching the redirects kick in :)

Awright, this is probably good enough for now. Thanks, really appreciate you doing this!

PR #4055, opened 2021-04-02
Rewrite Redux core error messages

markerikson:
---
---

This PR:

- Rewrites most of the Redux core error messages by:
    - Improving runtime typechecks to show a better description of the actual variable type
    - Clarifying wording
    - Adding links to relevant docs pages
- Overwrites the existing `errors.json` file with the updated errors, since none of our errors have been released yet
- Turns on CodeSandbox-CI
- Updates the error extraction script to ignore a Babel Runtime-specific error message


errors.json
```diff
@@ -1,19 +1,19 @@
...
-  "16": "Super expression must either be null or a function"
+  "10": "Expected the nextReducer to be a function. Instead, received: '",
+  "11": "Expected the observer to be an object. Instead, received: ''",
+  "12": "The slice reducer for key \"\" returned undefined during initialization. If the state passed to the reducer is undefined, you must explicitly return the initial state. The initial state may not be undefined. If you don't want to set a value for this reducer, you can use null instead of undefined.",
+  "13": "The slice reducer for key \"\" returned undefined when probed with a random type. Don't try to handle  or other actions in \"redux/*\" namespace. They are considered private. Instead, you must return the current state for any unknown actions, unless it is undefined, in which case you must return the initial state, regardless of the action type. The initial state may not be undefined, but can be null.",
```
msutkowski, line 15: > Don't try to handle  or other actions in \"redux/*\" namespace.

Is this missing a char or '' to get replaced that I'm not seeing?

src/createStore.ts
```diff
@@ -105,7 +114,7 @@ export default function createStore<
   let isDispatching = false
 
   /**
-   * This makes a shallow copy of currentListeners so we can use
+   * This makes a shallow copy of currentListeners so can use
    * nextListeners as a temporary list while dispatching.
```
msutkowski, line 118:
```suggestion
   * This makes a shallow copy of currentListeners so nextListeners
   * can be used as a temporary list while dispatching.
```

errors.json
markerikson: There's a template literal interpolation there:

```js
`Don't try to handle ${ActionTypes.INIT} or other actions
```

and that doesn't get included in the extracted message.

I'll add single quotes around it, though.

markerikson:
Going in! Next up, porting this whole set of changes to `4.x`

Issue #2751, opened 2017-12-07
should `dispatch` wait until the next event loop to notify its listeners?


warren-bank:
Hi. Today is my first day learning about redux, so maybe this is a dumb question. I was walking through the code in `applyMiddleware` to see how it all works. It wasn't immediately obvious to me, but I understand it now.

In tracing that code, I noticed something that seemed a bit off. So I just wanted to share the observation..

```javascript
function applyMiddleware(...middlewares) {
  let chain = middlewares.map(middleware => middleware({getState, dispatch}))
  // chain is now an Array of functions: (next) => (action) => {}

  let dispatch = compose(...chain)(store.dispatch)
```

To quickly summarize the path the code takes:
* assuming `applyMiddleware` is called with: `(middleware_1, middleware_2, ..., middleware_N)`
* `dispatch(action)` calls: `middleware_1(next_1)(action)`
  * where `next_1` is a reference to: `middleware_2(middleware_3(...(middleware_N(store.dispatch))))`
* `next_1(action)` calls: `middleware_2(next_2)(action)`
  * where `next_2` is a reference to: `middleware_3(...(middleware_N(store.dispatch)))`
* this iterative process continues until the last middleware in the chain is reached: `middleware_N`
  * where `next_N` is a reference to: `store.dispatch`

So:
* `middleware_1` executes its code up until it calls `next_1(action)`
* `middleware_2` executes its code up until it calls `next_2(action)`
* ...
* `middleware_N` executes its code up until it calls `next_N(action)`
* `store.dispatch`:
  * runs the root reducer
  * notifies its listeners of any state change
  * returns
* `middleware_N` completes the execution of its code resuming from after its call to `next_N(action)`
* ...
* `middleware_2` completes the execution of its code resuming from after its call to `next_2(action)`
* `middleware_1` completes the execution of its code resuming from after its call to `next_1(action)`

The possible issue that I see is that `store.dispatch` notifies its listeners before the middleware has the opportunity to complete its execution.
* if a listener dispatches additional actions in response to the new state, then the middleware begins to execute the new action before it has the opportunity to finish executing the previous action
* imho, it seems to me that there should really be a `nextTick()` or `setTimeout(0)` before the listeners are notified

_example_:

```javascript
function dispatch(action) {
  currentState = currentReducer(currentState, action)

  setTimeout(() => {
    for (let i = 0; i < nextListeners.length; i++) {
      nextListeners[i]()
    }      
  }, 0)
}
```

Thoughts?

[...]

PR #4061, opened 2021-04-06
Notify listeners after middleware

georeith:
---
about: Listeners being notified after 
---

I would classify it as a bug

See: https://github.com/reduxjs/redux/issues/4049

## Checklist

- [X] Have you added an explanation of what your changes do and why you'd like us to include them?
  - https://github.com/reduxjs/redux/issues/4049
- [X] Have the tests been updated to match the changes in the PR?
- [X] Have you run the tests locally to confirm they pass?

I don't believe any are required

See: https://github.com/reduxjs/redux/issues/4049

Currently listeners are notified when you call `next()` inside middleware, if one of those listeners leads to another action being dispatched then the middleware is nested and that action is dispatched before the rest of the middleware is completed.

As a result middleware is unable to with a guarantee access the state after an action was executed, and perhaps worse in the above case middleware is called twice with the same state.

See the regression test I have added here:

https://github.com/georeith/redux/blob/22e6c3ee1dc454e34d3e4c716cbc09333863f4fc/test/applyMiddleware.spec.ts#L151

Without this change the calls are as follows:

 ```ts
 [
  [
    [
      { id: 1, text: 'Use Redux' },
      { id: 2, text: 'Flux FTW!' }
    ]
  ],
  [
    [
      { id: 1, text: 'Use Redux' },
      { id: 2, text: 'Flux FTW!' }
    ]
  ]
]
 ```

A user will run into this issue if they dispatch from React's `useLayoutEffect` for instance, which they may need to do if they need to update the store based on new information from a browser layout cycle.

After the change the calls are:

```ts
[
  [
    [
      { id: 1, text: 'Use Redux' },
    ]
  ],
  [
    [
      { id: 1, text: 'Use Redux' },
      { id: 2, text: 'Flux FTW!' }
    ]
  ]
]
```

With each middleware receiving the state that resulted from the action that triggered it.

Exposes some internal methods under symbols for controlling the notification of listeners so that `applyMiddleware` is able to control this flow better.

I looked at two different ways of doing this, my first attempt was to expose `$$dispatchSilent` and `$$notifyListeners` and have `applyMiddleware` compose `$$dispatchSilent` which worked for the most part but broke if you use Redux Dev Tools as that attempts to override `store.dispatch` for middleware. So I arrived upon this solution which should not affect further composition above the middleware.

### DIFF
--- a/src/applyMiddleware.ts
+++ b/src/applyMiddleware.ts
@@ -8,6 +8,7 @@ import {
   StoreEnhancerStoreCreator
 } from './types/store'
 import { Reducer } from './types/reducers'
+import { $$notifyListeners, $$toggleListeners } from './createStore'
 
 /**
  * Creates a store enhancer that applies middleware to the dispatch method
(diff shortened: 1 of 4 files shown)


timdorr:
Thanks, but we've already been over this a while ago in #2751. This behavior is intentional and should not be changed.

If you need to get into the post-dispatch execution chain, you should just add yourself as a subscriber to the store. It's still discouraged to dispatch synchronously in a subscriber, but you can work around this by issuing that dispatch asychronously (setTimeout is a quick way to do that).

markerikson:
Ah, forgot we'd had that issue way back when.

Per my prior comments, I still think this can likely be done as a store enhancer in userland without needing to modify the store itself.

PR #4067, opened 2021-04-16
chore: Docusaurus with webpack 5

slorber:
Use an unreleased version of Docusaurus to test the webpack5 support works for this site.

See also https://github.com/facebook/docusaurus/pull/4089

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -9,5 +9,5 @@
   package = "netlify-plugin-cache"
   [plugins.inputs]
     paths = [
-      "website/node_modules/.cache",
+      "node_modules/.cache",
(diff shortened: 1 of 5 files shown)


msutkowski (commented):
I went through the majority of the content in both light & dark modes, and everything looks fine and nothing is obviously wrong. I spot-checked a handful of redirects, and no issues there either. All of the CSBs work as expected.

I'll give these changes a try against the RTK docs as well and see if there are any compatibility issues with the remark TS plugin.

timdorr:
LGTM

slorber:
thanks :)

PR #4072, opened 2021-04-21
[Doc] crossSliceReducer example should not update slice B twice

### LINKED ISSUE #4070: crossSliceReducer example is buggy-maybe
https://redux.js.org/recipes/structuring-reducers/beyond-combinereducers

- **Section**:  recipes
- **Page**: structuring-reducers/beyond-combinereducers

## What is the problem?
I believe the example of the third approach with the `crossSliceReducer` has a bug.
The `sliceReducerB` will be executed twice.

Here is the example from the doc with an added inline comment pointing the problem.
```
const combinedReducer = combineReducers({
  a: sliceReducerA,
  b: sliceReducerB
})

function crossSliceReducer(state, action) {
  switch (action.type) {
    case 'SOME_SPECIAL_ACTION': {
      return {
        // specifically pass state.b as an additional argument
        a: handleSpecialCaseForA(state.a, action, state.b),
        b: sliceReducerB(state.b, action)  //  ⚠🐛 state.b has already been updated by sliceReducerB in the combined reducer.
      }
    }
    default:
      return state
  }
}

function rootReducer(state, action) {
  const intermediateState = combinedReducer(state, action)
  const finalState = crossSliceReducer(intermediateState, action)
  return finalState
}
```

If we take the following simple reducer 
```
function sliceReducerB(stateB = {counter: 0}, _action) {
 return {counter: counter+1}
}
```

We will have the following state (only showing slice B) :
```
function rootReducer(state, action) {
  const intermediateState = combinedReducer(state, action)
  // intermediateState is {b: counter: 1,  }
  const finalState = crossSliceReducer(intermediateState, action)
  // finalState is {b: counter: 2} 😿
  return finalState
}

```

The crossSliceReducer should only update a.

lazybean:
We do not want the slice B to be updated two times per action.

---
---

## Checklist

  - https://github.com/reduxjs/redux/issues/4070
  - Not sure how to do that from github directlu

https://redux.js.org/recipes/structuring-reducers/beyond-combinereducers
- **Section**: recipes
- **Page**: structuring-reducers/beyond-combinereducers

## What is the problem?
The example of the 3rd approach is buggy.

We avoid updating the sliceB twice.

### DIFF
--- a/docs/recipes/structuring-reducers/BeyondCombineReducers.md
+++ b/docs/recipes/structuring-reducers/BeyondCombineReducers.md
@@ -28,7 +28,8 @@ function combinedReducer(state, action) {
       return {
         // specifically pass state.b as an additional argument
         a: sliceReducerA(state.a, action, state.b),
-        b: sliceReducerB(state.b, action)
+        // do not update the other slices
+        b: state.b
       }
     }
     case 'ANOTHER_SPECIAL_ACTION': {


lazybean:
Need coffee first.

PR #4073, opened 2021-04-21
[Doc] crossSliceReducer example should not update slice B twice

lazybean:
---
---

## Checklist

  - #4070
 - Not sure how to do that from github ui direclty

https://redux.js.org/recipes/structuring-reducers/beyond-combinereducers
- **Section**: recipes/structuring-reducers
- **Page**: beyond-combinereducers

## What is the problem?
The example of the 3rd approach is buggy.

We avoid updating the sliceB twice.


markerikson:
Closing per comment in the issue - the logic here is sufficiently reasonable as-is given the intent of the example.  Thanks anyway!

PR #4074, opened 2021-04-21
Tutorial Documentation edit - Being explicit about storeAPI object

jj05y:
Adding a small fix and some extra clarity to a point of friction I found in the tutorial

### DIFF
--- a/docs/tutorials/fundamentals/part-4-store.md
+++ b/docs/tutorials/fundamentals/part-4-store.md
@@ -452,6 +452,17 @@ Here's the same example as above, using arrow functions:
 ```js
 const anotherExampleMiddleware = storeAPI => next => action => {
   // Do something in here, when each action is dispatched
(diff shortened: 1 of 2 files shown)


docs/tutorials/fundamentals/part-4-store.md
```diff
@@ -452,6 +452,18 @@ Here's the same example as above, using arrow functions:
...
+  // Access the store with storeAPI.getState()
+
+  return next(action)
+}
+```
+
+Alternatively, storeAPI can be decomposed
+
+```js
+const anotherExampleMiddleware = {dispatch, getState} => next => action => {
```
markerikson, line 464:
```suggestion
const anotherExampleMiddleware = ({dispatch, getState}) => next => action => {
```

timdorr:
Thanks, but I think getting into destructuring isn't all that helpful here. We can keep the example fix, though.

Issue #4083, opened 2021-05-14
Broken in-page links in Immutable Data FAQ page


scitech:
## What docs page needs to be fixed?

- **Section**: FAQ
- **Page**: Immutable Data

## What is the problem?

All of the in-page anchor links on the first bullet point here are pointing to IDs that don't exist on the page:

<img width="400" alt="image" src="https://user-images.githubusercontent.com/4130722/118308842-0d955180-b4b2-11eb-8921-9c15d50ea620.png">

## What should be changed to fix the problem?

Links should be updated to the correct IDs

PR #4084, opened 2021-05-14
Fix broken in-page links in immutable data FAQ

scitech:
---
about: Broken in-page links on Immutable Data FAQ 
---

## Checklist

  - #4083 

- **Section**: FAQ
- **Page**: Immutable Data

## What is the problem?

All of the in-page anchor links on the first bullet point here are pointing to IDs that don't exist on the page:

image

It looks like these headings were rephrased at some point so I updated the links to point to the newer headings.


timdorr:
Thanks!

Issue #4089, opened 2021-05-22
Use of miniKindOf is not ES5-compliant


embeddedt:
### Prior Issues

None that I could find.

### What is the current behavior?

Redux causes my application to crash on some ES5 browsers because of a [nested function within an `if` statement in kindOf.ts](https://github.com/reduxjs/redux/blob/5f29b2667ff83f0189bc2aa304b214b4e19fa398/src/utils/kindOf.ts#L4-L6). This is not permitted in ES5 and Babel does not seem to provide a transform for this, probably because it would break existing code.

The exact console log is this: `SyntaxError: Nested functions cannot be declared in strict mode.`

### Steps to Reproduce

It's a bit hard to create a live repro case for this, since finding an ES5 browser is difficult nowadays. However, I can provide [a gist](https://gist.github.com/embeddedt/0f40117a6f83cffce012ff5783860905) with my Babel and browserslist configurations as well as a file highlighting the issue.

### What is the expected behavior?

Redux should run in an ES5 environment with a standard Babel configuration.

### Environment Details

Browser: Midori 0.5.11 on Ubuntu 18.04

markerikson:
Huh. This is the first I've ever heard that nested functions aren't legal in some cases. Can you point to some kind of a reference for where this restriction is described?

embeddedt:
> This is the first I've ever heard that nested functions aren't legal in some cases.

Same here until I ran into the error and did some research on it.

Here are some references:

* https://stackoverflow.com/a/10069457
* https://github.com/acornjs/acorn/issues/118
* http://linterrors.com/js/function-statements-should-not-be-placed-in-blocks (in particular, the section further down on syntax errors)

My understanding of this is that ES5 has extremely loose scoping rules. Normally, the function definition would be hoisted out of the if statement, so in strict mode, they chose to prevent confusion by refusing to allow functions to be declared inside if statements.

This error is rarely observed in practice for two reasons:

* Many ES5 browsers allow this type of declaration, but with an inconsistent scope. Nonetheless, for cases like this the inconsistent scoping probably wouldn't be apparent.
* ES6 allows this behavior as the scoping rules were changed there, so any browser which supports ES6 won't choke on this code.

markerikson:
Yeah, definitely a pretty rare edge case :)

tbh this is low priority for me atm, as we're focusing on getting the RTK Query APIs out the door.  If you'd like to submit a PR that rearranges the code to make this work as expected, we can look at getting that in.

[...]

PR #4090, opened 2021-05-22
Move miniKindOf out of if scope to fix ES5 compatibility issue

embeddedt:
---
---

It fixes a bug.

Currently redux cannot be run in ES5 environments even if Babel is used.

## Checklist

  - #4089

Please refer to #4089.

**TL;DR** There are functions declared inside an `if` statement, which ES5 forbids and Babel does not fix.

Redux should run without any syntax errors, if it's being bundled and transpiled in the standard manner for ES5.

It moves the aforementioned functions out of the if statement and into the file's scope.

The diff looks huge but it's mostly due to the indentation changing.

### DIFF
--- a/src/utils/kindOf.js
+++ b/src/utils/kindOf.js
@@ -1,68 +1,68 @@
+// Inlined / shortened version of `kindOf` from https://github.com/jonschlinkert/kind-of
+function miniKindOf(val) {
+  if (val === void 0) return 'undefined'
+  if (val === null) return 'null'
+
+  const type = typeof val
+  switch (type) {
+    case 'boolean':
+    case 'string':
+    case 'number':
+    case 'symbol':
+    case 'function': {
+      return type
+    }
+    default:
+      break
+  }
+
+  if (Array.isArray(val)) return 'array'
+  if (isDate(val)) return 'date'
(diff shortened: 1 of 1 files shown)


markerikson:
FYI, this really needs to be based against the `4.x` branch if we're going to release it.  `master` is an unreleased TypeScript conversion.

embeddedt:
> FYI, this really needs to be based against the 4.x branch if we're going to release it. master is an unreleased TypeScript conversion.

Thanks; I've force pushed a version based off 4.x.

> Also, I'd like to make sure that the function only actually exists in dev mode to save on byte size, but that itself requires an `if(process.env.NODE_ENV === 'development')` check. Not sure if there's a good way to resolve that.

Assuming that production environments are using something like terser, wouldn't it automatically see the function as dead code and remove it, since the `NODE_ENV` check will be replaced by `false`?

I've tested it locally by replacing the check with `'production' !== 'production'` and running `npx terser --module -c dead_code src/utils/kindOf.js` and it does get eliminated.

markerikson:
Okay, sounds good.  There's a couple other little bits that probably need to go into a 4.1.1, so I'll see if I can get this out in the next couple days.

PR #4117, opened 2021-06-25
[Docs] Remove dummy content and fix header issues

deminoth:
---
---

## Checklist

  - _link issue here_
- [v] Have the files been linted and formatted?

All

## What is the problem?

From Docusaurus v2, `hide_title: true` frontmatter hides h1 title from their page. 

Recent commits 264c21779f328093d6711d10f41140a35ca631dc and 757377326a48403203a44fa4d0f3b26072d002ea adds dummy content to work around this issue. But I recommend removing `hide_title: true` and using `sidebar_label` when we need it.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -1,11 +1,9 @@
 ---
 id: faq
 title: FAQ Index
-hide_title: true
+sidebar_label: FAQ Index
 description: 'FAQ Index: Frequently Asked Questions about Redux'
 ---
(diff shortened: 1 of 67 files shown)


timdorr:
IIRC, this was a temporary thing that was resolve in the final 2.0 release. I think. I could be misremembering.

Anyways, it looks like everything matches up between the preview and live sites. Thanks!

PR #4119, opened 2021-06-26
Rewrite docs content on selectors

markerikson:
This PR:

- Adds a new Style Guide rule emphasizing to keep state minimal and derive data when possible
- Adds a new "Deriving Data with Selectors" usage guide page that provides extensive info on what selectors are, why they're important, how to use Reselect, alternative selector libs, usage with React-Redux, and specific usage patterns
- Replaces links to the old "Computing Derived Data" page to point to the new "Deriving Data with Selectors" page, and removes the legacy page


docs/style-guide/style-guide.md
```diff
@@ -364,6 +364,18 @@ Many applications need to cache complex data in the store. That data is often re
...
+
+Whenever possible, **keep the actual data in the Redux store as minimal as possible, and _derive_ additional values from that state as needed**. This includes things like calculating filtered lists or summing up values. As an example, a todo app would keep an original list of todo objects in state, but derive a filtered list of todos outside the state whenever the state is updated. Similarly, a check for whether all todos have been completed, or number of todos remaining, can be calculated outside the store as well.
+
+This has several benefits:
+
+- The actual state is easier to read
+- Less logic is needed to calculate those additional values and keep them in sync with the rest of the data
+- The original state is still there as a reference and isn't being replaced
+
+Deriving data is often done in "selector" functions, which can encapsulate the logic for doing the derived data calculations. In order to improve performance, these selectors can be _memoized_ to cache previous results, using libraries like `reselect` and `proxy-memoize.
```
msutkowski, line 377:
```suggestion
Deriving data is often done in "selector" functions, which can encapsulate the logic for doing the derived data calculations. In order to improve performance, these selectors can be _memoized_ to cache previous results, using libraries like `reselect` and `proxy-memoize`.
```

docs/usage/deriving-data-selectors.md
```diff
@@ -0,0 +1,624 @@
...
+function selectSomeSpecificField(state) {
+  return state.some.deeply.nested.field
+}
+
+// Arrow function, deriving values from an array
+const selectItemsWhoseNamesStartWith = (items, namePrefix) =>
+  items.filter(item => item.name.startsWith(namePrefix))
+```
+
+A selector function can have any name you want. However, [**we recommend prefixing selector function names with the word `select` combiend with a description of the value being selected**](../style-guide/style-guide.md#name-selector-functions-as-selectthing). Typical examples of this would look like **`selectTodoById`**, **`selectFilteredTodos`**, and **`selectVisibleTodos**`.
```
msutkowski:
```suggestion
A selector function can have any name you want. However, [**we recommend prefixing selector function names with the word `select` combined with a description of the value being selected**](../style-guide/style-guide.md#name-selector-functions-as-selectthing). Typical examples of this would look like **`selectTodoById`**, **`selectFilteredTodos`**, and **`selectVisibleTodos**`.
```

docs/usage/deriving-data-selectors.md
```diff
@@ -0,0 +1,624 @@
...
+
+The first reason to use selector functions is for encapsulation and reusability when dealing with your Redux state shape.
+
+Let's say that one of your `useSelector` hooks makes a very specific lookup into part of your Redux state:
+
+```js
+const data = useSelector(state => state.some.deeply.nested.field)
+```
+
+That's a legal statement. But, imagine that you've got several components that need to access that field. What happens if you need to make a change to where that piece of state lives? You would now have to go change _every_ `useSelector` hook that references that value. So, in the same way that [we recommend using action creators to encapsulate details of creating actions](../style-guide/style-guide.md#use-action-creators), we recommend defining reusable selectors to encapsulate the knowledge of where a given piece of state lives. Then, you can use a given selector function many times in the codebase, anywhere that your app needs to retrieve that particular data.
```
msutkowski, line 127:
```suggestion
That's a valid selector. But, imagine that you've got several components that need to access that field. What happens if you need to make a change to where that piece of state lives? You would now have to go change _every_ `useSelector` hook that references that value. So, in the same way that [we recommend using action creators to encapsulate details of creating actions](../style-guide/style-guide.md#use-action-creators), we recommend defining reusable selectors to encapsulate the knowledge of where a given piece of state lives. Then, you can use a given selector function many times in the codebase, anywhere that your app needs to retrieve that particular data.
```

msutkowski: Just an idea - not sure if 'legal statement' would be confusing for translation purposes.

docs/usage/deriving-data-selectors.md
```diff
@@ -0,0 +1,638 @@
...
+```js
+import { createSelector } from 'reselect'
+
+const selectTodoDescriptionsReselect = createSelector(
+  [state => state.todos],
+  todos => todos.map(todo => todo.text)
+)
+```
+
+This will unfortunately recalculate the derived array if any other value inside of `state.todos` changes, such as toggling a `todo.completed` flag. The _contents_ of the derived array are identical, but because the input `todos` array changed, it has to calculate a new output array, and that has a new reference.
```
msutkowski, line 395:
```suggestion
Unfortunately, this will recalculate the derived array if any other value inside of `state.todos` changes, such as toggling a `todo.completed` flag. The _contents_ of the derived array are identical, but because the input `todos` array changed, it has to calculate a new output array, and that has a new reference.
```

docs/usage/deriving-data-selectors.md
```diff
@@ -0,0 +1,638 @@
...
+export const { todoAdded } = todosSlice.actions
+export default todosSlice.reducer
+
+// highlight-start
+// Export a reusable selector here
+export const selectTodos = state => state.todos
+// highlight-end
+```
+
+That way, if we happen to make an update the structure of the todos slice state, the relevant selectors are right here and can be updated at the same time, with minimal changes to any other parts of the app.
```
msutkowski, line 563:
```suggestion
That way, if we happen to make an update to the structure of the todos slice state, the relevant selectors are right here and can be updated at the same time, with minimal changes to any other parts of the app.
```

docs/usage/deriving-data-selectors.md
```diff
@@ -0,0 +1,638 @@
...
+
+### Balance Selector Usage
+
+It's possible to add _too many_ selectors to an application. **Adding a separate selector function for every single field is not a good idea!** That ends up turning Redux into something resembling a a Java class with getter/setter functions for every field. It's not going to _improve_ the code, and it's probably going to make the code _worse_ - maintaining all those extra selectors is a lot of additional effort, and it will be harder to trace what values are being used where.
```
msutkowski, line 567:
```suggestion
It's possible to add _too many_ selectors to an application. **Adding a separate selector function for every single field is not a good idea!** That ends up turning Redux into something resembling a Java class with getter/setter functions for every field. It's not going to _improve_ the code, and it's probably going to make the code _worse_ - maintaining all those extra selectors is a lot of additional effort, and it will be harder to trace what values are being used where.
```

PR #4121, opened 2021-06-27
[website] text color always white on darkness BG

ryota-murakami:
As same as https://github.com/reduxjs/redux-toolkit/pull/1200 I changed text color front of darkness background color.
This might be better to read.

BeforePR
before

AfterPR
after


timdorr:
Much better:

image

Thanks!

ryota-murakami:
@timdorr Thnaks too your review😄

PR #4130, opened 2021-07-07
Add thunks usage guide

markerikson:
This PR:

- Adds a brand new usage guide page on Thunks

### DIFF
--- a/docs/usage/redux-logic-thunks.md
+++ b/docs/usage/redux-logic-thunks.md
@@ -155,7 +155,7 @@ import thunkMiddleware from 'redux-thunk'
 
 const serviceApi = createServiceApi('/some/url')
 
-const thunkMiddlewareWithArg = thunkMiddleware.withExtraArgument(serviceApi)
+const thunkMiddlewareWithArg = thunkMiddleware.withExtraArgument({ serviceApi })
 ```
 
 Redux Toolkit's `configureStore` supports [this as part of its middleware customization in `getDefaultMiddleware`](https://redux-toolkit.js.org/api/getDefaultMiddleware):
@@ -166,13 +166,25 @@ const store = configureStore({
   middleware: getDefaultMiddleware =>
     getDefaultMiddleware({
       thunk: {
-        extraArgument: serviceApi
+        extraArgument: { serviceApi }
       }
     })
 })
 ```
 
 There can only be one extra argument value. If you need to pass in multiple values, pass in an object containing those.
+
+The thunk function will then receive that extra value as its third argument:
+
+```js title="Thunk function with extra argument"
+export const fetchTodoById =
+  todoId => async (dispatch, getState, extraArgument) => {
+    // In this example, the extra arg is an object with an API service inside
+    const { serviceApi } = extraArgument
+    const response = await serviceApi.getTodo(todoId)
+    dispatch(todosLoaded(response.todos))
+  }
+```
 
 ## Thunk Usage Patterns
 
--- a/docs/usage/redux-logic-thunks.md
+++ b/docs/usage/redux-logic-thunks.md
@@ -1,0 +1,443 @@
+---
+id: writing-logic-thunks
+title: Writing Logic with Thunks
+description: 'Usage > Redux Logic > Thunks: writing logic that interacts with the store'
+---
+
+:::tip What You'll Learn
+
+- What "thunks" are, and why they're used for writing Redux logic
+- How the thunk middleware works
+- Techniques for writing sync and async logic in thunks
+- Common thunk usage patterns
+
+:::
+
+## Thunk Overview
+
+### What is a "thunk"?
+
+The word "thunk" is a programming term that means ["a piece of code that does some delayed work"](https://en.wikipedia.org/wiki/Thunk). Rather than execute some logic _now_, we can write a function body or code that can be used to perform the work _later_.
+
+For Redux specifically, "thunks" are a pattern of writing logic that can interact with a Redux store, and require the [`redux-thunk` middleware](https://github.com/redujs/redux-thunk) to be added to the Redux store as part of its configuration.
+
+Thunks are [the standard approach for writing async logic in Redux apps](../style-guide/style-guide.md#use-thunks-for-async-logic), and are commonly used for data fetching. However, they can be used for a variety of tasks, and can contain both synchronous and asynchronous logic.
+
+### Writing Thunks
+
+A _thunk function_ is a function that accepts two arguments: the Redux store `dispatch` method, and the Redux store `getState` method. Thunk functions are not directly called by application code. Instead, they are passed to `store.dispatch()`:
+
+```js title="Dispatching thunk functions"
+const thunkFunction = (dispatch, getState) => {
+  // logic here that can dispatch actions or read state
+}
+
+store.dispatch(thunkFunction)
+```
+
+A thunk function may contain _any_ arbitrary logic, sync or async, and can call `dispatch` or `getState` at any time.
+
+In the same way that Redux code normally uses [action creators to generate action objects for dispatching](../tutorials/fundamentals/part-7-standard-patterns.md#action-creators) instead of writing action objects by hand, we normally use _thunk action creators_ to generate the thunk functions that are dispatched. A thunk action creator is a function that may have some arguments, and returns a new thunk function. The thunk typically closes over any arguments passed to the action creator, so they can be used in the logic:
+
+```js title="Thunk action creators and thunk functions"
+// fetchTodoById is the "thunk action creator"
+export function fetchTodoById(todoId) {
+  // fetchTodoByIdThunk is the "thunk function"
+  return async function fetchTodosThunk(dispatch, getState) {
+    const response = await client.get(`/fakeApi/todo/${todoId}`)
+    dispatch(todosLoaded(response.todos))
+  }
+}
+```
+
+Thunk functions and action creators can be written using either the `function` keyword or arrow functions - there's no meaningful difference here. The same `fetchTodoById` thunk could also be written using arrow functions, like this:
+
+```js title="Writing thunks using arrow functions"
+export const fetchTodoById = todoId => async dispatch => {
+  const response = await client.get(`/fakeApi/todo/${todoId}`)
+  dispatch(todosLoaded(response.todos))
+}
+```
+
+In either case, the thunk is dispatched by calling the action creator, in the same way as you'd dispatch any other Redux action:
+
+```js
+function TodoComponent({ todoId }) {
+  const dispatch = useDispatch()
+
+  const onFetchClicked = () => {
+    // Calls the thunk action creator, and passes the thunk function to dispatch
+    dispatch(fetchTodoById(todoId))
+  }
+}
+```
+
+### Why Use Thunks?
+
+Thunks allow us to write additional Redux-related logic separate from a UI layer. This logic can include side effects, such as async requests or generating random values, as well as logic that requires dispatching multiple actions or access to the Redux store state.
+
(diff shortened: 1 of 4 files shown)


docs/usage/redux-logic-thunks.md
```diff
@@ -0,0 +1,443 @@
...
+
+:::
+
+## Thunk Overview
+
+### What is a "thunk"?
+
+The word "thunk" is a programming term that means ["a piece of code that does some delayed work"](https://en.wikipedia.org/wiki/Thunk). Rather than execute some logic _now_, we can write a function body or code that can be used to perform the work _later_.
+
+For Redux specifically, "thunks" are a pattern of writing logic that can interact with a Redux store, and require the [`redux-thunk` middleware](https://github.com/redujs/redux-thunk) to be added to the Redux store as part of its configuration.
```
xjamundx, line 22: I think this should be split up into 2 sentences. Especially with RTK building thunks in, the second sentence about middleware seems like a minor distraction here

docs/usage/redux-logic-thunks.md
```diff
@@ -0,0 +1,443 @@
...
+
+Thunks are [the standard approach for writing async logic in Redux apps](../style-guide/style-guide.md#use-thunks-for-async-logic), and are commonly used for data fetching. However, they can be used for a variety of tasks, and can contain both synchronous and asynchronous logic.
+
+### Writing Thunks
+
+A _thunk function_ is a function that accepts two arguments: the Redux store `dispatch` method, and the Redux store `getState` method. Thunk functions are not directly called by application code. Instead, they are passed to `store.dispatch()`:
```
xjamundx, line 28: I have never heard this term used before, but I know what you mean. It's nice to see it defined!

docs/usage/redux-logic-thunks.md
```diff
@@ -0,0 +1,443 @@
...
+const thunkFunction = (dispatch, getState) => {
+  // logic here that can dispatch actions or read state
+}
+
+store.dispatch(thunkFunction)
+```
+
+A thunk function may contain _any_ arbitrary logic, sync or async, and can call `dispatch` or `getState` at any time.
+
+In the same way that Redux code normally uses [action creators to generate action objects for dispatching](../tutorials/fundamentals/part-7-standard-patterns.md#action-creators) instead of writing action objects by hand, we normally use _thunk action creators_ to generate the thunk functions that are dispatched. A thunk action creator is a function that may have some arguments, and returns a new thunk function. The thunk typically closes over any arguments passed to the action creator, so they can be used in the logic:
```
xjamundx, line 40: This is really good

docs/usage/redux-logic-thunks.md
```diff
@@ -0,0 +1,443 @@
...
+export function fetchTodoById(todoId) {
+  // fetchTodoByIdThunk is the "thunk function"
+  return async function fetchTodosThunk(dispatch, getState) {
+    const response = await client.get(`/fakeApi/todo/${todoId}`)
+    dispatch(todosLoaded(response.todos))
+  }
+}
+```
+
+Thunk functions and action creators can be written using either the `function` keyword or arrow functions - there's no meaningful difference here. The same `fetchTodoById` thunk could also be written using arrow functions, like this:
```
xjamundx, line 53: This seems almost like too much information, but might have a place. Not sure

docs/usage/redux-logic-thunks.md
```diff
@@ -0,0 +1,443 @@
...
+export const fetchTodoById = todoId => async dispatch => {
+  const response = await client.get(`/fakeApi/todo/${todoId}`)
+  dispatch(todosLoaded(response.todos))
+}
+```
+
+In either case, the thunk is dispatched by calling the action creator, in the same way as you'd dispatch any other Redux action:
+
+```js
+function TodoComponent({ todoId }) {
```
xjamundx, line 65: Great example

docs/usage/redux-logic-thunks.md
```diff
@@ -0,0 +1,443 @@
...
+}
+```
+
+### Why Use Thunks?
+
+Thunks allow us to write additional Redux-related logic separate from a UI layer. This logic can include side effects, such as async requests or generating random values, as well as logic that requires dispatching multiple actions or access to the Redux store state.
+
+Redux reducers [must not contain side effects](../tutorials/fundamentals/part-3-state-actions-reducers.md#rules-of-reducers), but real applications require logic that has side effects. Some of that may live inside components, but some may need to live outside the UI layer. Thunks (and other Redux middleware) give us a place to put those side effects.
+
+It's common to have logic directly in components, such as making an async request in a click handler or a `useEffect` hook and then processing the results. However, it's often necessary to move as much of that logic as possible outside the UI layer. This may be done to improve testability of the logic, to keep the UI layer as thin and "presentational" as possible, or to improve code reuse and sharing.
```
xjamundx, line 81: this is good

docs/usage/redux-logic-thunks.md
```diff
@@ -0,0 +1,443 @@
...
+  const onFetchClicked = () => {
+    // Calls the thunk action creator, and passes the thunk function to dispatch
+    dispatch(fetchTodoById(todoId))
+  }
+}
+```
+
+### Why Use Thunks?
+
+Thunks allow us to write additional Redux-related logic separate from a UI layer. This logic can include side effects, such as async requests or generating random values, as well as logic that requires dispatching multiple actions or access to the Redux store state.
```
xjamundx, line 77: dispatching multiple actions is a really important use case yes

docs/usage/redux-logic-thunks.md
markerikson: There's been a ridiculous amount of back and forth over "what's the thunk? is it the inner function, or the outer function?"  I finally settled on these.

docs/usage/redux-logic-thunks.md
```diff
@@ -0,0 +1,443 @@
...
+
+Redux reducers [must not contain side effects](../tutorials/fundamentals/part-3-state-actions-reducers.md#rules-of-reducers), but real applications require logic that has side effects. Some of that may live inside components, but some may need to live outside the UI layer. Thunks (and other Redux middleware) give us a place to put those side effects.
+
+It's common to have logic directly in components, such as making an async request in a click handler or a `useEffect` hook and then processing the results. However, it's often necessary to move as much of that logic as possible outside the UI layer. This may be done to improve testability of the logic, to keep the UI layer as thin and "presentational" as possible, or to improve code reuse and sharing.
+
+In a sense, a thunk is a loophole where **you can write any code that needs to interact with the Redux store, ahead of time, without needing to know _which_ Redux store will be used**. This keeps the logic from being bound to any specific Redux store instance and keeps it reusable.
```
xjamundx, line 83: Is it though? Reuable?

markerikson: Sometimes, yes. I've seen libs that export thunks, and many libs use thunks as part of their internal implementation. Like, say, RTK Query :)

docs/usage/redux-logic-thunks.md
```diff
@@ -0,0 +1,443 @@
...
+
+Historically, another reason to use thunks was to help keep React components "unaware of Redux". The `connect` API allowed passing action creators and "binding" them to automatically dispatch actions when called. Since components typically did not have access to `dispatch` internally, passing thunks to `connect` made it possible for components to just call `this.props.doSomething()`, without needing to know if it was a callback from a parent, dispatching a plain Redux action, dispatching a thunk performing sync or async logic, or a mock function in a test.
```
xjamundx, line 85: I think we've hit too many words and while this is interesting might be too much information, especially with connect not being recommended? Consider trimming?

markerikson: I can make it a "detailed explanation" expander, yeah

docs/usage/redux-logic-thunks.md
xjamundx: I like that and in the example you have `doThing` and `doThingThunk` which is how I've always thought of it as well

docs/usage/writing-logic-thunks.mdx
```diff
@@ -0,0 +1,463 @@
...
+
+### How Does the Middleware Work?
+
+[Redux middleware are all written as a series of 3 nested functions](../tutorials/fundamentals/part-4-store.md#writing-custom-middleware):
+
+- The outer function receives a "store API" object with `{dispatch, getState}`
+- The middle function receives the `next` middleware in the chain (or the actual `store.dispatch` method)
+- The inner function will be called with each `action` as it's passed through the middleware chain
+
+It's important to note that middleware can be used to allow passing values that are _not_ action objects into `store.dispatch()`, as long as the middleware intercepts those values and does not let them reach the reducers.
```
xjamundx, line 128: I think what you're saying here is .... "The  thunk middleware is able to handle thunks and not just action objects, because..." But instead you're speaking in a more generic way here and I found it confusing

docs/usage/writing-logic-thunks.mdx
```diff
@@ -0,0 +1,463 @@
...
+
+Dispatching thunk functions requires that the [`redux-thunk` middleware](https://github.com/redujs/redux-thunk) has been added to the Redux store as part of its configuration.
+
+### Adding the Middleware
+
+The [Redux Toolkit `configureStore` API automatically adds the thunk middleware during store creation](../tutorials/fundamentals/part-8-modern-redux.md#using-configurestore), so it should typically be available with no extra configuration needed.
+
+If you need to add the thunk middleware to a store manually, that can be done by [passing the thunk middleware to `applyMiddleware()`](..//tutorials/fundamentals/part-6-async-logic.md#configuring-the-store) as part of the setup process.
+
+### How Does the Middleware Work?
```
xjamundx, line 120: Honestly this is like an appendix item

docs/usage/writing-logic-thunks.mdx
```diff
@@ -0,0 +1,463 @@
...
+    return next(action)
+  }
+```
+
+In other words:
+
+- If you pass a function into `dispatch`, the thunk middleware sees that it's a function instead of an action object, intercepts it, and calls that function with `(dispatch, getState)` as its arguments
+- If it's a normal action object (or anything else), it's forwarded to the next middleware in the chain
+
+### Injecting Values Into Thunks
```
xjamundx, line 157: I thought this title meant `dispatch(fetchData(value))` but this is not what this was about. Consider renaming to something like `Injecting Arguments into Thunks Via Middleware Config` or something a bit more specific? The explanation of why is really good here, just thinking a diff title might help?

docs/usage/writing-logic-thunks.mdx
```diff
@@ -0,0 +1,463 @@
...
+    const response = await serviceApi.getTodo(todoId)
+    dispatch(todosLoaded(response.todos))
+  }
+```
+
+## Thunk Usage Patterns
+
+### Dispatching Actions
+
+Thunks have access to the `dispatch` method. This can be used to dispatch actions, or even other thunks. This can be useful for dispatching multiple actions in a row (although [this is a pattern that should be minimized]), or orchestrating complex logic that needs to dispatch at multiple points in the process.
```
xjamundx, line 201: `this is a pattern that should be minimized` is probably supposed to be a link to https://redux.js.org/style-guide/style-guide#avoid-dispatching-many-actions-sequentially

xjamundx: Wondering if this might be better with an example that is more encouraged (and ties into `createAsyncThunk`)
```
dispatch({ type: "fetchData/pending" })
try {
  const results = await thing();
  dispatch({ type: "fetchData/fulfilled", payload: results })
  } catch (e) {
    dispatch({ type: "fetchData/rejected", error: e })
  }
  ```
  
  Mine is a little bit long too though, so I'm not sure. Also are async funcs ok? 🤔

PR #4137, opened 2021-07-17
Added VS Code Badge

anilloutombam:
User can directly open to VS Code without cloning.


markerikson:
Appreciate the PR, but don't think this is something we need. Thanks anyway!

PR #4139, opened 2021-07-25
Add getReducer method to store

### LINKED ISSUE #4107: Expose currentReducer from store through getReducer()
This is not a duplicate of https://github.com/reduxjs/redux/issues/1246 because we have a different type of use case

## New Features

I would like to get the currentReducer using the store.getReducer()

### What is the new or updated feature that you are suggesting?

The createStore in https://github.com/reduxjs/redux/blob/master/src/createStore.ts will have a new method

```
function getReducer(): Reducer<S, A> {

  return currentReducer as Reducer<S, A>;

}
```

and get exposed in store object

### Why should this feature be included?

Our project uses a closed source proprietary library which initializes the redux context for us.
However we would like to override some part of the reducer logic, but we cannot modify the source because we don't have it or use other libraries to do it.

Since we cannot initialize the reducer by ourselves, we can only resort to extracting the reducer from the store. This is a different situation than https://github.com/reduxjs/redux/issues/1246 as they have the full access of the reducer when passing into createStore.
So we would like to add this method to get the currentReducer out, and wrap it to change some behavior, and pass it back to replaceReducer.

We will need to modify the Store Methods part of the document to explain the new method https://redux.js.org/api/store#store-methods

Also I can add some of the example code in the documentation to illustrate how to use together with replaceReducer to override some of the reducer logic.

smeng9:
---
---

New feature

Closes #4107

This PR wraps up the discussion in above and implements the feature for the use case

## Checklist

  - #4107

Add getReducer method to store

Document the new method and the example use cases

### DIFF
--- a/docs/api/Store.md
+++ b/docs/api/Store.md
@@ -19,6 +19,7 @@ To create it, pass your root [reducing function](../understanding/thinking-in-re
 ### Store Methods
 
 - [`getState()`](#getstate)
+- [`getReducer()`](#getreducer)
 - [`dispatch(action)`](#dispatchaction)
 - [`subscribe(listener)`](#subscribelistener)
 - [`replaceReducer(nextReducer)`](#replacereducernextreducer)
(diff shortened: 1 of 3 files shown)


timdorr:
Thanks, but this isn't something we're going to add right now. We don't have a solid use case for it and you can get this information in other ways (by tracking your store instance elsewhere).

PR #4150, opened 2021-08-05
Update Redux Essentials Part 2, app/store.js code

leochoo:
**I'm updating the DOC!, NOT the actual code!**
https://redux.js.org/tutorials/essentials/part-2-app-structure

In this Redux Essentials tutorial, app/store.js,
`export default configureStore` should be changed to
`export const store = configureStore` to match [the current code](https://github.com/reduxjs/cra-template-redux/blob/master/template/src/app/store.js)

Otherwise, it will throw ["Attempted import error"](https://stackoverflow.com/questions/53328408/receiving-attempted-import-error-in-react-app)

Thanks for the PR!

### DIFF
--- a/docs/tutorials/essentials/part-2-app-structure.md
+++ b/docs/tutorials/essentials/part-2-app-structure.md
@@ -125,14 +125,15 @@ Let's start by looking at how the Redux store is created.
 Open up `app/store.js`, which should look like this:
 
 ```js title="app/store.js"
-import { configureStore } from '@reduxjs/toolkit'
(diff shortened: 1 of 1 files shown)


markerikson:
Hmm. At the moment, the actual sandbox shown in this page is still using a default import in `index.js`, so for consistency we need to leave it the way it is.  It does look like we changed the template, so there's some differences between the template and what's in this page.

I'm going to close this, but I'm planning to update the Essentials tutorial in the near future anyway.  I'll aim to make it align better with the current contents of the template when I do.

PR #4153, opened 2021-08-21
Add new RTK Query pages to the "Essentials" tutorial

markerikson:
## Overview

This PR:

- Adds new pages to the Essentials tutorial covering how to use RTK Query
- Cleans up the Essentials tutorial to reflect updates in usage patterns (such as the "builder callback" syntax for `extraReducers`)
- Fixes several bits of explanation that had been skipped over during the original tutorial content for space and were identified as being confusing

## Status

~~This is a very rough first draft of the first half of the planned content. As in, I just got done writing for the last few hours :)~~

~~So far I've got a "Part 7: RTKQ Basics" page written up, as seen here.  I plan to add a "Part 8: RTKQ Advanced Patterns" page covering the rest of the example commits (page name TBD).~~ 

~~I've just added the "Advanced" page first draft to go along with the "Basics" page.~~

~~I've also got a third draft of the planned example app code completed, and have just pushed that up to https://github.com/reduxjs/redux-essentials-example-app/tree/tutorial-steps-2021-08-14.~~

~~I also will need to go through and do some more cleanup of the earlier sections of the Essentials tutorial, applying some of the feedback from existing issues.~~

I've got the content basically complete at this point. New pages should be done. Fixed up the existing pages to match the latest code and fill in some missing explanations, and addressed the review comments.

Only thing that should be left at this point is updating the example repo with the final code in the `tutorial-steps` branch, adding the right tags, and filling in the Sandboxes for the new pages.

The nearly-final example code is viewable at https://github.com/reduxjs/redux-essentials-example-app/tree/tutorial-steps-2021-08-28 .

The new content is previewable at https://deploy-preview-4153--redux-docs.netlify.app/tutorials/essentials/part-7-rtk-query-basics .


docs/tutorials/essentials/part-6-performance-normalization.md
```diff
@@ -23,9 +23,9 @@ import { DetailedExplanation } from '../../components/DetailedExplanation'
 
 ## Introduction
 
-In [Part 5: Async Logic and Data Fetching](./part-5-async-logic.md), we saw how to write async thunks to fetch data from a server API, patterns for handling async request loading state, and use of selector functions for encapsulating lookups of data from the Redux state.
+In [Part 5: Async Logic and Data Fetching](./part-5-async-logic.md), we saw to write async thunks to fetch data from a server API, patterns for handling async request loading state, and use of selector functions for encapsulating lookups of data from the Redux state.
```
phryneas, line 26: > we saw to write

not sure if this change is intentional

docs/tutorials/essentials/part-7-rtk-query-basics.md
```diff
@@ -0,0 +1,692 @@
...
+      query: () => '/posts'
+    })
+  })
+})
+
+// Export the auto-generated hook for the `getPost` query endpoint
+export const { useGetPostsQuery } = apiSlice
+```
+
+RTK Query's functionality is based on a single method, called `createApi`. All of the Redux Toolkit APIs we've seen so far are UI-agnostic, and could be used with _any_ UI layer. The RTK Query core logic is the same way. However, RTK Query also includes a React-specific version of `createApi`, and we need to use that. So, we import from `'@reduxjs/toolkit/query/react'` specifically.
```
phryneas, line 139: > and we need to use that

maybe *want* instead of *need*?

docs/tutorials/essentials/part-7-rtk-query-basics.md
```diff
@@ -0,0 +1,692 @@
...
+
+#### API Slice Parameters
+
+When we call `createApi`, there are three fields that are required:
+
+- `reducerPath`: defines the expected top-level state slice field. For our other slices like `postsSlice`, there's no guarantee that it will be used to update `state.posts` - we _could_ have attached the reducer anywhere in the root state, like `someOtherField: postsReducer`. Here, `createApi` expects us to tell it where the cache state will exist when we add the cache reducer to the store.
```
phryneas, line 145: `reducerPath` is actually optional, defaulting to `api`, so the "required" in the last sentence is not fully true.

docs/tutorials/essentials/part-7-rtk-query-basics.md
```diff
@@ -0,0 +1,692 @@
...
+
+#### Defining Endpoints
+
+The first part of the URL for all requests is defined as `'/fakeApi'` in the `fetchBaseQuery` definition.
+
+For our first step, we want to add an endpoint that will return the entire list of posts from the fake API server. We'll include an endpoint called `getPosts`, and define it as a **query endpoint** using `builder.query()`. This method accepts many options for configuring how to make the request and process the response. For now, all we need to do is supply the remaining piece of the URL path by defining a `query` option, with a callback that returns the URL string: `() => '/posts'`.
+
+By default, query endpoints will use a `GET` HTTP request, but you can override that by returning an object like `{url: '/posts', method: 'POST', body}` instead of just the URL string itself.
```
phryneas, line 155: since `body` is not an established variable, maybe `body: { name: "Tom" }` or something?

docs/tutorials/essentials/part-7-rtk-query-basics.md
```diff
@@ -0,0 +1,692 @@
...
+  )
+}
+```
+
+Conceptually, `<PostsList>` is still doing all the same work it was before, but we were able to replace the multiple `useSelector` calls and the `useEffect` dispatch with a single call to `useGetPostsQuery()`.
+
+Each generated query hook returns an object containing several fields, including:
+
+- `data`: the actual response contents from the server. **This field will be `undefined` until the response is received**.
+- `isLoading`: a boolean indicating if the hook is currently making the _first_ request to the server for this data
```
phryneas, line 282: "first request from this hook instance", but not necessarily "for this data".
If you switch from page 1 to page 2 and page 2 needs to be fetched, `isLoading` will not reactivate.

docs/tutorials/essentials/part-7-rtk-query-basics.md
```diff
@@ -0,0 +1,692 @@
...
+- `isFetching`: a boolean indicating if the hook is currently making _any_ request to the server
+- `isSuccess`: a boolean indicating if the hook has made a successful request and has cached data available (ie, `data` should be defined now)
+- `isError`: a boolean indicating if the last request had an error
+- `error`: a serialized error object
+
+Typically, you'll want to destructure fields from the result object, and possibly rename `data` to a more specific variable like `posts` to describe what it contains.
```
phryneas, line 288: Not 10000% sure we should encourage destructuring *too* much as it has drawbacks using TS (if destructured, `isSuccess` will no longer indicate that `data` is not `undefined`)

phryneas (commented):
Read through part 7, all I have are nitpicks. This is already very polished, like always :)
Probably gonna read through part 8 tomorrow.

Shrugsy (commented):
Awesome content @markerikson!

I've left some nitpicks/questions/spelling-fixes to look at

docs/tutorials/essentials/part-7-rtk-query-basics.md
markerikson: We're using RTKQ + React, so definitely "need" :)

docs/tutorials/essentials/part-6-performance-normalization.md
markerikson: Nope, fixed.

docs/tutorials/essentials/part-7-rtk-query-basics.md
markerikson: Will this be fixed in TS 4.4?

I find destructuring to be very natural here, and I would find it annoying to always have to do `res.isSuccess` all the time.

Shrugsy: > Will this be fixed in TS 4.4?

Not to my knowledge, 4.4 doesn't change anything about that. It still needs a single object to maintain a relationship between properties

mini-example on 4.4.2 here: https://tsplay.dev/NBP1xW

docs/tutorials/essentials/part-8-rtk-query-advanced.md
```diff
@@ -0,0 +1,1074 @@
...
+
+  // Use the same posts query, but extract only part of its data
+  const { postsForUser } = useGetPostsQuery(undefined, {
+    // highlight-start
+    selectFromResult: result => ({
+      ...result,
+      // Include a field called `postsForUser` in the hook result object,
+      // which will be filtered list of posts
+      postsForUser: selectPostsForUser(result, userId)
+    })
```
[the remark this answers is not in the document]
markerikson: Looking at the source, pretty sure it runs into this `shallowEqual` check here:

```ts
      const currentState = useSelector(
        (state: RootState<Definitions, any, any>) =>
          querySelector(state, lastValue.current),
        shallowEqual
      )
```

so spread + add field with a memoized selector should work, I think.

docs/tutorials/essentials/part-8-rtk-query-advanced.md
```diff
@@ -0,0 +1,1074 @@
...
+    }),
+    addNewPost: builder.mutation({
+      query: initialPost => ({
+        url: '/posts',
+        method: 'POST',
+        body: initialPost
+      }),
+      // highlight-next-line
+      invalidatesTags: ['Post']
+    }),
```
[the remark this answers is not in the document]
markerikson: Yeah, I may punt on this one and link to the "automated refetching" page

docs/tutorials/essentials/part-8-rtk-query-advanced.md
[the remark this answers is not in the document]
Shrugsy: It definitely works either way, but what I mean is that it will better optimized for renders if you exclude the rest of `result`, because including it means that more things are being checked in the `shallowEqual` comparison, which in this case isn't being used.

So for this case:
```ts
    selectFromResult: result => ({
      ...result,
      postsForUser: selectPostsForUser(result, userId)
    })
```
If *any part* of result changes, the hook will re-render. i.e. if posts for other users update, or even when changing status between `pending` & `fulfilled`.

Whereas when removing the spread `...result` like this case:
```ts
    selectFromResult: result => ({
      postsForUser: selectPostsForUser(result, userId)
    })
```
This will only re-render specifically if `postsForUser` changes, without caring about the rest of the values in `...result`. Since `postsForUser` is the *only* thing being used by the caller, this is a strictly more optimized version with no downside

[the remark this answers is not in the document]
Shrugsy: FYI here's a contrived example that demonstrates the behaviour: https://codesandbox.io/s/rtk-query-fiddlebox-selectfromresult-dx17l?file=/src/components/TodoApp.tsx

- Two components, one with each scenario above:
  - Using a list query, one spreading the entire result, one without, both with an added field
- The added field picks out a particular item (todo with ID 2)
- Each logs something in the render body
- Use the UI to update the message for a different todo (with ID 1) - i.e. part of the result that the hook caller is not using
- Observe that the version with the spread result will re-render while the one without does not re-render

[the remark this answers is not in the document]
markerikson: Ok, I'll add a note along those lines.

docs/tutorials/essentials/part-8-rtk-query-advanced.md
```diff
@@ -0,0 +1,1092 @@
...
+
+export const EditPostForm = ({ match }) => {
+  const { postId } = match.params
+
+  // highlight-next-line
+  const { data: post } = useGetPostQuery(postId)
+  // highlight-next-line
+  const [updatePost, { isLoading }] = useEditPostMutation()
+
+  const [title, setTitle] = useState(post.title)
```
phryneas, line 107: Not sure if relevant for the tutorial: This would cause an `undefined` error on page refresh while in edit mode

docs/tutorials/essentials/part-8-rtk-query-advanced.md
```diff
@@ -0,0 +1,1092 @@
...
+
+We could add a `'Post'` tag to both the `getPost` query and the `editPost` mutation, but that would force all the other individual posts to be refetched as well. Fortunately, **RTK Query lets us define specific tags, which let us be more selective in invalidating data**. These specific tags look like `{type: 'Post', id: 123}`.
+
+Our `getPosts` query defines a `providesTags` field that is an array of strings. The `providesTags` field can also accept a callback function that receives the `result` and `arg`, and returns an array. This allows us to create tag entries based on IDs of data that is being fetched. Similarly, `invalidatesTags` can be a callback as well.
+
+In order to get the right behavior, we need to set up each endpoint with the right tags:
+
+- `getPosts`: provides a general `'Post'` tag for the whole list, as well as a specific `{type: 'Post', id}` tag for each received post object
+- `getPost`: provides a specific `{type: 'Post', id}` object for the individual post object
+- `addNewPost`: invalidates the general `'Post'` tag, to refetch the whole list
```
phryneas, line 157: This would also refetch all `{type: 'Post', id}` tags. 

See https://redux-toolkit.js.org/rtk-query/usage/automated-refetching#tag-invalidation-behavior - maybe adding a link to that table would be a good idea here anyways.

docs/tutorials/essentials/part-8-rtk-query-advanced.md
```diff
@@ -0,0 +1,1092 @@
...
+When we save the edited post this time, we should see two requests happen back-to-back:
+
+- The `POST /posts/:postId` from the `editPost` mutation
+- A `GET /posts/:postId` as the `getPost` query is refetched
+
+Then, if we click back to the main "Posts" tab, we should also see:
+
+- A `GET /posts` as the `getPosts` query is refetched
+
+Because we provided the relationships between the endpoints using tags, **RTK Query knew that it needed to refetch the individual post and the list of posts when we made that edit and the specific tag with that ID was invalidated** - no further changes needed! Also, while RTK Query knew that the posts list data was invalid, it did _not_ refetch it right away because the component with `useGetPostsQuery()` wasn't being shown when we edited the post. When we opened the `<PostsList>` component again, RTK Query saw that the data was stale and refetched it.
```
phryneas, line 218: Regarding the last sentence: actually, it removed that "unused stale data" from the cache and a new request for a new cache entry was made as it was required again.

docs/tutorials/essentials/part-8-rtk-query-advanced.md
```diff
@@ -0,0 +1,1092 @@
...
+
+There is one caveat here. By specifying a plain `'Post'` tag in `getPosts` and invalidating it in `addNewPost`, we actually end up forcing a refetch of all _individual_ posts as well. If we really want to just refetch the list of posts for the `getPost` endpoint, you can include an additional tag with an arbitrary ID, like `{type: 'Post', id: 'LIST'}`, and invalidate that tag instead.
```
phryneas, line 220: Ah. Yeah. Disregard my pre-previous comment then. Apart from the table link thing.

docs/tutorials/essentials/part-8-rtk-query-advanced.md
```diff
@@ -0,0 +1,1092 @@
...
+```jsx title="index.js"
+// omit other imports
+// highlight-next-line
+import { apiSlice } from './features/api/apiSlice'
+
+// Start our mock API server
+worker.start({ onUnhandledRequest: 'bypass' })
+
+// highlight-next-line
+store.dispatch(apiSlice.endpoints.getUsers.initiate())
```
phryneas, line 295: Maybe add an info box that describes that this would need to be unsubscribed or will never be cache-collected?

docs/tutorials/essentials/part-8-rtk-query-advanced.md
```diff
@@ -0,0 +1,1092 @@
...
+})
+
+export const { useGetUsersQuery } = extendedApiSlice
+
+export const selectUsersResult = extendedApiSlice.endpoints.getUsers.select()
+// highlight-end
+```
+
+`injectEndpoints()` **mutates the original API slice object to add the additional endpoint definitions, and then returns it**. The actual caching reducer and middleware that we originally added to the store still work okay as-is. At this point, `apiSlice` and `extendedApiSlice` are the same object, but
+it can be helpful to refer to the `extendedApiSlice` object instead of `apiSlice` here as a reminder to ourselves. (This is more important if you're using TypeScript, because only the `extendedApiSlice` value is has the added types for the new endpoints.)
```
phryneas, line 399:
```suggestion
it can be helpful to refer to the `extendedApiSlice` object instead of `apiSlice` here as a reminder to ourselves. (This is more important if you're using TypeScript, because only the `extendedApiSlice` value has the added types for the new endpoints.)
```

docs/tutorials/essentials/part-8-rtk-query-advanced.md
```diff
@@ -0,0 +1,1092 @@
...
+
+At the moment, the only file that references the `getUsers` endpoint is our index file, which is dispatching the `initiate` thunk. We need to update that to import the extended API slice instead:
+
+```js title="index.js"
+// highlight-next-line
+import { extendedApiSlice } from './features/users/usersSlice'
```
phryneas, line 405: I would even suggest doing
```js
export const { getUsers } = extendedApiSlice.endpoints
```
in `'./features/users/usersSlice'` and importing only that.  
Might give better auto-import suggestions than having multiple `extendedApiSlice` imports in the end.

I would probably never export `extendedApiSlice` itself.

docs/tutorials/essentials/part-8-rtk-query-advanced.md
```diff
@@ -0,0 +1,1092 @@
...
+        body: post
+      }),
+      // highlight-next-line
+      invalidatesTags: (result, error, arg) => [{ type: 'Post', id: arg.id }]
+    })
+  })
+})
+```
+
+It's possible for the `result` argument in these callbacks to be undefined if the response has no data or there's an error, so we have to handle that safely. For `getPosts` we can do that by using a default argument array value to map over, and for `getPost` we just return an empty array if `result` isn't there. For `editPost`, we know the ID of the post from the partial post object that was passed into the trigger function, so we can read it from there.
```
Shrugsy, line 203: > and for `getPost` we just return an empty array if `result` isn't there

I'm not sure of the best way to re-word it, but `getPost` now always provides tags, so this line should be altered to say that it uses the ID of the post from the argument

docs/tutorials/essentials/part-8-rtk-query-advanced.md
```diff
@@ -0,0 +1,1092 @@
...
+      providesTags: (result = [], error, arg) => [
+        'Post',
+        ...result.map(({ id }) => ({ type: 'Post', id }))
+      ]
+      // highlight-end
+    }),
+    getPost: builder.query({
+      query: postId => `/posts/${postId}`,
+      // highlight-start
+      providesTags: (result, error, arg) => [{ type: 'Post', id: arg.id }]
```
Shrugsy, line 178:
```suggestion
      providesTags: (result, error, arg) => [{ type: 'Post', id: arg }]
```
✍️ 
`arg` in this case is already the ID

docs/tutorials/essentials/part-8-rtk-query-advanced.md
markerikson: Yeah, there was also a similar issue with `<SinglePostPage>` before the RTKQ section. Not worth messing with for now.

docs/tutorials/essentials/part-8-rtk-query-advanced.md
markerikson: Depends on how fast the clicks happened :)  But yeah, most likely this.

markerikson:
Ready to roll!

PR #4154, opened 2021-08-22
Update quick-start.md

dagolinuxoid:
It's actually a reducer but not reducers field at -> Add Slice Reducers to the Store section
Deleted nonsense sentence at -> Full Counter App Example section

### DIFF
--- a/docs/tutorials/quick-start.md
+++ b/docs/tutorials/quick-start.md
@@ -120,7 +120,7 @@ export default counterSlice.reducer
 
 ### Add Slice Reducers to the Store
 
-Next, we need to import the reducer function from the counter slice and add it to our store. By defining a field inside the `reducer` parameter, we tell the store to use this slice reducer function to handle all updates to that state.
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks, but that was the correct parameter. But the other dangling sentence should go.

dagolinuxoid:
@timdorr I know the sentence is confusing but this is my understanding of the concept and why I think the *reducer* should be used as a key value. When we are using createSlice function it accepts an object as its argument which indeed should have a field/key *reducers* however in case of using configureStore function its object argument must have *reducer* property. Honestly I don't like the paragraph since it's sloppy I would write it like ... idk something like this: By defining a field (counter) inside of a reducer property, we tell the store to use a value of the field (which is a slice reducer function) to handle all updates to that state (counter).  Well I bet you can enhance it further .  BTW this is a screenshot of redux and redux toolkit docs reducers vs reducer:
[](https://postimg.cc/K15btz1d)

timdorr:
Oh, my bad. I was thinking this was about createSlice, not configureStore. I'll fix that up now.

PR #4165, opened 2021-09-03
Fix code snippet in tutorial (async logic)

arielbk:
The mock server expects _just_ the `initialPost` object, rather than an object with a `post` field.


timdorr:
Thanks!

markerikson:
Whoops, missed updating that one when I updated things. Thank you!

PR #4166, opened 2021-09-05
Fix code snippet and typo in tutorial

arielbk:
The action payload first needs to be pushed to the state object before the `isNew` field is changed, otherwise messages are all always marked as 'read'.

Trivially fixes a couple of references to `useEffect` where it should be `useLayoutEffect` in the `NotificationsList` component.


markerikson:
Yep, looks like I missed a couple of those in the updates. Thanks!

PR #4177, opened 2021-09-26
📝 Update MSW usage

Shrugsy:
- show MSW usage as awaiting service worker registration
  in case of race conditions with other workers
- minor text updates

---
---

## Checklist

  N/A - No existing issue

- **Section**: Tutorials/Redux-Essentials
- **Page**: 
  - Redux Essentials, Part 5: Async Logic and Data Fetching
  - Redux Essentials, Part 8: RTK Query Advanced Patterns

## What is the problem?

MSW by default will defer network requests while starting the worker (see [start#waituntilready](https://mswjs.io/docs/api/setup-worker/start#waituntilready)). However, @msutkowski has pointed out if service worker registration fights with another, it won't kick in. The result is flaky behaviour - sometimes the requests will be attempted prior to the service worker registration, causing failed requests.

e.g. the 'posts' in the corresponding codesandbox examples won't necessarily be mocked: https://codesandbox.io/s/github/reduxjs/redux-essentials-example-app/tree/checkpoint-6-rtkqConversion/?from-embed

This PR updates the tutorial snippets to show `awaiting` the service worker registration prior to firing initial requests or mounting the app.

Note that this does _not_ inherently update the codesandbox examples themselves, which are instead hosted on the [redux-essentials-example-app](https://github.com/reduxjs/redux-essentials-example-app) repo. @markerikson; given that the codesandbox demos use tag-based references for multiple stages of the example app, I'm not sure what the best way to update each of them would be.

### DIFF
--- a/docs/tutorials/essentials/part-5-async-logic.md
+++ b/docs/tutorials/essentials/part-5-async-logic.md
@@ -31,7 +31,7 @@ In this section, we'll convert our social media app to fetch the posts and users
 
 ### Example REST API and Client
 
-To keep the example project isolated but realistic, the initial project setup already included a fake in-memory REST API for our data (configured using [the Mock Service Worker mock API tool](https://mswjs.io/)). The API uses `/fakeApi` as the base URL for the endpoints, and supports the typical `GET/POST/PUT/DELETE` HTTP methods for `/fakeApi/posts`, `/fakeApi/users`, and `fakeApi/notifications`. It's defined in `src/api/server.js`.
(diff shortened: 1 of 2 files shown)


Shrugsy:
If anyone is interested in bikeshedding, rather than

```js
async function main() {
  await worker.start()
  // do stuff
}
main()
```

It might instead be more conventional to show like this:
```js
worker.start().then(() => {
  // do stuff
});
```

bybruno (approved):

PR #4178, opened 2021-09-26
recommend use of an api cache library in the style guide

phryneas:
---
name: :book: Recommend use of an api cache library in the style guide
about: Add a section to the style guide to use RTK Query or an alternative library.
---

## PR Type

**Update an _existing_ page**

## Checklist

No, the linter runs amok on my local machine and tries to format everything differently.

- **Section**: Style Guide

Add a section to the style guide to use RTK Query or an alternative library.

They add more emphasis that api cache logic should usually not be written by hand.

@markerikson generally I am very open to pretty much rewording *anything* in here, but we should probably get RTK-Q mentioned in the style guide.

### DIFF
--- a/docs/style-guide/style-guide.md
+++ b/docs/style-guide/style-guide.md
@@ -96,6 +96,12 @@ Writing immutable update logic by hand is frequently difficult and prone to erro
 
 <a id="structure-files-as-feature-folders-or-ducks"></a>
 
+### Avoid hand-writing API Caching Logic
+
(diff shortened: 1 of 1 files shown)


3imed-jaberi (approved):

EskiMojo14:
this seems to have been superceded by https://github.com/reduxjs/redux/pull/4460 - (https://redux.js.org/style-guide/#use-rtk-query-for-data-fetching) 

okay to close?

markerikson:
Yep!

PR #4186, opened 2021-10-10
docs: Remove line that breaks code from Redux Essentials Tutorial

ivansvlv:
---
---

## Checklist

  - _link issue here_

- **Page**: Part 8: RTK Query Advanced Patterns

## What is the problem?
In this part: https://redux.js.org/tutorials/essentials/part-8-rtk-query-advanced#streaming-cache-updates , there is a line (see below) in that transform data received in query response, but we receive a plain array with notification objects, there is no nesting.

```javascript
transformResponse: (res) => res.notifications
```

It breaks code because it will return `undefined`.

It removes `transformResponse` method from the query builder.


markerikson:
Yep, that's another one of those that I missed when rewriting the tutorial.  Thanks!

PR #4209, opened 2021-11-03
Fix typo in part-5-async-logic.md

KonstHardy:
I removed the redundant word

Fix typo:
that that -> that


timdorr:
I'm guessing that was supposed to be a "the"? 🤷 Well, that works too. Thanks!

markerikson:
The sentence was grammatically and semantically correct as written, but I guess this works.

timdorr:
Huh, I guess it was. Kinda awkward though, so this is less awkwarder.

PR #4217, opened 2021-11-10
refactor: removed duplicated export default in createStore

sameroso:
I did some refactor on createStore that seemed duplicated


markerikson:
Hi. Appreciate the PR, but that's not "duplicated". Those are TypeScript function overloads, each of which represents a different possible set of arguments that can be passed in to the same function, and they're necessary for this to work correctly.

PR #4218, opened 2021-11-11
Update part-7-rtk-query-basics.md

bamukhier:
Fix a comment in a code snippet in part-7-rtk-query-basics.md

In line 135, It should have been `getPosts` instead of `getPost`. I think the og author mixed it with another endpoint called `getPost` that was introduced further down the page.


markerikson:
Just a typo, but yeah.  Thanks!

PR #4219, opened 2021-11-15
Convert postId param to number before using

jferriday:
---
---

## Checklist

  - _link issue here_
- [ x] Have the files been linted and formatted? 

- **Section**: Redux Essentials Tutorial
- **Page**: Part 4

## What is the problem?
Going through this tutorial in the docs (Redux Essentials Part 4), reading the code example for the SinglePostPage component, I found that `match.params.postId` is a string, whereas the postId on the post object is a number (line 252).

e.g `2 === '2' `returns false in line 255.

Propose wrapping this in the `Number()` function to avoid `state.posts.find(post => post.id === postId)` returning undefined.

### DIFF
--- a/docs/tutorials/essentials/part-4-using-data.md
+++ b/docs/tutorials/essentials/part-4-using-data.md
@@ -249,7 +249,7 @@ import { useHistory } from 'react-router-dom'
 import { postUpdated } from './postsSlice'
 
 export const EditPostForm = ({ match }) => {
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

jferriday:
I've just realised that this doesn't work for nanoid generated IDs, and that the tutorial actually uses string IDs previously. I'll open another PR reverting my changes!

PR #4233, opened 2021-12-08
Docs: Redux [FAQ: Organizing State] remove You Might Not Need Redux

ryota-murakami:
Because this was written at 2016, the content seems like counter part for over engineering with verbosity [Boilerplate](https://github.com/ryota-murakami/trueblue/tree/master/app/Resources/js).  

But we are living post Redux-Toolkit,「You Might Not Need Redux」is not tell us the Author's original meant by context lost.

### DIFF
--- a/docs/faq/OrganizingState.md
+++ b/docs/faq/OrganizingState.md
@@ -44,7 +44,6 @@ There are a number of community packages that implement various approaches for s
 **Articles**
 
 - [When (and when not) to reach for Redux](https://changelog.com/posts/when-and-when-not-to-reach-for-redux)
-- [You Might Not Need Redux](https://medium.com/@dan_abramov/you-might-not-need-redux-be46360cf367)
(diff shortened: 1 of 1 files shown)


timdorr:
You might not need this blog post. 

Thanks!

markerikson:
I... actually would prefer to leave that link in, tbh.

I know it's old, but Dan did write it, and the arguments he made about "tradeoffs" and the use cases he described are still meaningful.

ryota-murakami:
@markerikson Yeah, I know that feeling. I thought it was great post at the 2017.
> I know it's old, but Dan did write it, and the arguments he made about "tradeoffs" and the use cases he described are still meaningful.

I agree with you the post fundamental still correct think.  
But why we could have such as opinion? Because we are veteran of Redux(actually your level is god of Redux) so, 
I think this is because we can extract and evaluate the essential purpose of this article, taking into account the differences between then and now.

But it's not friendly for new leaner.  

original Tradeoff

PR #4236, opened 2021-12-15
Add a tip pointing to code splitting

Faithfinder:
---
name: :book: Add a tip pointing to code splitting
about: Updating content in an existing docs page
---

## PR Type

**update an _existing_ page**

## Checklist

  - [Discord thread](https://discord.com/channels/102860784329052160/103538784460615680/920489794331611136)
  - No, since I've edited directly in Github, but LGTM?

- **Page**: RTK Query Advanced Patterns

The main goal is to highlight the fact that you're expect to have only one `createApi` call in the app. (Maybe it's rather one per API, as in, if it's a different server, it's a different `createApi`?). I'm also putting a link to the advanced concept of `Injecting endpoints` that people most likely want when calling `createApi` multiple times.

I don't think so.

### DIFF
--- a/docs/tutorials/essentials/part-7-rtk-query-basics.md
+++ b/docs/tutorials/essentials/part-7-rtk-query-basics.md
@@ -140,7 +140,9 @@ RTK Query's functionality is based on a single method, called `createApi`. All o
 
 :::tip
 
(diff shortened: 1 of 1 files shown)


markerikson:
Thanks!

PR #4243, opened 2021-12-22
Update part-8-rtk-query-advanced.md

jbuhl-augeo:
Currently the app will error if it tries to run the selector before posts are cached. This fixes it by returning an empty array in that case.


markerikson:
Hmm. Strictly speaking this isn't ideal - returning a new array reference from that input selector every time will cause the selector to never memoize properly and always re-filter.

PR #4273, opened 2022-01-26
Update devtools extension npm install guide

DuBento:
Thanks for the PR!

### DIFF
--- a/docs/introduction/Installation.md
+++ b/docs/introduction/Installation.md
@@ -48,7 +48,7 @@ Most likely, you'll also need [the React bindings](https://github.com/reduxjs/re
 
 ```bash
 npm install react-redux
-npm install --save-dev redux-devtools-extension
+npm install --save-dev @redux-devtools/extension
 ```
 
 Note that unlike Redux itself, many packages in the Redux ecosystem don't provide UMD builds, so we recommend using CommonJS module bundlers like [Webpack](https://webpack.js.org/) and [Browserify](http://browserify.org/) for the most comfortable development experience.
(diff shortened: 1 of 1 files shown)


markerikson:
Actually, I think as of the recent DevTools 3.0 release, it's now a separate scoped package name, per https://github.com/reduxjs/redux-devtools/releases/tag/remotedev-redux-devtools-extension%403.0.0 and https://github.com/reduxjs/redux-devtools/pull/948/files#diff-46ac21909169cd3c3c528d28cca98e1fb32578556269dd06149c3ddb95e95ce2 .

Could you update the PR to match that?

Methuselah96:
Note that `redux-devtools` is now `@redux-devtools/core` and that `redux-devtools-extension` is now `@redux-devtools/extension`. I'm guessing that this should should refer to `@redux-devtools/extension`, but I'm not certain.

DuBento:
> Note that `redux-devtools` is now `@redux-devtools/core` and that `redux-devtools-extension` is now `@redux-devtools/extension`. I'm guessing that this should should refer to `@redux-devtools/extension`, but I'm not certain.

I think you are correct. Updated patch.

timdorr:
@DuBento I believe it's supposed to be `@redux-devtools/core` now.

Methuselah96:
It seems a little out-of-place and outdated to recommend installing `@redux-devtools/core` without any context on how to use it (and to say "Most likely, you'll also need the React bindings and the developer tools."). It seems to me that `@redux-devtools/extension` should be the recommended way to use the developer tools.

DuBento:
> @DuBento I believe it's supposed to be `@redux-devtools/core` now.

Since I was just looking for extension support, `/extension` made more sense at the time. But to keep consistent with what was previously in the installation guide maybe I could change to `@redux-devtools`.
Also since someone is mainly looking for the extension support there could be a note suggesting to install only `@redux-devtools/extension`. Could also use your thoughts on the matter @Methuselah96.

PR #4277, opened 2022-02-05
initial spanish translation folder and core concepts translated

moctezumadaniel:
Full copy of docs added for future translation.
Core concepts translated


moctezumadaniel:
Core concept's original text was not deleted to keep as a quick reference

timdorr:
Thanks, but this should be kept external to the repo. My Spanish knowledge is limited to ordering off menus and I'm not sure if we have any active Spanish speakers here. If you want to set up an external repo and site, we can link to it here: https://redux.js.org/introduction/learning-resources#redux-docs-translations

The current Spanish translation is outdated, so a more recently version would be great!

moctezumadaniel:
Thanks for replying, I'll try to complete the translation and contact you afterwards

PR #4282, opened 2022-02-08
Copying an array using spread operator

AbderrahmanAboumouslim:
---
---
redux contribution spread operator way

- **Section**:
- 
## What is the problem?
Docs use old way of copying an array. I assume the spread operator is the way to go.

Make the Docs up to date.


markerikson:
`array.slice()` and `[...array]` are identical for all intents and purposes, yeah.

AbderrahmanAboumouslim:
As you said it won 't make a difference, but I assume for beginner JS dev will be confused at first time. Thanks for taking time to check my useless PR :)

PR #4297, opened 2022-02-21
📝 Fix rtkq selectFromResult example

Shrugsy:
---
---

## Checklist

  - No existing issue
  Yes

- **Section**: Tutorials -> Redux Essentials
- **Page**: RTK Query Advanced Patterns -> [Selecting Values from Results​](https://redux.js.org/tutorials/essentials/part-8-rtk-query-advanced#selecting-values-from-results)

## What is the problem?

In the example snippet, `res.data` will be undefined initially, before the query returns any data. Calling `data.filter` would result in a runtime error.

This PR alters the example snippet to account for `data` being undefined, and falls back to return an empty array in that case instead.


timdorr:
Thanks!

PR #4312, opened 2022-03-11
experiment: clean $CombinedState out from getState

phryneas:
Hey @markerikson take a look - does this solve https://github.com/reduxjs/redux-toolkit/issues/2068 ?

### DIFF
--- a/src/types/store.ts
+++ b/src/types/store.ts
@@ -40,6 +40,7 @@ declare const $CombinedState: unique symbol
 interface EmptyObject {
   readonly [$CombinedState]?: undefined
 }
+export type CleanState<T> = T extends CombinedState<infer S> ? { [K in keyof S]: CleanState<S[K]> } : T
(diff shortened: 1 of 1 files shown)


markerikson:
Just tried this. It _does_ fix the issue... _if_ you infer `type RootState = ReturnType<typeof store.getState>`.  However, **it does _not_ fix the issue if you infer directly from `ReturnType<typeof rootReducer>`**.  Which is what I suspected might happen.

So clearly something is going on with Reselect and that `$CombinedState` type.

mpeyper:
This would likely be a workable solution for my issue, as just having an option to erase the `$CombinedState` would at lease keep me moving, but I agree with @markerikson that it doesn't actually solve the root of the issue.

markerikson:
Where does this idea stand? I think I saw us talking recently about dropping `$CombinedState` completely?  What does that even _do_, anyway?

PR #4318, opened 2022-03-23
Add explanation for returning payload

tbrlpld:
This is to address the difference between the reducer used in the
"loading users" section and the rest of the reducers used in the
tutorial up to that point.

See the discussion in:
https://github.com/reduxjs/redux/issues/4317

---
---

## Checklist

  - https://github.com/reduxjs/redux/issues/4317

- **Section**: Loading users
- **Page**:  https://github.com/reduxjs/redux/blob/master/docs/tutorials/essentials/part-5-async-logic.md

## What is the problem?

The reducer in this section returns a value instead of updating the existing state. This is the first time this technique is used in the tutorial, all previously used reducers update the existing state. While this works and is the correct solution, it is used without any comment as of why this approach is chosen over the update. To a beginner without much Redux and especially Immer experience this can be quite surprising and confusing. 

This PR adds a short explanation pointing out that a different approach was taken and why.

### DIFF
--- a/docs/tutorials/essentials/part-5-async-logic.md
+++ b/docs/tutorials/essentials/part-5-async-logic.md
@@ -702,12 +702,15 @@ const usersSlice = createSlice({
 export default usersSlice.reducer
 ```
 
-You may noticed that our reducer is not using the `state` variable at all.
-Instead, we are returning the action payload directly.
-This has the effect that the existing state is completely replaced with whatever we are returning.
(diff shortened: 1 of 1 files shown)


markerikson:
Tweaked the wording a bit. Thank you!

tbrlpld:
Yours is so much better! Thank you for fixing this!

PR #4329, opened 2022-04-15
fixed typo in redux essential tutorial part-1-overview-concepts

uzlash:
---
---

## Checklist

  -  I don't think so
  - Yes

- **Section**:
- Tutorials
- **Page**:
- part-1-overview-concepts

## What is the problem?
Just a quick typo fix no problem
Changed the array value from d to b

### DIFF
--- a/docs/tutorials/essentials/part-1-overview-concepts.md
+++ b/docs/tutorials/essentials/part-1-overview-concepts.md
@@ -165,7 +165,7 @@ obj.b = 3
 const arr = ['a', 'b']
 // In the same way, we can change the contents of this array
 arr.push('c')
-arr[1] = 'd'
+arr[1] = 'b'
 ```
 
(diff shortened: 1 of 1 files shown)


markerikson:
Hiya, thanks for the PR.  Looking at it, I think the current contents are correct and not a typo.  Thanks anyway!

uzlash:
Hello mark, the content isn't correct because this  is what is in the docs:

const arr = ['a', 'b']
// In the same way, we can change the contents of this array
arr.push('c')
arr[1] = 'd'

and it is supposed to be arr[1] = 'b' and not 'd'

Best regards

markerikson:
That's actually my point.  `'b'` is what's _already_ in the array.  But, this snippet is showing how you can _change_ what's in the array.  In other words, it's intentionally showing "this is how you can overwrite `'b'` with `'d'` instead".

uzlash:
Ohhhh, thank you for the clarification then, I appreciate it.

Warm regards

PR #4330, opened 2022-04-15
Issues/3900

phapdinh:
---
---

## Checklist

- [x ] Is there an existing issue for this PR?
  - https://github.com/reduxjs/redux/issues/3900
- [x ] Have the files been linted and formatted?

- **structuring-reducers**:
- **reusing-reducer-logic**:

## What is the problem?
Provide an example of how to reuse reducer logic with action creators

Add an example of using reusing redux with action creators


timdorr:
Cool. Thanks!

Issue #4325, opened 2022-04-09
Mark `createStore` as deprecated


markerikson:
We don't want anyone using the core `createStore` method directly in their apps today. We want them using `configureStore` from RTK instead.

We should tag `createStore` as `@deprecated` on both the 4.x and 5.x branches, and do another 4.x release.

**edit**

After a bunch of discussion, it's become clear that a lot of people aren't familiar with what RTK actually _is_, or what the technical differences are between the core `createStore` API and RTK's `configureStore`.  Let me provide some background resources.

First, I had a chance to do a livestream with Jason Lengstorf where we explained what people disliked about Redux historically, and how RTK was created to solve those issues.  We then built a small example app with RTK+TS:

- https://www.youtube.com/watch?v=9zySeP5vH9c

Next, the "Modern Redux with RTK" tutorial page shows the differences in usage between "vanilla" Redux and RTK for things like store setup and writing reducers:

- https://redux.js.org/tutorials/fundamentals/part-8-modern-redux

And for a lot of additional background on why RTK was created and how it evolved, see [my "Redux Toolkit 1.0" announcement post](https://blog.isquaredsoftware.com/2019/10/redux-toolkit-1.0/).

For the actual technical differences, [Lenz had a very good explanation on Twitter today]():

> In short: `configureStore` calls `applyMiddleware` and `combineReducers` for you and sets up the devtools, all while having a more "sane" api than `createStore`. For that reason alone, it should be a technical successor of `createStore`.
> In dev, it also adds two additional middlewares per default (middleware added by `configureStore` is freely configurable): One checks for accidental store mutations, both between renders as well as inside the reducer.
> That means it eliminates a whole class of bugs that only surface in "UI sometimes not rerendering" or things "being out of order some time later" - hard and frustrating to find. The other warns if you put non-(de)serializable values like classes into the store.
> Not only do these make a lot of problems with the devtools or libraries like redux-persist, even something as simple as a Date instance can mutate itself out of the normal Redux data flow.
> So, even if you were not to use the *rest* of RTK (which still makes you life a lot easier beyond that, removing the need for action types and manual immutable logic), it would still be a good idea to install RTK and use only `configureStore` out of it, instead of `createStore`.

There's some comparisons of the before and after setup here:

- https://redux.js.org/usage/configuring-your-store
- https://redux.js.org/tutorials/fundamentals/part-8-modern-redux#store-setup

phryneas:
Just a first wording suggestion:

```

[...]

PR #4336, opened 2022-04-18
Mark `createStore` as deprecated, and add `legacy_createStore` alias

markerikson:
This PR marks the `createStore` API as deprecated, and adds a new `legacy_createStore` API as an alias without the deprecation notice.

- Updated `createStore`'s doc description to mark it as `@deprecated`,
  so that it shows up with a visual strikethrough in editors.
- Added new descriptive text to `createStore` to encourage users to
  migrate to RTK, and pointing to the "RTK is Redux Today" docs page
- Rewrote TS typedefs for `createStore` to define it as a function with
  overloads, rather than a varible, to get correct docs tooltips when
  hovering over variable usages
- Added `legacy_createStore` API as an alias without the deprecation

## Goal

Redux Toolkit (the `@reduxjs/toolkit` package) is the right way for Redux users to write Redux code today:

https://redux.js.org/introduction/why-rtk-is-redux-today

Unfortunately, many tutorials are still showing legacy "hand-written" Redux patterns, which result in a much worse experience for users. New learners going through a bootcamp or an outdated Udemy course just follow the examples they're being shown, don't know that RTK is the better and recommended approach, and don't even think to look at our docs.

Given that, the goal of this PR is to provide them with a visual indicator in their editor, like ~~createStore~~ .  When users hover over the `createStore` import or function call, the doc tooltip recommends using `configureStore` from RTK instead, and points them to that docs page.  We hope that new learners will see the strikethrough, read the tooltip, read the docs page, learn about RTK, and begin using it.

To be _extremely_ clear:

**WE ARE _NOT_ GOING TO ACTUALLY REMOVE THE `createStore` API, AND ALL YOUR EXISTING CODE WILL STILL CONTINUE TO WORK AS-IS!**

We are just marking `createStore` as ["deprecated"](https://en.wikipedia.org/wiki/Deprecation#Software):

> "the discouragement of use of some feature or practice, typically because it has been superseded or is no longer considered efficient or safe, without completely removing it or prohibiting its use"

## Rationale

- RTK provides a vastly improved Redux usage experience, with APIs that simplify standard usage patterns and eliminate common bugs like accidental mutations
- We've had suggestions to merge all of RTK into the `redux` core package, or fully deprecate the entire `redux` package and rename it to `@reduxjs/core`.  Unfortunately, those bring up too many complexities:
  - We already had a package rename from `redux-starter-kit` to `@reduxjs/toolkit`, and all of our docs and tutorials have pointed to it for the last three years. I don't want to put users through another whiplash package transition for no real benefit
  - Merging or rearranging our packages would effectively require merging all of the Redux repos into a single monorepo.  That would require hundreds of hours of effort from us maintainers, including needing to somehow merge all of our docs sites together.  We don't have the time to do that.
- I don't want to add _runtime_ warnings that would be really annoying

So, this is the minimum possible approach we can take to reach out to users who otherwise would never know that they are following outdated patterns, while avoiding breaking running user code or having to completely rewrite our package and repo structure.

## Results

When a user imports `createStore` in their editor, they will see a visual strikethrough.  Hovering over it will show a doc tooltip that encourages them to use `configureStore` from RTK, and points to an explanatory docs page:

image

Again, _no_ broken code, and _no_ runtime warnings.  

If users do not want to see that strikethrough, they have three options:

- Follow our suggestion to switch over to Redux Toolkit and `configureStore`
- Do nothing. It's just a visual strikethrough, and it doesn't affect how your code behaves. Ignore it.
- Switch to using the `legacy_createStore` API that is now exported, which is the exact same function but with no `@deprecation` tag. The simplest option is to do an aliased import rename:

image

## Plan

I intend to release this as Redux 4.2.0 as soon as this is merged.

## More Details

See the extensive additional discussion in #4325 .

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -372,34 +372,134 @@ export interface StoreCreator {
 }
 
 /**
+ * @deprecated
+ *
+ * **We recommend using the `configureStore` method
+ * of the `@reduxjs/toolkit` package**, which replaces `createStore`.
+ *
+ * Redux Toolkit is our recommended approach for writing Redux logic today,
+ * including store setup, reducers, data fetching, and more.
+ *
+ * **For more details, please read this Redux docs page:**
+ * **https://redux.js.org/introduction/why-rtk-is-redux-today**
+ *
+ * `configureStore` from Redux Toolkit is an improved version of `createStore` that
+ * simplifies setup and helps avoid common bugs.
+ *
+ * You should not be using the `redux` core package by itself today, except for learning purposes.
+ * The `createStore` method from the core `redux` package will not be removed, but we encourage
+ * all users to migrate to using Redux Toolkit for all Redux code.
+ *
+ * If you want to use `createStore` without this visual deprecation warning, use
+ * the `legacy_createStore` import instead:
+ *
+ * `import { legacy_createStore as createStore} from 'redux'`
+ *
+ */
+export declare function createStore<S, A extends Action, Ext, StateExt>(
+  reducer: Reducer<S, A>,
+  enhancer?: StoreEnhancer<Ext, StateExt>
+): Store<S & StateExt, A> & Ext
+/**
+ * @deprecated
+ *
+ * **We recommend using the `configureStore` method
+ * of the `@reduxjs/toolkit` package**, which replaces `createStore`.
+ *
+ * Redux Toolkit is our recommended approach for writing Redux logic today,
+ * including store setup, reducers, data fetching, and more.
+ *
+ * **For more details, please read this Redux docs page:**
+ * **https://redux.js.org/introduction/why-rtk-is-redux-today**
+ *
+ * `configureStore` from Redux Toolkit is an improved version of `createStore` that
+ * simplifies setup and helps avoid common bugs.
+ *
+ * You should not be using the `redux` core package by itself today, except for learning purposes.
+ * The `createStore` method from the core `redux` package will not be removed, but we encourage
+ * all users to migrate to using Redux Toolkit for all Redux code.
+ *
+ * If you want to use `createStore` without this visual deprecation warning, use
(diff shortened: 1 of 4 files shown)


budarin:
I'm not agree with the decision -we will never use redux toolkit because of redundant dependencies  

you need to leave createStore for those who know what to do and how to do it, and let beginners use the redux toolkit

phryneas:
@budarin you did read the deprecation note that said that it will not go away, right?

markerikson:
@budarin we made the release, and we're not changing it at this point.

That said, I am genuinely curious on what your reasons are to avoid using RTK. Can you give some more details on why?

budarin:
@markerikson 

because we want to use a unified approach in changing data in redux and in react.

Mutating data in redux will form a bad habit of mutating data in React, which will lead to errors in the code

Plus, we are not satisfied with another dependency that increases the size of the application

markerikson:
@budarin : as I've said repeatedly in the other thread: it's our job as maintainers to make the decisions that we think will have the best results for our users, and we made this decision because we think it will help the most people.

As for understanding state update syntax and bundle size, reasons for wanting people to use Immer are listed here:

https://redux-toolkit.js.org/usage/immer-reducers#why-immer-is-built-in

markerikson:
@budarin : this feels really hyperbolic tbh :(

We most definitely did not "criminalize" anything. We are intentionally keeping all of our users' code running as-is!  We didn't even add a runtime warning.  In fact, the release notes explicitly say one of the options is "ignore this entirely, it doesn't affect how your code runs".

All of we've done is add a visual marker and a descriptive note saying "we recommend using RTK".

At this point there's nothing else worth debating here, because we've made the change, are not undoing it, and I'm getting tired of having our plainly stated reasoning ignored.

PR #4339, opened 2022-04-19
Fix style guide page URL

markerikson:
The Style Guide page is _supposed_ to be at `/style-guide/style-guide` , but that seems to have changed to just `/style-guide/` as of the recent Docusaurus upgrade.

~~Adding a specific slug _may_ fix that.~~

Apparently this is an intentional thing:

https://docusaurus.io/docs/sidebar/autogenerated#category-index-convention

Switching to redirects instead - we'll keep the new URLs, but have to handle them okay for SEO and user experience.


Josh-Cena:
Try clearing Netlify cache. Looks like https://github.com/facebook/docusaurus/issues/5757

Josh-Cena:
Technically there's https://redux.js.org/introduction (previously with `README` slug) but it seems to be hidden from the user

markerikson:
Grr. I added a pair of Netlify redirects:

```
/usage/index                               /usage/
/style-guide/style-guide                  /style-guide/
```

The Style Guide one works:

https://deploy-preview-4339--redux-docs.netlify.app/style-guide/style-guide

But the Usage Guide one does not:

https://deploy-preview-4339--redux-docs.netlify.app/usage/index

(404s)

Uh... @Josh-Cena , I know this is probably more likely a Netlify thing than a Docusaurus thing, but any idea why that one isn't working?

Josh-Cena:
I think this is a Netlify quirk. When you look at the response, you see that https://deploy-preview-4339--redux-docs.netlify.app/usage/index actually already returns 304 instead of 301. My guess is that Netlify believes `/usage/index` and `/usage/` are the same thing.

markerikson:
Hrm. Does that mean that a redirect entry isn't going to work? Any suggestions for avoiding a 404 here?

Josh-Cena:
I'm experimenting. Could you try redirecting `/usage/index/` instead?

markerikson:
Yep, pushing...

Josh-Cena:
Mmm, still does work. cc @slorber since you are probably more familiar with Netlify

website/_redirects
```diff
@@ -188,4 +188,11 @@
...
 
-/recipes/*  /usage/:splat
\ No newline at end of file
+/recipes/*  /usage/:splat
+
+
+# 2022 Docusaurus "index file" update:
+# https://docusaurus.io/docs/sidebar/autogenerated#category-index-convention
+
+/usage/index/                               /usage/
```
Josh-Cena, line 197:
```suggestion
/usage/index                               /usage/   301!
```

https://docs.netlify.com/routing/redirects/rewrites-proxies/#shadowing

markerikson:
Given that I don't think I've ever linked directly to the "Usage Guide Index" page externally, but I _have_ frequently linked to the Style Guide, I'm going to go ahead and merge this PR now as-is even though it doesn't fix the `/usage/index` problem.

I'd really appreciate any further suggestions you can offer to fix that issue, though!

markerikson:
HAH I shoulda waited another 30 seconds before hitting merge :)

Okay, lemme put up another PR with that

PR #4366, opened 2022-06-09
(docs) minor improve

camiaei:
💪`because we have the store right there:`
😢but the store isn't right there
💪so I added the store right there


markerikson:
Hah, thanks!

PR #4379, opened 2022-06-27
Update `test-utils.tsx` location

nvh95:
---
---

## Checklist

- **Section**: Code Quality
- **Page**: Writing Tests

## What is the problem?

- `utils/test-utils.tsx` should be `src/test-utils.tsx`. Since people can use absolute import and they just need to do:
```diff
-import { render, screen } from 'utils/test-utils';
+import { render, screen } from 'test-utils';
```

I guess this is a typo in docs @markerikson

### DIFF
--- a/docs/usage/WritingTests.md
+++ b/docs/usage/WritingTests.md
@@ -170,7 +170,7 @@ The custom render function should let us:
 
 A typical custom render function setup could look like this:
 
-```tsx title="utils/test-utils.tsx"
+```tsx title="src/test-utils.tsx"
 import React, { PropsWithChildren } from 'react'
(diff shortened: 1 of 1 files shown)


markerikson:
In this case it's not really a typo, for two reasons:

- The filename is just for display purposes over the codeblock. There's no _real_ file here.
- Notionally, I wrote those file paths assume that we're already inside of `src`.  So, there's an `app/store.ts` block, and if it were a real repo the path would be `src/app/store.ts`, but I didn't think it's worth including `src` in here.

Given that, I'm going to close this, but thanks for helping trying to fix this!

PR #4383, opened 2022-07-06
Enable prefer type-only imports

raulsanchez1024:
This PR really doesn't fit under any of the categories available. Sorry if this kind of PR is not accepted, I will close if it's not.

This PR enables prefer type-only imports eslint rule. I noticed that `redux-toolkit` has this rule enabled so I thought it'd be good to port it over `redux` as well. 

Related to issue: https://github.com/reduxjs/redux/issues/4382


raulsanchez1024:
@markerikson - I forgot about this PR. Should be good to merge now

timdorr (approved):

EskiMojo14:
thanks!

PR #4390, opened 2022-07-21
Add status and todoList to initialState

gotcadumitru:
Inside the "todosLoading" action creator, the status value is set to 'loading' but the initial state is an array, so I propose the following initialState:
```
const initialState = {
    todoList: [],
    status: null,
}
```
Type of PR:


timdorr:
Thanks!

PR #4391, opened 2022-07-23
Fix lint error

enriquejr99:
Fixes lint error in counter example.
Use `npm run examples:lint` to reproduce.
See [here](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/prefer-screen-queries.md) for error details.


timdorr:
Thanks!

PR #4395, opened 2022-07-28
fixed spelling error

asifurrahaman754:
Thanks for the PR!

### DIFF
--- a/docs/faq/ReactRedux.md
+++ b/docs/faq/ReactRedux.md
@@ -206,7 +206,7 @@ With Redux, you get the power of [Redux Dev Tools Extension](https://github.com/
 
 With React's Context API, you deal with a pair of components speaking only to each other. This gives you nice isolation between irrelevant data. You also have the flexibility of how you may use the data with your components, i.e., you can provide the state of a parent component, and you may pass context data as props to wrapped components.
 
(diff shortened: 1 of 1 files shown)


markerikson:
Huh. I'm... not sure that _either_ of those words are right. And I'm fairly sure I _wrote_ that FAQ entry in the first place :)

Let me tweak it a bit more now.

docs/faq/ReactRedux.md
```diff
@@ -206,7 +206,7 @@ With Redux, you get the power of [Redux Dev Tools Extension](https://github.com/
...
 
-There is a key difference in how Redux and React's Context treat data. Redux maintains the data of your whole app in a giant, stateful object. It deduces the changes of your data by running the reducer function you provide, and returns the next state that corresponds to every action dispatched. React Redux then optimizes component rendering and makes sure that each component re-renders only when the data it needs change. Context, on the other hand, does not hold any state. It is only a conduit for the data. To express changes in data you need to rely on the state of a parent component.
+There is a key difference in how Redux and React's Context treat data. Redux maintains the data of your whole app in a giant, stateful object. It reduces the changes of your data by running the reducer function you provide, and returns the next state that corresponds to every action dispatched. React Redux then optimizes component rendering and makes sure that each component re-renders only when the data it needs change. Context, on the other hand, does not hold any state. It is only a conduit for the data. To express changes in data you need to rely on the state of a parent component.
```
markerikson, line 209:
```suggestion
There is a key difference in how Redux and React's Context treat data. Redux maintains the data of your whole app in a single `state` object. It updates that data by running the reducer function you provide, and returns the next state that corresponds to every action dispatched. React Redux then optimizes component rendering and makes sure that each component re-renders only when the data it needs change. Context, on the other hand, does not hold any state. It is only a conduit for the data, and is used to avoid "prop-drilling" data through each level of components. The actual state management occurs in a parent component via `useState/useReducer`.  Additionally, passing a new value through a context provider causes _all_ components that read the context to re-render.
```

Issue #4295, opened 2022-02-20
Redux Essentials docs error (Memoizing Selector Functions)


bruce-c-liu:
## What docs page needs to be fixed?
https://redux.js.org/tutorials/essentials/part-6-performance-normalization

- **Section**: [Memoizing Selector Functions](https://redux.js.org/tutorials/essentials/part-6-performance-normalization#memoizing-selector-functions)

## What is the problem?
Below is a code snippet from that section.
```js
dispatch(reactionAdded())
const state2 = getState()
// Output selector does not run, because `posts` and `userId` are the same
selectPostsByUser(state2, 'user2')
```
Two issues:
1. `dispatch(reactionAdded())` would throw an error because `reactionAdded()` expects an argument of the form:
```js 
{ postId: 'abcde', reaction: 'heart' }
```
2. Assuming the above were fixed and dispatched correctly, then `selectPostsByUser`'s output selector *will* rerun, because the `reactionAdded` action mutates `posts`.

## What should be changed to fix the problem?
Option 1: Fix the issues mentioned in the code snippet.
Option 2: Dispatch a different action that does not mutate `posts` or `userId`. (e.g. `allNotificationsRead` from the `notificationsSlice`)

timdorr:
Your second point isn't correct. The posts slice *isn't* mutated by the reactionAdded action, just the reactions for that post are. Immer handles this for us and keeps it performant.

As for the first, we're just trying to show that those actions were dispatched. What was provided to them is irrelevant. We're talking about the effects, not writing real code.

PR #4396, opened 2022-08-03
docs: fix memoized selectors code snippet

jeferson-sb:
---
about: Fixing a problem in Memoizing Selector Functions page
---

## Checklist

  - Issue #4295 

- **Section**: https://redux.js.org/tutorials/essentials/part-6-performance-normalization#memoizing-selector-functions
- **Page**: https://redux.js.org/tutorials/essentials/part-6-performance-normalization

## What is the problem?

Dispatching the post reactionAdded action with no arguments makes it so it throws an error, so the user can't see
the output of a memoizing selector.

- Add arguments to `reactionAdded` and `addNewPost` actions

### DIFF
--- a/docs/tutorials/essentials/part-6-performance-normalization.md
+++ b/docs/tutorials/essentials/part-6-performance-normalization.md
@@ -509,7 +509,7 @@ export const Navbar = () => {
 
 ## Improving Render Performance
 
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks, but this section is for us to show the effects of dispatching those actions, not for real code. Adding too much example code makes it harder to read.

PR #4401, opened 2022-08-10
Remove unnecessary key attribute

wardxela:
Using `key` attribute here is a mistake. It makes sense only in context of an array according to official React documentation (see [here](https://reactjs.org/docs/lists-and-keys.html#extracting-components-with-keys)).


markerikson:
Keys _can_ make sense in non-array contexts, such as recreating a form if the user selects a different item.

But yeah, in this case I copy-pasted some JSX from an earlier snippet where this _was_ in an array loop, so the key doesn't make sense in this case.

PR #4404, opened 2022-08-17
Docs: delete extraneous full stop after exclamation mark

komali2:
---
---

## Checklist

  - _link issue here_

- **Section** Fundamentals
- **Page**: 3

## What is the problem?

There's [no situation](https://www.thepunctuationguide.com/terminal-points.html) in which a period should follow an exclamation point, including when inside quotes, regardless of text styling.

Change "it's okay to have other state values outside of Redux!." to "it's okay to have other state values outside of Redux!"


markerikson:
Thanks!

PR #4405, opened 2022-08-18
Docs: Update RTK Query basics to replace single quotes with double

komali2:
---
about: Proper usage of quotation marks
---

## Checklist

  - _link issue here_

- **Page**: Part 7: RTK Query Basics

## What is the problem?

The sentence fragment

>  (no 's' this time)

improperly uses single quotes. Double quotes should be used in almost all instances, except when already within a fragment wrapped in double quotes: https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Quotations/faq0001.html

Changes to 

> (no "s" this time)


markerikson:
Appreciate the PR, but I don't think this change is necessary.  (We use single quotes in our code style anyway, and this is also an example of a single character.)

PR #4408, opened 2022-09-02
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [master]
 
+permissions:
+  contents: read
+
 jobs:
   build:
     name: Test Suite


sashashura:
The workflow is started on `pull_request` **and** on `push`. Every time GitHub generates a temporary repository token for the workflow. `On pull_request` from forks it has only content read permissions. This is logical because it builds and runs (like tests) a potentially untrusted code. `On push` happens when the changes are accepted or maintainers directly push commit to the repository. In this case the temporary token has read/write permissions to the repository, but this is OK since the changes are trusted/reviewed by maintainers. However when building it is common to download and install various third party libraries and tools. They potentially can be forged at some moment and may use the github token to modify the repository.
The section tells GitHub workflow runner to issue `contents: read` in both case. It is a safe change, like not running things with sudo unnecessarily.

timdorr:
Thanks, but this is already set at the org level: 

image

sashashura:
Great, sorry for the hassle. No way to see it externally.

PR #4411, opened 2022-09-05
Update all GitHub actions

ddzz:
No.

Actions used in GitHub Actions should be updated to the latest version and not use any beta versions. Dependabot could be used to automate this.


timdorr:
Thanks!

PR #4412, opened 2022-09-07
Middlewares TypeChecking problem and Redux Toolkit 

andra961:
---
---

## Checklist

  - _link issue here_

- **Section**: usage
- **Page**: UsageWithTypescript.md

## What is the problem?
As a Redux Toolkit user, when i stumbled upon the circular depency problem when TypeChecking the middlwares and read this page, i was confused by the current snippet and explanation. By setting a store with the Redux Toolkit quick start example (https://redux-toolkit.js.org/tutorials/quick-start), you never use `combineReducers` and so i did not really know what to change (i had to do further research). 

I added a code snippet (how i solved the problem in my app) for Redux Toolkit hoping that it can be more clear for users them and save some searches on google. Let me know what you think please!

### DIFF
--- a/docs/usage/UsageWithTypescript.md
+++ b/docs/usage/UsageWithTypescript.md
@@ -267,6 +267,21 @@ const rootReducer = combineReducers({ ... });
 type RootState = ReturnType<typeof rootReducer>;
 ```
 
+Switching the type definition of `RootState` with Redux Toolkit example:
+
+```ts
+//instead of defining the reducers in the reducer field of configureStore, combine them here:
+const rootReducer = combineReducers({ counter: counterReducer });
+
+//then set rootReducer as the reducer object of configureStore
+const store = configureStore({
+  reducer: rootReducer,
+  middleware: [...],
+});
+
+type RootState = ReturnType<typeof rootReducer>;
+```
+
 ### Type Checking Redux Thunks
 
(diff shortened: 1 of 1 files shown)


markerikson:
Thanks!

docs/usage/UsageWithTypescript.md
```diff
@@ -267,6 +267,21 @@ const rootReducer = combineReducers({ ... });
...
+Switching the type definition of `RootState` with Redux Toolkit example:
+
+```ts
+//instead of defining the reducers in the reducer field of configureStore, combine them here:
+const rootReducer = combineReducers({ counter: counterReducer });
+
+//then set rootReducer as the reducer object of configureStore
+const store = configureStore({
+  reducer: rootReducer,
+  middleware: [...],
```
phryneas, line 279: Middleware here should almost never be an array though - and if passed in as an array probably with `as const`. Could you switch this up with the `getDefaultMiddleware` notation?

andra961: sure! would you be ok with this syntax?

```ts
middleware: (getDefaultMiddleware) => getDefaultMiddleware().concat(yourMiddleware),
```

phryneas: That would be fine :)

andra961: i submitted the PR with the agreed changes! here the [request](https://github.com/reduxjs/redux/pull/4414)

PR #4417, opened 2022-09-12
fix: type error in `UserDisplay` component example

MikePrus:
There is a type error in the `UserDisplay` component example because we don't display the name but the object instead.

### DIFF
--- a/docs/usage/WritingTests.mdx
+++ b/docs/usage/WritingTests.mdx
@@ -567,13 +567,13 @@ import { fetchUser, selectUserName, selectUserFetchStatus } from './userSlice'
 
 export default function UserDisplay() {
   const dispatch = useAppDispatch()
-  const user = useAppSelector(selectUserName)
+  const userName = useAppSelector(selectUserName)
   const userFetchStatus = useAppSelector(selectUserFetchStatus)
 
   return (
     <div>
       {/* Display the current user name */}
-      <div>{user.name}</div>
+      <div>{userName}</div>
       {/* On button click, dispatch a thunk action to fetch a user */}
       <button onClick={() => dispatch(fetchUser())}>Fetch user</button>
       {/* At any point if we're fetching a user, display that on the UI */}
--- a/docs/usage/WritingTests.mdx
+++ b/docs/usage/WritingTests.mdx
@@ -125,7 +125,7 @@ const userSlice = createSlice({
   }
 })
 
-export const selectUser = (state: RootState) => state.user.name
+export const selectUserName = (state: RootState) => state.user.name
 export const selectUserFetchStatus = (state: RootState) => state.user.status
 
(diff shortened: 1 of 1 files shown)


markerikson:
Hi! In this case the example is actually correct.  If you look at the `WritingTests.mdx` file, there's a "hidden" `userSlice.ts` file defined in the code snippet, with these lines:

```ts
export const selectUser = (state: RootState) => state.user.name
export const selectUserFetchStatus = (state: RootState) => state.user.status
export default userSlice.reducer
```

So, it _is_ getting the `user.name` string already.

Could you change the selector name to be `selectUserName` and the variable name to be `userName` instead?

MikePrus:
> Hi! In this case the example is actually correct. If you look at the `WritingTests.mdx` file, there's a "hidden" `userSlice.ts` file defined in the code snippet, with these lines:
> 
> ```ts
> export const selectUser = (state: RootState) => state.user.name
> export const selectUserFetchStatus = (state: RootState) => state.user.status
> export default userSlice.reducer
> ```
> 
> So, it _is_ getting the `user.name` string already.
> 
> Could you change the selector name to be `selectUserName` and the variable name to be `userName` instead?

Sure, I did this in the second commit please review.

docs/usage/WritingTests.mdx
```diff
@@ -563,17 +563,17 @@ export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector
...
   const dispatch = useAppDispatch()
-  const user = useAppSelector(selectUser)
+  const user = useAppSelector(selectUserName)
   const userFetchStatus = useAppSelector(selectUserFetchStatus)
 
   return (
     <div>
       {/* Display the current user name */}
-      <div>{user}</div>
+      <div>{user.name}</div>
```
markerikson, line 576:
```suggestion
      <div>{userName}</div>
```

docs/usage/WritingTests.mdx
```diff
@@ -563,17 +563,17 @@ export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector
 // file: features/users/UserDisplay.tsx noEmit
 import React from 'react'
 import { useAppDispatch, useAppSelector } from '../../app/hooks'
-import { fetchUser, selectUser, selectUserFetchStatus } from './userSlice'
+import { fetchUser, selectUserName, selectUserFetchStatus } from './userSlice'
 
 export default function UserDisplay() {
   const dispatch = useAppDispatch()
-  const user = useAppSelector(selectUser)
+  const user = useAppSelector(selectUserName)
```
markerikson, line 570:
```suggestion
  const userName = useAppSelector(selectUserName)
```

markerikson (commented):
Fix userName naming

vincentCheng:
I also found the error.

PR #4418, opened 2022-09-13
Update part-6-performance-normalization.md

anthonyss09:
line 480 the original filter function returned an array notifications where !notification.read was the condition. Due to the second render of the NotificationList component allNotificationsRead is being called in useLayoutEffect and notifications.filter(notification=>!notification.read).length becomes 0. So we should return a notifications  array from the filter function with the condition notifications.isNew.

---
---

## Checklist

- [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
  - _link issue here_
- [ ] Have the tests been updated to match the changes in the PR?
- [ ] Have you run the tests locally to confirm they pass?

### DIFF
--- a/docs/tutorials/essentials/part-6-performance-normalization.md
+++ b/docs/tutorials/essentials/part-6-performance-normalization.md
@@ -477,7 +477,7 @@ export const Navbar = () => {
   const dispatch = useDispatch()
   // highlight-start
   const notifications = useSelector(selectAllNotifications)
-  const numUnreadNotifications = notifications.filter(n => !n.read).length
+  const numUnreadNotifications = notifications.filter(n => n.isNew).length
   // highlight-end
   // omit component contents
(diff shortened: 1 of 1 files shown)


timdorr:
This is correct as-is. The intended UX is the number of "unread" notifications are in the NavBar, not the number of "new" notifications. This should go to 0 when you click on the tab, displaying NotificationList.

anthonyss09:
> This is correct as-is. The intended UX is the number of "unread" notifications are in the NavBar, not the number of "new" notifications. This should go to 0 when you click on the tab, displaying NotificationList.

I see... I didn't realize the notifications badge was not intended to be present whilst viewing the notifications, but that makes sense. Thanks for your time!

markerikson:
@anthonyss09 thank you, glad to hear it was helpful!

PR #4419, opened 2022-09-13
Update combineReducers.md

eperrier:
App.js calls 'reducer' but index.js exports combineReducers, not reducer. Added const to remedy.

Thanks for the PR!


phryneas:
That's a default export, it gets its name on `import`, not on `export`. That code is correct.

Generally, all over the docs Redux reducers are default exports, so we are going to keep that pattern. 

Thanks a lot for the PR though!

PR #4420, opened 2022-09-20
Ignore "Format all the things!" from git blame

devinrhode2:


phryneas:
Thanks, I'm sure that will come in handy from time to time :)

PR #4425, opened 2022-10-15
remove listener with more semantic code

Einstellung:
## Little Change
A little change for source code. I think use `filter` can make the code more
semantic and easier to read than using `slice`


markerikson:
Appreciate the PR, but this code works and there's no immediate reason to change it.

Issue #4431, opened 2022-11-11
Codesandbox embeds mobile scrolling broken


JetUni:
## What docs page needs to be fixed?

All sections and pages that have a codesandbox.io embed

## What is the problem?

When viewing the codesandbox on mobile (my phone is a Samsung Galaxy S22 Ultra), the embeds will not scroll which makes them utterly useless as you can then only view the code that is pasted in snippets throughout the tutorials, but not the whole code sample

## What should be changed to fix the problem?

Enable the CodeMirror editor for all of the embeds. That editor doesn't have this problem and the editor will also reduce the size of the embeds significantly as a bonus

PR #4432, opened 2022-11-11
fix: codesandbox embeds mobile scrolling

JetUni:
Currently the embeds for codesandbox.io do not allow you to scroll on mobile devices (tested on Samsung Galaxy S22 Ultra). This is a problem because users are unable to view the whole project as they are reading through the amazing documentation and examples that have been created. Fix this by updating the src url to include the codemirror option which uses the CodeMirror editor instead of Monaco and also reduces the embed size as well

---
---

## Checklist

  - #4431

All sections and pages that use the codesandbox.io embeds

## What is the problem?

Users are unable to scroll within files in the embeds so they are unable to view the entire file

Switch to using the CodeMirror editor instead of the default Monaco editor from codesandbox.io

### DIFF
--- a/docs/introduction/Examples.md
+++ b/docs/introduction/Examples.md
@@ -21,7 +21,7 @@ open index.html
 
 Or check out the [sandbox](https://codesandbox.io/s/github/reduxjs/redux/tree/master/examples/counter-vanilla):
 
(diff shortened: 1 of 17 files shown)


JetUni:
Tested and working locally as well as on the netlify link above using my Samsung Galaxy S22 Ultra

timdorr:
Thanks for looking into this!

PR #4433, opened 2022-11-12
Update typescript.md

Klotz95:
---
about: Added missing export statement
---

## Checklist

- [] Is there an existing issue for this PR?
  - _link issue here_

- **Section**: tutorials
- **Page**: typescript.md

## What is the problem?
I think there is an export statement missing. Otherwise you can't access the store in other ts-files.

Added the missing export statement


timdorr:
Thanks, but that's not what we're trying to explain here. We're only talking about types.

PR #4434, opened 2022-11-13
Remove outdated content in "Do I have to put all my state into Redux?"

jshoung:
---
name: :book: Updated Documentation Content
about: Remove outdated content in "Do I have to put all my state into Redux?"
---

## Checklist

  - https://github.com/reduxjs/redux/issues/4421

- **Section**: Do I have to put all my state into Redux? Should I ever use React's `setState()`?
- **Page**: Redux FAQ: Organizing State

- There's a mix of outdated links that mention either class components and/or unmaintained libraries, which can be confusing for new developers
- I went through each link and only kept the ones that had answers which were agnostic about the version of React used (or used React hooks)
  - I had a minor concern about keeping [this link](https://stackoverflow.com/questions/35328056/react-redux-should-all-component-states-be-kept-in-redux-store/35328862#35328862) because it mentions `this.state` - I submitted an edit that rewords the sentence more generically to "You're better off using _component state_ and calling it a day."

### DIFF
--- a/docs/faq/OrganizingState.md
+++ b/docs/faq/OrganizingState.md
@@ -37,7 +37,6 @@ Some common rules of thumb for determining what kind of data should be put into 
 - Do you want to cache the data (ie, use what's in state if it's already there instead of re-requesting it)?
 - Do you want to keep this data consistent while hot-reloading UI components (which may lose their internal state when swapped)?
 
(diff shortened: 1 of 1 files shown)


jshoung:
I'm in the process of helping to sweep through the FAQ entries and update things - do let me know if you'd prefer the PRs to be less granular. E.g. one page per PR.

I felt like doing it section-by-section would be easier to review but happy to adjust to whatever works best!

timdorr:
Thanks! This format works well so we don't have small things holding up a larger PR.

PR #4446, opened 2022-11-28
Dead link for ryanFunduk removed.

KunwarVSuryavanshi:
Name: :memo: Documentation Fix
About: Link for dead website removed from **Prerequisite Reducer Concepts** section.


markerikson:
Thanks! Yeah, we really ought to replace those links with https://daveceddia.com/react-redux-immutability-guide/ instead.

PR #4449, opened 2022-12-10
an Update createStore.ts

Kananmas:
on line 208 // before change //  it does not check whether or not the value of index is greater or equal to zero meaning that if index is equal to -1 it will carry out the splice opreation on the array  and it will delete the last element of the array instead of removing the element that we intended to remove . 

**example**:consider an array [1,2,3,4,5] and use splice(-1,1) on it

 
let arr = [1,2,3,4,5]

arr.splice(-1,1)

console.log(arr); // [1,2,3,4]

arr.splice(-1,1)
 
console.log(arr);//[1,2,3,]

Capture

// after change // we check if  value of index was bigger or equal to zero then carry out the splice opreation to prevent any unwanted changes

Thanks for the PR!

### DIFF
--- a/src/createStore.ts
+++ b/src/createStore.ts
@@ -205,7 +205,7 @@ export default function createStore<
 
       ensureCanMutateNextListeners()
       const index = nextListeners.indexOf(listener)
-      nextListeners.splice(index, 1)
+      if(index >= 0) nextListeners.splice(index, 1)
       currentListeners = null
     }
   }


phryneas:
It should be impossible that that line of code is ever reached if `listener` is not in `nextListeners`.  
Can you provide a test that would fail without this change?

timdorr:
Thanks, but as Lenz said, this wouldn't be possible to reach without a valid index value. There is essentially a proof for this via Typescript, which is doing this type checking for us.

PR #4451, opened 2022-12-13
bug: splice index error handle

shohreh-zarkesh:
Hi guys, 

When I follow my application error in my case, I found the error in splice index.

Thanks for the PR!

### DIFF
--- a/src/createStore.ts
+++ b/src/createStore.ts
@@ -205,7 +205,10 @@ export default function createStore<
 
       ensureCanMutateNextListeners()
       const index = nextListeners.indexOf(listener)
-      nextListeners.splice(index, 1)
+      if(index >= 0)
+      {
+        nextListeners.splice(index, 1)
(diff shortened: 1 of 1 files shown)


phryneas:
It should be impossible that that line of code is ever reached if listener is not in nextListeners.
Can you provide a test that would fail without this change?

timdorr:
This is the same change as #4449, which is not an actual issue.

phryneas:
Pretty much the reason why I copied my comment over from there.

@shohreh-zarkesh we would accept a PR for this if you can prove that this happens - but we really need proof. From reading the code, it should be 100% impossible and never happen.

Does this actually happen to you or do you have another reason to open this PR?

markerikson:
Vaguely related: for Redux 5.0 we should convert `listeners` into a `Set`. We can safely assume ES6 support by now :)

PR #4452, opened 2022-12-14
change reducer type validation place

Ahmed-Hakeem:
change reducer type validation to be at the very start of createStore function since all operations after this condition is met are neglected, and this also forces users who pass their own enhancers to pass the reducer as a function

### DIFF
--- a/src/createStore.ts
+++ b/src/createStore.ts
@@ -68,6 +68,15 @@ export default function createStore<
   preloadedState?: PreloadedState<S> | StoreEnhancer<Ext, StateExt>,
   enhancer?: StoreEnhancer<Ext, StateExt>
 ): Store<ExtendState<S, StateExt>, A, StateExt, Ext> & Ext {
+  
+  if (typeof reducer !== 'function') {
+    throw new Error(
+      `Expected the root reducer to be a function. Instead, received: '${kindOf(
+        reducer
+      )}'`
+    )
+  }
+
   if (
     (typeof preloadedState === 'function' && typeof enhancer === 'function') ||
     (typeof enhancer === 'function' && typeof arguments[3] === 'function')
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks, but this functionally doesn't do anything. In addition, the type check would fail for the wrong kind of reducer.

Ahmed-Hakeem:
> Thanks, but this functionally doesn't do anything. In addition, the type check would fail for the wrong kind of reducer.

It does, check the below cases :

1.  function will return before throwing this error, check the comment below 
```js
if (
    (typeof preloadedState === 'function' && typeof enhancer === 'function') ||
    (typeof enhancer === 'function' && typeof arguments[3] === 'function')
  ) {
    throw new Error(
      'It looks like you are passing several store enhancers to ' +
        'createStore(). This is not supported. Instead, compose them ' +
        'together to a single function. See https://redux.js.org/tutorials/fundamentals/part-4-store#creating-a-store-with-enhancers for an example.'
    )
  }
  if (typeof preloadedState === 'function' && typeof enhancer === 'undefined') {
    enhancer = preloadedState as StoreEnhancer<Ext, StateExt>
    preloadedState = undefined
  }

  if (typeof enhancer !== 'undefined') {
    if (typeof enhancer !== 'function') {
      throw new Error(
        `Expected the enhancer to be a function. Instead, received: '${kindOf(
          enhancer
        )}'`
      )
    }
    /**function will return before throwing reducer's type error,
    but the returned createStore call or its sub calls from the enhancer will throw if you pass it wrong reducer type    
   */
    return enhancer(createStore)(
      reducer,
      preloadedState as PreloadedState<S>
    ) as Store<ExtendState<S, StateExt>, A, StateExt, Ext> & Ext
  }

  if (typeof reducer !== 'function') {
    throw new Error(
      `Expected the root reducer to be a function. Instead, received: '${kindOf(
        reducer
      )}'`
    )
  }
```

```js
//so if you pass these types function will return before throwing this error
const store = createStore(undefined, {} ,enhancer); 
```

so I think throwing reducer's type error from the very start will be better since there will be no cost from putting it at the very start.

also, doing all these conditions before throwing the reducer's type error is not the best even if it's low cost, as passing the wrong reducer type will in the end throw an error and break all that as you have said.

timdorr:
Hmm, I guess that does make sense. I'm so used to TS now that I don't really think about these things anymore! 😄

Ahmed-Hakeem:
> Hmm, I guess that does make sense. I'm so used to TS now that I don't really think about these things anymore! 😄

Never mind ..... thank you ❤️

PR #4466, opened 2023-01-18
Update title: Why Redux Tookit

aghArdeshir:
Another alternative could be: What Redux Toolkit is and How To Use Rdux Today

image


markerikson:
Hi. Appreciate the PR, but the current title is correct and intended.

PR #4476, opened 2023-01-30
Replace listeners array with a Map for better performance

markerikson:
Redux has always used an array of listener callbacks internally.  However, as we found out with React-Redux, using `listeners.findIndex(callback)` is bad for performance ( https://github.com/reduxjs/react-redux/pull/1523 , https://github.com/reduxjs/react-redux/issues/1869 ).

React-Redux uses a custom linked list internally.  That's way more than we need here.  But, now that we're assuming modern browsers, it's safe to assume ES2015+ support, and that `Map/Set`s are available.  

I initially tried `Set<ListenerCallback>`, but found that one test failed... because it was passing in the same `jest.fn()` instance _twice_.  A `Set` compares by reference, so the callback got removed in the first `unsubscribe()` call and was no longer around for the second.

We don't have a formal statement that we support passing in the same callback more than once, but that's the implicit semantics thus far.

Given that, a `Map<number, ListenerCallback>` works equivalently, and we'll just have an incrementing counter for the key.

Also, I double-checked perf for cloning `Set`s before I switched over to `Map`s.  Somewhat surprisingly, `existing.forEach(item => newSet.add(item))` runs _faster_ than `new Set(existing)`, or using `for..of`, so I assume the same is true for `Map`s.  (I will note that I was creating `Set`s with 1M items, and the difference was like 150ms to 120ms. So I don't expect that to have any meaningful cost here.)

In practice, I don't expect this to make a lot of difference, because React-Redux's `<Provider>` normally subscribes directly to the store, and all child components subscribe to that `<Provider>`'s internal `Subscription` instance.

But getting rid of this little perf footgun is nice.

### DIFF
--- a/src/createStore.ts
+++ b/src/createStore.ts
@@ -6,7 +6,8 @@ import {
   StoreEnhancer,
   Dispatch,
   Observer,
-  ExtendState
+  ExtendState,
+  ListenerCallback
 } from './types/store'
 import { Action } from './types/actions'
 import { Reducer } from './types/reducers'
@@ -119,8 +120,9 @@ export function createStore<S, A extends Action, Ext = {}, StateExt = never>(
 
   let currentReducer = reducer
   let currentState = preloadedState as S
-  let currentListeners: (() => void)[] | null = []
+  let currentListeners: Map<number, ListenerCallback> | null = new Map()
   let nextListeners = currentListeners
+  let listenerIdCounter = 0
   let isDispatching = false
 
   /**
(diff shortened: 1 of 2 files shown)


src/createStore.ts
```diff
@@ -132,7 +134,10 @@ export function createStore<S, A extends Action, Ext = {}, StateExt = never>(
    */
   function ensureCanMutateNextListeners() {
     if (nextListeners === currentListeners) {
-      nextListeners = currentListeners.slice()
+      nextListeners = new Map()
+      currentListeners.forEach((listener, key) => {
+        nextListeners.set(key, listener)
+      })
```
nstepien, line 140: Why not do this?
```suggestion
      nextListeners = new Map(currentListeners)
```

nstepien: Huh, sounds like a bug in the js engine.

romellem:
I've been looking at a lot of the react-redux and redux code, and was curious about this implementation split between a `Map` and the doubly-linked list used in https://github.com/reduxjs/react-redux/pull/1523. 

You mention

> [a custom linked list is] way more than we need here

Do you just mean that the linked list is probably the fastest, while using a `Map` is easier since you can rely on native data structures vs rolling your own?

PR #4482, opened 2023-02-12
documentation-edit.md

elite16op:
---
---

## Checklist

  - _link issue here_

- **Section**: Tutorials/Quick Start
- **Page**:  Create a redux state slice

## What is the problem?
At line 93 we created counterSlice and gave a reducers as key on line 98
at line 118 we export default counterSlice.reducer but shouldn't it be export default counterSlice.reducers

I changed it to counterSlice.reducers


markerikson:
Sorry, the current code is correct as-is.

You pass _in_ an object called `reducers`, full of "case reducer" functions.

It _generates_ a `slice.reducer` field, which is the one "slice reducer" function that has all of the "case reducers" inside.

Thanks anyway!

PR #4486, opened 2023-02-12
ref: Update IsPlainObject

Liknox:
Update isPlainObject in order to increase speed and better understanding.

[:bug: Bug fix or new feature](?template=bugfix.md)


markerikson:
Hi. Appreciate the PR, but the current implementation works as-is, and that `while` loop was added years ago to handle some specific edge cases.  Thanks anyway!

PR #4489, opened 2023-02-15
Type-check tests

Methuselah96:
In https://github.com/reduxjs/redux/pull/4487 we switched from ts-jest (which runs type-checking on test files) to Vitest (which does not).

This PR runs type-checking on the tests to make sure they pass.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -14,6 +14,7 @@ jobs:
         with:
           filters: |
             src:
+              - '.github/**'
               - 'src/**'
               - 'test/**'
 
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -98,7 +98,7 @@ jobs:
       fail-fast: false
       matrix:
         node: ['16.x']
-        ts: ['4.2', '4.3', '4.4', '4.5', '4.6', '4.7', '4.8', '4.9.2-rc']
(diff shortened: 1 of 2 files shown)


markerikson:
Hmm. Apparently the optimization I added / copied that tries to skip tests if there are no `src` changes is skipping the tests here too :(

While you're at it, can you try tweaking the condition to run if there's changes to `.github`?

Worst case, just kill the condition and always run the tests for now.

Methuselah96:
Odd test failure for TypeScript 4.1. I'll take a look later.

Methuselah96:
rxjs types only support >=4.2 👀 

```
  "name": "rxjs",
  ...
  "typesVersions": {
    ">=4.2": {
      "*": [
        "dist/types/*"
      ]
    }
  },
```

Methuselah96:
I think we're good to drop testing against TypeScript 4.1? It [came out on 11/19/20](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/) so it's more than the 2-year support window [defined by DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped#support-window).

markerikson:
Yep, absolutely!  We've generally done rolling TS version compat in our minor releases anyway, so let's drop 4.1 from the test matrix.  (Might as well add 5.0-beta while we're at it?)

PR #4491, opened 2023-02-17
Add PreloadedState generic

Methuselah96:
## Summary

The main change to pay attention to here is the update to the `Reducer` type. Before this PR `Reducer` has a type of:

```
type Reducer<S, A extends Action> = (
  state: S | undefined,
  action: A
) => S
```

After this PR `Reducer` has this type:

```
type Reducer<S, A extends Action, PreloadedState = S> = (
  state: S | PreloadedState | undefined,
  action: A
) => S
```

### Explanation

Why the need for this change? When the store is first created by `createStore`, the initial state is set to whatever is passed as the `preloadedState` argument (or `undefined` if nothing is passed). That means that the first time that the reducer is called, it is called with the `preloadedState`. After the first call, the reducer is always passed the current state (which is `S`).

For most normal reducers, `S | undefined` accurately describes what can be passed in for the `preloadedState`. However the `combineReducers` function allows for a preloaded state of `Partial<S> | undefined`.

The solution is to have a separate generic that represents what the reducer accepts for its preloaded state. That way `createStore` can then use that generic for its `preloadedState` argument.

### Previous work

Up until now, this has been accounted for with the `$CombinedState` type which is a way to "mark" a type as returned by `combineReducers`. The `PreloadedState` type then looks to see if the type has that "mark" and then accepts `Partial<S>` if it has that "mark."

One of the problems this can cause is that if there are multiple copies of Redux in `node_modules`, then the "mark" for one copy of Redux is different than the "mark" for the other copy, and therefore they won't recognize each other.

This also seems to be causing some problems in https://github.com/reduxjs/redux-toolkit/issues/2068.

This PR removes the need for `$CombinedState` altogether.

## Upgrade Impact

This change does include some breaking changes, but overall should not have a huge impact on users upgrading in user-land.

### Additional generic arguments

The `Reducer`, `ReducersMapObject`, and `createStore` types/function take an additional `PreloadedState` generic which defaults to `S`.

### `combineReducers`

The overloads for `combineReducers` are removed in favor of a single function definition that takes the `ReducersMabObject` as its generic parameter. Removing the overloads was necessary with these changes, since sometimes it was choosing the wrong overload.

### Enhancers

Enhancers that explicitly list the generics for the reducer will need to add the third generic.

### DIFF
--- a/src/createStore.ts
+++ b/src/createStore.ts
@@ -2,7 +2,6 @@ import $$observable from './utils/symbol-observable'
 
 import {
   Store,
-  PreloadedState,
   StoreEnhancer,
   Dispatch,
   Observer,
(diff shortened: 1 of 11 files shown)


src/createStore.ts
```diff
@@ -76,21 +75,23 @@ export function createStore<
 export function createStore<
   S,
   A extends Action,
+  PreloadedState,
```
timdorr, line 78: Why not add this as the last param and make it have a default? That would be less breaking.

Methuselah96: I'm open to it. It feels better to keep the generics for the reducer next to each other and then have the generics for the enhancer next to each other. Considering that we don't encourage direct use of `createStore` or manually specifying generics, it feels worth it to me, but I'm open to whatever you, @markerikson, and @phryneas think is best.

phryneas: While *hopefully*, nobody would ever write those out, but have them inferred by parameter usage, I agree - new arguments should be added at the end, not somewhere in the middle.

Methuselah96:
@markerikson I think this is ready to merge. I've created https://github.com/reduxjs/redux-toolkit/pull/3198 to show that RTK should be able to consume this update fairly easily, which I think was @phryneas's only remaining concern. Let me know if you need anything else from me.

PR #4498, opened 2023-03-08
Add documentation page "Writing Custom Middleware"

### LINKED ISSUE #4472: Share expectations/best practices for Redux middleware
As discussed in an [rtk query issue](https://github.com/reduxjs/redux-toolkit/issues/3118#issuecomment-1403784368), there are some implicit assumptions about how middleware should be written to be compatible with other Redux libraries. Even though it is possible to write any kind of middleware, it would be helpful to document these assumptions so that middleware written by different people is compatible.

Reference

Devs writing their own middleware

How to write middleware

How to write middleware that is compatible with other redux projects like redux toolkit.

That the middleware should not change the return value (even to a Promise) and needs to call next with an action.

dbartholomae:
Fixes #4472

---
---

## PR Type

Add a page

## Checklist

  - #4472 

- **Section**: Usage Guide/Using Redux/Redux Logic and Patterns
- **Page**: Writing Custom Middleware

How-to

Developers thinking about writing their own middleware with low to medium experience with Redux.

How to use Redux and middleware in general, but not how to write middleware.

- Understanding what middleware can do
- Understanding when to write custom middleware vs. using existing packages
- Understanding how to write middleware that is compatible with other Redux middleware like RTK Query middleware

- That they most likely shouldn't write custom middleware

### DIFF
--- a/docs/usage/WritingCustomMiddleware.md
+++ b/docs/usage/WritingCustomMiddleware.md
@@ -97,7 +97,7 @@ More subtly, some middlewares expect that the middleware is called on the same t
 
 ### Returning the dispatch return value
 
-Unless the middleware needs to explicitly modify the return value of `dispatch`, just return what you get from `next`. It you do need to modify the return value, then your middleware anyway will need to sit in a very specific spot in the middleware chain to be able to do what it is supposed to, so you will need to check compatability with all other middlewares manually and decide how they could work together.
+Unless the middleware needs to explicitly modify the return value of `dispatch`, just return what you get from `next`. If you do need to modify the return value, then your middleware will need to sit in a very specific spot in the middleware chain to be able to do what it is supposed to - you will need to check compatibility with all other middlewares manually and decide how they could work together.
 
 This has a tricky consequence:
 
--- a/docs/usage/WritingCustomMiddleware.md
+++ b/docs/usage/WritingCustomMiddleware.md
@@ -15,12 +15,15 @@ title: Writing Custom Middleware
 
 Middleware in Redux can be mainly used to either
 
-- create side effects for actions, or to
-- modify or cancel actions.
+- create side effects for actions,
+- modify or cancel actions, or to
+- modify the input accepted by dispatch.
 
 Most use cases fall into the first category: For example [Redux-Saga](https://github.com/redux-saga/redux-saga/), [redux-observable](https://github.com/redux-observable/redux-observable), and [RTK listener middleware](https://redux-toolkit.js.org/api/createListenerMiddleware) all create side effects that react to actions. These examples also show that this is a very common need: To be able to react to an action other than with a state change.
 
-The most obvious example for modifying actions is [Redux Thunk](https://github.com/reduxjs/redux-thunk), which transforms a function returning an action into an action by calling it. But it can also be used to e.g. enhance an action with information from the state or from an external input.
+Modifying actions can be used to e.g. enhance an action with information from the state or from an external input, or to throttle, debounce or gate them.
+
+The most obvious example for modifying the input of dispatch is [Redux Thunk](https://github.com/reduxjs/redux-thunk), which transforms a function returning an action into an action by calling it.
 
 ## When to use custom middleware
 
--- a/docs/usage/WritingCustomMiddleware.md
+++ b/docs/usage/WritingCustomMiddleware.md
@@ -93,7 +93,7 @@ There are two contact points between our middleware and the other middlewares:
 
 When you call `next`, the middleware will expect some form of action. Unless you want to explicitly modify it, just pass through the action that you received.
 
-More subtly, some middlewares except that the middleware is called on the same tick as `dispatch` is called, so next should be called synchronously by your middleware.
+More subtly, some middlewares expect that the middleware is called on the same tick as `dispatch` is called, so `next` should be called synchronously by your middleware.
 
 ### Returning the dispatch return value
 
--- a/docs/usage/WritingCustomMiddleware.md
+++ b/docs/usage/WritingCustomMiddleware.md
@@ -1,0 +1,146 @@
+---
+id: writing-custom-middleware
+title: Writing Custom Middleware
+---
+
+# Writing Custom Middleware
+
+:::tip What You'll Learn
+
+- When to use custom middleware
+- Standard patterns for middleware
+- How to make sure that your middleware is compatible with other Redux projects
+
+:::
+
+Middleware in Redux can be mainly used to either
+
+- create side effects for actions, or to
+- modify or cancel actions.
+
+Most use cases fall into the first category: For example [Redux-Saga](https://github.com/redux-saga/redux-saga/), [redux-observable](https://github.com/redux-observable/redux-observable), and [RTK listener middleware](https://redux-toolkit.js.org/api/createListenerMiddleware) all create side effects that react to actions. These examples also show that this is a very common need: To be able to react to an action other than with a state change.
+
+The most obvious example for modifying actions is [Redux Thunk](https://github.com/reduxjs/redux-thunk), which transforms a function returning an action into an action by calling it. But it can also be used to e.g. enhance an action with information from the state or from an external input.
+
+## When to use custom middleware
+
+Most of the time, you won't actually need custom middleware. The most likely use case for middleware is side effects, and there is plenty of packages who nicely package side effects for Redux and have been in use long enough to get rid of the subtle problems you would run into when building this yourself. A good starting point is [RTK Query](https://redux-toolkit.js.org/rtk-query/overview) for managing server-side state and [RTK listener middleware](https://redux-toolkit.js.org/api/createListenerMiddleware) for other side effects.
+
+You might still want to use custom middleware in one of two cases:
+
+1. If you only have a single, very simple side effect, it might not be worth it to add a full additional framework. Just make sure that you switch to an existing framework once your application grows instead of growing your own custom solution.
+2. If you need to modify or cancel actions.
+
+## Standard patterns for middleware
+
+Middleware roughly works like this:
+
+```ts
+const middleware: Middleware = api => next => action => {
+  // Do something before the action hits the reducer and the next middlewares
+  const beforeState = api.getState()
+  beforeSideEffect(action, beforeState)
+  // Do something with the action before passing it along
+  const enhancedAction = enhance(action, beforeState, api)
+  let response: unknown
+  // Pass the action along, or cancel it by not passing it
+  if (!shouldCancelAction(enhancedAction, beforeState)) {
+    response = next(enhancedAction)
+  }
+  // Do something after the action hits the reducer
+  const afterState = api.getState()
+  afterSideEffect(action, afterState, api)
+
+  // Do something with the dispatch function return value
+  return enhanceDispatch(response, api)
+}
+```
+
+The different points to hook in (as marked by the comments) combined with the main usage for side-effects leads to two main patterns that are used in many different libraries.
+
+### Early side effect
+
+```ts
+const middleware: Middleware = api => next => action => {
+  // Do something before the action hits the reducer and the next middlewares
+  const beforeState = api.getState()
+  beforeSideEffect(action, beforeState, api)
+  return next(action)
+}
+```
+
+### Late side effect
+
+```ts
+const middleware: Middleware = api => next => action => {
+  const response = next(action)
+
+  // Do something after the action hits the reducer
+  const afterState = api.getState()
+  afterSideEffect(action, afterState, api)
+
+  return response
+}
+```
+
+## Rules to make compatible middleware
+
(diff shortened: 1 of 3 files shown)


docs/usage/WritingCustomMiddleware.md
```diff
@@ -0,0 +1,146 @@
...
+:::
+
+Middleware in Redux can be mainly used to either
+
+- create side effects for actions, or to
+- modify or cancel actions.
+
+Most use cases fall into the first category: For example [Redux-Saga](https://github.com/redux-saga/redux-saga/), [redux-observable](https://github.com/redux-observable/redux-observable), and [RTK listener middleware](https://redux-toolkit.js.org/api/createListenerMiddleware) all create side effects that react to actions. These examples also show that this is a very common need: To be able to react to an action other than with a state change.
+
+The most obvious example for modifying actions is [Redux Thunk](https://github.com/reduxjs/redux-thunk), which transforms a function returning an action into an action by calling it. But it can also be used to e.g. enhance an action with information from the state or from an external input.
```
markerikson, line 23: :hand: I'd add a third category - "teaching `dispatch` how to understand values that are _not_ plain action objects", and put thunks into that.  "Modifying an action" to me would be something like debouncing / throttling / gating an action, or literally mutating it.

docs/usage/WritingCustomMiddleware.md
```diff
@@ -0,0 +1,146 @@
...
+  // Pass the action along, or cancel it by not passing it
+  if (!shouldCancelAction(enhancedAction, beforeState)) {
+    response = next(enhancedAction)
+  }
+  // Do something after the action hits the reducer
+  const afterState = api.getState()
+  afterSideEffect(action, afterState, api)
+
+  // Do something with the dispatch function return value
+  return enhanceDispatch(response, api)
```
markerikson, line 55: :question: not sure I understand what `enhanceDispatch` is here, or where it's coming from

docs/usage/WritingCustomMiddleware.md
```diff
@@ -0,0 +1,146 @@
...
+
+In principle, middleware is a very powerful pattern and can do whatever it wants with an action. Existing middleware might have assumptions about what happens in the middleware around it, though, and being aware of these assumptions will make it easier to ensure that your middleware works well with existing commonly used middleware.
+
+There are two contact points between our middleware and the other middlewares:
+
+### Calling the next middleware
+
+When you call `next`, the middleware will expect some form of action. Unless you want to explicitly modify it, just pass through the action that you received.
+
+More subtly, some middlewares except that the middleware is called on the same tick as `dispatch` is called, so next should be called synchronously by your middleware.
```
markerikson, line 96:
```suggestion
More subtly, some middlewares expect that the middleware is called on the same tick as `dispatch` is called, so `next` should be called synchronously by your middleware.
```

docs/usage/WritingCustomMiddleware.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Returning the dispatch return value
+
+Unless the middleware needs to explicitly modify the return value of `dispatch`, just return what you get from `next`. It you do need to modify the return value, then your middleware anyway will need to sit in a very specific spot in the middleware chain to be able to do what it is supposed to, so you will need to check compatability with all other middlewares manually and decide how they could work together.
```
markerikson, line 100:
```suggestion
Unless the middleware needs to explicitly modify the return value of `dispatch`, just return what you get from `next`. If you do need to modify the return value, then your middleware will need to sit in a very specific spot in the middleware chain to be able to do what it is supposed to - you will need to check compatibility with all other middlewares manually and decide how they could work together.
```

docs/usage/WritingCustomMiddleware.md
```diff
@@ -0,0 +1,146 @@
...
+  if (action.type === 'REQUEST_FETCH') {
+    const data = await fetchData()
+    api.dispatch(dataFetchedAction(data))
+  }
+
+  return response
+}
+```
+
+Even though it looks like we didn't modify the response, we actually did: Due to async-await, it is now a promise. This will break some middlewares like e.g. the one from RTK query.
```
markerikson, line 119:
```suggestion
Even though it looks like we didn't modify the response, we actually did: Due to async-await, it is now a promise. This will break some middlewares like the one from RTK Query.
```

docs/usage/WritingCustomMiddleware.md
```diff
@@ -0,0 +1,146 @@
...
+
+So how to write this middleware?
```
markerikson, line 121:
```suggestion
So, how can we write this middleware instead?
```

docs/usage/WritingCustomMiddleware.md
```diff
@@ -0,0 +1,146 @@
...
+
+```ts
+const middleware: Middleware = api => next => action => {
+  const response = next(action)
+
+  // Do something after the action hits the reducer
+  const afterState = api.getState()
+  if (action.type === 'REQUEST_FETCH') {
+    void loadData(api)
```
markerikson, line 130: :hand: Don't think the `void` is needed here

docs/usage/WritingCustomMiddleware.md
```diff
@@ -0,0 +1,146 @@
...
+
+So how to write this middleware?
+
+```ts
+const middleware: Middleware = api => next => action => {
+  const response = next(action)
+
+  // Do something after the action hits the reducer
+  const afterState = api.getState()
+  if (action.type === 'REQUEST_FETCH') {
```
markerikson, line 129: :hand: I'd suggest using RTK-style `"some/action"` constants throughout the page for consistency

docs/usage/WritingCustomMiddleware.md
```diff
@@ -0,0 +1,146 @@
...
+  return response
+}
+
+async function loadData(api) {
+  const data = await fetchData()
+  api.dispatch(dataFetchedAction(data))
+}
+```
+
+Just move out the async logic into a separate function so that you can still use async-await, but don't actually wait for the promise to resolve in the middleware. `void` indicates to others reading the code that you decided to not await the promise explicitly without having an effect on the code.
```
markerikson, line 142: :speech_balloon: hmm. tbh I've never seen that `void` pattern used before 🤷‍♂️

docs/usage/WritingCustomMiddleware.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Next Steps
+
+If you haven't yet, take a look at [the Middleware section in Understanding Redux](../understanding/history-and-design/middleware.md) to understand how middleware works below the hood.
```
markerikson, line 146:
```suggestion
If you haven't yet, take a look at [the Middleware section in Understanding Redux](../understanding/history-and-design/middleware.md) to understand how middleware works under the hood.
```

markerikson (changes requested):
Great work!  Left some requests for tweaks

docs/usage/WritingCustomMiddleware.md
```diff
@@ -0,0 +1,146 @@
...
+- When to use custom middleware
+- Standard patterns for middleware
+- How to make sure that your middleware is compatible with other Redux projects
+
+:::
+
+Middleware in Redux can be mainly used to either
+
+- create side effects for actions, or to
+- modify or cancel actions.
```
dbartholomae, line 19:
```suggestion
- create side effects for actions,
- modify or cancel actions, or to
- modify the input accepted by dispatch.
```

docs/usage/WritingCustomMiddleware.md
dbartholomae, line 23:
```suggestion
Modifying actions can be used to e.g. enhance an action with information from the state or from an external input, or to throttle, debounce or gate them.

The most obvious example for modifying the input of dispatch is [Redux Thunk](https://github.com/reduxjs/redux-thunk), which transforms a function returning an action into an action by calling it.
```

docs/usage/WritingCustomMiddleware.md
dbartholomae:
```suggestion
  return enhanceDispatchReturnValue(response, api)
```

Maybe this works better? Basically you can modify the return value of a dispatch function.

docs/usage/WritingCustomMiddleware.md
dbartholomae: It's recommended e.g. by [this eslint rule](https://typescript-eslint.io/rules/no-floating-promises/) that is part of eslint's configuration set [`recommended-requiring-type-checking`](https://typescript-eslint.io/linting/configs/#recommended-requiring-type-checking). I've also seen it in other codebases as a nice convention.

docs/usage/WritingCustomMiddleware.md
dbartholomae: See discussion below. I'll resolve this one to avoid having to look in two places.

docs/usage/WritingCustomMiddleware.md
```diff
@@ -0,0 +1,146 @@
...
+
+So, how can we write this middleware instead?
+
+```ts
+const middleware: Middleware = api => next => action => {
+  const response = next(action)
+
+  // Do something after the action hits the reducer
+  const afterState = api.getState()
+  if (action.type === 'REQUEST_FETCH') {
```
dbartholomae, line 129:
```suggestion
  if (action.type === 'some/action') {
```

docs/usage/WritingCustomMiddleware.md
```diff
@@ -0,0 +1,146 @@
...
+
+This has a tricky consequence:
+
+```ts
+const middleware: Middleware = api => next => async action => {
+  const response = next(action)
+
+  // Do something after the action hits the reducer
+  const afterState = api.getState()
+  if (action.type === 'REQUEST_FETCH') {
```
dbartholomae, line 110:
```suggestion
  if (action.type === 'some/action') {
```

dbartholomae (commented):
I've fixed most of these. Open questions are
- how to name the pattern of a middleware modifying its return value
- whether to use the `void` convention for unawaited promises

dbartholomae:
@markerikson do you have feedback for the last two questions?

docs/usage/WritingCustomMiddleware.md
markerikson: I think there's still a couple things that bother me about this example.

One is that there's no indication what any of these functions do or where they're coming from. I think I understand what you're _trying_ to show - that you can run side effects before and after calling `next()`, etc.  I'm just not sure this example as written is the best way to explain that.

Second is that the `enhance` and `enhanceDispatch` functions in particular are very confusing even after re-reading this a few times.  I _think_ that `enhance` is trying to indicate that you can mutate the contents of the action.  I still don't know what `enhanceDispatchReturnValue` is supposed to do.

I think I'd much rather have this whole example explicitly show "real" examples of each kind of a thing, with comments right above each line explaining "here's an example of a side effect before calling `next(action)`", etc.  In other words, instead of calling a made-up function called `enhance(action)`, I'd rather see something like `action.meta.someField = 123`.  Similar for the side effect checks.

I don't immediately have good suggestions for what those examples ought to be.

markerikson (changes requested):

markerikson:
Per the "naming" question, I think I'd just refer to it as "overriding" or "replacing" the return value.

docs/usage/WritingCustomMiddleware.md
dbartholomae: This example tries to cover all the points where you can "hook" into the lifecycle. E.g. `enhanceDispatchReturnValue` is used by redux-thunk to make dispatch awaitable. But most middlewares will only use one or maybe two of these points.
I like the idea of concrete examples. Maybe we could split it into the following:

1. `afterSideEffect` is very common, we could e.g. use RTK listener middleware for this.
2. `beforeSideEffect` is very similar to `afterSideEffect`, but I haven't seen this in use anywhere yet, as it doesn't allow access to the state after the action went through the reducer, so I would need help to find an example
3. `enhance` and `enhanceDispatch` can both be shown with Redux Thunk middleware
4. For stopping execution, I don't have a good example. Maybe something from RTK query? I don't have deep enough understanding of that code base for a good example, though.

markerikson: I'm okay with miniature examples, I'm just thinking it would be helpful if they showed actually _doing_ something instead of calling functions that don't actually exist in the code block :)

dbartholomae: I'll add a separate comment with a suggestion.

dbartholomae:
@markerikson I've rewritten the [standard examples](https://github.com/reduxjs/redux/pull/4498/files#diff-7b29e8f8d69967db16bc79197d05294f8761883dd5303e7ab50d34bfe1af9522R37-R140) with concrete examples from rtk.

dbartholomae:
@markerikson any feedback?

markerikson:
Okay, let's go with this. Thanks!

PR #4502, opened 2023-03-11
Update App.test.tsx (better to use screen.getByText rather than from …

mod7ex:
better to use screen.getByText rather than distruct the return of render


EskiMojo14:
out of curiosity, why is it better?

timdorr:
Thanks, but this is just a preference thing. Using the return of `render` is arguably _slightly_ better because it works in environments without a global DOM.

PR #4517, opened 2023-04-09
Update workflow badge

sh-cho:
- ref: https://github.com/badges/shields/issues/8671

shields.io badge needs to be updated

image
(left(as-is), right(to-be))


timdorr:
Thanks!

Issue #4541, opened 2023-05-04
[Docs] Exported variable store has or is using name CounterState from external module "/somepath" but cannot be named.


Nightvision53:
## What docs page needs to be fixed?

- **tutorials**:
- **typescript-quick-start**:

## What is the problem?

The problem is when I try to implement redux with TypeScript starter I always get this error on store type:
```
Exported variable store has or is using name _AccountState_ from external module "_mypath_" but cannot be named.
```

Here AccountState is the initial state interface.

## What should be changed to fix the problem?

Adding export to slices Initial State interface is fixing the issue.

PR #4542, opened 2023-05-04
add export to typescript tutorial

Nightvision53:
---
---

## Checklist

- [ x ] Is there an existing issue for this PR?
  - #4541 
- [ x ] Have the files been linted and formatted?

- **Section**: tutorials
- **Page**: typescript-quick-start

Add `export` to typescript quick start so the store object does not give same error.

### DIFF
--- a/docs/tutorials/typescript.md
+++ b/docs/tutorials/typescript.md
@@ -96,7 +96,7 @@ import type { RootState } from '../../app/store'
 
 // highlight-start
 // Define a type for the slice state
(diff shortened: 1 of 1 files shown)


timdorr:
This isn't needed unless you are explicitly referencing that interface/type elsewhere. The code still works in the embedded sandbox if I remove the export. 

Nonetheless, it's not hurting anything by having it exported, so I suppose we can add this in. Thanks!

PR #4555, opened 2023-06-07
Fix example in why-rtk-is-redux-today.md

someden:
---
---

- **Page**: https://redux.js.org/introduction/why-rtk-is-redux-today

According to the usage [action.payload.id](https://github.com/reduxjs/redux/blob/4a6d2fb227ba119d3498a43fab8f53fe008be64c/docs/introduction/why-rtk-is-redux-today.md?plain=1#L94) payload should contain object.


timdorr:
Thanks!

PR #4560, opened 2023-06-11
Update example code in why-rtk-is-redux-today.md

aliansyahFirdaus:
Hi!

I hope you're doing well. I've created this pull request to address an issue in the Redux Toolkit documentation. The current code example contains an error, and I've made the necessary correction to ensure accuracy and clarity.

I've thoroughly tested the modified code to ensure its correctness. Your review and feedback on this pull request would be greatly appreciated. Your attention to detail will help improve the quality of the Redux Toolkit documentation.

Thank you for your time and consideration. I look forward to working together to improve the Redux Toolkit documentation for the community.

### DIFF
--- a/docs/introduction/why-rtk-is-redux-today.md
+++ b/docs/introduction/why-rtk-is-redux-today.md
@@ -138,6 +138,8 @@ const todosSlice = createSlice({
     todoToggled(state, action) {
       const todo = state.find(todo => todo.id === action.payload)
       todo.completed = !todo.completed
+      
+      state = todo
     }
   }
 })


EskiMojo14:
Hi, thanks for the PR!

Unfortunately, I think you're misunderstanding a key part of how Immer works (which is used for RTK's reducers).

As covered in [Writing Reducers with Immer](https://redux-toolkit.js.org/usage/immer-reducers#resetting-and-replacing-state), `state =` will *never* replace state. All you're doing is reassigning the local variable inside your reducer, and losing access to the Immer draft you're allowed to mutate.

If you want to completely replace state, you should return a value instead.

With that said - `state = todo` doesn't seem to be what you want in this case anyway? `state` is an array of Todos, and `todo` is just a single Todo. replacing an array with a single item seems like a mistake.

PR #4573, opened 2023-06-23
Update UsageWithTypescript.md

cyogian:
UnknownAction is no more a type offered by Redux Library, so it should be AnyAction ?
[:memo: Documentation Fix](?template=documentation-edit.md)

### DIFF
--- a/docs/usage/UsageWithTypescript.md
+++ b/docs/usage/UsageWithTypescript.md
@@ -207,10 +207,10 @@ You could add this to your ESLint config as an example:
 
 [Reducers](../tutorials/fundamentals/part-3-state-actions-reducers.md) are pure functions that receive the current `state` and incoming `action` as arguments, and return a new state.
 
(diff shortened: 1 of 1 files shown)


EskiMojo14:
UnknownAction is introduced in the v5 alphas/beta - I'm not sure how we want to handle docs for that as v5 things are in master 🤔

cyogian:
> UnknownAction is introduced in the v5 alphas/beta - I'm not sure how we want to handle docs for that as v5 things are in master 🤔

Shouldn't the primary Documentation be about Stable version instead of versions still under alpha/beta testing.
As when we install redux@latest we get v4 right now.

EskiMojo14:
i agree, that makes the most sense - i'm just unsure how best to handle that from a source control perspective as we're using master as the working branch for v5, but the current docs are also deployed from there

timdorr:
We're pushing towards a 5.0 release (finally, OMG!), so this will be taken care of Soon™

PR #4578, opened 2023-07-02
Minor RTK Mutation Query Function Documentation Fix

weixin-tan:
Fix mutation function usage, according to part 7 an `.unwrap()` call is needed.

---
---

## Checklist

  - https://github.com/reduxjs/redux/issues/4577

- **Page**: Part 8 RTK Query Advanced Patterns

## What is the problem?
Seems like a minor typo/missing function call in the usage of mutation function. In part 7, it is stated that the `.unwrap()` function needs to be called for the RTK mutation query.

> As with the thunk dispatch, we call addNewPost with the initial post object. This returns a special Promise with a .unwrap() method, and we can await addNewPost().unwrap() to handle any potential errors with a standard try/catch block.

Whereas in part 8 code snippet for Editing Post, the `.unwrap()` call is missing for a very similar mutation function.
```
  const [updatePost, { isLoading }] = useEditPostMutation()

  /* code omitted */

  const onSavePostClicked = async () => {
    if (title && content) {
      await updatePost({ id: postId, title, content }) // <- this should have a .unwrap() ?
      history.push(`/posts/${postId}`)
    }
  }
```

Add `.unwrap()` to the code snippet

### DIFF
--- a/docs/tutorials/essentials/part-8-rtk-query-advanced.md
+++ b/docs/tutorials/essentials/part-8-rtk-query-advanced.md
@@ -115,7 +115,7 @@ export const EditPostForm = ({ match }) => {
   const onSavePostClicked = async () => {
     if (title && content) {
       // highlight-next-line
-      await updatePost({ id: postId, title, content })
+      await updatePost({ id: postId, title, content }).unwrap()
       history.push(`/posts/${postId}`)
     }
   }


markerikson:
Strictly speaking, the `.unwrap()` isn't needed here. Dispatching the mutation trigger will return a promise either way - it's just that it always resolves.  `.unwrap()` is only needed if you want to convert that back into a potential thrown error:

- https://redux-toolkit.js.org/api/createAsyncThunk#unwrapping-result-actions

PR #4582, opened 2023-07-24
Unify parameter/variable names to plural forms in applyMiddleware.md

codersjj:
Also keep consistency with [the existing definition in the source code](https://github.com/reduxjs/redux/blob/master/src/applyMiddleware.ts#L54).


timdorr:
Unfortunately, the function name itself is inconsistent and can't be changed. But "ware" is a word that is ambiguous as far as count goes, so both "ware" and "wares" are a valid plural form. Consistency is cool, but isn't needed in this case.

PR #4588, opened 2023-08-08
chore: remove useless parameters

li-jia-nan:
chore: remove useless parameters


timdorr:
Thanks, but this doesn't really matter one way or another.

Issue #4538, opened 2023-05-02
[Docs] Add/Update better dark theme palette for "DetailedExplanation.jsx"


cweiqi27:
The detailed explanation component on dark theme is too bright and looks out of place compare to the rest of the theme.

![image](https://user-images.githubusercontent.com/97137887/235633929-c61297a0-da24-4e53-af83-626bd51e3103.png)

PR #4589, opened 2023-08-14
Add color property to 'code' css declaration

BarokDG:
---
---

## Checklist

  - #4538 

- The inline code blocks used inside DetailedExplanation.tsx

## What is the problem?

It's difficult to see the text in inline code blocks inside the DetailedExplanation component in dark mode. The css doesn't match the styles used for inline code blocks on the rest of the documentation.

I added a css 'color' property to the code declaration inside website\src\css\custom.css. The value is consistent with what's used outside the DetailedExplanation component.

### Before
image

### After
image


markerikson:
Thanks!

PR #4593, opened 2023-08-28
Update vanilla example

stuymedova:
It is, to my taste, quite a bit cleaner to have all constants and operations be on top. What do you think?


timdorr:
Thanks, but this doesn't really increase readability. In fact, it makes it slightly harder because there are additional variables to track here and you have to go back and forth through the code to read it.

PR #4594, opened 2023-08-28
Update vanilla example

stuymedova:
This is an extension of my previous PR (#4593). What I was suggesting was not about increasing the number of variables, it was about reordering some lines of code, so the important ones, the ones you'd need to read first are on top. The variables was just something else I did in addition to the idea, in addition to using default parameters, for instance. I don't mind if this gets cut off as well. I believe this idea was misunderstood and deserves a fair second chance. And if the contributors think that's not the case, I'd like to learn why, perhaps I can learn something from this.

It always feels nice to contribute to a big project, even in a small way.

### DIFF
--- a/examples/counter-vanilla/index.html
+++ b/examples/counter-vanilla/index.html
@@ -16,11 +16,6 @@
     </div>
     <script>
       const valueEl = document.getElementById('value')
-      const incrementButtonEl = document.getElementById('increment')
-      const decrementButtonEl = document.getElementById('decrement')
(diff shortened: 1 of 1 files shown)


timdorr:
This still isn't an improvement in readability. You declare functions _after_ their first use, relying on hoisting to make them available. That will likely confuse new users by having things out of order and relying on features of the language they may not know about. 

I appreciate what you're trying to do here, and while this example may not be as perfect as it can be, it's pretty close and probably best left alone.

PR #4600, opened 2023-09-13
Use default Node version in CI

nickmccurdy:
Deprecated Node versions are used in CI, potentially causing security and reliability issues. Instead, it's better to use GitHub's default Node version, which also doesn't require additional downloads or installations.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -33,7 +33,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Use node
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
           cache: 'yarn'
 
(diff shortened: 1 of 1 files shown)


nickmccurdy:
The size action is warning because of https://github.com/preactjs/compressed-size-action/issues/93 but it seems like it's safe to ignore as it already runs on Node 16:

> The following actions uses node12 which is deprecated and will be forced to run on node16: preactjs/compressed-size-action@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

nickmccurdy:
Blocked by reduxjs/redux-toolkit#3715

timdorr:
This should be superseded by #4680

PR #4601, opened 2023-09-19
Adding a missing dot

Arthur-Milchior:
It seems your list elements ends with dots. This ensure consistency.

---
---

## Checklist

No

I use the tool provided by "Edit this page" at the bottom of https://redux.js.org/tutorials/essentials/part-2-app-structure , so I just used the tool you provide. I do not know whether any extra lint is done. At least, I didn't apply it.
Given the commits only add a character, I doubt it is an issue here.

- **Section**: Rules of Reducers
- **Page**: https://redux.js.org/tutorials/essentials/part-2-app-structure

Do you really need this ? It seems the PR content would be enough information.

## What is the problem?

Not consistent with punctuation in list

Add the missing dot

### DIFF
--- a/docs/tutorials/essentials/part-2-app-structure.md
+++ b/docs/tutorials/essentials/part-2-app-structure.md
@@ -280,7 +280,7 @@ console.log(newState)
 
 We said earlier that reducers must **always** follow some special rules:
 
-- They should only calculate the new state value based on the `state` and `action` arguments
+- They should only calculate the new state value based on the `state` and `action` arguments.
 - They are not allowed to modify the existing `state`. Instead, they must make _immutable updates_, by copying the existing `state` and making changes to the copied values.
 - They must not do any asynchronous logic or other "side effects"
 


docs/tutorials/essentials/part-2-app-structure.md
```diff
@@ -280,7 +280,7 @@ console.log(newState)
 
 We said earlier that reducers must **always** follow some special rules:
 
-- They should only calculate the new state value based on the `state` and `action` arguments
+- They should only calculate the new state value based on the `state` and `action` arguments.
 - They are not allowed to modify the existing `state`. Instead, they must make _immutable updates_, by copying the existing `state` and making changes to the copied values.
 - They must not do any asynchronous logic or other "side effects"
```
EskiMojo14, line 285:
```suggestion
- They must not do any asynchronous logic or other "side effects".
```

EskiMojo14: actually, it seems quite a few of the other list items also don't end in a full stop 😕

timdorr:
Yeah, this is quite yak shave-y. We're not consistent elsewhere, but it really doesn't matter. Sorry, but this is fine as-is.

PR #4606, opened 2023-10-20
Update ConfiguringYourStore.md

koechkevin:
Prefer to use the callback notation for the middleware option in configureStore to access a pre-typed getDefaultMiddleware instead.

---
---

## Checklist

  - _link issue here_

- **Section**:  https://redux.js.org/usage/configuring-your-store#simplifying-setup-with-redux-toolkit
- **Page**: https://redux.js.org/usage/configuring-your-store

## What is the problem?
`getDefaultMiddleware` import from `reduxjs/toolkit` has been marked as deprecated. The documentation however still uses  import.

 This PR removes the imported `getDefaultMiddleware` and uses the callback notation for the middleware option.

### DIFF
--- a/docs/usage/ConfiguringYourStore.md
+++ b/docs/usage/ConfiguringYourStore.md
@@ -356,7 +356,7 @@ import rootReducer from './reducers'
 export default function configureAppStore(preloadedState) {
   const store = configureStore({
     reducer: rootReducer,
-    middleware: (getDefaultMiddleware) => [loggerMiddleware, ...getDefaultMiddleware()],
+    middleware: (getDefaultMiddleware) => [loggerMiddleware, ...getDefaultMiddleware().prepend()],
     preloadedState,
     enhancers: [monitorReducersEnhancer]
   })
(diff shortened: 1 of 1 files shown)


markerikson:
Thanks! Could you make one more tweak? Use the specialized `gDM().prepend()` method.  It doesn't do much with JS, but with TS it correctly preserves the types of the middleware.  (I know the snippet is JS and not TS, but if we're going to tweak it, might as well show the method we really want folks to use.)

docs/usage/ConfiguringYourStore.md
```diff
@@ -356,7 +356,7 @@ import rootReducer from './reducers'
 export default function configureAppStore(preloadedState) {
   const store = configureStore({
     reducer: rootReducer,
-    middleware: [loggerMiddleware, ...getDefaultMiddleware()],
+    middleware: (getDefaultMiddleware) => [loggerMiddleware, ...getDefaultMiddleware().prepend()],
```
EskiMojo14, line 359:
```suggestion
    middleware: (getDefaultMiddleware) => getDefaultMiddleware().prepend(loggerMiddleware),
```

EskiMojo14:
thanks!

PR #4608, opened 2023-10-28
add links to egghead RTK Query tutorial

phryneas:
@markerikson also here - can you take a look if this looks right for you?


phryneas:
Disclaimer: This PR also adds my affiliate code to all the other egghead courses, since there was no such code on them (so it doesn't take away anything from anyone) and we might at least make something off the traffic we give to egghead, right? :)

PR #4612, opened 2023-11-02
Update part-1-overview-concepts.md

BryanMyersss:
It might not seem like much, but I've spent 40 minutes trying to figure out how exactly was immutability being enforced here, and overall got confused. This will surely help newbies like me.


timdorr:
Thanks, but I think the existing comments are sufficient. If someone doesn't understand the spread operator's function, that should be learned elsewhere. These comments are not intended to teach the purpose of that operator (nor are they sufficient for that task)

PR #4627, opened 2023-12-01
Don't notify subscribers when after action state hasn't changed, add test

gentlee:
---
name: Optimization of subscriptions
about: Performance
---

No.

**Problem**

Up to 50% of all dispatched actions in the apps that use `redux-saga` or `redux-thunk` could be ones that don't change the redux state but start some async logic.

Current implementation notifies subscribers for each such action even if state hasn't changed, causing all active selectors to be reevaluated without any good reason.

In huge apps there can be hundreds of subscribed selectors and dozens of actions per second, so this leads to additional CPU usage for no good reason.

**Solution**

Don't notify subscribers when action did not change the state.

**Can it break existing code?**

As far as I can see, the only reason to subscribe to the store is to get the new state using store.getState(). There is no sense in subscribing to get the called action because there is no way to get the latest action called from subscription. So it should not break any existing code.

But I would vote for raising major version here.

## Checklist

 No, as far as I know.

### DIFF
--- a/src/createStore.ts
+++ b/src/createStore.ts
@@ -294,6 +294,7 @@ export function createStore<
       throw new Error('Reducers may not dispatch actions.')
     }
 
+    const previousState = currentState
     try {
       isDispatching = true
       currentState = currentReducer(currentState, action)
@@ -302,9 +303,12 @@ export function createStore<
     }
 
     const listeners = (currentListeners = nextListeners)
-    listeners.forEach(listener => {
-      listener()
-    })
+    if (previousState !== currentState) {
+      listeners.forEach(listener => {
+        listener()
+      })
+    }
+
     return action
   }
 
--- a/test/createStore.spec.ts
+++ b/test/createStore.spec.ts
@@ -207,19 +207,28 @@ describe('createStore', () => {
     ])
   })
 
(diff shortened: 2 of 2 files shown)


markerikson:
Hi! I recognize your username from some recent discussions in the React-Redux repo.

This is actually an intentional design decision, and _not_ something we will be changing:

- https://redux.js.org/faq/design-decisions#why-doesnt-redux-pass-the-state-and-action-to-subscribers

However, you can implement this same logic as a Redux store enhancer, and add it to your own store in your app.

gentlee:
As for the layer where to fix this issue, it makes sense to me that it is better to remove the root cause of a problem which is here. I don't understand what lies beyond that design decision to notify subscribers when state hasn't changed considering that we don't pass state and action to the subscriber.

markerikson:
Hmm. It _used_ to bail out. Maybe we lost something in the transition to `useSyncExternalStore` in v8.

I'm not questioning whether the change here makes a difference. I'm saying it's a piece of core behavior we're not going to alter, and that it _can_ be overridden at the enhancer level.

markerikson:
Filed https://github.com/reduxjs/react-redux/issues/2089 and I'll try to look at it soon.

PR #4633, opened 2023-12-05
fix(isPlainObject):  support check Object.create(null)

zhe-he:
https://github.com/reduxjs/redux/issues/4632


timdorr:
Added a test for this as well. Thanks!

PR #4638, opened 2023-12-16
fix(types/store): Unexpectedly narrowed return type of function `Store['getState']`

exuanbo:
---
name: :bug: Fix the return type of function `Store['getState']`
about: Unexpectedly narrowed type
---

This PR fixes a bug in the type definitions.

#### What is the current behavior, and the steps to reproduce the issue?

The return type of function `Store['getState']` is unexpectedly narrowed to non-nullable.

```ts
import { type Reducer, createStore } from 'redux'

type NonNullableState = { count: number }
const nullableReducer: Reducer<NonNullableState | null> = () => null

const storeWithNullableState = createStore(nullableReducer)
const nullableState = storeWithNullableState.getState()
//    ^? const nullableState: NonNullableState
```

TS Playground: <https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzjAnmApnASugJgVwGN0oAaOQqdAQxnQGUZpMBfOAMyghDgHIqCAD14AoEagxwAchAB2U-ABtF1AEaKGMWpgC8iChHyyYALjiz8IVSTgsRhOQGd4F5Wo04CxKGc9ESADwy8koq6pracAA+5qEAfHB6ABQAlIkJropiDrLOcM7MAOrAMAAWCm7hjJF6lDR0jMxJme7oft4p9k4uoa3VdIn5TFTFZRVhGv3oAHQA5ugwU6kiAPQrcBtwAHoA-GJAA>

#### What is the expected behavior?

If `StateExt` type parameter is not provided, `Store['getState']` should return the untouched type `S`.

#### How does this PR fix the problem?

The problem is in this line:

https://github.com/reduxjs/redux/blob/7876f8e99ac041a6084c0495c5f4fbc1553dda30/src/types/store.ts#L114-L119

If `StateExt` is not given a specific value, the type will be `S & {}` which narrows the type to non-nullable.

The behavior of intersection with `{}` is changed in TypeScript 4.8, ["Improved Intersection Reduction, Union Compatibility, and Narrowing"](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-8.html#improved-intersection-reduction-union-compatibility-and-narrowing).

In this PR:

- The default value of the type parameter `StateExt` in the `Store` interface is widened to `unknown`. This can ensure that `S & unknown` is still `S`.
- In other types for creating store, the type arguments `StateExt` passed to `Store` is widened using a utility type `UnknownIfNonSpecific`.

```ts
type UnknownIfNonSpecific<T> = {} extends T ? unknown : T
```

## Checklist

  - No.

### DIFF
--- a/src/createStore.ts
+++ b/src/createStore.ts
@@ -5,7 +5,8 @@ import {
   StoreEnhancer,
   Dispatch,
   Observer,
-  ListenerCallback
+  ListenerCallback,
+  UnknownIfNonSpecific
 } from './types/store'
 import { Action } from './types/actions'
 import { Reducer } from './types/reducers'
@@ -46,7 +47,7 @@ export function createStore<
 >(
   reducer: Reducer<S, A>,
   enhancer?: StoreEnhancer<Ext, StateExt>
-): Store<S, A, StateExt> & Ext
+): Store<S, A, UnknownIfNonSpecific<StateExt>> & Ext
 /**
  * @deprecated
  *
@@ -82,7 +83,7 @@ export function createStore<
   reducer: Reducer<S, A, PreloadedState>,
   preloadedState?: PreloadedState | undefined,
   enhancer?: StoreEnhancer<Ext, StateExt>
-): Store<S, A, StateExt> & Ext
+): Store<S, A, UnknownIfNonSpecific<StateExt>> & Ext
 export function createStore<
   S,
   A extends Action,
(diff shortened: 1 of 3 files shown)


EskiMojo14:
good catch, thanks! admittedly i don't think stores with a nullable state are very common, but you're correct that this is an issue.

any chance you could check whether the same issue needs fixing on the RTK side of things too?

EskiMojo14:
thanks!

PR #4660, opened 2024-01-10
fixed typo in nextjs docs

nsklyarov:
When creating an application via `npx create-next-app --example with-redux my-app` (the method described in the documentation), the application does not create the `src` directory. 

[Related Issue](https://github.com/reduxjs/redux/issues/4656#issue-2068940198)

### DIFF
--- a/docs/usage/nextjs.mdx
+++ b/docs/usage/nextjs.mdx
@@ -63,7 +63,7 @@ These recommendations are specific to applications written with the Next.js App 
 
 ### Folder Structure
 
-Next apps can be created to have the `/app` folder either at the root, or nested under `/src/app`. Your Redux logic should go in a separate folder, alongside the `/app` folder. It's common to put the Redux logic in a folder named `/lib`, but not required.
+Next apps can be created to have the `/app` folder either at the root, or nested under `/app`. Your Redux logic should go in a separate folder, alongside the `/app` folder. It's common to put the Redux logic in a folder named `/lib`, but not required.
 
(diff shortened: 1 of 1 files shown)


docs/usage/nextjs.mdx
```diff
@@ -63,7 +63,7 @@ These recommendations are specific to applications written with the Next.js App
 
 ### Folder Structure
 
-Next apps can be created to have the `/app` folder either at the root, or nested under `/src/app`. Your Redux logic should go in a separate folder, alongside the `/app` folder. It's common to put the Redux logic in a folder named `/lib`, but not required.
+Next apps can be created to have the `/app` folder either at the root, or nested under `/app`. Your Redux logic should go in a separate folder, alongside the `/app` folder. It's common to put the Redux logic in a folder named `/lib`, but not required.
```
EskiMojo14, line 66:
```suggestion
Next apps can be created to have the `/app` folder either at the root, or nested under `/src/app`. Your Redux logic should go in a separate folder, alongside the `/app` folder. It's common to put the Redux logic in a folder named `/lib`, but not required.
```
this isn't a typo, NextJS does support putting your app directory under a src directory: https://nextjs.org/docs/app/building-your-application/configuring/src-directory

The codeblock title makes sense to adjust as in most cases it won't be, but this sentence is correct as is.

markerikson: Yep, that part was intentional.

timdorr:
Thanks!

PR #4677, opened 2024-02-20
Update docs: Add s to come

Thalmann:
---
---

## Checklist

  - _link issue here_

- **Section**: https://redux.js.org/tutorials/essentials/part-5-async-logic#using-middleware-to-enable-async-logic
- **Page**: https://redux.js.org/tutorials/essentials/part-5-async-logic#using-middleware-to-enable-async-logic

## What is the problem?
Missing s

Adds an s


Thalmann:
@timdorr Fair enough. I was thinking "comes in" because we were referring to the middleware concept or system as a singular entity.
Not a big deal at all ofc.

PR #4679, opened 2024-02-21
Add `.gitattributes` file to make line endings more consistent

aryaemami59:
**This PR**:

  - [X] Adds `.gitattributes` file to make line endings more consistent.


timdorr:
I would advise against this, for the same reasons that we don't have a formatting check anymore. It becomes a big, annoying barrier to contributions by others who don't set up their tooling the right way before writing some code. 

Also, git's autocrlf stuff is godawful. This is basically punishing all Windows devs 😆

EskiMojo14:
@timdorr doesn't prettier already enforce  LF EOL anyway? I'm open to ditching the gitattributes file but the prettierrc we use seems to already have this same opinion

timdorr:
Not in `auto` mode. [It just uses whatever the first line of the file uses.](https://prettier.io/docs/en/options.html#end-of-line) 

Honestly, that's fine because nearly any editor outside of Notepad handles differing linebreak styles automatically. It also doesn't matter to end users because everything is run through our tooling, which normalizes everything anyways.

EskiMojo14:
prettierrc doesn't specify an option, and the default has been LF since Prettier 2

Methuselah96:
I develop entirely on Windows and afaik `* text=auto eol=lf` actually makes it easier for Windows newcomers, especially since Prettier expects LF line endings. It makes it more consistent by cloning using LF line endings.

timdorr:
> prettierrc doesn't specify an option, and the default has been LF since Prettier 2

Apologies, I read the PR change backwards. Yeah, we're on LF for folks using Prettier.

Nonetheless, I'm not a huge fan of "big bang", large reformat PRs. I'd just set the setting for changes going forward and leave the existing formatting alone.

EskiMojo14:
@timdorr my concern with that is that you then get spam in unrelated PRs as formatting gets updated 😕

aryaemami59:
@timdorr Would it be easier if I split this into multiple PRs?

timdorr:
Let's keep it simple and just make this PR the singular commit of 1d83caa826ecf14ee88e23715c0b74e8bbfbe1f0. No need to complicate things by also doing a formatting run. We can make that a separate PR so the noise is kept down.

Basically separate out the policy decision from the effects of that policy. Sort of equivalent to an RFC vs. implementation.

aryaemami59:
@timdorr Awesome, thanks Tim! I'm gonna do the same with the othe repos.

timdorr:
Thanks!

PR #4683, opened 2024-03-01
add reducer in nextJs doc

pikooli:
---
---

## Checklist

  - _link issue here_

- **Section**:  usage
- **Page**: nextjs

## What is the problem?
Didn't have a reducer in the store configuration

- Add the reducer counterReducer in the store configuration
- Use formatted script to format repo.


timdorr:
Thanks, but that's not really needed at that point. We're just showing some of the setup code and not getting into using specific reducers at that point. It's fine as-is.

PR #4685, opened 2024-03-18
Update part-3-data-flow.md

monchipi:
I would like to fix the sample code in the file "store.js" to fix a problem that is not working properly.

--- name: :memo: Documentation Fix about: fix existing documentation page problem ---.

## Checklist.

- [No] Is there an existing problem with this PR?

## Is there a document page that needs to be fixed?

- Section : part3

## What is the problem?
It does't work.

## What changes will this PR make to correct the problem?
Change the function to variable.

### DIFF
--- a/docs/tutorials/essentials/part-3-data-flow.md
+++ b/docs/tutorials/essentials/part-3-data-flow.md
@@ -139,7 +139,7 @@ import { configureStore } from '@reduxjs/toolkit'
 
 import postsReducer from '../features/posts/postsSlice'
 
-export default configureStore({
+export const store = configureStore({
   reducer: {
     posts: postsReducer
   }


phryneas:
That is currently right, your change would break the tutorial.

In [Part 2](https://redux.js.org/tutorials/essentials/part-2-app-structure), a default import is used to import the store:

```js
// this is a default import like in part 2
import store from './app/store'
// this would be a named import, but we don't show that anywhere for `store`
import { store } from './app/store'
```

That also means that here in part 3, a default export has to be used, or it wouldn't be working.

Did you see another kind of import for `store`  somewhere else in the tutorials?

monchipi:
Thank you for your review and comment. Regarding your question about the import for 'store' in the tutorials, I didn't see any other type of import. If we don't modify it, the code on line 6 in 'index.js' won't work correctly, which is why nothing appears on my screen.

Have you completed the Part 3 tutorial?

EskiMojo14:
are you sure that you didn't accidentally write a named import in your local project?

monchipi:
@EskiMojo14 
Thank you for your review and comment. I didn't write accidentally. I have cloned this project and modify two files named "features/posts/postsSlice.js" and "app/store.js" according to the article.

timdorr:
If you're following through, this is all covered in Part 2, specifically with the Provide The Store section: https://redux.js.org/tutorials/essentials/part-2-app-structure#providing-the-store

You can see those exports from `app/store.js` are all `default`, not named. You have copied something incorrectly if you need to switch to a named export at that point.

Regardless, this doc is correct as-is. Thanks, but this PR will break the tutorial.

PR #4687, opened 2024-03-21
Fix webpack 4 support by setting `target: es2017` for `redux.legacy-esm.js`

aryaemami59:
## This PR:

  - [X] Fixes webpack 4 support by setting `target: es2017` for `redux.legacy-esm.js` which is related to https://github.com/reduxjs/redux-toolkit/issues/4282.


timdorr:
Bonus: we also get a sourcemap. Thanks!

PR #4691, opened 2024-03-29
Add a note to ignore `@typescript-eslint/ban-types` rule when typing middlewares

renchap:
## PR Type

Update an existing page

## Checklist

  - Not an issue, but a discussion: https://github.com/reduxjs/redux-toolkit/discussions/4312

- **Section**: Using Redux
- **Page**: Usage with Typescript

This adds a warning to the docs because a very commonly used ESLint rule (`@typescript-eslint/ban-types`) will recommend switching the `{}` dispatch argument for `Middleware` to something else `(like `Record<string, never>`), which will then break the store's types. As per the discussion linked above, the rule should be ignored for this line and you *need* to use `{}` here.

This might add an information not useful to some readers, but this page is for Typescript users, and typing middlewares is not very common, so the audience for this part should already be targeted, and there is a high probability of them using `typescript-eslint`

### DIFF
--- a/docs/usage/UsageWithTypescript.md
+++ b/docs/usage/UsageWithTypescript.md
@@ -280,6 +280,12 @@ export const exampleMiddleware: Middleware<
 }
 ```
 
+:::caution
+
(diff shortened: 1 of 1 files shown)


EskiMojo14:
thanks!

Issue #4393, opened 2022-07-26
Rewrite "Redux Essentials" tutorial to be TS-first and show RTK 2.0 usage


markerikson:
The "Essentials" tutorial currently uses _only_ plain JS for all of the examples and content.

If you want to learn how to use Redux with TS, we have two separate pages:

- "TS Quick Start", which just gives quick examples of inferring types from the store, defining typed hooks, declaring slice state and action types, and using the hooks
- The longer "TS Usage" guide page, which is a laundry list of various use cases

Unfortunately neither of those have any larger context for how TS gets used in practice.

I _want_ the "Essentials" tutorial to cover TS usage.  However, some of the suggested approaches are impractical:

- If we rewrite the entire tutorial to be TS-first/only, then it adds an extra barrier to learning for anyone who doesn't know TS yet.  Since we do have a lot of folks still being pushed into Redux fairly early on, this is likely to be a significant number of people who would now have to decipher TS on top of Redux basics
- The next obvious approach is "use a TS/JS language toggle".  We already do this in the RTK docs for the examples, and we _did_ just add that same capability to the Redux core docs.  _However_... the tutorial is way more than just some example snippet blocks.  It's an entire app repo with a commit history.  In addition, there's a ton of explanations of "here's what this code is doing", and the point of this is to explain how to use TS for as much of the actual app code as possible. If we were to make all the code toggleable, then we'd somehow have to start toggling large chunks of explanation, too.  Technically feasible, but this would be extremely hard to write coherently.

I think the best approach is to add a "Redux Essentials, Part 9: Redux and TypeScript" page that would cover the TS material as a capstone to the tutorial sequence.  I'm picturing a page that would run through some of the key code sections and API examples from the earlier pages in the tutorial, and explain how you'd write that same code from scratch with TS.

There was also a great suggestion on Twitter here at https://twitter.com/MikeLunarLandis/status/1551777390447079425 :

> I can see prefacing the JS docs with a "We'll cover the TS-specific stuff at the end" note. Then, where it's important in the JS docs for TS folks to know about, say, inference adapters, add a "TS" badge that references the relevant section at the end.

I think that strikes a good balance between showing meaningful examples of TS usage, and not having to maintain multiple copies or weird back-and-forths in the main tutorial sequence.

zvolcsey:
Hey, it is an interesting task. I am taking a stab at it. :)

markerikson:
[...]

PR #4706, opened 2024-05-12
Revamp "Redux Essentials" tutorial to be TS-first and update contents

markerikson:
Actual content changes for #4393 , at long last!

The current WIP example code is over in:

- https://github.com/reduxjs/redux-essentials-example-app/tree/feature/tutorial-steps-ts-revamped

I'm doing another round of revisions and step-by-step checking to those code commits as I rework the tutorial content, but that should be the progression and code content I want to show off in the tutorial.

Big picture summary:

- The entire example app is TS, start to finish
- Improve some existing explanations
- Cover some concepts that weren't included originally, like having multiple reducers handle one action
- Cover newer APIs we didn't have in 2020, like `createListenerMiddleware` and thunks in `createSlice`
- While not Redux-related, update some of the usage patterns like React Router 5 -> 6 and using uncontrolled inputs in forms instead of controlled


docs/tutorials/essentials/part-3-data-flow.md
```diff
@@ -304,38 +467,65 @@ Our posts slice is responsible for handling all updates to the posts data. Insid
...
+
+:::
+
+When we write the `postAdded` reducer function, `createSlice` will automatically generate an ["action creator" function](../fundamentals/part-7-standard-patterns.md#action-creators) with the same name. We can export that action creator and use it in our UI components to dispatch the action when the user clicks "Save Post".
 
-When we write the `postAdded` reducer function, `createSlice` will automatically generate an "action creator" function with the same name. We can export that action creator and use it in our UI components to dispatch the action when the user clicks "Save Post".
+```ts title="features/posts/postsSlice.ts"
+// highlight-start
+// Import the `PayloadAction` TS type
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
```
aryaemami59, line 485:
```suggestion
import type { PayloadAction } from '@reduxjs/toolkit'
import { createSlice } from '@reduxjs/toolkit'
``` 

Explicit type imports could help better distinguish between TS/JS related material on a quick glance.

docs/tutorials/essentials/part-3-data-flow.md
```diff
@@ -345,55 +535,59 @@ We'll talk more about generating IDs and dispatching actions in [Part 4: Using R
...
-import React, { useState } from 'react'
+import React from 'react'
 // highlight-start
-import { useDispatch } from 'react-redux'
 import { nanoid } from '@reduxjs/toolkit'
 
-import { postAdded } from './postsSlice'
+import { useAppDispatch } from '@/app/hooks'
+
+import { type Post, postAdded } from './postsSlice'
```
aryaemami59, line 549:
```suggestion
import type { Post } from './postsSlice'
import { postAdded } from './postsSlice'
```

More of a personal nitpick, feel free to ignore :)

docs/tutorials/essentials/part-3-data-flow.md
```diff
@@ -345,55 +535,59 @@ We'll talk more about generating IDs and dispatching actions in [Part 4: Using R
...
+In order to dispatch actions from a component, **we need access to the store's `dispatch` function**. We get this by calling the `useDispatch` hook from React-Redux. Since we're using TypeScript, that means that we should actually **import the `useAppDispatch` hook with the right types**. We also need to import the `postAdded` action creator into this file.
 
-Once we have the `dispatch` function available in our component, we can call `dispatch(postAdded())` in a click handler. We can take the title and content values from our React component `useState` hooks, generate a new ID, and put them together into a new post object that we pass to `postAdded()`.
+Once we have the `dispatch` function available in our component, we can call `dispatch(postAdded())` in a click handler. We can take the title and content values from our form, generate a new ID, and put them together into a new post object that we pass to `postAdded()`.
 
 ```jsx title="features/posts/AddPostForm"
```
aryaemami59, line 542:
```suggestion
```tsx title="features/posts/AddPostForm"
```

Not 100% sure if this should be `tsx`, but thought I'd mention it just in case.

docs/tutorials/essentials/part-4-using-data.md
```diff
@@ -195,31 +190,44 @@ First, we need to update our `postsSlice` to create a new reducer function and a
...
 We also know that the reducer is responsible for determining how the state should actually be updated when an action is dispatched. Given that, we should have the reducer find the right post object based on the ID, and specifically update the `title` and `content` fields in that post.
 
 Finally, we'll need to export the action creator function that `createSlice` generated for us, so that the UI can dispatch the new `postUpdated` action when the user saves the post.
 
 Given all those requirements, here's how our `postsSlice` definition should look after we're done:
 
-```js title="features/posts/postsSlice.js"
+```ts title="features/posts/postsSlice.ts"
+// highlight-next-line
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
```
aryaemami59, line 218:
```suggestion
import type { PayloadAction } from '@reduxjs/toolkit'
import { createSlice } from '@reduxjs/toolkit'
```

docs/tutorials/essentials/part-4-using-data.md
```diff
@@ -413,31 +422,44 @@ const postsSlice = createSlice({
...
+You can't have a "social media" app if there aren't any other people involved! Let's add the ability to keep track of a list of users in our app, and update the post-related functionality to make use of that data.
 
 ### Adding a Users Slice
 
 Since the concept of "users" is different than the concept of "posts", we want to keep the code and data for the users separated from the code and data for posts. We'll add a new `features/users` folder, and put a `usersSlice` file in there. Like with the posts slice, for now we'll add some initial entries so that we have data to work with.
 
-```js title="features/users/usersSlice.js"
-import { createSlice } from '@reduxjs/toolkit'
+```ts title="features/users/usersSlice.ts"
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
```
aryaemami59, line 455:
```suggestion
import type { PayloadAction } from '@reduxjs/toolkit'
import { createSlice } from '@reduxjs/toolkit'
```

docs/tutorials/essentials/part-4-using-data.md
```diff
@@ -619,39 +626,68 @@ Like with the `post.user` field, we'll update our `postAdded` prepare callback t
...
 
-**Redux actions and state should only contain plain JS values like objects, arrays, and primitives. Don't put class instances, functions, or other non-serializable values into Redux!**.
+**Redux actions and state should only contain plain JS values like objects, arrays, and primitives. Don't put class instances, functions, `Date/Map/Set` instances, or other non-serializable values into Redux!**.
 
 :::
 
-Since we can't just put a `Date` class instance into the Redux store, we'll track the `post.date` value as a timestamp string:
+Since we can't just put a `Date` class instance into the Redux store, we'll track the `post.date` value as a timestamp string. We'll add it to the initial state values (using `date-fns` to subtract a few minutes from the current date and time), and also add it to each new post in the prepare callback
+
+```ts title="features/posts/postsSlice.ts"=
```
aryaemami59, line 635:
```suggestion
```ts title="features/posts/postsSlice.ts"
```

docs/tutorials/essentials/part-4-using-data.md
```diff
@@ -675,169 +711,514 @@ Typically, social media feeds show the newest posts first, and you scroll down t
...
+Then, we can define a new reducer that will handle updating the reaction count for a post when a user clicks the reaction button.
 
-Like with post authors and timestamps, we want to use this everywhere we show posts, so we'll create a `<ReactionButtons>` component that takes a `post` as a prop. We'll start by just showing the buttons inside, with the current reaction counts for each button:
+Like with editing posts, we need to know the ID of the post, and which reaction button the user clicked on. We'll have our `action.payload` be an object that looks like `{id, reaction}`. The reducer can then find the right post object, and update the correct reactions field.
 
-```jsx title="features/posts/ReactionButtons.js"
-import React from 'react'
+```ts
+import { createSlice, nanoid, PayloadAction } from '@reduxjs/toolkit'
```
aryaemami59, line 742:
```suggestion
import type { PayloadAction } from '@reduxjs/toolkit'
import { createSlice, nanoid } from '@reduxjs/toolkit'
```

docs/tutorials/essentials/part-4-using-data.md
```diff
@@ -675,169 +711,514 @@ Typically, social media feeds show the newest posts first, and you scroll down t
...
+For this example, we'll just add an `auth` slice that tracks `state.auth.username` so we know who the user is. Then, we can use that information whenever they add a post to automatically add the right user ID to the post.
+
+### Adding an Auth Slice
+
+The first step is to create the `authSlice` and add it to the store. This is the same pattern we've seen already - define the initial state, write the slice with a couple of reducers to handle updates for login and logout, and add the slice reducer to the store.
+
+In this case, our auth state is really just the current logged-in username, and we'll reset it to `null` if they log out.
+
+```ts title="features/auth/authSlice.ts"
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
```
aryaemami59, line 882:
```suggestion
import type { PayloadAction } from '@reduxjs/toolkit'
import { createSlice } from '@reduxjs/toolkit'
```

docs/tutorials/essentials/part-4-using-data.md
```diff
@@ -675,169 +711,514 @@ Typically, social media feeds show the newest posts first, and you scroll down t
...
+The `extraReducers` field is a function that receives a parameter named `builder`. The `builder` object has three methods attached, each of which lets the slice listen for other actions and do its own state updates:
+
+- `builder.addCase(actionCreator, caseReducer)`: listens for one specific action type
+- `builder.addMatcher(matcherFunction, caseReducer)`: listens for any one of multiple action types, using [a Redux Toolkit "matcher" function](https://redux-toolkit.js.org/api/matching-utilities) for comparing action objects
+- `builder.addDefaultCase(caseReducer)`: adds a case reducer that runs if nothing else in this slice matched the action (equivalent to a `default` case inside of a `switch`).
+
+Given that, we can import the `userLoggedOut` action from `authSlice.ts` into `postsSlice.ts`, listen for that action inside of `postsSlice.extraReducers`, and return an empty posts array to reset the posts list on logout:
+
+```ts title="features/posts/postsSlice.ts"
+import { createSlice, nanoid, PayloadAction } from '@reduxjs/toolkit'
```
aryaemami59, line 1178:
```suggestion
import type { PayloadAction } from '@reduxjs/toolkit'
import { createSlice, nanoid } from '@reduxjs/toolkit'
```

docs/tutorials/essentials/part-4-using-data.md
markerikson: I may apply the `type` import change near the end, but I'm honestly not worried about that for now

markerikson:
**THE MAIN DRAFT IS DONE!**

I've finished updating the entire example app and all the content in Parts 3-8.

I do want to review Parts 1 and 2 just to see if there's anything in those worth updating.  (I may want to convert the example app in Part 2 to TS as well, but that's a single sandbox.)

At this point, it should be a matter of reviewing the content changes, and then finalizing the PR:

- Taking the rewritten app code in https://github.com/reduxjs/redux-essentials-example-app/tree/feature/tutorial-steps-ts-checked , creating a new initial commit, and squashing any leftover intermediate commits to form the final commit sequence
- Tagging the right commits for each page
- Updating the CodeSandboxes to point to the tags

Dan503:
I'm working through this tutorial today.

I downloaded the code here thinking it was the starter code:
https://github.com/reduxjs/redux-essentials-example-app/tree/feature/tutorial-steps-ts-revamped

It has the store and everything already set up though. How do I download the starter code intended as the starting point for the tutorial?

markerikson:
@Dan503 yeah, I need to finish preparing the branch. If you go back far enough, there should be a couple of commits labeled "SQUASHME", with the commit adding the store right after that. You'd want to make a new branch off of that commit. 

On mobile atm, but let me see if I can find the right commit and paste it here.

Dan503:
I'll post issues here as I run into them.

I wanted to use `pnpm` to install the project. I ran into this error when I ran `pnpm install`:

```
$ pnpm install
 ERR_PNPM_OTHER_PM_EXPECTED  This project is configured to use yarn
```

`pnpm install` works on the version of the tutorial in master and the instructions in the docs explicitly say:

> The project is configured to use [Yarn 4](https://yarnpkg.com/) as the package manager, but you can use any package manager ([NPM](https://docs.npmjs.com/cli/v10), [PNPM](https://pnpm.io/), or [Bun](https://bun.sh/docs/cli/install)) as you prefer.

markerikson:
I do have Yarn set up already, yeah, but there's nothing specific about the project that requires _only_  Yarn.

But, sounds like PNPM tried to warn you if it detects other package manager config files.

If you delete the Yarn config file and lock file, PNPM would likely install without warnings.

Dan503:
image

This code example is misleading as to what a reducer is. There is going to be a lot of people who will skip the previous documentation and want to just jump straight into the code tutorial so it's important not to mislead people and teach the concepts as they work through the exercises.

This code sample tells me this:
"A reducer is a function for... um... retrieving a single value from the state? ...but the function just returns the exact value you pass into it.... so a reducer is a function that does nothing?"

I think it is better to just set up the store and leave a comment saying that slice reducer functions will go into the reducer property

```ts
export const store = configureStore({
  // Pass in the root reducer setup as the `reducer` argument
  reducer: {
    // Slice reducer functions will go here, more on these later
  },
})
```

Dan503:
[Creating the Posts Slice section](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-3-data-flow#creating-the-posts-slice)

default export

I don't think default export should be encouraged. It makes it more difficult to import the reducer into the `store.ts` file which is the main (if not only) reason why you are exporting the reducer in the first place.

I think that line should be replaced with this:

```ts
// Export the generated reducer function as `postReducer`
export const postReducer = postSlice.reducer
```

Dan503:
I'm going to submit further feedback as PR review comments. It is easier to have further discussion that way.

markerikson:
@Dan503 fwiw the `export default someReducer` is a pattern that's existed in the Redux ecosystem since 2016:

- https://github.com/erikras/ducks-modular-redux

and it's something we've taught pretty consistently with RTK ever since it came out:

- https://redux.js.org/tutorials/essentials/part-2-app-structure#creating-slice-reducers-and-actions

Agreed that default exports don't help with intellisense, and I've seen plenty of good arguments against them.  If you _want_ to do a named export like `export const postsReducer = postsSlice.reducer`, definitely nothing stopping you there.  But this is also a pattern we've had pretty consistently and I'm not looking to change it atm.

Dan503:
[Showing the Posts List](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-3-data-flow#showing-the-posts-list)

It is important to highlight that `PostList.tsx` is a `tsx` file and not a `ts` file. 

This is the first `tsx` file that the user has to create themselves while following the tutorial. I can easily see someone not noticing the `x` at the end of the file name, creating a `PostList.ts` file, then running into all sorts of confusing errors that they have no idea how to debug because they are putting JSX code in a `ts` file.

markerikson:
@Dan503 (sorry if it sounds like I'm nitpicking or counteracting your feedback here! Just seeing the notifications come in and responding while it's on my mind.)

Might be worth saying something, but at the same time, I'm trying to keep the tutorial itself focused on Redux concepts.  If you look at Page 1, I specifically added a section saying "here's a list of things you should already know before you go through this tutorial", so that I can make the assumption that "the reader is already familiar with these things" and I _don't_ have to try to explain them.  (If you take things far enough, you end up with "for this Redux tutorial, we will first explain how physics work, so we can explain how CPUs work, so we can explain assembly, all the way up to JS".  Gotta draw the line somewhere! :)  )

That said, I should at least add a "you should be familiar with TS syntax and TS+React basics" line to that callout block.

Dan503:
@markerikson 

I get your point. At the same time, this seems like a very minimal amount of effort to cater to a larger audience.

My suggestion is to essentially add one more sentence to this paragraph:

> Now that we have some posts data in our store, we can create a React component that shows the list of posts. All of the code related to our feed posts feature should go in the `posts` folder, so go ahead and create a new file named `PostsList.tsx` in there.

The new sentence being:
"Note that the file has a `tsx` extension instead of `ts` as this file will contain JSX code in it."

So the full paragraph becomes:

> Now that we have some posts data in our store, we can create a React component that shows the list of posts. All of the code related to our feed posts feature should go in the `posts` folder, so go ahead and create a new file named `PostsList.tsx` in there. Note that the file has a `tsx` extension instead of `ts` as this file will contain JSX code in it.

Dan503:
[Adding the Single Post Route](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-4-using-data#adding-the-single-post-route)

This code block has an accessibility issue in it:

```jsx
 const renderedPosts = posts.map(post => (
    <article className="post-excerpt" key={post.id}>
      <h3>{post.title}</h3>
      <p className="post-content">{post.content.substring(0, 100)}</p>

      {/* This link is an accessibility issue */}
      <Link to={`/posts/${post.id}`} className="button muted-button">
        View Post
      </Link>

    </article>
  ))
```

Any screen reader user tabbing through the posts will just hear "link view post, link view post, link view post" over and over with zero context around what is on the other side of that link.

To fix this accessibility issue link the title instead:

```jsx
 const renderedPosts = posts.map(post => (
    <article className="post-excerpt" key={post.id}>
      <h3>
        <Link to={`/posts/${post.id}`} className="button muted-button">
          {post.title}
        </Link>
      </h3>
      <p className="post-content">{post.content.substring(0, 100)}</p>
    </article>
  ))
```

Dan503:
I got up to here: https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-4-using-data#adding-user-login

I'll go through more of it tomorrow

markerikson:
FWIW accessibility isn't my top priority for this tutorial - my real concerns are around actually teaching Redux concepts and usage patterns.

_If_ I can make a couple of those tweaks I'll try, although it'll mean having to redo the entire commit stack _again_.

Dan503:
I like the new login section of the tutorial. It teaches a new concept at the end that the current live site tutorial doesn't cover (The using actions from different slice files aspect)

markerikson:
Yep, that's the _entire_ reason I came up with the whole login feature :)  It's literally just to show off `extraReducers` and handling other actions, as its own concept, and separate from using `extraReducers` to handle thunks.

Dan503:
The current tutorial names the root redux state object `RootState`

When I am typing `RootState` and I use the intellisense auto-import feature in VS code, the default import statement it adds to the top of the file is this:

```js
import { RootState } from '@reduxjs/toolkit/query'
```

Idealy it would import the local `RootState` type by default instead.

This ends up leading to confusion where you are getting a TS error and don't understand why, then you eventually realize that it imported the wrong `RootState` type.

I suggest changing the type name in the tutorial to `ReduxState` or `AppRootState` to avoid this name conflict with the `RootState` type in `'@reduxjs/toolkit/query'`.

Dan503:
The location of the [Extracting Selectors for Slices](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-5-async-logic#extracting-selectors-for-slices) in the tutorial is very odd.

We are introducing the concept of thunks and async logic so I'm expecting this section to be all about exploring how to do async logic in Redux... then we suddenly go on a massive tangent explaining reusable selectors that have nothing to do with thunks and async logic. It's quite jarring. 

I believe that the perfect place to put the reusable selectors tutorial would be when implementing the `EditPostForm.tsx` part of the tutorial. `EditPostForm.tsx` is the first time we need to reuse the `selectPostById` logic.

The `selectPostById` logic is somewhat complicated and so it is easy to justify placing the tutorial here. 

**Present the problem:** "We need to select a post by the ID again here for the second time in our app. The logic to select a post by an ID is slightly complicated and something we are likely to need to do in multiple places throughout our app. We don't want to have to duplicate this code every time we want select a post by its ID."

**Provide the solution:** "Let's use **reusable selector functions** in our `postsSlice.tsx` file to avoid writing this code over and over."

Then the Selector tutorial you have already written can slot in perfectly right there.

markerikson:
Yeah, that's reasonable. I wasn't terribly _thrilled_ about adding all the selector stuff at the beginning of Part 5.  Part of my thought process was that Part 4 is already 9500 words and I was trying to keep page length _somewhat_ balanced.

(I've also briefly considered altering the tutorial page structure to insert a "part 4.5" - put selectors and `extraReducers` in a page in between the existing Part 4 for data usage and Part 5 for async logic.  But I also don't want to have to go reworking all the page links / invalidating existing links.)

Dan503:
The current code for `selectPostById` is this:

```ts
export const selectPostById = (state: RootState, postId: string) =>
  state.posts.find(post => post.id === postId)
```

`postId` is of type `string`.

In the [`SinglePostPage.tsx` tutorial](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-4-using-data#creating-a-single-post-page) you retrieve `postId` from `useParams`

```ts
  const { postId } = useParams()
  const post = useAppSelector((state) => selectPostById(state, postId))
```

`postId` from `useParams()` is of type `string | undefined`. So when you refactor the `const post` line into this:

```ts
  const post = useAppSelector((state) => selectPostById(state, postId))
```

You get a TS error:
```
Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.
```

To avoid getting the TS error, I recommend `postId` be typed as `string | undefined` instead of type `string`.

```ts
export const selectPostById = (state: RootState, postId: string | undefined) =>
  state.posts.find(post => post.id === postId)
```

markerikson:
I think the example code does `postId!` to tell TS it _is_ defined, but yeah, that's another option.

markerikson:
(btw, _thank you_ for the _very_ thorough review work here! Really appreciate you going through and actually thinking about what's there - very helpful!)

Dan503:
> I think the example code does `postId!` to tell TS it _is_ defined, but yeah, that's another option.

I haven't seen `!` used in any types throughout the documentation so far. `!` is also a more advanced TS concept that beginner TS users are likely to not understand. I would avoid using that.

In the [`selectUserById`](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-5-async-logic#extracting-auth-and-users-selectors) code, you type `userId` as `userId?: string`.

```ts
export const selectUserById = (state: RootState, userId?: string) => {
  return state.users.find(user => user.id === userId)
}
```

I don't recommend that either. `userId` is a required parameter for that function. It should never be excluded when using the  function.

That code allows you to write the selector like this:

```ts
const user = selectUserById(state)
```

That is an invalid use of the `selectUserById` function. Typing `userId` as `userId: string | undefined` forces people to use the selector correctly by passing in an ID (even if the ID is currently undefined).

Dan503:
I ran into a type error with the `postUpdated` function.

`postUpdated` reducer added in the [Updating Post Entries](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-4-using-data#updating-post-entries) section. 

The action is typed as `PayloadAction<Post>` for that code block (makes sense, that is all the Post has in it at this stage).

```ts
postUpdated(state, action: PayloadAction<Post>) {
    // code omitted
}
```

The problem comes in when we start adding new fields like date, author and reactions to the `Post` interface.

The code `type PostUpdate = Pick<Post, 'id' | 'title' | 'content'>` is introduced in the [Tracking Reactions Data in Posts](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-4-using-data#tracking-reactions-data-in-posts) section. 

This resolves the type conflict issue however there are two problems with it currently:

1. The code is not highlighted even though this is the first time we see that line of code.
2. The code that uses the `PostUpdate` type is omitted so the tutorial never actually tells you where to use that type in the code.

I think the `date` field is the first field added that breaks the ts type for the `postUpdated` function.

I recommend having an explicit section for fixing the type conflict on `postUpdated` once the `date` field is added to the `Post` interface.

Dan503:
[Loading State for Requests](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-5-async-logic#loading-state-for-requests)

You have this code sample already use TS notation:

```ts
{
  // Multiple possible status enum values
  status: 'idle' | 'loading' | 'failed' | 'succeeded',
  error: string | null
}
```

Make it an explicit TS interface and store it somewhere global (not sure where it would make sense to store it, maybe in an apiTypes.ts file in the api directory)

```ts
export interface LoadingState {
  // Multiple possible status enum values
  status: 'idle' | 'loading' | 'failed' | 'succeeded',
  error: string | null
}
```

Then the `PostState` interface can extend the `LoadingState` interface to add the loading state.

```ts
interface PostsState extends LoadingState {
  posts: Array<Post>
}
```

There are multiple slice files that need to load data so it doesn't make sense to hardcode the loading state into PostsState.

Dan503:
Another point on [Loading State for Requests](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-5-async-logic#loading-state-for-requests)

The status values in the loading pattern demonstration are inconsistent.

The loading pattern demonstration has a `'loading'` value.

The `PostsState` interface has a `'pending'` value instead.

This is a bit confusing, it would be better if the values were consistent.

Dan503:
This is probably out of scope for this PR but I'll mention it anyway.

The dark-mode styling of the optional extra info sections has a couple issues.

1. The link text is difficult to read, it is very light and it is against a light background
2. The background color of the details box is very bright for a large dark-mode content area. It is like having a light-bulb shone in your face while working at night.

Both of these issues can be fixed with a darker background color.

image

Dan503:
[Sending Data with Thunks](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-5-async-logic#sending-data-with-thunks)

The `features/posts/postsSlice.ts` code sample has incorrect types.

The code sample in the tutorial says to write this:

```ts
export const addNewPost = createAsyncThunk(...)
```

To avoid a TS error, I needed to write this:

```ts
// Placed next to the PostUpdate type
type PostAddNew = Pick<Post, 'title' | 'content' | 'user'> 

// code omitted

// Added explicit <Post, PostAddNew> generics
export const addNewPost = createAsyncThunk<Post, PostAddNew>(...)
```

Dan503:
Another point on this section: [Sending Data with Thunks](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-5-async-logic#sending-data-with-thunks)

Highlight this comment line to ensure that the tutorial user reads it (I accidentally left the `postAdded` reducer in and ended up having it post twice)
```js
// The existing `postAdded` reducer and prepare callback were deleted
```

I also suggest adding a highlighted comment line saying to delete this bit of code:
```
dispatch(postAdded(title, content, userId))
```

I also think the input fields and submit button should become disabled while the form is submitting. This prevents the user from submitting the form twice or thinking that they can modify the data before it finishes submitting.

Dan503:
[Thunk Functions](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-5-async-logic#thunk-functions)

I went through the entire Part 5 Async Logic and Data Fetching tutorial.

I never ended up using this anywhere in my code:
```ts
// Export a reusable type for handwritten thunks
export type AppThunk = ThunkAction<void, RootState, unknown, Action>
```

So there either needs to be an example added on when `AppThunk` should be used, or it should be removed from the tutorial.

Dan503:
[Sending Login Requests to the Server](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-6-performance-normalization#sending-login-requests-to-the-server)

> Along with that, we'll update `<Navbar>` and `<LoginPage>` to import and dispatch the new thunks instead of the previous action creators.

The `<Navbar>` and `<LoginPage>` updates need to show code samples showing what to do.

Dan503:
[Simplifying Thunk Types](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-6-performance-normalization#simplifying-thunk-types)

I'm not sure why this section is so late in the tutorial. Please don't force the reader to have to go back and do massive refactors of all their previous work.

This should just be how thunks are introduced to the reader from the beginning. Like what you did when teaching `useSelector`. You did not get the reader to write all their `useSelector()` code and then tell them to go back and refactor it all in the next part of the tutorial. You introduced the concept of `useAppSelector()` from the very start so that they were able to use that method from the very beginning.

Dan503:
More on [Simplifying Thunk Types](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-6-performance-normalization#simplifying-thunk-types)

I see `createAppAsyncThunk` as belonging alongside `useAppDispatch` and `useAppSelector`. So instead of having `useAppDispatch` and `useAppSelector` in a `hooks.ts` file, I think they should all live in the `withTypes.ts` file.

Dan503:
This is likely out of scope for this PR...

`createAppAsyncThunk` appears to be able to detect the first generic quite well automatically. If a parameter is needed, the parameter generic type needs to always be explicitly declared. Unfortunately the Parameter generic is the 2nd Generic so to reach it you have to provide the return type first (the 1st Generic). 

An example of what I mean:

```ts
// The only reason I am providing Post here is because I need to provide the PostAddNew type for the parameter
export const addNewPost = createAppAsyncThunk<Post, PostAddNew>(
  'posts/addNewPost',
  async (initialPost) => {
    // omitted
  },
)
```

What I would prefer to do:

```ts
// Post is able to be inferred while a type for the parameter is explicitly defined
export const addNewPost = createAppAsyncThunk<PostAddNew>(
  'posts/addNewPost',
  async (initialPost) => {
    // omitted
  },
)
```
It would be good if there was a way to make `createAppAsyncThunk` able to optionally accept only a parameter generic if one is needed.

Dan503:
[Notifications Slice](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-6-performance-normalization#notifications-slice)

There is a native `Notification` type. When using VS Code auto-import it tends to prefer just using the native `Notification` type instead of the custom one. Changing the name will help avoid the user accidentally using the incorrect type when importing it. 

I suggest "`AppNotification`" as the type name.

Dan503:
[Adding the Notifications List](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-6-performance-normalization#adding-the-notifications-list)

`UNKNOWN_USER` is currently written like this:

```ts
const UNKNOWN_USER = {
  name: 'Unknown User'
}
```

I think this is a better way of handling `UNKNOWN_USER` since it enforces consistency with all other User values.

```ts
const UNKNOWN_USER: User = {
  name: 'Unknown User',
  id: nanoid(),
}
```

Dan503:
[Adding the Notifications List](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-6-performance-normalization#adding-the-notifications-list)

I did my NotificationsList component quite differently.

This is mine for comparison:

```tsx
import { nanoid } from '@reduxjs/toolkit'
import { useAppSelector } from '../../app/withTypes'
import { TimeAgo } from '../posts/TimeAgo'
import { User, selectUserById } from '../users/usersSlice'
import { AppNotification, selectAllNotifications } from './notificationsSlice'
import { Link } from 'react-router-dom'

const UNKNOWN_USER: User = {
  name: 'Unknown User',
  id: '',
}

export function NotificationsList() {
  const notifications = useAppSelector(selectAllNotifications)

  return (
    <section className="notificationsList">
      <h2>Notifications</h2>
      {notifications.map((n) => (
        <NotificationItem notification={n} key={n.id} />
      ))}
    </section>
  )
}

interface NotificationItemProps {
  notification: AppNotification
}
function NotificationItem({ notification }: NotificationItemProps) {
  const user = useAppSelector((state) => selectUserById(state, notification.user)) || UNKNOWN_USER
  return (
    <article key={notification.id}>
      <p>
        <b>{user.id ? <Link to={`/users/${user.id}`}>{user.name}</Link> : user.name}</b>
        {` ${notification.message} - `}
        <TimeAgo isoTime={notification.date} />
      </p>
    </article>
  )
}
```

Dan503:
[Tracking Notification Status](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-6-performance-normalization#tracking-notification-status)

This should be highlighted in the code preview:

```ts
state.push(...notificationsWithMetadata)
```

Dan503:
[Updating the Posts Slice](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-6-performance-normalization#updating-the-posts-slice) (`createEntityAdapter`)

If I don't make any changes to the PostsState interface then I get this TS error on the `initialState` variable declaration:

> Property 'posts' is missing in type 'EntityState<Post, string> & { status: "idle"; error: null; }' but required in type 'PostsState'.

The `PostsState` interface needs to be updated and the docs do not mention this at all. The interface is also omitted in the code sample.

This is what I updated mine to:

```ts
// src/api/api.types.ts

export interface LoadingState {
  status: LoadingStatusString
  error: string | null
}

export type LoadingStatusString = 'idle' | 'pending' | 'failed' | 'succeeded'
```

```ts
// src/features/posts/postsSlice.ts

import { EntityState } from '@reduxjs/toolkit'
import { LoadingState } from '../../api/api.types'

type PostsState = LoadingState & EntityState<Post, string>
```

Dan503:
[Updating the Posts Slice](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-6-performance-normalization#updating-the-posts-slice) (createEntityAdapter)

This code in the preview example triggers a TS error:

```ts
export const { ... } = postsAdapter.getSelectors(state => state.posts)
```

> 'state' is of type 'unknown'.

It needs to be fixed by doing this:
```ts
export const { ... } = postsAdapter.getSelectors<RootState>(state => state.posts)
```

It would be good if I could do `createEntityAdapter.withTypes()` to avoid this need. This doesn't seem to be possible at the moment though.

Dan503:
[Showing Toasts for New Posts](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-6-performance-normalization#showing-toasts-for-new-posts)

The instructions currently don't say where you need to call `addPostListeners()`, so readers get to the end of the tutorial and it doesn't work because the function is never called.

I called it in `main.ts` just under the `const root = createRoot(document.getElementById('root')!)` line. That seemed to work.

EskiMojo14:
> This is likely out of scope for this PR...
> 
> `createAppAsyncThunk` appears to be able to detect the first generic quite well automatically. If a parameter is needed, the parameter generic type needs to always be explicitly declared. Unfortunately the Parameter generic is the 2nd Generic so to reach it you have to provide the return type first (the 1st Generic).
> 
> An example of what I mean:
> 
> ```ts
> // The only reason I am providing Post here is because I need to provide the PostAddNew type for the parameter
> export const addNewPost = createAppAsyncThunk<Post, PostAddNew>(
>   'posts/addNewPost',
>   async (initialPost) => {
>     // omitted
>   },
> )
> ```
> 
> What I would prefer to do:
> 
> ```ts
> // Post is able to be inferred while a type for the parameter is explicitly defined
> export const addNewPost = createAppAsyncThunk<PostAddNew>(
>   'posts/addNewPost',
>   async (initialPost) => {
>     // omitted
>   },
> )
> ```
> 
> It would be good if there was a way to make `createAppAsyncThunk` able to optionally accept only a parameter generic if one is needed.

```ts
export const addNewPost = createAppAsyncThunk(
  'posts/addNewPost',
  async (initialPost: PostAddNew) => {
    // omitted
  },
)
```

Dan503:
Thanks @EskiMojo14 

@markerikson This needs to be documented as the recommended method for using `createAppAsyncThunk`

```ts
export const addNewPost = createAppAsyncThunk(
  'posts/addNewPost',
  async (initialPost: PostAddNew) => {
    // omitted
  },
)
```

Dan503:
[Prefer a video course?](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-7-rtk-query-basics)

Should probably mention in the text description introducing the video that the video course is written in plain JavaScript, not TypeScript.

Dan503:
[Motivation](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-7-rtk-query-basics#motivation)

This quote textbox needs to be made dark-mode friendly. It is far too bright.

image

Dan503:
[Configuring the Store](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-7-rtk-query-basics#configuring-the-store)

You said `store.js` instead of `store.ts` here.

Dan503:
[Using Query Hooks in Components](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-7-rtk-query-basics#using-query-hooks-in-components)

> `isLoading`: a boolean indicating if this hook is currently making the first request to the server. (Note that if the parameters change to request different data, isLoading will remain false.)
>
> `isFetching`: a boolean indicating if the hook is currently making any request to the server

This is all the description we are given on the difference between `isLoading` and `isFetching` (based on what I've read so far). When I used Redux in my own projects I went with `isFetching` because I didn't want to run into a situation where Redux was loading data from the server and wasn't displaying the loading indicator. I was essentially taking an "it's better safe than sorry" approach since `isFetching` will be guaranteed to be true if data is being fetched. I didn't trust `isLoading` to behave the way I want since I don't fully understand when it is and is not appropriate to use `isLoading`.

What the tutorial needs is a clear demo showing when it is appropriate to use `isLoading` and when it is more appropriate to use `isFetching`. `isLoading` is too scary and confusing to use at the moment since it is only true _some_ of the time, not _all_ of the time.

Dan503:
[Sorting Posts](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-7-rtk-query-basics#sorting-posts)

Instead of creating a copy of the array, you could use the new `toSorted` array method which returns a sorted array copy instead of mutating the original. It is supported in all modern browsers.

```ts
// .sort() method
const sortedPosts = useMemo(() => {
    const sortedPosts = posts.slice()
    // Sort posts in descending chronological order
    sortedPosts.sort((a, b) => b.date.localeCompare(a.date))
    return sortedPosts
}, [posts])
```
```ts
// .toSorted() method
const sortedPosts = useMemo(() => {
    // Sort posts in descending chronological order
    return posts.toSorted((a, b) => b.date.localeCompare(a.date))
}, [posts])
```

Dan503:
[Cache Data Subscription Lifetimes](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#cache-data-subscription-lifetimes)

At this point in the tutorial, author names on posts are undefined since author names haven't been converted to RTK Query yet. Since the author id on the post (`undefined`) does not match the users author ID, the user is not able to edit any posts unless they create a new one.

I suggest having the reader update the author names code first so that they can edit a post that is already assigned to themself.

Dan503:
[Cache Data Subscription Lifetimes](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#cache-data-subscription-lifetimes)

These are the instructions you provide:

> Open up your browser's DevTools, go to the Network tab, and refresh the main page. You should see a GET request to /posts as we fetch the initial data.

When you refresh the main page you get logged out, so you need to mention a couple more steps.

> Open up your browser's DevTools, go to the Network tab, refresh the page, clear the network tab, then login. You should see a GET request to "/posts" as we fetch the initial data.

I suggest clearing the network tab because it is far easier to see the important requests if you clear it beforehand.

Dan503:
[Invalidating Specific Items](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#invalidating-specific-items)

The basic string `'Post'` is used with no security around ensuring that the value is typed correctly. It is very easy to type `"Post"` as `"post"` or `"pst"` or `"P0st"`.

To avoid accidentally misspelling words and reduce repetition, I used enums and a `TAGS` constant.

This was my final code:

```ts
enum TagType {
  post = 'post',
}
enum TagId {
  list = 'list',
}

const TAGS = {
  postList: { type: TagType.post, id: TagId.list },
}

// Define our single API slice object
export const apiSlice = createApi({
  // The cache reducer expects to be added at `state.api` (already default - this is optional)
  reducerPath: 'api',
  // All of our requests will have URLs starting with '/fakeApi'
  baseQuery: fetchBaseQuery({ baseUrl: '/fakeApi' }),
  // Used for invalidating cached data triggering an immediate auto-refetch of data
  tagTypes: [TagType.post],
  // The "endpoints" represent operations and requests for this server
  endpoints: (builder) => ({
    // The `getPosts` endpoint is a "query" operation that returns data.
    // The return value is a `Post[]` array, and it takes no arguments.
    getPosts: builder.query<Array<Post>, void>({
      // The URL for the request is '/fakeApi/posts'
      query: () => '/posts',
      providesTags: (result = [], _error, _arg) => [
        TAGS.postList,
        ...result.map(({ id }) => ({ type: TagType.post, id }) as const),
      ],
    }),
    getPostById: builder.query<Post, string | undefined>({
      query: (postId) => `/posts/${postId}`,
      providesTags: (_result, _error, arg) => [{ type: TagType.post, id: arg }],
    }),
    addNewPost: builder.mutation<Post, PostAddNew>({
      query: (newPostData) => ({
        url: '/posts',
        method: 'POST',
        // Include the entire post object as the body of the request
        body: newPostData,
      }),
      invalidatesTags: [TAGS.postList],
    }),
    editPost: builder.mutation<Post, PostUpdate>({
      query: (post) => ({
        url: `posts/${post.id}`,
        method: 'PATCH',
        body: post,
      }),
      invalidatesTags: (_result, _error, arg) => [{ type: TagType.post, id: arg.id }],
    }),
  }),
})
```

EskiMojo14:
> The basic string 'Post' is used with no security around ensuring that the value is typed correctly. It is very easy to type "Post" as "post" or "pst" or "P0st".
your "security" is Typescript. it will explicitly complain if you type the wrong string.

Dan503:
I just realized that I still have all of my manual fetching code in my `postsSlice.ts` file. There should be a reminder at the end of each conversion-to-RTK-Query section to remove the specific parts of the slice file that have been made obsolete.

Dan503:
> your "security" is Typescript. it will explicitly complain if you type the wrong string.

Ok yep. You're right, it will complain, and autocomplete does work in most circumstances which is good.

The one situation where it didn't provide autocomplete functionality was on this line:

```ts
...result.map(({ id }) => ({ type: 'post', id }) as const),
```

`providesTags` still errors in that situation though so I'll mark that comment as resolved.

Dan503:
[Selecting Users Data](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#selecting-users-data)

At the end of this tutorial, the Login page doesn't work because the Login Page uses this selector that has no data:

```ts
const userStatus = useAppSelector((state) => state.users.status)
```

A small tutorial needs to be added to the end of this section explaining how to correctly update `LoginPage.tsx`

Dan503:
I've gotten to [this point](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#advanced-cache-updates) in the tutorial and the author names are not being rendered on my posts.

I probably did something wrong at the [Transforming Responses](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#transforming-responses) part of the tutorial.

I'm going to debug and then update this comment with how the tutorial can be improved to avoid this from happening.

Dan503:
[Selecting Values from Results](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#selecting-values-from-results)

You pass `undefined` into this code example however it is never explained why. I have no idea from the tutorial why I am passing `undefined` into that 1st parameter. 

When exactly am I supposed to pass `undefined`? What should I pass into that 1st parameter if it isn't `undefined`? I have no idea.

```ts
const { postsForUser } = useGetPostsQuery(undefined, {
```

Dan503:
[Streaming Cache Updates](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#streaming-cache-updates)

Typing `event` as `any` here seems like a bad use case for using `any`.

```ts
const listener = (event: any) => {
```

The correct typing for `event` here appears to be this:

```ts
const listener = (event: MessageEvent<string>) => {
```

Dan503:
I've gotten to the end of the Advanced RTK Query tutorial now. Overall this Redux tutorial project is a massive improvement compared to the original.

By the end of the [notifications tutorial][1] section my notifications are still not working. I'm going to do some debugging and see if I can figure out if there is something in the tutorial that needs updating or if I just missed something.

The [notifications tutorial][1] is the most complex and difficult to follow tutorial out of all of them. It would be good if there was a purpose built API on the horizon that can simplify this right down like how RTK Query simplified data fetching in Redux.

[1]: https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#tracking-client-side-state

Dan503:
This seems to be the core of my issue:
```ts
export const selectNotificationsResult = notificationApiSlice.endpoints.getNotifications.select()

export const selectNotificationsData = createSelector(selectNotificationsResult, (result) => result.data ?? [])

export const fetchNotificationsWebsocket = (): AppThunk => (_dispatch, getState) => {
  const state = getState()
  const allNotifications = selectNotificationsData(state)
  const [latestNotification] = allNotifications
  const latestTimeStamp = latestNotification?.date ?? ''
  // Hardcode a call to the mock server to simulate a server push scenario over websockets
  forceGenerateNotifications(latestTimeStamp)
}
```
`selectNotificationsData` is where the issue appears to come from.

`getState()` is working. It comes back with the full state object.

`selectNotificationsData(state)` is not working. It always returns an empty array.

markerikson:
@Dan503 Thanks again for all the review comments! Sorting through these now.

> The [notifications tutorial](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#tracking-client-side-state) is the most complex and difficult to follow tutorial out of all of them

Anything specific about this that felt "complex" and "difficult to follow"?  Was it the idea of the websocket setup? The way the cache lifetime callbacks work? Something about the explanation itself?

> selectNotificationsData(state) is not working. It always returns an empty array.

If you can push up your repo, I could take a look and see what's going on.

markerikson:
Okay, pushed content updates for parts 3, 4, and half of 5.

The new example code branch is **[`feature/tutorial-steps-ts-post-review`](https://github.com/reduxjs/redux-essentials-example-app/tree/feature/tutorial-steps-ts-post-review)** - pushed what I have so far.

Dan503:
> If you can push up your repo, I could take a look and see what's going on.

[My GitHub repo](https://github.com/Dan503/playtest-ts-redux-example-app)

docs/tutorials/essentials/part-4-using-data.md
```diff
@@ -441,6 +441,150 @@ const handleSubmit = (e: React.FormEvent<AddPostFormElements>) => {
...
+const postsSlice = createSlice({
+  name: 'posts',
+  initialState,
+  reducers: {
+    /* omit reducer logic */
+  },
+  // highlight-start
+  selectors: {
+    // Note that `state` here is just the `PostsState`!
+    selectAllPosts: state => state,
```
Dan503, line 562: I think it would be clearer for people if the variable here was named `postsState` instead of `state`.

People have associated the name `state` in selectors with `rootState` now. Calling it just "`state`" hear makes it confusing since it is the same name for a different concept.

docs/tutorials/essentials/part-4-using-data.md
```diff
@@ -441,6 +441,150 @@ const handleSubmit = (e: React.FormEvent<AddPostFormElements>) => {
...
+export const EditPostForm = () => {
+  const { postId } = useParams()
+
+  // highlight-next-line
+  const post = useAppSelector(state => selectPostById(state, postId!))
+  // omit component logic
+}
+```
+
+Note that the `postId` we get from `useParams()` is typed as `string | undefined`, but `selectPostById` expects a valid `string` as the argument. We can use the TS `!` operator to tell the TS compiler this value will not be `undefined` at this point in the code. (This can be dangerous, but we can make the assumption because we know the routing setup only shows `<EditPostForm>` if there's a post ID in the URL.)
```
Dan503, line 526: Glad you added this explanation and highlight of the TS `!` operator.

docs/tutorials/essentials/part-3-data-flow.md
```diff
@@ -212,7 +232,7 @@ export type RootState = ReturnType<typeof store.getState>
 // highlight-end
 ```
 
-If you hover over the `RootState` type in your editor, you should see `type RootState = { value: number; }`. Since this type is automatically derived from the store definition, all the future changes to the `reducer` setup will automatically be reflected in the `RootState` type as well. This way we only need to define it once, and it will always be accurate.
+If you hover over the `RootState` type in your editor, you should see `type RootState = { counter: CounterState; }`. Since this type is automatically derived from the store definition, all the future changes to the `reducer` setup will automatically be reflected in the `RootState` type as well. This way we only need to define it once, and it will always be accurate.
```
Dan503, line 235: Please rename `RootState` to `AppRootState` to avoid people accidentally importing the wrong `RootState` type by accident when auto-importing and to be consistent with the other App specific redux names.

docs/tutorials/essentials/part-4-using-data.md
```diff
@@ -498,13 +649,29 @@ Now, the root state looks like `{posts, users}`, matching the object we passed i
 
 ### Adding Authors for Posts
 
-Every post in our app was written by one of our users, and every time we add a new post, we should keep track of which user wrote that post.
+Every post in our app was written by one of our users, and every time we add a new post, we should keep track of which user wrote that post. This will need changes for both the Redux state and the `<AddPostForm>` component.
 
-To keep things simpler for this example, we'll update our `<AddPostForm>` component so that we can select a user from a dropdown list, and we'll include that user's ID as part of the post. Once our post objects have a user ID in them, we can use that to look up the user's name and show it in each individual post in the UI.
+First, we need to update the existing `Post` data type to include a `user: string` field that contains the user ID that created the post. We'll also update the existing post entries in `initialState` to have a `post.user` field with one of the example user IDs.
```
Dan503, line 654: I would prefer the key name to be `userId` instead of `user` since when I see `user` it sounds like it contains the full user object.

Dan503 (commented):
Lots of good changes in this update :)

Dan503:
[Tracking Client-Side State](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#tracking-client-side-state)

> export const extendedApi = apiSlice.injectEndpoints({

Instead of calling these injected API endpoints `extendedApi`, it would be better to call them `[componentName]ApiSlice` like `notificationApiSlice`. This avoids name conflict confusion when importing them into components.

Dan503:
> Anything specific about this that felt "complex" and "difficult to follow"? Was it the idea of the websocket setup? The way the cache lifetime callbacks work? Something about the explanation itself?

It mostly comes from the size of the code block. It needs to be broken up into smaller chunks.

There are 137 lines of code in that one code block with no explanation helping the reader be guided through what is happening or what is being changed until the very end. I need to scroll 4 times to get to the bottom of it.

This massive code block is overwhelming and intimidating. As a learner, I read through the code block to see if I can understand what it is doing without needing to read the explanation. This code block just keeps going and going with seemingly no end in sight. If there is something in the code that is weird or hard to understand, there is no available explanation to look for answers until you reach the very end.

Through all of the tutorial, I felt guided through step by step. When I reached this part of the tutorial I felt like I was left on my own to try and figure out this massive pile of code all by myself.

Having the full explanation at the very end is also problematic since it makes it difficult to understand what part of the explanation refers to what part of the code.

docs/tutorials/essentials/part-3-data-flow.md
markerikson: We've been pretty consistent with using `RootState` in all of our documentation, so I'd prefer not to change it here. (Really, the better answer is to change the export from RTKQ, but we can't do that until the next major version.)

docs/tutorials/essentials/part-4-using-data.md
markerikson: Agreed, but it's not critical, and it would require some additional tweaks to the mock server setup as well.

markerikson:
Updated Parts 5 and 6! Also pushed the code updates to https://github.com/reduxjs/redux-essentials-example-app/tree/feature/tutorial-steps-ts-post-review as well.

markerikson:
**DONE!**

I just pushed what _should_ be the final content update, barring any last-minute edits and revisions.

I significantly rewrote the entire "Optimistic Updates" and "Streaming Updates" section:

- Added sections explaining what the `onQueryStarted` and `onCacheEntryAdded` lifecycles are, what they're for, and what the different arguments / extra promises are
- Added a section describing the cache update utils
- Broke the "streaming updates" section into multiple discrete steps and reordered the changes:
  - First add a `getNotifications` endpoint and use the query hook in `<NotificationsList>`
  - Then switch to watching for `getNotifications.matchFulfilled` in the slice and track the client-side metadata
  - Set up the thunk that forces pushing updates over the websocket
  - And then _finally_ add `onCacheEntryAdded`, create the websocket, handle messages, and update the cache

A lot more words, but hopefully it's now a _much_ better explanation, there isn't one _giant_ code block, and the code snippets are a lot more readable.

The only other changes that _should_ be needed at this point are around the example app and sandboxes:

- figuring out final setup with the CodeSandbox embeds (talking with Ives about this)
- swapping the branches in the example app repo
- updating the sandbox links

and then we should be ready to launch!

Dan503:
Since my last attempt at implementing notifications using the old instructions failed I'll retry creating the feature using the [new instructions](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#streaming-updates-for-notifications).

Dan503:
The current method of creating the root reducer in the tutorial documentation is this:

```ts
export const store = configureStore({
  reducer: {
    [apiSlice.reducerPath]: apiSlice.reducer,
    counter: counterReducer
  }
})
```

When creating a project using  `npx create-next-app --example with-redux my-app` the root reducer is formed using this method instead:

```ts
const rootReducer = combineSlices(
  counterSlice,
  apiSlice,
)

const store = configureStore({
  reducer: rootReducer,
})
```

I see the `combineSlices` method as being the far superior method of forming the root reducer. It is much cleaner and it is easier to maintain. You don't even need to export the reducer from the Redux slice file for it to work.

There is no mention of this method anywhere in the tutorial at the moment. This should be the primarily endorsed and used method all throughout the tutorial. If not that, then there should at least be a mention of it around where [the concept of the root reducer is introduced](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-3-data-flow#creating-the-store). This will make that option available to learners so that they can decide for themselves which method they would prefer to use.

markerikson:
Yeah, I've been having discussions with the other maintainers about this.

`combineSlices` is a new method that was introduced in RTK 2.0, and shortly after that release we modified the templates to use it.

I specifically noted the difference between the templates and the tutorial a few weeks ago, and we had some discussions about teaching approaches.  

There's some similarities here between "legacy Redux vs RTK", in the sense that there's now multiple ways to write equivalent code.  This is true for using `createSlice` with the callback syntax for both synchronous reducers + actions, and defining thunks inside of `createSlice`, as well as with `combineSlices`.

> I see the combineSlices method as being the far superior method of forming the root reducer. It is much cleaner and it is easier to maintain. You don't even need to export the reducer from the Redux slice file for it to work.

_Unlike_ legacy Redux vs RTK, I don't see the new RTK 2.0 methods as being unambiguously better:

- `combineSlices` maybe saves a couple lines of code in terms of defining the `reducer` object inside of `configureStore`, but it also hides the direct correlation of field name and reducer function, like `posts: postsReducer`.  The primary benefit is really about allowing code splitting at runtime, and that A) isn't something the tutorial covers other than the RTKQ `api.injectEndpoints()` example, and B) isn't something that needs to be taught up front.
- The callback syntax for `createSlice` can save on needing to write `PayloadAction<SomeType>`, but it's also somewhat "noisier" visually and doesn't specifically have more advantages by itself
- Defining thunks inside of `createSlice` saves on having to write an action type prefix and the `extraReducers` field with `builder.addCase` calls, but also requires the extra `buildCreateSlice()` call first

There's reasons why you might want to use these, but I don't see any of them as being such a huge advantage that we ought to switch the defaults we teach. Instead, to me they make more sense as _optional_ features you _can_ use if you want to, but shouldn't be the primary approach taught.

Given that, we've got a PR up to alter the templates by dropping `combineSlices`, and to have both "standard" and "advanced" versions of the example `counterSlice` to contrast how you can do things.

Dan503:
I just completed the [Pushing Notifications Via Websocket](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#pushing-notifications-via-websocket) section of the tutorial.

[Current state of my code base](https://github.com/Dan503/playtest-ts-redux-example-app/tree/8c5c8ecc5d5aee336eb0c3e22f1d150cbfb9c36a)

```
TypeError: Cannot read properties of undefined (reading 'send')
    at sendMessage (server.ts:258:10)
    at sendRandomNotifications (server.ts:268:3)
    at forceGenerateNotifications (server.ts:272:3)
    at notificationsSlice.ts:24:5
```

Error encountered at server.ts:258:10

```
const sendMessage = (socket, obj) => {
  socket.send(JSON.stringify(obj));
        ^
};
```

If it isn't supposed to be functional yet then the wording at the end of that section needs to be updated:

> Almost there! We're fetching initial notifications via RTK Query, tracking read status on the client side, and we've got the infrastructure set up to force new notifications via a websocket.

This wording suggests that clicking the "refresh notifications" button will, in it's current form, fetch a new set of notifications without issue.

markerikson:
It's _supposed_ to be working :)

I did see that error happen a couple times as I was testing it. `server.ts` is supposed to be saving the current websocket reference when the client establishes the connection.  I'll have to dig into it a bit more and figure out when/why that isn't happening.  But it definitely _was_ working overall when I wrote and testing that code today.

Dan503:
[Implementing Streaming Updates](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#implementing-streaming-updates)

This line needs to be highlighted:

```ts
const notificationsReceived = createAction<ServerNotification[]>('notifications/notificationsReceived')
```

I didn't see that I needed to add it, then I ran into an error when forming the `matchNotificationsReceived` variable.

Dan503:
[Implementing Streaming Updates](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#implementing-streaming-updates)

```ts
const message: {
  type: 'notifications'
  payload: ServerNotification[]
} = JSON.parse(event.data)
```

I'm not sure why you are making a custom type for the `message` variable inside `listener()` when you could use `PayloadAction<ServerNotification[]>` as the type instead.

```ts
const message: PayloadAction<ServerNotification[]> = JSON.parse(event.data)
```

Dan503:
[Implementing Streaming Updates](https://deploy-preview-4706--redux-docs.netlify.app/tutorials/essentials/part-8-rtk-query-advanced#implementing-streaming-updates)

```ts
catch {
  // no-op in case `cacheEntryRemoved` resolves before `cacheDataLoaded`,
  // in which case `cacheDataLoaded` will throw
}
```

The wording on this error handler comment is confusing.

This is my understanding of that comment:

> Ignore any errors caused by the cache entry being removed (cacheEntryRemoved) before the cache data has finished loading (cacheDataLoaded)

Dan503:
I've completed the tutorial now. Due to [the error](https://github.com/reduxjs/redux/pull/4706#issuecomment-2257468097), I can't tell if I've completed it successfully or not. The new tutorial was much easier to follow though and I felt far less intimidated working through it this time. 

The web-socket portion is still complex. The comments throughout that code block go a long way to walking the reader through that bit of code though. The code-block sizes are much more reasonable now as well. I felt guided through the tutorial this time which is a huge improvement.

markerikson:
> If it isn't supposed to be functional yet then the wording at the end of that section needs to be updated:

@dan503 : Ah, yeah, the issue is that it _isn't_ supposed to be functional yet. The websocket connection doesn't get activated until the section after this, so it's expected that the button is actually "broken" at this point in the sequence.

I can try to clarify the instructions here, as well as maybe add an alert or log message if the socket doesn't exist yet.

markerikson:
> I'm not sure why you are making a custom type for the message variable inside listener() when you could use PayloadAction<ServerNotification[]> as the type instead.

Mostly because I wasn't thinking of it that way :)

The mock server backend _could_ send any arbitrary data over the socket.  I just happened to structure its message similarly to how a Redux action looks out of habit. Since I wasn't thinking of it as an _actual_ action, it didn't occur to me to reuse `PayloadAction` here.

Dan503:
@markerikson 

> It still needs the onCacheEntryAdded logic from the final commit / instruction sequence, which is where the Websocket itself is getting created and told to connect to the mock server.

Ok... I was still getting the error by the time I reached the end of the tutorial though 😕
The problem should have fixed itself when I reached the end if that is the only issue. 

[My code at the end of the tutorial](https://github.com/Dan503/playtest-ts-redux-example-app/blob/master/src/features/notifications/notificationsSlice.ts)

markerikson:
@Dan503 : checked out your code. I _don't_ see the `Cannot read properties of undefined (reading 'send')` error, and logging shows that the websocket is getting set up correctly.  

I _suspect_ that the issue might have something to do with hot module reloading, but I'm not entirely sure.

Behavior-wise, I see a bug in the `onCacheEntryAdded` handler - the `lifecycleApi.updateCachedData` thunk needs to be _dispatched_, and it isn't, so the new notification entries from the socket aren't actually getting added to the store.

So, I _think_ the example repo code and instructions are correct at this point, but I'll give them another look-through.

Dan503:
> Behavior-wise, I see a bug in the onCacheEntryAdded handler - the lifecycleApi.updateCachedData thunk needs to be dispatched, and it isn't, so the new notification entries from the socket aren't actually getting added to the store.

@markerikson I'm not sure what you mean by that.

The code written in the tutorial doesn't wrap `lifecycleApi.updateCachedData()` in a dispatch function.

It writes it like this:

```ts
switch (message.type) {
   case 'notifications': {
      lifecycleApi.updateCachedData((draft) => {
         // Insert all received notifications from the websocket
         // into the existing RTKQ cache array
         draft.push(...message.payload)
         draft.sort((a, b) => b.date.localeCompare(a.date))
      })
      break
   }
   default:
      break
}
```

My code:

```ts
if (message.type === 'notifications') {
   lifecycleApi.updateCachedData((draft) => {
      // Insert all received notifications from the websocket
      // into the existing RTKQ cache array
      draft.push(...message.payload)
      draft.sort((a, b) => b.date.localeCompare(a.date))
   })
}
```

markerikson:
Ah, yeah. Sorry, getting things mixed up :) `apiSlice.utils.updateCachedData` requires all the params and a separate dispatch, `lifecycleApi.updateCachedDate` does not because it has those wrapped in already.

I'll have to look further to see why your code isn't getting the notifications saved right. I see the websocket messages going through...

markerikson:
Ah, I see it! The _notifications_ are getting added fine, but `onCacheEntryAdded` is missing this:

```ts
 				// Dispatch an additional action so we can track "read" state
                lifecycleApi.dispatch(notificationsReceived(message.payload))
                break
```

so the `extraReducers` code isn't running and we don't get the metadata items added for the notification entries.

Dan503:
I noticed a bug in my code. I pushed up a fix for it. I didn't store the current user correctly when logging in so `currentUser` on the server side was always undefined.

That might have been causing some issues.

> lifecycleApi.dispatch(notificationsReceived(message.payload))

That is not written anywhere in the tutorial...

Dan503:
@markerikson  I've updated my listener code to this now and it's working now :)

```
// when data is received from the socket connection to the server,
// update our query result with the received message
function listener(event: MessageEvent<string>) {
   const message: PayloadAction<Array<ServerNotification>> = JSON.parse(
      event.data,
   )
   if (message.type === 'notifications') {
      lifecycleApi.updateCachedData((draft) => {
         // Insert all received notifications from the websocket
         // into the existing RTKQ cache array
         draft.push(...message.payload)
         draft.sort((a, b) => b.date.localeCompare(a.date))
      })

      // Dispatch an additional action so we can track "read" state
      lifecycleApi.dispatch(notificationsReceived(message.payload))
   }
}
```

So the instructions were missing this line of code
```ts
   // Dispatch an additional action so we can track "read" state
   lifecycleApi.dispatch(notificationsReceived(message.payload))
 break
```

markerikson:
Yep, my bad, I see I did forget to include those lines there!  Updating that locally.

Dan503:
So the [instruction clarification](https://github.com/reduxjs/redux/pull/4706#issuecomment-2258838341)

> I can try to clarify the instructions here, as well as maybe add an alert or log message if the socket doesn't exist yet.

And the [`notificationsReceived` dispatch](https://github.com/reduxjs/redux/pull/4706#issuecomment-2259476951)

> lifecycleApi.dispatch(notificationsReceived(message.payload))

Are the only two changes that the documentation needs (along with any "TODO" items you may have left for applying images or other things like that)

Dan503:
Oh also the highlighting of the following line of code as mentioned in https://github.com/reduxjs/redux/pull/4706#issuecomment-2257527908

```ts
const notificationsReceived = createAction<ServerNotification[]>('notifications/notificationsReceived')
```

That is a small though important change or it will confuse a huge number developers doing the tutorial.

markerikson:
Okay, **I _think_ this is done!**

- Updated all screenshots
- Fixed the missing bits of code in Part 8
- Added an alert from `server.ts` if the socket isn't initialized yet
- All sandboxes should be working. They're currently pointing to fixed commit IDs - I'll point them to the final tags once I switch the branches in the Essentials example repo

I'm sure there's leftover typos and things that could be improved, but I think this is good enough to merge and launch at this point!

@Dan503 **THANK YOU _SO MUCH_ FOR YOUR DETAILED REVIEW FEEDBACK!**  You made this _drastically_ better!

Dan503:
@markerikson the timing ended up working out ridiculously well... like _insanely_ well...

1. I'm in-between projects at work at the moment so my only task is self-directed study
2. I chose Redux as one of my study topics because I am using it in a side project and I only knew the basics, I wanted to know it in depth
3. The portability bug hadn't been fixed yet so I encountered it when trying to do the tutorial in TS on my own (which lead me to that original github issue)
4. None of the existing answers were working for me so I created my own solution and posted what I did to help future developers (so if any of the other answers worked for me this feedback wouldn't have happened)
6. The documentation for the TS version was already almost complete, just not published yet, allowing me to redo the full tutorial in TS with the correct types info this time
7. Since the tutorial was in PR stage and not published yet (and you explicitly asked for feedback) I felt empowered to comment. (If it had already been published I probably wouldn't have said anything)
8. Being in-between projects at work gave me the time needed to redo the tutorial and provide feedback and this seems like something important for the betterment of the wider web development community so I set it as a high priority

So yeah, a lot of converging coincidences to make this thing happen 🤯

PR #4708, opened 2024-05-29
chore: updated docs to remove extra heading tag

Zain-Bin-Arshad:
---
---

## Checklist

  - _link issue here_

- **Section**: Redux Fundamentals
- **Page**: All 8 pages

## What is the problem?
We are using docusaurus for the docs and it will automatically add a header tag from the `title` attribute. So there is no need to add a redundant `h1` tag containing the same title. Currently in the [docs here ](https://redux.js.org/tutorials/fundamentals/part-1-overview) it will display 2 titles with same text which is not a good user experience.

image

image

I have removed the extra `h1` tag and now we leave it to docusaurus to render the page's title. This is how it has been done 
on the sections i.e. [Redux Essentials](https://redux.js.org/tutorials/essentials/part-1-overview-concepts)


EskiMojo14:
thanks!

PR #4733, opened 2024-08-17
docs: Add extra reducers tutorial link to allow many reducers style guide

Lambdaphile:
I was reading the style guide the other day and noticed that "Allow Many Reducers to Respond to the Same Action" wasn't clear enough, especially for those who are not familiar with the never APIs like extraReducers and createSlice. Linking the extra reducers tutorial here, I think will be useful.


markerikson:
I've actually mostly tried to _not_ add links to the style guide, in contrast to the other docs page.

Could be worth reconsidering, but for now let's hold off on that.

PR #4737, opened 2024-09-05
fix(types): getState is not bound to MiddlewareAPI

janeklb:

Fixes a type bug

The `getState` function provided by `MiddlewareAPI` is not bound to any object, and as such it can be used freely as in the examples.

The `MiddlewareAPI` type does not make this explicit, and typescript linters can report this as problematic usage.

## Checklist

  - _link issue here_
  - there's a file that needs formatting, but is outside the scope of my change: `docs/usage/deriving-data-selectors.md`
  - No updates needed IMO
  - No new tests added - the linting rules required to 

- Enable the [@typescript-eslint/unbound-method](https://typescript-eslint.io/rules/unbound-method/) linting rule
- Load up one of the examples
  ```ts
  const middleware: Middleware<unknown, unknown> =
    ({ getState }) =>
    (next) =>
    (action) => {
      // ...
    };
  ```
- The linter will report that the rule is being violated

`getState` can be used with the linting rule enabled and not produce errors

Fixes the signature of `getState`

### DIFF
--- a/src/types/middleware.ts
+++ b/src/types/middleware.ts
@@ -2,7 +2,7 @@ import type { Dispatch } from './store'
 
 export interface MiddlewareAPI<D extends Dispatch = Dispatch, S = any> {
   dispatch: D
-  getState(this: void): S
+  getState: () => S
 }
 
 /**
--- a/src/types/middleware.ts
+++ b/src/types/middleware.ts
@@ -2,7 +2,7 @@ import type { Dispatch } from './store'
 
 export interface MiddlewareAPI<D extends Dispatch = Dispatch, S = any> {
   dispatch: D
-  getState(): S
+  getState(this: void): S
 }
 
 /**
--- a/test/typescript/middleware.test-d.ts
+++ b/test/typescript/middleware.test-d.ts
@@ -89,6 +89,18 @@ describe('type tests', () => {
     }
   })
 
+  test('middleware has unbound getState function.', () => {
+    type State = { field: 'string' }
+
+    const customMiddleware: Middleware<{}, State> = api => next => action => {
+      const { getState } = api
+
+      expectTypeOf(getState).thisParameter.toEqualTypeOf()
+
+      return next(action)
+    }
+  })
+
   test('middleware that expects custom dispatch.', () => {
     type MyAction = { type: 'INCREMENT' } | { type: 'DECREMENT' }
 
--- a/test/typescript/middleware.test-d.ts
+++ b/test/typescript/middleware.test-d.ts
@@ -89,18 +89,6 @@ describe('type tests', () => {
     }
   })
 
-  test('middleware has unbound getState function.', () => {
-    type State = { field: 'string' }
-
-    const customMiddleware: Middleware<{}, State> = api => next => action => {
-      const { getState } = api
-
-      expectTypeOf(getState).thisParameter.toEqualTypeOf()
-
-      return next(action)
-    }
-  })
-
   test('middleware that expects custom dispatch.', () => {
     type MyAction = { type: 'INCREMENT' } | { type: 'DECREMENT' }
 


src/types/middleware.ts
```diff
@@ -2,7 +2,7 @@ import type { Dispatch } from './store'
 
 export interface MiddlewareAPI<D extends Dispatch = Dispatch, S = any> {
   dispatch: D
-  getState(): S
+  getState(this: void): S
```
EskiMojo14, line 5:
```suggestion
  getState: () => S
```

timdorr:
This should have some sort of test to go along with it. That will prevent a regression in the future.

janeklb:
> This should have some sort of test to go along with it. That will prevent a regression in the future.

I'm able to create a test if we revert to
```ts
export interface MiddlewareAPI<D extends Dispatch = Dispatch, S = any> {
  dispatch: D
  getState(this: void): S
}
```

however, with the current signature (`getState: () => S`), I can't find a way to do it. @EskiMojo14 can you advise?

janeklb:
> > This should have some sort of test to go along with it. That will prevent a regression in the future.
> 
> I'm able to create a test if we revert to
> 
> ```ts
> export interface MiddlewareAPI<D extends Dispatch = Dispatch, S = any> {
>   dispatch: D
>   getState(this: void): S
> }
> ```
> 
> however, with the current signature (`getState: () => S`), I can't find a way to do it. @EskiMojo14 can you advise?

Found something that covers the bases, I think.

test/typescript/middleware.test-d.ts
```diff
@@ -89,6 +89,18 @@ describe('type tests', () => {
     }
   })
 
+  test('middleware has unbound getState function.', () => {
+    type State = { field: 'string' }
+
+    const customMiddleware: Middleware<{}, State> = api => next => action => {
+      const { getState } = api
+
+      expectTypeOf(getState).thisParameter.toEqualTypeOf()
```
EskiMojo14, line 98: this doesn't actually fail if the signature is changed back

if we really need a test, then we can switch back to the this: void method, but i'm not convinced that we should worry too much about testing a change solely to satisfy a lint rule, and I think this: void is unnecessary noise for a codebase that doesn't actually use the lint rule it's meant to satisfy

janeklb: I swear I tested this.. thanks for double checking it.

EskiMojo14: that would be my preference - @timdorr?

timdorr: I would add the test as one commit, which should fail, and then add the fix in the next commit to show that this satisfies the condition we're looking to correct.

EskiMojo14: right, but the point is that this is to satisfy a lint rule we're not using - so short of enabling the rule (which i would object to, and would require significant changes to our lint config to enable) I'm not sure how that would work

timdorr: Oh, I see what you're saying. Sorry, I was thinking about the actual test passing. 

As long as the test essentially functions as a lint check, I think that's sufficient. Can we get a spot check by importing this version into a project and seeing if that does pass the lint (vs the current release failing)?

janeklb: I've tested it out in my project and it fixes the lint violation.

Should I remove the test?

timdorr: That seems good enough to me. Leaving the test seems fine, since it at least validates the behavior and will trip if someone breaks this again later on.

EskiMojo14: the test doesn't actually verify anything, and will not break if anything is changed.

phryneas: It's an eslint rule fix. Even if this breaks, it will not cause any bugs in any production environment but only give someone red squigglies. I don't think it needs a test honestly.

timdorr:
Thanks!

PR #4744, opened 2024-09-26
Create reduceReducers util

EskiMojo14:
Based on https://github.com/redux-utilities/reduce-reducers/, but with (hopefully) better typings.


EskiMojo14:
**Size Change:** +99 B (+2.13%) 

**Total Size:** 4.76 kB

| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/redux.legacy-esm.js` | 4.76 kB | +99 B (+2.13%) |

<a href="https://github.com/preactjs/compressed-size-action"><sub>compressed-size-action</sub></a>

test/reduceReducers.spec.ts
```diff
@@ -0,0 +1,34 @@
+import reduceReducers from "@internal/reduceReducers";
```
aryaemami59, line 1:
```suggestion
import reduceReducers from 'redux'
```

aryaemami59:
Excited for this. Do you think this needs a type test as well?

EskiMojo14:
> Excited for this. Do you think this needs a type test as well?

can do - combineReducers doesn't really have one as far as i can tell, so i was following that precedent

PR #4748, opened 2024-10-31
Add redux.dev course callouts

lynnandtonic:
Hi Mark! This PR:

- Updates the docs website
- Adds three SVG images for redux.dev callouts (`narrow`, `mid`, and `wide` sizes) 
- Adds the mid size callout to the homepage

I wasn’t sure how to add the callouts to the internal docs pages (I’m not super familiar with docusaurus), is that something you could help me with?

Here is where I was thinking we could place them (either `mid` at the bottom of the main column or `narrow` in the right sidebar). Also open to your suggestions here!

callout-placement-docs

We could also potentially add the `wide` one above the nav (only on the homepage) if we wanted it higher up on the page. (Though not great for mobile and we’d probably want to hide or swap out there)

callout-placement-home

Let me know what you think works and if there’s anything else I can provide. Thanks!


markerikson:
Thanks! Since the PR is from your `master` branch, I'll just merge this and then fiddle with it a bit - probably going to add the "wide" version as a sitewide banner for the next few weeks. I'll also add it to the RTK and React-Redux sites as well.

markerikson:
Done! Updated all 3 docs sites ( https://redux.js.org, https://redux-toolkit.js.org , https://react-redux.js.org ) with the same top banner and front-page image.

Thanks!

PR #4749, opened 2024-11-01
Fix example in tutorial

dalvine:
Updated the example in the documentation to better reflect the current functionality more clearly and concisely for users

- [:book: Updated Documentation Content](?template=[[docs/tutorials/essentials/part-8-rtk-query-advanced.md](https://github.com/reduxjs/redux/compare/master...dalvine:redux:patch-1#diff-f8a620a1df128001cde78273f5ac84097d5bb8e2e44cd567af36692b08e82e6d)])


timdorr:
While this does appear to be correct, I don't think that selector is exported from the authSlice module. We should add that as well.

markerikson:
Hmm. Best to check this against the actual example code at that point in time.

PR #4761, opened 2024-12-16
exclude menu and TOC items from the visited link styles for better UX

Ahmed-Hakeem:
for better UX, menu and TOC items style should indicate only one thing, which is the current read item

When you click on an item and then scroll to another .... notice: the prev item still having the same style as the current one.

image


Ahmed-Hakeem:
I guess we need the same for the navbar menu... I will include it as well

timdorr:
Thank you!

PR #4765, opened 2025-01-17
Fix issue with `combineReducer` returning incorrect type when given a custom action

OliverJAsh:
Fixes https://github.com/reduxjs/redux/issues/4764. See issue for details.

Note I had to enable `exactOptionalPropertyTypes` to reproduce the bug in the test.

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -20,6 +20,7 @@
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
+    "exactOptionalPropertyTypes": true,
     "target": "ESnext",
     "types": ["vitest/globals", "vitest/importMeta"]
   },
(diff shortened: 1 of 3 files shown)


tsconfig.base.json
```diff
@@ -20,6 +20,7 @@
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
+    "exactOptionalPropertyTypes": true,
```
timdorr, line 23: I don't have a strong opinion, but would it make sense to move this into tsconfig.test.json instead?

OliverJAsh: I considered that but then I couldn't think of a reason for it not to live here. It's probably best here because it increases strictness for the whole repo.

timdorr (approved):

tsconfig.base.json
aryaemami59: This might cause issues with the type tests at some point.

timdorr: It doesn't now, though 🤷

aryaemami59: Yeah it's probably okay for now but it might cause a hiccup or 2 at some point.

[Reference](https://github.com/mmkal/expect-type/issues/128)

timdorr:
Welp, no one else has yelled at me, so I'll merge it in then. Thanks!

PR #4768, opened 2025-01-26
fix(part-4-store): misprint in text

lerfich:


markerikson:
"Pane" is valid and correct here, but "panel" is also reasonable. Thanks!

PR #4772, opened 2025-02-11
fix(rtkq-advanced-patterns): endpoint missprint

lerfich:
If I understood correctly, we want just to refetch getPosts via additional id in tag, but disable the refresh of each one by getPost endpoint (which is different request).


markerikson:
Thanks!

PR #4788, opened 2025-04-21
Update Docusaurus to v3 and switch search to Canary + Pagefind

markerikson:
This PR:

- Updates Docusaurus to v3
- Switches our search setup from Algolia to Canary + Pagefind


markerikson:
Skipped fixing some of the broken anchor looks that DS is reporting, but those don't seem to block the build

PR #4805, opened 2025-08-05
Docs: add link alongside `createSelector` memoization behavior

brennarvo:
When explaining the memoization behavior of `createSelector`, a link to help developers implement desired behavior should be provided.

---
---

## Checklist

  - _link issue here_

- **Section**: [`createSelector` Behavior](https://redux.js.org/usage/deriving-data-selectors#createselector-behavior)
- **Page**: [Deriving Data with Selectors](https://redux.js.org/usage/deriving-data-selectors)

## What is the problem?

The `createSelector` behavior explains memoization behavior, but does not provide a solution for enhancing said memoization behavior.

Adds a link to existing documentation that explains an enhancement to memoization behavior.

### DIFF
--- a/docs/usage/deriving-data-selectors.md
+++ b/docs/usage/deriving-data-selectors.md
@@ -265,6 +265,8 @@ const c = someSelector(state, 2) // different inputs, not memoized
 const d = someSelector(state, 1) // different inputs from last time, not memoized
 ```
 
+To address this, you can create [unique selector instances](https://redux.js.org/usage/deriving-data-selectors#creating-unique-selector-instances).
+
 Also, you can pass multiple arguments into a selector. Reselect will call all of the input selectors with those exact inputs:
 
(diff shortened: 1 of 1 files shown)


brennarvo:
> hmm, that section only applies when using `lruMemoize` anyway - `reselect` now defaults to `weakMapMemoize` which has no limit on cache size, so the docs might need a rethink.

@EskiMojo14 does this mean, then, that this code snippet in the docs is no longer accurate?

```js
const a = someSelector(state, 1) // first call, not memoized
const b = someSelector(state, 1) // same inputs, memoized
const c = someSelector(state, 2) // different inputs, not memoized
const d = someSelector(state, 1) // different inputs from last time, not memoized
```

I'm only wondering so that I can determine if I need to account for situations like `c` and `d` (from code snippet above) in my application.

PR #4844, opened 2026-02-06
chore: exclude build artifacts from tsconfig and add legacy_createSto…

Divyapahuja31:
## Description

This PR addresses two maintenance issues to improve repository health and developer experience:

1.  **Added tests for `legacy_createStore`**:
    *   The `legacy_createStore` export was previously uncovered by tests, leaving a gap in coverage for `createStore.ts`.
    *   I added `test/legacy_createStore.spec.ts` to verify the legacy alias functions correctly and exposes the same public API as `createStore`.
    *   **Result:** This brings test coverage for `createStore.ts` to **100%**.

2.  **Updated `tsconfig.json` exclusions**:
    *   Previously, running `yarn test` (which triggers `vitest --typecheck`) would fail with hundreds of type errors if the `website/`, `examples/`, or `coverage/` directories were present (e.g., after building the site or running coverage). This was because the root TypeScript config was attempting to check these directories which have their own unrelated configurations.
    *   I added `website`, `examples`, `docs`, and `coverage` to the `exclude` list in `tsconfig.json`.
    *   **Result:** `yarn test` now passes cleanly regardless of the state of these folders.

3.  **Fixed missing dependency**:
    *   Detailed coverage reports were failing because `@vitest/coverage-v8` was missing. I added it to `devDependencies`.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes


Divyapahuja31:
Thanks for the clarification, that makes sense. I agree that dedicated tests for legacy_createStore aren’t really needed since it’s just an alias and has been stable for a long time. I’ll remove the standalone test file and updated the PR accordingly. 
Appreciate the guidance.

PR #4864, opened 2026-04-21
Fix: Add missing `StateExt` generic to `StoreCreator` second overload

Olexandr88:
The enhancer parameter in the second overload of `StoreCreator` was typed as `StoreEnhancer<Ext>`, omitting `StateExt`. This caused TypeScript to never infer `StateExt` from the enhancer, always falling back to `{}`, despite it being declared in the overload's generics and used in the return type.


aryaemami59 (approved):

timdorr:
Thanks!

PR #4877, opened 2026-06-05
Update homepage URL to use HTTPS

okuramasafumi:
---
---

## Checklist

  - _link issue here_

- `package.json`

## What is the problem?

It uses `http` version of the website, which is a redirect to `https`.

It directly uses `https` as a homepage url.


timdorr:
Thanks://!

PR #4888, opened 2026-06-22
fix: use String() to safely serialize Symbol action.type in dispatch error message

JSap0914:
## What

When `dispatch()` receives an action whose `type` property is a `Symbol`, the error-message template literal

```ts
// src/createStore.ts  (~L289)
`…Value was: '${action.type}' (stringified)`
```

throws a **`TypeError: Cannot convert a Symbol value to a string`** before the intended `Error` can be constructed and thrown.  The developer sees the wrong exception with no indication that the real problem is a non-string action type.

## Why

Template-literal interpolation calls the abstract `ToString` operation, which is specified to throw a `TypeError` for Symbol values (ECMAScript §7.1.17).  Every other primitive (`boolean`, `number`, `null`, `bigint`) coerces fine, which is why the bug wasn't caught when PR #4544 added the check.

## Fix

Replace the bare interpolation with `String(action.type)`.  `String(Symbol('foo'))` returns `'Symbol(foo)'` without throwing, and is a no-op for every other non-string value.

```diff
- `…Value was: '${action.type}' (stringified)`
+ `…Value was: '${String(action.type)}' (stringified)`
```

## Test

Added a `Symbol` case to the existing *"throws if action type is not string"* test (alongside `false`, `0`, and `null` from #4544):

```ts
expect(() => store.dispatch({ type: Symbol('MY_ACTION') })).toThrow(
  /the actual type was: 'symbol'.*Value was: 'Symbol\(MY_ACTION\)'/
)
```

Before the fix the test fails with `"Cannot convert a Symbol value to a string"`; after the fix all 120 tests pass.

---
> AI-assisted contribution (GitHub Copilot / Claude)

### DIFF
--- a/src/createStore.ts
+++ b/src/createStore.ts
@@ -286,7 +286,7 @@ export function createStore<
       throw new Error(
         `Action "type" property must be a string. Instead, the actual type was: '${kindOf(
           action.type
-        )}'. Value was: '${action.type}' (stringified)`
+        )}'. Value was: '${String(action.type)}' (stringified)`
       )
     }
 
--- a/test/createStore.spec.ts
+++ b/test/createStore.spec.ts
@@ -606,6 +606,10 @@ describe('createStore', () => {
     )
     // @ts-expect-error
     expect(() => store.dispatch({ type: '' })).not.toThrow()
+    // @ts-expect-error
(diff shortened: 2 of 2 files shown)


timdorr:
Thanks!

PR #4902, opened 2026-07-25
chore: migrate ESLint to flat config

veksa:

Neither — it's a tooling update for linting.

ESLint 9 dropped the old `.eslintrc` format and ESLint 10 removed it for good, so the config had to move to flat config. Everything now lives in `eslint.config.mjs` on ESLint 10 with typescript-eslint 8 and refreshed React/hooks/jsx-a11y plugins. The old `.eslintrc.cjs` and `.eslintignore` are gone, and I leaned on the plugins' presets so the config stays small.

## Checklist

  - No
  - Not needed, tooling only.

Nothing at runtime. Just flat config plus the newer plugin rule sets.

None.

Not a bug. The new baseline just surfaced a couple of small things I fixed along the way.

Everything lints and runs the same, on current tooling.

A few never-reassigned `let`s became `const`, and one test now uses rest params instead of `arguments`. The single spot in `createStore` that truly needs `arguments` keeps a narrow inline disable with a comment. Lint and tests (including the type tests) pass locally.


markerikson:
Honestly, if we ever actually touch our lint configs, I'd rather we just switch to `oxlint` at this point :)

veksa:
@markerikson Yeah, oxlint's tempting for sure. I'd just do it as a separate PR though — flat config is something we need regardless (eslintrc is gone in ESLint 10), and this change is low-risk with no rule changes. oxlint's type-aware support is still pretty early, and we rely on the typescript-eslint rules a lot here, so swapping fully would need its own look at coverage first. Can spin up a follow-up to try it side-by-side if you want.

markerikson:
That's my point. I don't see a reason to _upgrade_ ESLint as this time, and especially in this repo where we don't have any actual active _development_ work going on atm.  What we've got works, so there's no immediate need to bump the linting setup.

veksa:
Fair enough, no urgency here and "it works" is a good enough reason to leave it as is for now.

For what it's worth, I went ahead and threw together the oxlint version on a separate [branch](https://github.com/reduxjs/redux/pull/4903) just to see how it'd shake out, so it's there if the topic ever comes back around. Drops the entire ESLint dep chain and keeps the same rule coverage. No need to look at it now though, happy to leave the whole thing until there's an actual reason to touch linting.

PR #4903, opened 2026-07-26
chore: migrate lint tooling from ESLint to oxlint

veksa:

Neither, really. It's a tooling change that swaps our linter from ESLint over to oxlint.

This came out of the flat-config discussion in #4902. The takeaway there was that if we're going to touch the lint setup anyway, we might as well move to oxlint instead of investing more into ESLint. So this is that move.

oxlint runs the whole lint in a fraction of the time, and switching to it lets us drop the entire ESLint dependency chain (`eslint`, the typescript-eslint packages, `eslint-config-react-app` and the various plugins) along with the `resolutions` pins we needed to keep them compatible. That's around 230 fewer transitive dev dependencies. The config also collapses into a single `.oxlintrc.json` at the root, so we no longer need `.eslintrc.cjs`, `.eslintignore`, and the separate `.eslintrc.cjs` override that lived under the type tests.

I kept the rule coverage in line with what we had before. For `src` and `test` we still enforce `no-unused-vars` (with the same ignore patterns) and `consistent-type-imports` as errors. The example apps keep getting linted with the React, react-hooks and jsx-a11y rules, surfaced as warnings the same way they were, since that code is intentionally kept simple for teaching.

## Checklist

  - No standalone issue, but it follows directly from the discussion in #4902.
  - Not needed, this is tooling only. The `CONTRIBUTING.md` lint instructions still hold since `yarn lint` is unchanged.

Nothing at runtime. It only changes how we lint.

None.

Not a bug. The new baseline did surface two small things worth fixing, though.

Everything lints and runs the same, just on oxlint.

The two findings oxlint flagged are fixed inline: `warning.ts` now uses an optional catch binding instead of an unused `catch (e)`, and one test variable that's intentionally left undefined gets a narrow disable comment explaining why. `yarn lint`, the full test suite (including the type tests) and `format:check` all pass locally.


markerikson:
Superseded by the "Modern Tooling" PR stack, thanks!

PR #4904, opened 2026-07-26
docs: fix Store API links

ANSHUL-REAL:
## Checklist

  - No existing issue found for these broken links.

- **Section**: Understanding Redux and Usage
- **Page**: `Glossary.md` and `Troubleshooting.md`

## What is the problem?

Several Store API links use paths relative to `docs/`, even though the pages that contain them are nested further down the documentation tree. Those paths resolve to files that do not exist.

Updates 14 links so they resolve to the existing Store API and related API reference pages.

## Validation

- Verified every relative Markdown target in `docs/` resolves to an existing file
- `yarn prettier --check docs/understanding/thinking-in-redux/Glossary.md docs/usage/Troubleshooting.md`
- `yarn lint`
- `yarn test` (119 tests)

### DIFF
--- a/docs/understanding/thinking-in-redux/Glossary.md
+++ b/docs/understanding/thinking-in-redux/Glossary.md
@@ -13,7 +13,7 @@ This is a glossary of the core terms in Redux, along with their type signatures.
 type State = any
 ```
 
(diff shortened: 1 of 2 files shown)


EskiMojo14:
i don't think this is necessary, the links work - without the leading `/`, they're resolved against the root, not relatively

ANSHUL-REAL:
Thanks for clarifying, I misunderstood how the docs link resolution works. I’ll close this PR.

PR #4905, opened 2026-07-27
chore: switch from Prettier to oxfmt

veksa:
## PR Type

This is a tooling/chore change: it swaps out our code formatter, but doesn't touch runtime behavior.

This switches the formatter for `src` and `test` from Prettier to oxfmt, the Oxc formatter. The main win is speed: it's dramatically faster while aiming to match Prettier's output, so we get the same style with much quicker runs (formatting the whole project finishes in well under 100ms here).

The nice part is that oxfmt matches Prettier v3's output, and since we were already on Prettier v3, none of the source files needed reformatting. The diff is limited to config and tooling.

## What changed

- Removed `.prettierrc.json` and added `.oxfmtrc.json` with the same style settings we had before (no semicolons, single quotes, two-space indent, no trailing commas, avoid arrow parens) plus `printWidth: 80` to keep things explicit.
- Updated the `format` and `format:check` scripts to call oxfmt instead of Prettier.
- Replaced `prettier` with `oxfmt` in devDependencies.
- Added a "Check formatting" step to CI right after the lint step so formatting is enforced on every run.

One thing worth calling out: the old `format` script also ran Prettier over `**/*.md`. oxfmt only handles JS/TS/JSON, so markdown is no longer auto-formatted. If we want to keep formatting the docs, we can wire up a separate step for that — happy to do it either way.

## Checklist

  - No existing issue.


markerikson:
Superseded by the "Modern Tooling" PR stack, thanks!

PR #4912, opened 2026-08-01
docs: clarify normalized relationship storage

### LINKED ISSUE #4234: Normalizing State Shape - Relationships and Tables​ - Join/associative table example
First of all, I love the new documentation, last time I checked it was years ago and was way more confusing.

Using Redux > Redux logic and patterns > Structuring reducers > Nomalizing State Shape:  
- Relationships and Tables​ > Join/associative table example
https://redux.js.org/usage/structuring-reducers/normalizing-state-shape#relationships-and-tables

## What is the problem?

From the documentation, we have this given code:
```javascript
{
    entities: {
        authors : { byId : {}, allIds : [] },
        books : { byId : {}, allIds : [] },
        authorBook : {
            byId : {
                1 : {
                    id : 1,
                    authorId : 5,
                    bookId : 22
                },
                2 : {
                    id : 2,
                    authorId : 5,
                    bookId : 15,
                },
                3 : {
                    id : 3,
                    authorId : 42,
                    bookId : 12
                }
            },
            allIds : [1, 2, 3]

        }
    }
}
```
It is stated:
> Operations like "Look up all books by this author", can then be accomplished easily with a single loop over the join table.

If I understood well, the goal is to loop over `authorBook` keys to save all `bookId`s where `authorId` is the one we want.  
I struggle to understand how it is better than retrieving `bookId`s directly with `authors.byId[id].books`?
However, if we didn't want to store `bookId`s inside the `author` schema, we could have used a look up table like this one:
```javascript
booksByAuthorId : {
                5 : {
                    authorId : 5,
                    bookIds : [22, 23]
                },
           }
```
Wouldn't it make more sense?

I asked my backend team about it and they were also confused about it.

It would be really appreciated if someone could enlighten us.

If we couldn't understand it, maybe it's a sign it needs to be more explicit (e.g. more/better examples).

ThomasConway01:
## PR Type

Updates an existing page.

## Checklist

  - Fixes #4234
  - `git diff --check` passes. The repository's frozen Yarn install stops because it would modify the checked-in lockfile in this environment, so I kept the page's existing formatting outside this focused section.

- **Section**: Usage > Structuring Reducers
- **Page**: Normalizing State Shape

Clarifies when a related-ID array is the simplest relationship representation, when a join table is useful, and when an additional lookup index is worth maintaining. The join-table example now includes relationship metadata so it explains a case that an ID list alone cannot represent.

No. The existing audience stays the same, with more guidance for readers deciding how to model relational state.

### DIFF
--- a/docs/usage/structuring-reducers/NormalizingStateShape.md
+++ b/docs/usage/structuring-reducers/NormalizingStateShape.md
@@ -173,7 +173,39 @@ This could be expanded in a number of ways. For example, an application that doe
 
 ## Relationships and Tables
 
-Because we're treating a portion of our Redux store as a "database", many of the principles of database design also apply here as well. For example, if we have a many-to-many relationship, we can model that using an intermediate table that stores the IDs of the corresponding items (often known as a "join table" or an "associative table"). For consistency, we would probably also want to use the same `byId` and `allIds` approach that we used for the actual item tables, like this:
+Because we're treating a portion of our Redux store as a "database", many of the principles of database design also apply here as well. There is no single required way to store relationships. Choose the shape that makes the reads and updates your application needs straightforward.
+
+### Storing related IDs on an entity
+
+If you usually navigate a relationship in one direction and the relationship has no data of its own, an array of related IDs on the entity is often the simplest option. For example, an author record can store the IDs of that author's books:
+
+```js
+{
+    authors: {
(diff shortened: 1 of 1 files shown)


ThomasConway01:
Glad to help! Thanks for the review.

PR #4914, opened 2026-08-16
Fixes #4662: Remove manual reducer recipes in favour of RTK utilities

### LINKED ISSUE #4662: Add `combineSlices` section to Code Splitting page

- **Section**: Usage
- **Page**: Code Splitting

Now that RTK has its own code-splitting utilities (combineSlices and createDynamicMiddleware), it would be very useful to have an in-depth of how these are intended to be used, separate from the API reference.

This would avoid the standard "recipe" approach in the current docs in favour of depending on an actual library.

TheRodzz:
Fixes #4662 by removing the older, manual 'Reducer Injection Approaches' recipes, relying entirely on the new `combineSlices` and `createDynamicMiddleware` utilities from Redux Toolkit.


timdorr:
Thanks!

Issue #4464, opened 2023-01-16
Document usage of selectFromState thunk


EskiMojo14:
## What docs page is being updated?

- **Section**: Using Selectors with React-Redux
- **Page**: Deriving Data with Selectors

### What updates should be made to the page?
Adding documentation regarding the `selectFromState` thunk for selecting the most recent value from state when you only have access to `dispatch`, or want a one-time read from state without a subscription.
```ts
const selectFromState = <Selected>(selector: (state: RootState) => Selected): AppThunk<Selected> => (dispatch, getState) => selector(getState());

const users = dispatch(selectFromState(selectUsers));
```

### Do these updates change any of the assumptions or target audience? If so, how do they change?
I don't think so - this is a pattern often advised to use to help people who need a one-shot selection rather than maintaining a subscription.

raising after discussion with @markerikson in Reactiflux :)

PR #4931, opened 2026-09-14
docs: describe one-time state reads with selector thunks

Hugohong258:
## PR Type

Updates an existing page.

## Checklist

  - #4464

- **Section**: Usage > Redux Logic > Selectors
- **Page**: Deriving Data with Selectors

Add a short section showing how code that only has access to `dispatch` can perform a one-time selector read through a typed thunk. The text distinguishes this snapshot read from subscription APIs and directs React components that need updates to `useSelector`.

No. The addition uses the existing `RootState` and `AppThunk` types documented elsewhere and targets readers already learning selector usage.

Verification: `pnpm format:check` passed for all 119 matched files.


timdorr:
Thanks!

PR #4899, opened 2026-07-22
docs: recommend `createEntityAdapter` for normalized state

Hashim1999164:
## Summary
- Update the Normalizing Nested Data section to recommend Redux Toolkit's `createEntityAdapter` for storing and updating normalized entity collections.
- Keep Normalizr as an option for complex relational API transforms, with a note that it is stable but no longer actively maintained.
- Add matching links on the Prerequisite Concepts reading list.

Fixes #4699

## Test plan
- [ ] Preview the Normalizing State Shape page and confirm the new guidance and links render correctly
- [ ] Confirm the Essentials tutorial and `createEntityAdapter` API links resolve

Issue #4421, opened 2022-09-23
Redux FAQ: Organizing State Outdated


wadsworj:
## What docs page needs to be fixed?

- **Faq**:
- **Redux FAQ: Organizing State**:

https://redux.js.org/faq/organizing-state

## What is the problem?

I was looking at this today and this FAQ seems outdated.

In particular, it links to two community packages that have been out of date since 2015:

[redux-component](https://github.com/tomchentw/redux-component), [redux-react-local](https://github.com/threepointone/redux-react-local)

It also makes mention of this.setState().

There is no mention of RTK, useSelector, or react hooks.

As a beginner, this is a very confusing page to come across via Google.

## What should be changed to fix the problem?

Ideally - it should be updated. If not - it should be marked as depreciated or removed.

markerikson:
Yeah, I wrote the original FAQ page back in 2016, and most of it hasn't been updated since then. Unfortunately, I'm busy enough that I haven't had time to go through and clean it up.  Sorry!

The general principles in the FAQ are still 100% valid, but agreed that some of the _specific_ bits of advice and links are out of date.

If anyone else would like to help sweep through the FAQ entries and update things, I'd really appreciate it!

dhananjaysa92:
Looking into this issue

PR #4932, opened 2026-09-17
docs: recommend createEntityAdapter in the normalized state FAQ

ryanchou1994:
## Checklist

  - Refs #4421

- **Section**: FAQ
- **Page**: Organizing State — How do I organize nested or duplicate data in my state?

## What is the problem?

The normalized-data answer still points readers only to Normalizr and Redux-ORM. It does not mention Redux Toolkit's `createEntityAdapter`, which the normalization guide now recommends following #4899.

Keep the existing normalization principles and recommend `createEntityAdapter` for managing normalized collections. Explain that its reducers and selectors work with `{ ids, entities }`, while nested API responses still need to be normalized separately. Link to the API reference and the existing nested-data guide.

Validation: targeted `oxfmt --check` and full Docusaurus production builds passed on macOS and Linux. The generated FAQ text and added link targets match on both platforms, including the internal destination anchor. The builds report warnings from an unchanged dependency and anchors elsewhere in the site; the new internal link was checked directly. No runtime code changes.


EskiMojo14 (approved):

EskiMojo14:
thanks!

PR #4952, opened 2026-09-22
docs: fix empty link in counter-vanilla README

haimingZZ:

- **Section**: Examples
- **File**: `examples/counter-vanilla/README.md`

## What is the problem?

The README has a markdown link whose target URL was never filled in:

```md
As Dan said [in the original PR for this example]():
```

The `()` after the link text renders as an empty link on the docs site. The README was added in #3320 (2018) without the URL, and the broken link has been sitting there ever since.

Point the link at the PR that actually introduced the example: #1289 (Dan Abramov, 2016-01-27, "Add counter-vanilla example"). After this change:

```md
As Dan said [in the original PR for this example](https://github.com/reduxjs/redux/pull/1289):
```

The FAQ at `docs/faq/General.md` already references the same quote via a different link; this fix only restores the missing source URL on the example README.

## Checklist

- [x] Is there an existing issue for this PR? Not that I could find — this is a small, low-risk docs fix.
- [x] Have the files been linted and formatted? Yes — single line change in a Markdown file, no formatting impact.
- [x] Have the changes been tested? The change is a single markdown URL fix; verified with `git diff` that only the link target was touched.


EskiMojo14 (approved):

EskiMojo14:
thanks!

